PR #2, opened 2016-07-06
Add lein-cljsbuild and adapt datascript's release-js vehicle.

ncalexan:
@rnewman for your interest.


ncalexan:
This has landed.

PR #4, opened 2016-07-06
Isqliteconnection

ncalexan:
@rnewman built on top of old, 'cuz lazy.

The idea here is to work out some minimal SQLite interface that can be provided by all of `sqlite3` (although I found `promise-sqlite` easier), `SQLite.jsm`, and `jdbc`.

I think it's possible to implement the protocol directly on the JS object, but I wasn't confident, and this is  pretty straightforward.


ncalexan:
A version of this has landed.

PR #5, opened 2016-07-07
Query parsing

rnewman:


ncalexan:
Looks good!  We'll see about that `reduce` vs. `recur`.  Recur plays better with go blocks; wonder how much that'll matter...

rnewman:
db68a71

PR #7, opened 2016-07-13
Part 1: Implement database on top of SQLite connection.

ncalexan:
We would prefer to talk about a knowledge base on top of a database, but
all the Datomic and DataScript code (and symbols, like :db/add, etc)
refer to the "database of datoms", so let's roll with that nomenclature
and try to be specific that the persistent storage-layer is SQLite.
This will become more clear when we actually use SQLite's unique
capabilities for text indexing.

### DIFF
--- a/src/datomish/db.cljc
+++ b/src/datomish/db.cljc
@@ -1,0 +1,38 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+(ns datomish.db
+  #?(:cljs
+     (:require-macros
+      [datomish.pair-chan :refer [go-pair <?]]
+      [cljs.core.async.macros :refer [go]]))
+  #?(:clj
+     (:require
+      [datomish.sqlite :as s]
+      [datomish.sqlite-schema :as sqlite-schema]
+      [datomish.pair-chan :refer [go-pair <?]]
+      [clojure.core.async :refer [go <! >!]])
+     :cljs
+     (:require
+      [datomish.sqlite :as s]
+      [datomish.sqlite-schema :as sqlite-schema]
+      [datomish.pair-chan]
+      [cljs.core.async :as a :refer [<! >!]])))
+
+(defprotocol IDB
+  (close
+    [db]
+    "Close this database. Returns a pair channel of [nil error]."))
+
+(defrecord DB [sqlite-connection]
+  IDB
+  (close [db] (close (.-sqlite-connection db))))
+
+(declare ensure-current-version)
+
+(defn with-sqlite-connection [sqlite-connection]
+  (go-pair
+    (when-not (= sqlite-schema/current-version (<? (sqlite-schema/ensure-current-version sqlite-connection)))
+      (throw (Exception. "badness ensued"))) ;; TODO: raise
+    (->DB [sqlite-connection])))
--- a/src/datomish/sqlite_schema.cljc
+++ b/src/datomish/sqlite_schema.cljc
@@ -1,0 +1,61 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+(ns datomish.sqlite-schema
+  #?(:cljs
+     (:require-macros
+      [datomish.pair-chan :refer [go-pair <?]]
+      [cljs.core.async.macros :refer [go]]))
+  #?(:clj
+     (:require
+      [datomish.sqlite :as s]
+      [datomish.pair-chan :refer [go-pair <?]]
+      [clojure.core.async :refer [go <! >!]])
+     :cljs
+     (:require
+      [datomish.sqlite :as s]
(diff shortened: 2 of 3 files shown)


src/datomish/db.cljc
```diff
@@ -0,0 +1,38 @@
...
+    "Close this database. Returns a pair channel of [nil error]."))
+
+(defrecord DB [sqlite-connection]
+  IDB
+  (close [db] (close (.-sqlite-connection db))))
+
+(declare ensure-current-version)
+
+(defn with-sqlite-connection [sqlite-connection]
+  (go-pair
```
rnewman: Do we need a naming convention for these functions that return a promise channel?

src/datomish/sqlite_schema.cljc
```diff
@@ -0,0 +1,61 @@
...
+     (:require
+      [datomish.sqlite :as s]
+      [datomish.pair-chan]
+      [cljs.core.async :as a :refer [<! >!]])))
+
+(def current-version 1)
+
+(def v1-statements
+  ["CREATE TABLE datoms (e INTEGER NOT NULL, a TINYINT NOT NULL, v BLOB NOT NULL, tx INTEGER NOT NULL, index_avet TINYINT NOT NULL DEFAULT 0, index_vaet TINYINT NOT NULL DEFAULT 0)"
+   "CREATE TABLE transactions (e INTEGER NOT NULL, a TINYINT NOT NULL, v BLOB NOT NULL, tx INTEGER NOT NULL, added TINYINT NOT NULL DEFAULT 1)"
```
rnewman: We also need `CREATE TABLE attributes (name TEXT NOT NULL PRIMARY KEY, a UNIQUE INT NOT NULL)`, and an in-memory slurp/flush of that, I think.

src/datomish/sqlite_schema.cljc
```diff
@@ -0,0 +1,61 @@
...
+     :cljs
+     (:require
+      [datomish.sqlite :as s]
+      [datomish.pair-chan]
+      [cljs.core.async :as a :refer [<! >!]])))
+
+(def current-version 1)
+
+(def v1-statements
+  ["CREATE TABLE datoms (e INTEGER NOT NULL, a TINYINT NOT NULL, v BLOB NOT NULL, tx INTEGER NOT NULL, index_avet TINYINT NOT NULL DEFAULT 0, index_vaet TINYINT NOT NULL DEFAULT 0)"
```
rnewman: `a` should probably be a `SMALLINT`.

src/datomish/sqlite_schema.cljc
```diff
@@ -0,0 +1,61 @@
...
+   "CREATE TABLE transactions (e INTEGER NOT NULL, a TINYINT NOT NULL, v BLOB NOT NULL, tx INTEGER NOT NULL, added TINYINT NOT NULL DEFAULT 1)"
+   "CREATE INDEX eavt ON datoms (e, a)" ;; No v -- that's an opt-in index.
+   "CREATE INDEX aevt ON datoms (a, e)" ;; No v -- that's an opt-in index.
+   "CREATE INDEX avet ON datoms (a, v, e) WHERE index_avet = 1" ;; Opt-in index: only if a has :db/index true.
+   "CREATE INDEX vaet ON datoms (v, a, e) WHERE index_vaet = 1" ;; Opt-in index: only if a has :db/valueType :db.type/ref
```
rnewman: We might consider doing more here, given that we're able — e.g., it's possible to have a partial index where `v = 1` to catch boolean true, perhaps even with a particular attribute. Using SQL for the underlying storage gives us a lot of options for defining special-purpose indices.

src/datomish/sqlite_schema.cljc
```diff
@@ -0,0 +1,61 @@
...
+   "CREATE INDEX tx ON transactions (tx)"])
+
+(defn create-current-version
+  [db]
+  (go-pair
```
rnewman: This should be in a transaction.

src/datomish/db.cljc
ncalexan: I'd be fine with one.  Somebody suggest `<foo`, so `(<? (<foo))`.

src/datomish/sqlite_schema.cljc
ncalexan: This is specifically for interning, right?  I have ideas about persisting the schema.

src/datomish/sqlite_schema.cljc
ncalexan: I wondered about putting this in a transaction, but observed it wasn't in a transaction in Tofino.  Was that an oversight?

src/datomish/db.cljc
rnewman: Yeah, that's the one I saw. Seems fine to me.

src/datomish/sqlite_schema.cljc
rnewman: Yeah. But there's a shorter-term version of persisting the schema, which is just persisting the mapping between keyword and `SMALLINT`, which we would be able to roll with in the mean time.

src/datomish/sqlite_schema.cljc
rnewman: I think some libraries (e.g., Android) call `onCreate` and friends in transactions. I don't know if ours does — if it doesn't, we should do it explicitly here.

PR #9, opened 2016-07-14
Initial work on executing queries

rnewman:
Without having consumers, it's hard to tell which patterns make the most sense — returning a channel, a lazy sequence, and eager sequence, and promise variations of these. Still, this is enough to get an idea. Opinions please, @ncalexan.

### DIFF
--- a/src/datomish/exec.cljc
+++ b/src/datomish/exec.cljc
@@ -1,3 +1,7 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
 (ns datomish.exec
   #?(:cljs
      (:require-macros
--- a/src/datomish/exec.cljc
+++ b/src/datomish/exec.cljc
@@ -8,11 +8,11 @@
      [datomish.sqlite-schema :as ss]
      [datomish.query :as dq]
      #?@(:clj
-           [[datomish.jdbc-sqlite :as j]
+           [[datomish.jdbc-sqlite]
             [datomish.pair-chan :refer [go-pair <?]]
             [clojure.core.async :refer [go <!! <! >! chan close! take!]]])
      #?@(:cljs
-           [[datomish.promise-sqlite :as j]
+           [[datomish.promise-sqlite]
             [datomish.pair-chan]
             [cljs.core.async :as a :refer [<!! <! >! chan close! take!]]])))
 
@@ -46,7 +46,7 @@
    and return the results as a sequence wrapped in a promise."
   [path find]
   (go-pair
-    (let [d (<? (j/open path))]
+    (let [d (<? (s/<sqlite-connection path))]
       (try
         (<? (ss/<ensure-current-version d))
         (doall (run-to-seq d find))
--- a/src/datomish/exec.cljc
+++ b/src/datomish/exec.cljc
@@ -5,6 +5,7 @@
 (ns datomish.exec
   #?(:cljs
      (:require-macros
+      [datomish.util :refer [while-let]]
       [datomish.pair-chan :refer [go-pair <?]]
       [cljs.core.async.macros :refer [go]]))
   (:require
@@ -14,12 +15,18 @@
      #?@(:clj
            [[datomish.jdbc-sqlite]
             [datomish.pair-chan :refer [go-pair <?]]
-            [clojure.core.async :refer [go <!! <! >! chan close! take!]]])
+            [datomish.util :refer [while-let]]
+            [clojure.core.async :refer
+             [go                   ; macro in cljs.
+              <! >! chan close! take!]]])
      #?@(:cljs
            [[datomish.promise-sqlite]
             [datomish.pair-chan]
-            [cljs.core.async :as a :refer [<!! <! >! chan close! take!]]])))
+            [datomish.util]
+            [cljs.core.async :as a :refer
+             [<! >! chan close! take!]]])))
 
+#_
 (defn <run
   "Execute the provided query on the provided DB.
    Returns a transduced channel of results.
--- a/src/datomish/exec.cljc
+++ b/src/datomish/exec.cljc
@@ -1,0 +1,67 @@
+(ns datomish.exec
+  #?(:cljs
+     (:require-macros
+      [datomish.pair-chan :refer [go-pair <?]]
+      [cljs.core.async.macros :refer [go]]))
+  (:require
+     [datomish.sqlite :as s]
+     [datomish.sqlite-schema :as ss]
+     [datomish.query :as dq]
+     #?@(:clj
+           [[datomish.jdbc-sqlite :as j]
+            [datomish.pair-chan :refer [go-pair <?]]
+            [clojure.core.async :refer [go <!! <! >! chan close! take!]]])
+     #?@(:cljs
+           [[datomish.promise-sqlite :as j]
+            [datomish.pair-chan]
+            [cljs.core.async :as a :refer [<!! <! >! chan close! take!]]])))
+
+(defn <run
+  "Execute the provided query on the provided DB.
+   Returns a transduced channel of results.
+   Closes the channel when fully consumed."
+  [db find]
+  (let [context (dq/find->prepared-context (dq/parse find))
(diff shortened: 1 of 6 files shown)


src/datomish/query.cljc
```diff
@@ -29,7 +29,7 @@
 ;; `:bindings` is a map from var to qualified columns.
 ;; `:wheres` is a list of fragments that can be joined by `:and`.
```
ncalexan: Update comment.

src/datomish/query.cljc
```diff
@@ -53,7 +53,7 @@
       (raise (str "Couldn't find variable " variable))))
 
 (defn make-context []
-  (->Context [] {} []
+  (->Context [] {} [] []
```
ncalexan: Consider `map->Context` for clarity.

src/datomish/query.cljc
```diff
@@ -147,19 +150,30 @@
...
+  [context]
+  (let [elements (:elements context)]
+    (when-not (every? #(instance? Variable %1) elements)
+      (raise "Unable to :find non-variables."))
+    (map (fn [elem]
+           (let [var (:symbol elem)]
+             [(lookup-variable context var) (var->sql-var var)]))
+         elements)))
+
+(defn row-transducer [context projection rf]
```
ncalexan: Why do you want the `context` here?  Is this just `(map row-mapper)`?  Will it grow?

src/datomish/query.cljc
```diff
@@ -174,21 +188,28 @@
...
+    (apply-elements-to-context
       (expand-where-from-bindings
         (patterns->context where))    ; 'where' here is the Datalog :where clause.
       (:elements find))))
 
+(defn find->sql-clause
+  "Take a parsed `find` expression and turn it into a structured SQL
+   expression that can be formatted by honeysql."
+  [find]
+  ;; There's some confusing use of 'where' and friends here. That's because
```
ncalexan: Drop misleading comment.

src/datomish/exec.cljc
```diff
@@ -0,0 +1,67 @@
...
+  #?(:cljs
+     (:require-macros
+      [datomish.pair-chan :refer [go-pair <?]]
+      [cljs.core.async.macros :refer [go]]))
+  (:require
+     [datomish.sqlite :as s]
+     [datomish.sqlite-schema :as ss]
+     [datomish.query :as dq]
+     #?@(:clj
+           [[datomish.jdbc-sqlite :as j]
```
ncalexan: With the newest factory work, you should be able to drop the `:as j` here and below, and use `s/<sqlite-connection` in both situations.  Try it!

src/datomish/query.cljc
```diff
@@ -207,8 +204,7 @@
   ;; There's some confusing use of 'where' and friends here. That's because
   ;; the parsed Datalog includes :where, and it's also input to honeysql's
   ;; SQL formatter.
-  (context->sql-clause
-    (find->prepared-context find)))
+  (-> find find->prepared-context context->sql-clause))
```
ncalexan: I find this visually hard to parse; prefer newlines between sexps.

src/datomish/query.cljc
```diff
@@ -161,16 +160,11 @@
...
   ;; our transducer is trivial.
-  (let [columns-in-order (map second projection)
-        row-mapper (fn [row] (map columns-in-order row))]
-    (fn
-      ([] (rf))
-      ([result] (rf result))
-      ([result input]
-       (rf result (row-mapper input))))))
+  (let [columns-in-order (map second projection)]
+    (map (fn [row] (map row columns-in-order)))))
```
ncalexan: Oh hey, you did this :)

src/datomish/exec.cljc
```diff
@@ -0,0 +1,67 @@
...
+            [datomish.pair-chan :refer [go-pair <?]]
+            [clojure.core.async :refer [go <!! <! >! chan close! take!]]])
+     #?@(:cljs
+           [[datomish.promise-sqlite :as j]
+            [datomish.pair-chan]
+            [cljs.core.async :as a :refer [<!! <! >! chan close! take!]]])))
+
+(defn <run
```
ncalexan: I'm not convinced streaming is the right approach.  I think I'd rather see us pull into memory; streaming across channels is fraught.

src/datomish/exec.cljc
```diff
@@ -0,0 +1,67 @@
...
+        chan (chan 50 row-transducer)]
+
+    (s/<all-rows db (dq/context->sql-string context) chan)
+    chan))
+
+(defn channel->lazy-seq
+  "Returns a blocking lazy sequence of items taken from the provided channel."
+  [channel]
+  (lazy-seq
+   (when-let [v (<!! channel)]
```
ncalexan: `<!!` is not supported under CLJS.  (If it was, we wouldn't be doing any of this!  We'd just hide our channel use inside `(<!! (go ...))` blocks and have a better life.

src/datomish/exec.cljc
```diff
@@ -0,0 +1,67 @@
...
+     (cons v (seq!! channel)))))
+
+(defn run-to-seq
+  "Given an open database, returns a lazy sequence of results.
+   When fully consumed, underlying resources will be released."
+  [db find]
+  (channel->lazy-seq (<run db find)))
+
+(defn <open-and-run-to-seq-promise
```
ncalexan: Let's break some of these things into a `repl` module, since they're very unlikely to be the final API, but are super useful in hacking and testing.

src/datomish/sqlite.cljc
```diff
@@ -50,6 +50,15 @@
           [nil err]
           [@acc nil])))))
 
+(defn <all-rows
```
ncalexan: You can do this with
```clojure
(let [pair-chan (chan)]
  (reduce-rows db rest pair-chan
    (fn [result input] (put! pair-chan result) result)))
```
can you not?

src/datomish/query.cljc
```diff
@@ -72,15 +72,15 @@
                 (print-str pattern))))
 
   (let [table :datoms
-        alias (gensym table)
+        alias (gensym (name table))
```
ncalexan: Can you fold these tweaks into the previous commit?  It's confusing as is.

src/datomish/query.cljc
```diff
@@ -179,6 +173,9 @@
             nil
             (cons :and (:wheres context)))})
 
+(defn context->sql-string [context]
+  (-> context context->sql-clause (sql/format :quoting sql-quoting-style)))
```
ncalexan: Same: I find this visually hard to parse; prefer newlines between sexps.

ncalexan:
I expect you'll find a better expression of the last commit by adding `q`, in some form to `DB`.  I'm okay with whatever hacky approach you want for testing right now, though, since we're both fleshing out the API for testing, etc.

src/datomish/exec.cljc
rnewman: I think results will *ultimately* end up collected in memory; we won't hand a channel back to JS consumers, and we need a bounded lifetime. There are good reasons to work with streams within that:

- Capped memory use, modulo buffering. The current implementation should block the sqlite producer once the buffer is full, keeping the cursor open until exhausted; this channel is a lot like getting a cursor instance in a traditional sqlite library. IMO that's better than slurping the entire cursor into in-memory maps.
- Transducing a stream is a good model of what we're going to have to do to implement `pull` in a single query: roll up sequential cursor results from a `GROUP BY`, eventually emitting a single compound result. The row transducer is currently simple, but will eventually map M rows to N results in this way.
- It's a foundation of making limit queries work efficiently, when those queries depend on a join or a condition outside of SQL: the storage cursor needs to be ready to return lots of results, but we probably won't need them all.

src/datomish/sqlite.cljc
rnewman: I'm confused by your use of the name `pair-chan` there: `-each` calls its callback with just the row. The result of `reduce-rows` is a channel that results in a pair (I think?), but `pair-chan` in your example is a channel of rows.

Besides, `reduce-rows` uses an `atom`. Not only do we not need that — a channel is a queue; we don't need another mechanism for collection — but `swap!` can result in the same item being put into the channel multiple times. It's not safe to supply a side-effecting function as the argument to `swap!`.

rnewman:
This'll do for now: d695554

PR #10, opened 2016-07-14
Transact

ncalexan:
@rnewman this isn't ready to land, but you can at least see the approach coming together.

I'll start knocking up the schema support after lunch.

### DIFF
--- a/src/datomish/query.cljc
+++ b/src/datomish/query.cljc
@@ -52,10 +52,16 @@
   (or (-> context :bindings variable first)
       (raise (str "Couldn't find variable " variable))))
 
-(defn make-context []
-  (->Context [] {} [] []
-             transforms/attribute-transform-string
-             transforms/constant-transform-default))
+(defn make-context
+  ([]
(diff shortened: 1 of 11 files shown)


src/datomish/query.cljc
```diff
@@ -127,10 +133,10 @@
 (defn apply-elements-to-context [context elements]
   (assoc context :elements elements))
 
-(defn patterns->context
-  "Turn a sequence of patterns into a Context."
-  [patterns]
-  (reduce apply-pattern-to-context (make-context) patterns))
+(defn patterns->into-context
```
rnewman: These should be `apply-patterns-to-context`, I think. Certainly not both `->` ("transform") and `into`.

rnewman:
@ncalexan: I cherry-picked a bunch of these commits onto master so that I can safely work on `query.cljc`.

rnewman:
I ported all of this work on top of #16. Tests pass.

ncalexan:
This is happening over in #18.

PR #17, opened 2016-07-26
Implement simple 'or' clauses.

rnewman:
This is sufficient to handle the parameter-forking kind of `or` that's prevalent in Datomic's docs.

### DIFF
--- a/src/datomish/query/clauses.cljc
+++ b/src/datomish/query/clauses.cljc
@@ -13,14 +13,14 @@
      [datascript.parser :as dp
       #?@(:cljs
             [:refer
-             [PlainSymbol Predicate Not Pattern DefaultSrc Variable Constant Placeholder]])]
+             [PlainSymbol Predicate Not Or Pattern DefaultSrc Variable Constant Placeholder]])]
      [honeysql.core :as sql]
      [clojure.string :as str]
      )
   #?(:clj
        (:import
           [datascript.parser
-           PlainSymbol Predicate Not Pattern DefaultSrc Variable Constant Placeholder])))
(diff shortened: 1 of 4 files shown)


rnewman:
Paging @ncalexan.

src/datomish/query/clauses.cljc
```diff
@@ -62,6 +62,12 @@
                           (attribute-in-source (:source cc) value)
                           (constant-in-source (:source cc) value))]))
 
+(defn merge-ccs [left right]
+  (assoc left
+         :from (concat (:from left) (:from right))
```
ncalexan: I somewhat expect these to be ordered sets rather than lists.  Just be certain (to yourself) that `concat` is appropriate here.

src/datomish/query/clauses.cljc
```diff
@@ -268,3 +316,76 @@
...
+
+(defn simple-or->cc
+  "The returned CC has not yet had bindings expanded."
+  [source external-bindings orc]
+  (validate-or-clause orc)
+
+  ;; We 'fork' a CC for each pattern, then union them together.
+  ;; We need to build the first in order that the others use the same
+  ;; column names.
+  (let [cc (->ConjoiningClauses source [] (or external-bindings {}) {} [])
```
ncalexan: Prefer `map->ConjoiningClauses`; already this is hard to read :(

ncalexan:
Nifty!  Rock on.

src/datomish/query/clauses.cljc
rnewman: Each one is a unique table alias, so we just use a sequence. They're pairs, `[[:datoms datoms123], [:datoms datoms124]]`, etc.

rnewman:
8a77dcd

PR #18, opened 2016-07-27
Transact rewrite

ncalexan:
@rnewman okay, first commit here is ready for first review.  Still lots to do, but it's a start.  I'm happy to land something like this as a first step, mostly so you can use the bits here to generate test data; or I'm happy to build it up further before landing.

Some of the stuff I'll be working on next, some of which are noted with TODOs:

- [ ] Implement parts, or revert to DataScript-like negative integers.
- [ ] Persist idents, parts, entity id counters, transaction indices in SQLite.
- [ ] Implement `:db/value` type checks throughout.
- [ ] Implement lookup refs for real.
- [ ] Finish implementing map properties (reverse refs, sequential values, etc).
- [ ] Set `:db/txInstant` correctly; ensure tx timestamps are reasonable.
- [ ] Work out the story around using `#dm/id[:db.part/tx]` to refer to the current transaction, or revert to DataScript-like ":db/currentTx".


rnewman:
One thing I wanted to note, before even getting to the PR: the clock for `txInstant` and otherwise should be an input to the system. Needed for testing/mocking/smart drift handling/etc.

ncalexan:
> One thing I wanted to note, before even getting to the PR: the clock for txInstant and otherwise should be an input to the system. Needed for testing/mocking/smart drift handling/etc.

I concur.  One way this could be handled is to allow setting `:db/txInstant` directly; alternately, we supply a clock (which I've stubbed in for now).

src/datomish/datom.cljc
```diff
@@ -0,0 +1,136 @@
...
+       (assoc [d k v] (assoc-datom d k v))
+       ]))
+
+(defn ^Datom datom
+  ([e a v tx]       (Datom. e a v tx true))
+  ([e a v tx added] (Datom. e a v tx added)))
+
+(defn datom? [x] (instance? Datom x))
+
+(defn- hash-datom [^Datom d]
```
rnewman: Can you explain why you haven't included `tx` and `added` in the hash?

src/datomish/db.cljc
```diff
@@ -12,31 +12,312 @@
...
+    [db]
+    "TODO: document this interface."))
+
+;; TODO: handle _?
+(defn search->sql-clause [pattern]
+  (merge
+    {:select [:*] ;; e :a :v :tx] ;; TODO: generalize columns.
+     :from [:datoms]}
+    (if-not (empty? pattern)
+      {:where (cons :and (map #(vector := %1 (if (keyword? %2) (str %2) %2)) [:e :a :v :tx] pattern))} ;; TODO: use schema to process v.
```
rnewman: Doesn't that `keyword?` assume a lot? Might want some validation in here…

src/datomish/db.cljc
```diff
@@ -12,31 +12,312 @@
...
+
+  ;; TODO: use q for searching?  Have q use this for searching for a single pattern?
+  (<eavt [db pattern]
+    (go-pair
+      ;; TODO: find a better expression of this pattern.
+      (let [rows (<? (->>
+                       (search->sql-clause pattern)
+                       (sql/format)
+                       (s/all-rows (:sqlite-connection db))))]
+        (mapv #(Datom. (:e %) (:a %) (:v %) (:tx %) true) rows)))) ;; TODO: map values according to schema.
```
rnewman: Ugh. Really need a positional SQL interface.

src/datomish/db.cljc
```diff
@@ -12,31 +12,312 @@
...
+            rows (<? (->>
+                       {:select [:*] :from [:datoms] :where [:and [:= :a a] [:= :v v]]}
+                       (sql/format)
+                       (s/all-rows (:sqlite-connection db))))]
+        (mapv #(Datom. (:e %) (:a %) (:v %) (:tx %) true) rows)))) ;; TODO: map values according to schema.
+
+  (<apply-datoms [db datoms]
+    (go-pair
+      (let [exec (partial s/execute! (:sqlite-connection db))]
+        ;; TODO: batch insert, batch delete.
```
rnewman: Also SQL transactions, no? Otherwise the KB transaction log can get out of sync with the datoms table.

rnewman:
Wow, there's a lot of code here. Lots of TODOs, too. Happy for this to land in-tree and iterate there. Nice.

rnewman:
(Also that was a super-skim review 'cos I'm heading afk. More thoroughness next week.)

src/datomish/datom.cljc
ncalexan: Copy-pasta from DataScript.  I think this will be because in the DS store, Datoms compare equal without regard to `tx` (and `added` is implicitly `true`).  I'll just cull pieces of this that aren't used.

src/datomish/db.cljc
ncalexan: Oh yes, yes it does.  9c8b116 and 5c564bc make this sensible.

src/datomish/db.cljc
ncalexan: The entire thing runs in a SQL transaction, managed in `<transact!`.  I'll make that more clear in some technical notes.

rnewman: That implies that this should take a connection (or transaction) arg instead/as well — transactions are scoped to connections, so one should make that explicit. Better to address this kind of thing now rather than find some surprising bug where two transacts are running at the same time with one DB transaction, or one transact somehow spans two DB connections by virtue of running in a worker thread pool. (Some of that would require running in a JVM, or with a socket-based module in node, granted.)

rnewman: Unless you're implying that DB ~= connection, of course.

ncalexan: I'm somewhere in the middle, which is an awkward place to be :(

At the moment, `DB` has a direct handle to the SQLite connection, and doesn't think about transactions.  `<transact!` is two-phase; it first does a bunch of reads (via DB), and then bulk applies updates-and-delete.  I think `<apply-datoms` might move to `IConnection`, but I don't know the final form for "stale" DBs, where the underlying store has moved on.

src/datomish/db.cljc
```diff
@@ -77,7 +77,7 @@
   (let [e (:e row)
         a (:a row)
         v (:v row)]
-    (Datom. e a (ds/<-SQLite schema a v) (:tx row) (:added row))))
+    (Datom. e a (ds/<-SQLite schema a v) (:tx row) (boolean (:added row)))))
```
rnewman: Not sure this does what you want:

```
user=> (boolean 1)
true
user=> (boolean 0)
true
```

src/datomish/db.cljc
```diff
@@ -130,22 +148,31 @@
...
+                      ["INSERT INTO datoms VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)" e a v tx
+                       (ds/indexing? schema a) ;; index_avet
+                       (ds/ref? schema a) ;; index_vaet
+                       fulltext? ;; index_fulltext
+                       (ds/unique-value? schema a) ;; unique_value
+                       (ds/unique-identity? schema a) ;; unique_identity
+                       ])))
+              (if fulltext?
+                (<? (exec
+                      ;; TODO: in the future, purge fulltext values from the fulltext_datoms table.
```
rnewman: You could have a foreign key relation to the transaction log with `ON DELETE CASCADE`, I suppose.

This reminds me to file an issue…

src/datomish/d.clj
```diff
@@ -0,0 +1,196 @@
+(ns datomish.d
```
rnewman: License block, even for this hack code.

rnewman: Unless you're going to rebase this away, of course.

src/datomish/datom.cljc
```diff
@@ -0,0 +1,137 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+;; Purloined from DataScript.
```
rnewman: @ncalexan: are we allowed to use Eclipse Public License code? Does it affect whether we are MPL or Apache licensed?

src/datomish/db.cljc
```diff
@@ -12,31 +12,599 @@
...
+  ;; TODO: use q for searching?  Have q use this for searching for a single pattern?
+  (<eavt [db pattern]
+    (let [[e a v] pattern
+          v (and v (ds/->SQLite schema a v))] ;; We assume e and a are always given.
+      (go-pair
+        (->>
+          {:select [:e :a :v :tx [1 :added]] ;; TODO: generalize columns.
+           :from [:all_datoms]
+           :where (cons :and (map #(vector := %1 %2) [:e :a :v] (take-while (comp not nil?) [e a v])))} ;; Must drop nils.
+          (sql/format)
```
rnewman: `:quoting` (or just use `datomish.query` or `datomish.clauses` directly).

src/datomish/db.cljc
```diff
@@ -12,31 +12,599 @@
...
+
+          (s/all-rows (:sqlite-connection db))
+          (<?)
```
rnewman: `->>` accepts individual symbols, so you can just use `sql/format` and `<?` instead of `(sql/format)` etc.

src/datomish/db.cljc
```diff
@@ -12,31 +12,599 @@
...
+
+          (mapv (partial row->Datom (.-schema db))))))) ;; TODO: understand why (schema db) fails.
+
+  (<avet [db pattern]
+    (let [[a v] pattern
+          v (ds/->SQLite schema a v)]
+      (go-pair
+        (->>
+          {:select [:e :a :v :tx [1 :added]] ;; TODO: generalize columns.
```
rnewman: You can format this query (and the other fixed queries above) once, rather than once per execution. That'll save a lot of runtime.

src/datomish/db.cljc
```diff
@@ -12,31 +12,599 @@
...
+           :from [:all_datoms]
+           :where [:and [:= :a a] [:= :v v] [:= :index_avet 1]]}
```
rnewman: … because these will end up as bound variables anyway.

src/datomish/db.cljc
```diff
@@ -12,31 +12,599 @@
...
+      (let [exec (partial s/execute! (:sqlite-connection db))]
+        ;; TODO: batch insert, batch delete.
+        (doseq [datom datoms]
+          (let [[e a v tx added] datom
+                schema (.-schema db) ;; TODO: understand why (schema db) fails.
+                v (ds/->SQLite schema a v)
+                fulltext? (ds/fulltext? schema a)]
+            ;; Append to transaction log.
+            (<? (exec
+                  ["INSERT INTO transactions VALUES (?, ?, ?, ?, ?)" e a v tx (if added 1 0)]))
```
rnewman: You'll get better locality and less fragmentation if you insert everything into `transactions` then everything into `datoms`, rather than interleaving in this way. So I'd `mapcat` or `reduce` across `datoms` to yield collections of statements to execute.

src/datomish/db.cljc
```diff
@@ -12,31 +12,599 @@
...
+              (let [v (if fulltext?
+                        (<? (<insert-fulltext-value db v))
+                        v)]
+                (<? (exec
+                      ["INSERT INTO datoms VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)" e a v tx
+                       (ds/indexing? schema a) ;; index_avet
+                       (ds/ref? schema a) ;; index_vaet
+                       fulltext? ;; index_fulltext
+                       (ds/unique-value? schema a) ;; unique_value
+                       (ds/unique-identity? schema a) ;; unique_identity
```
rnewman: Is it worth memoizing any of these?

src/datomish/db.cljc
```diff
@@ -12,31 +12,599 @@
...
+
+(defn <current-tx [sqlite-connection]
+  "Find the largest tx written to the SQLite store.
+  Returns an integer, -1 if no transactions have been written yet."
+
   (go-pair
-    (when-not (= sqlite-schema/current-version (<? (sqlite-schema/<ensure-current-version sqlite-connection)))
-      (raise-str "Could not ensure current SQLite schema version."))
-    (->DB sqlite-connection)))
+    (let [rows (<? (s/all-rows sqlite-connection ["SELECT COALESCE(MAX(tx), -1) AS current_tx FROM transactions"]))]
```
rnewman: Perhaps make this a view or a counter table.

rnewman: Also make sure that `tx` is an indexed column!

src/datomish/db.cljc
```diff
@@ -12,31 +12,599 @@
...
+  {:db/ident {:db/cardinality :db.cardinality/one}}."
+
+  (let [<-SQLite (get-in ds/value-type-map [:db.type/keyword :<-SQLite])] ;; TODO: make this a protocol.
+    (go-pair
+      (->>
+        (->>
+          {:select [:ident :attr :value] :from [:schema]}
+          (sql/format)
+          (s/all-rows sqlite-connection))
+        (<?)
```
rnewman: Same comment as earlier.

src/datomish/db.cljc
```diff
@@ -12,31 +12,599 @@
...
+(defn <symbolic-schema [sqlite-connection]
+  "Read the schema map materialized view from the given SQLite store.
+  Returns a map (keyword ident) -> (map (keyword attribute -> keyword value)), like
+  {:db/ident {:db/cardinality :db.cardinality/one}}."
+
+  (let [<-SQLite (get-in ds/value-type-map [:db.type/keyword :<-SQLite])] ;; TODO: make this a protocol.
+    (go-pair
+      (->>
+        (->>
+          {:select [:ident :attr :value] :from [:schema]}
```
rnewman: `GROUP BY ident`?

src/datomish/db.cljc
```diff
@@ -12,31 +12,599 @@
...
+
+      ;; Explode map shorthand, such as {:db/id e :attr value :_reverse ref},
+      ;; to a list of vectors, like
+      ;; [[:db/add e :attr value] [:db/add ref :reverse e]].
+      (->> (explode-entities (schema db)))
+
+      (update :entities (partial map ensure-entity-form))
+
+      ;; Replace idents with entids where possible, using db* to capture :db/tx.
+      (update :entities (partial map (partial maybe-ident->entid db*)))
```
rnewman: You should compose these stages together to try to reduce garbage. That is: take the `:entities` list and walk it to produce exploded, entid-translated vectors. Each of these `update`+`map` calls is producing a new map and a new array.

(Yes, it'll be less readable.)

src/datomish/db.cljc
```diff
@@ -67,4 +635,391 @@
...
+
+(defn <resolve-lookup-refs [db report]
+  {:pre [(db? db) (report? report)]}
+
+  (go-pair
+    (->>
+      (vec (for [[op & entity] (:entities report)]
+             (into [op] (for [field entity]
+                          (if (lookup-ref? field)
+                            (first (<? (<eavt db field))) ;; TODO improve this -- this should be avet, shouldn't it?
```
rnewman: A lookup ref *must* have a fixed and unique `a`, and a fixed and unique `v`, so yes, `avet` is a better choice.

src/datomish/db.cljc
```diff
@@ -12,31 +12,599 @@
...
+                          (if (first (filter (partial tx-instant? db*) entities))
+                            entities
+                            (conj entities (tx-entity db report)))))
+
+      ;; Extract the current txInstant for the report.
+      (->> (update-txInstant db*)))))
+
+(defn- lookup-ref? [x]
+  (and (sequential? x)
+       (= (count x) 2)))
```
rnewman: And

```
(keyword-or-numeric-entid? (first x))
(not (var-or-placeholder? (second x)))
```

src/datomish/db.cljc
```diff
@@ -67,4 +635,391 @@
...
+                          (->
+                            report
+                            (assoc-in [:tempids tempid] upserted-eid)
+                            (assoc-in [:entities] es)))))
+
+(defn- transact-entity [report entity]
+  (update-in report [:entities] conj entity))
+
+(defn <resolve-id-literals
+  "Upsert uniquely identified literals when possible and allocate new entids for all other id literals.
```
rnewman: Note to self: I haven't looked at this yet.

src/datomish/db.cljc
```diff
@@ -67,4 +635,391 @@
...
+
+  [db report]
+  {:pre [(db? db) (report? report)]}
+
+  ;; TODO: consider accumulating errors to show more meaningful error reports.
+  ;; TODO: constrain entities; constrain attributes.
+
+  (go-pair
+    (doseq [[op e a v] (:entities report)]
+      (ds/ensure-valid-value (schema db) a v))
```
rnewman: Compute `(schema db)` once.

src/datomish/db.cljc
```diff
@@ -67,4 +635,391 @@
...
+  ;; TODO: constrain entities; constrain attributes.
+
+  (go-pair
+    ;; TODO: comment on applying datoms that violate uniqueness.
+    (let [unique-datoms (transient {})] ;; map (nil, a, v)|(e, a, nil)|(e, a, v) -> datom.
+      (doseq [[e a v tx added :as datom] (:tx-data report)]
+
+        (when added
+          ;; Check for violated :db/unique constraint between datom and existing store.
+          (when (ds/unique? (schema db) a)
```
rnewman: Same.

src/datomish/schema.cljc
```diff
@@ -0,0 +1,168 @@
...
+                     :attribute a
+                     :key k
+                     :value v}))))
+
+;; TODO: consider doing this with a protocol and extending the underlying Clojure(Script) types.
+(def value-type-map
+  {:db.type/ref     { :valid? #(and (integer? %) (pos? %)) :->SQLite identity :<-SQLite identity }
+   :db.type/keyword { :valid? keyword? :->SQLite str :<-SQLite #(keyword (subs % 1)) }
+   :db.type/string  { :valid? string? :->SQLite identity :<-SQLite identity }
+   :db.type/boolean { :valid? #?(:clj #(instance? Boolean %) :cljs #(= js/Boolean (type %))) :->SQLite #(if % 1 0) :<-SQLite #(if (= % 1) true false) }
```
rnewman:
```
(if (= % 1) true false)
```

is `(= % 1)`.

src/datomish/schema.edn
```diff
@@ -0,0 +1,59 @@
+[
+ ;; news
```
rnewman: Does this have to live in the main source dir?

src/datomish/sqlite_schema.cljc
```diff
@@ -18,21 +18,70 @@
 (def current-version 1)
 
 (def v1-statements
-  ["CREATE TABLE datoms (e INTEGER NOT NULL, a SMALLINT NOT NULL, v BLOB NOT NULL, tx INTEGER NOT NULL, index_avet TINYINT NOT NULL DEFAULT 0, index_vaet TINYINT NOT NULL DEFAULT 0)"
+  ["CREATE TABLE datoms (e INTEGER NOT NULL, a SMALLINT NOT NULL, v BLOB NOT NULL, tx INTEGER NOT NULL,
+                         index_avet TINYINT NOT NULL DEFAULT 0, index_vaet TINYINT NOT NULL DEFAULT 0,
+                         index_fulltext TINYINT NOT NULL DEFAULT 0,
+                         unique_value TINYINT NOT NULL DEFAULT 0, unique_identity TINYINT NOT NULL DEFAULT 0)"
    "CREATE INDEX eavt ON datoms (e, a)" ;; No v -- that's an opt-in index.
```
rnewman: Think about this a little. Doesn't an index includes all of the columns of the table? If so, is it worth including `v` as part of the index key, if the SQL query engine can use it for querying?

rnewman: That is: not including `v` here doesn't make the index any smaller (this isn't a partial index), it just makes the order of rows less constrained.

src/datomish/sqlite_schema.cljc
```diff
@@ -18,21 +18,70 @@
...
+   ;; prefix='2,3'
+   ;; By default we use Unicode-aware tokenizing (particularly for case folding), but preserve
+   ;; diacritics.
+   "CREATE VIRTUAL TABLE fulltext_values
+     USING FTS4 (text NOT NULL, tokenize=unicode61 \"remove_diacritics=0\")"
+
+   ;; A view transparently interpolating fulltext indexed values into the datom structure.
+   "CREATE VIEW fulltext_datoms AS
+     SELECT e, a, fulltext_values.text AS v, tx, index_avet, index_vaet, index_fulltext, unique_value, unique_identity
+       FROM datoms JOIN fulltext_values ON datoms.v = fulltext_values.rowid
```
rnewman: SQLite doesn't make `JOIN ON` more efficient than `WHERE AND`, but the latter gives us control over which condition applies first. Consider turning this into the `WHERE AND` formulation so we can experiment with different clause orders more easily.

src/datomish/sqlite_schema.cljc
```diff
@@ -18,21 +18,70 @@
...
+   ;; TODO: possibly remove this index.  :db.unique/value should be asserted by the transactor in
+   ;; all cases, but the index may speed up some of SQLite's query planning.  For now, it services
+   ;; to validate the transactor implementation.
+   "CREATE UNIQUE INDEX unique_value ON datoms (v) WHERE unique_value = 1"
+   ;; TODO: possibly remove this index.  :db.unique/identity should be asserted by the transactor in
+   ;; all cases, but the index may speed up some of SQLite's query planning.  For now, it serves to
+   ;; validate the transactor implementation.
+   "CREATE UNIQUE INDEX unique_identity ON datoms (a, v) WHERE unique_identity = 1"
+
+   "CREATE INDEX fulltext ON datoms (v) WHERE index_fulltext = 1"
```
rnewman: Consider simply rolling this into `vaet` — `WHERE index_vaet = 1 OR index_fulltext = 1`. It's the same index but with more control.

src/datomish/sqlite_schema.cljc
```diff
@@ -18,21 +18,70 @@
...
    "CREATE INDEX aevt ON datoms (a, e)" ;; No v -- that's an opt-in index.
-   "CREATE INDEX avet ON datoms (a, v, e) WHERE index_avet = 1" ;; Opt-in index: only if a has :db/index true.
-   "CREATE INDEX vaet ON datoms (v, a, e) WHERE index_vaet = 1" ;; Opt-in index: only if a has :db/valueType :db.type/ref
+   "CREATE UNIQUE INDEX avet ON datoms (a, v, e) WHERE index_avet = 1" ;; Opt-in index: only if a has :db/index true.
+   "CREATE UNIQUE INDEX vaet ON datoms (v, a, e) WHERE index_vaet = 1" ;; Opt-in index: only if a has :db/valueType :db.type/ref
+
+   ;; TODO: possibly remove this index.  :db.unique/value should be asserted by the transactor in
+   ;; all cases, but the index may speed up some of SQLite's query planning.  For now, it services
+   ;; to validate the transactor implementation.
+   "CREATE UNIQUE INDEX unique_value ON datoms (v) WHERE unique_value = 1"
```
rnewman: I have a gentle preference for naming indices `idx_datoms_unique_value` etc.

src/datomish/sqlite_schema.cljc
```diff
@@ -18,21 +18,70 @@
...
+   "CREATE VIEW fulltext_datoms AS
+     SELECT e, a, fulltext_values.text AS v, tx, index_avet, index_vaet, index_fulltext, unique_value, unique_identity
+       FROM datoms JOIN fulltext_values ON datoms.v = fulltext_values.rowid
+       WHERE datoms.index_fulltext = 1"
+
+   ;; A view transparently interpolating all entities (fulltext and non-fulltext) into the datom structure.
+   "CREATE VIEW all_datoms AS
+     SELECT e, a, v, tx, index_avet, index_vaet, index_fulltext, unique_value, unique_identity
+       FROM datoms
+       WHERE index_fulltext != 1
```
rnewman: I have a gentle preference for `IS NOT` instead of `!=`, if only to avoid arguments about `<>`!

src/datomish/db.cljc
ncalexan: You did file the issue (https://github.com/mozilla/datomish/issues/19) and I'd prefer to leave it as is for now.  Excision is a future issue: https://github.com/mozilla/datomish/issues/21.

src/datomish/d.clj
ncalexan: Yeah, this was never supposed to be here.  I've removed it.

src/datomish/datom.cljc
ncalexan: I don't know, and neither did @Mossop.  I'll email @tonsky about a license grant, or we can rewrite this.

src/datomish/db.cljc
ncalexan: I agree, and earlier versions did this with a single composed transducer.  We can move back to that style (here and later on, after the retry-with-tempids logic) eventually.  I'm viewing this as "getting it working" before "making it fast".

src/datomish/db.cljc
ncalexan: Yeah -- I should really just remove this, since it's not complete and not tested.  I might get to it before I go.

src/datomish/db.cljc
ncalexan: I've added the first check, but in general these look like `[:attr value]` and `value` is arbitrary.

src/datomish/db.cljc
ncalexan: Done, here and throughout.

src/datomish/db.cljc
ncalexan: Done.

src/datomish/schema.cljc
ncalexan: Oops!

src/datomish/sqlite_schema.cljc
ncalexan: Done.

src/datomish/sqlite_schema.cljc
ncalexan: Done.  I standardized on `IS 0` for false, `IS NOT 0` for true.

src/datomish/sqlite_schema.cljc
ncalexan: In the views I create below, there's an explicit condition `index_fulltext IS NOT 0`.  Is it your contention that SQLite will filter the rows where `index_vaet IS NOT 0` and `index_fulltext IS 0` so fast that it's not worth having the separate index?

src/datomish/sqlite_schema.cljc
ncalexan: Done.

src/datomish/sqlite_schema.cljc
ncalexan: I've done this.

src/datomish/db.cljc
ncalexan: We `group-by` just below, so I don't think there's value here.

src/datomish/db.cljc
ncalexan: This max transaction logic is not finalized, so I'll think more about this.  `tx` is indexed, so the select should be fast.

src/datomish/db.cljc
ncalexan: Probably worth memoizing all of them (at the schema level).  I need to extract `ident` and `entid` for real; this can fit into that refactoring.

src/datomish/db.cljc
ncalexan: OK.  I'm not going to try to squash these all at once, but I'll adopt this style and we can fix things as we see them.

src/datomish/db.cljc
rnewman: Oh yeah, you can't use variables here.

src/datomish/sqlite_schema.cljc
rnewman: The situation in which they will not be the same thing: `v` is a row ID that collides with an integer value for a non-fulltext but value-indexed predicate.

We might expect the predicate to be known, in which case it goes back to being equivalent.

And if not — as in the case of a plain ol' fulltext search yielding `rowid`s — yes, we can walk `vaet` and filter for fulltextness.

But given that `index_fulltext` and `index_vaet` (value indexing) are mutually exclusive, it might be fine to just make them separate index chunks. In that case, make this `ON datoms (v, a, e)`.

src/datomish/db.cljc
rnewman: Yeah, the idea is that it might make the `group-by` more efficient, because you get an ordered walk. Don't know if that'll happen in practice.

ncalexan:
I pushed this to https://github.com/ncalexan/datomish/tree/master; we'll keep working there.

ncalexan:
To summarize:

> Implement parts, or revert to DataScript-like negative integers.

This isn't done.

>     Persist idents, parts, entity id counters, transaction indices in SQLite.

Idents and transaction indices are done; entity IDs are not done (those are under the banner of parts above).

>     Implement :db/value type checks throughout.

Done!

>     Implement lookup refs for real.

Not done.  This is somewhat orthogonal to the rest of the work; it's a nice to have.

>     Finish implementing map properties (reverse refs, sequential values, etc).

Done!

>     Set :db/txInstant correctly; ensure tx timestamps are reasonable.

Done.

>     Work out the story around using #dm/id[:db.part/tx] to refer to the current transaction, or revert to DataScript-like ":db/currentTx".

I ended doing a DS-like `:db/tx` rather than copy Datomic's `:db.part/tx`, which seemed to me very inconsistent.

PR #22, opened 2016-08-04
Turn Source into a protocol. Allow source->from to switch on attribute.

rnewman:
Simple prep for handling `fulltext` attributes. @ncalexan, rubberstamp!


ncalexan:
This looks fine, although I'm having trouble understanding where to glue the schema in here.  Tomorrow...

rnewman:
Presumably the schema is accessible via the `db`, which we have at this point.

rnewman:
Landed in b1e6ffc.

PR #23, opened 2016-08-05
Implement rudimentary fulltext query expansion

rnewman:
This builds SQL for the simplest `fulltext` queries. Waddaya think, @ncalexan?

### DIFF
--- a/src/datomish/query/functions.cljc
+++ b/src/datomish/query/functions.cljc
@@ -1,0 +1,164 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+(ns datomish.query.functions
+  (:require
+     [honeysql.format :as fmt]
+     [datomish.query.cc :as cc]
+     [datomish.query.source :as source]
+     [datomish.util :as util #?(:cljs :refer-macros :clj :refer) [raise raise-str cond-let]]
+     [datascript.parser :as dp
+      #?@(:cljs
+            [:refer
+             [
+              BindColl
+              BindScalar
+              BindTuple
+              BindIgnore
+              Constant
+              Function
+              PlainSymbol
+              SrcVar
+              Variable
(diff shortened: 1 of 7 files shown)


src/datomish/query/source.cljc
```diff
@@ -35,13 +35,17 @@
 (defprotocol Source
   (source->from [source attribute]
     "Returns a pair, `[table alias]` for a pattern with the provided attribute.")
+  (source->non-fts-from [source])
```
ncalexan: Would you mind always using `fulltext` in names?  Might as well make it easy to grep.

src/datomish/query/source.cljc
```diff
@@ -72,6 +76,14 @@
             (:fts-view source))]
       [table ((:table-alias source) table)]))
 
+  (source->non-fts-from [source]
```
ncalexan: I was thinking, late last night, about special representations for certain data types.  Is it worth trying to generalize a little here?  For example, we might want to add value hashes to certain rows, in order to have fast equality comparisons for large values; or we might want to represent URLs interned (like fulltext, but perhaps stored in some other representation).

src/datomish/query/functions.cljc
```diff
@@ -0,0 +1,164 @@
...
+    (when-not (fulltext-attribute? (:source cc) (:value attr))
+      (raise-str "Attribute " (:value attr) " is not a fulltext-indexed attribute."))
+
+    (when-not (and (instance? BindColl (:binding function))
+                   (instance? BindTuple (:binding (:binding function)))
+                   (every? (fn [s]
+                             (or (instance? BindIgnore s)
+                                 (and (instance? BindScalar s)
+                                      (instance? Variable (:variable s)))))
+                           (:bindings (:binding (:binding function)))))
```
ncalexan: This is pretty hard to read :)

ncalexan:
@rnewman this is looking really good!  It would be nice to simplify some of the DS parser destructuring, but this is surprisingly easy for me to follow.

rnewman:
Landed in b1e6ffc with review comments addressed.

PR #55, opened 2016-09-26
Allow sets of attributes in fulltext expressions. Fixes #54.

### LINKED ISSUE #54: Allow to search set of fields using `fulltext`
Right now, `fulltext` supports a single field (`:for/example`) or the special symbol `:all` to search all fields.  This ticket tracks generalizing to allow a set of fields, like `#{:for/example :and/another}`.

rnewman:
Over to you, @ncalexan.

### DIFF
--- a/src/common/datomish/query/functions.cljc
+++ b/src/common/datomish/query/functions.cljc
@@ -97,12 +97,32 @@
         ;;   [(fulltext $ _ "foo") [[?x]]]
         ;;
         ;; so we instead have a placeholder attribute. Sigh.
-        attr-constant (or
-                        (and (instance? Constant attr)
-                             (not (= :any (:value attr)))
-                             (source/attribute-in-source (:source cc) (:value attr)))
-                        (and (instance? Variable attr)
-                             (cc/binding-for-symbol-or-throw cc (:symbol attr))))
+        ;; We also support sets of attributes, so you can write
+        ;;
+        ;;   [(fulltext $ #{:foo/bar :foo/baz} "Noo") [[?x]]]
+        ;;
+        ;; which involves some tomfoolery here.
+        attr-constants (or
+                         (and (instance? Constant attr)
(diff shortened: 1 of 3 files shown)


src/common/datomish/query/functions.cljc
```diff
@@ -97,12 +97,32 @@
...
+        ;;   [(fulltext $ #{:foo/bar :foo/baz} "Noo") [[?x]]]
+        ;;
+        ;; which involves some tomfoolery here.
+        attr-constants (or
+                         (and (instance? Constant attr)
+                              (let [attr (:value attr)
+                                    intern (partial source/attribute-in-source (:source cc))]
+                                (when-not (= :any attr)
+                                  (cond
+                                    (set? attr)
```
ncalexan: `sequential?`  Not sure how string we want to be here.

test/datomish/tofinoish_test.cljc
```diff
@@ -460,3 +466,24 @@
         (is (= apricot-url "http://example.com/apricots/1"))
         (is (= apricot-title "A page about apricots."))
         (is (= apricot-excerpt "")))))))
+
+(deftest-db test-fulltext-set-attribute conn
+  (<? (d/<transact! conn tofino-schema))
+  (<? (d/<transact! conn
+                    [{:db/id             999
```
ncalexan: Do we hit the case where `:a` is `:db/fulltext true`, the value would match, but it's excluded from the search?  I don't think we do here.

src/common/datomish/query/functions.cljc
rnewman: I kinda like that sets enforce no ordering and no duplicates…

test/datomish/tofinoish_test.cljc
rnewman: Added.

rnewman:
Landed in 1296b80.

PR #60, opened 2016-09-28
Support complex `or` expressions.

### LINKED ISSUE #57: Support non-trivial `or` expressions
This will involve generating a `UNION` of two or more conjoining clauses with the same projection.

rnewman:
Fixes #57.

This is _almost_ done — it passes a test, but that test has the `or` at the top level, and thus doesn't exercise one spot in which I left a `TODO`. More tomorrow, but feel free to skim already, @ncalexan.

### DIFF
--- a/src/common/datomish/query/clauses.cljc
+++ b/src/common/datomish/query/clauses.cljc
@@ -6,6 +6,7 @@
   (:require
    [datomish.query.cc :as cc]
    [datomish.query.functions :as functions]
+   [datomish.query.projection :refer [sql-projection-for-simple-variable-list]]
    [datomish.query.source
     :refer [pattern->schema-value-type
             attribute-in-source
@@ -18,6 +19,7 @@
     #?@(:cljs
         [:refer
          [
+          And
           Constant
           DefaultSrc
           Function
@@ -35,6 +37,7 @@
   #?(:clj
      (:import
       [datascript.parser
+       And
        Constant
        DefaultSrc
        Function
@@ -50,6 +53,8 @@
 ;; Pattern building is recursive, so we need forward declarations.
 (declare
   Not->NotJoinClause not-join->where-fragment
+  expand-pattern-clauses
+  complex-or->cc
   simple-or? simple-or->cc)
 
 (defn- check-or-apply-value-type [cc value-type pattern-part]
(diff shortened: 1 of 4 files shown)


src/common/datomish/query/clauses.cljc
```diff
@@ -245,27 +245,37 @@
...
-         :extracted-types {}
-         :external-bindings (or external-bindings {})
-         :bindings {}
-         :ctes {}
-         :wheres []})
+      (make-cc source known-types external-bindings)
       patterns)))
 
 (defn cc->partial-subquery
   "Build part of a honeysql query map from a CC: the `:from` and `:where` parts.
```
ncalexan: Update docstring to reflect `select`.

src/common/datomish/query/clauses.cljc
```diff
@@ -6,6 +6,7 @@
   (:require
    [datomish.query.cc :as cc]
    [datomish.query.functions :as functions]
+   [datomish.query.projection :refer [sql-projection-for-simple-variable-list]]
```
ncalexan: Consider folding the previous commit into this one; it's factored but hard to understand without a use (or test).

src/common/datomish/query/clauses.cljc
```diff
@@ -204,17 +209,16 @@
   ;; Otherwise -- perhaps each leg of the `or` binds different variables, which
```
ncalexan: Might as well add comment about complex `or` here.  It's not clear to me that it's worth separating simple from complex; the former must be a subset of the latter.  But that's for you to decide.

src/common/datomish/query/clauses.cljc
```diff
@@ -391,3 +398,90 @@
...
+          ;; Fourth: map this query to an alias in `:from`, and establish bindings
+          ;; so that the enclosing query and projection know which names to use.
+          ;; Finally, return a CC that can be merged.
+          alias ((:table-alias source) :orjoin)
+          bindings (into {} (map (fn [var]
+                                   (let [sym (:symbol var)]
+                                     [sym [(sql/qualify alias (util/var->sql-var sym))]]))
+                                 free-vars))]
+
+      (cc/map->ConjoiningClauses
```
ncalexan: Consider extracting another `make-cc`-like function.

rnewman:
Comments addressed, folded and landed: 6ab9320

PR #66, opened 2016-10-05
Make tempids work from JS

rnewman:


src/common/datomish/js.cljs
```diff
@@ -73,10 +73,20 @@
         ;; We pickle the connection as a thunk here so it roundtrips through JS
         ;; without incident.
         {:conn (fn [] c)
-         :roundtrip (fn [x] (clj->js (cljify x)))
          :db (fn [] (transact/db c))
```
ncalexan: I feel like these should be `d/db`, to use the public API as much as possible.

src/common/datomish/js.cljs
```diff
@@ -73,10 +73,20 @@
...
+         :path path
+
+         ;; Primary API.
          :ensureSchema (fn [simple-schema] (ensure-schema c simple-schema))
          :transact (fn [tx-data] (transact c tx-data))
+         :q (fn [find opts] (q (transact/db c) find opts))
          :close (fn [] (db/close-db db))
+
+         ;; Some helpers for testing the bridge.
```
ncalexan: Is there no better way to expose these?

src/node/datomish/cljify.cljs
```diff
@@ -1,7 +1,27 @@
...
 (defn cljify
-  "In node, equivalent to `(js->clj o :keywordize-keys true).
-  See <http://dev.clojure.org/jira/browse/CLJS-439?focusedCommentId=43909>."
-  [o]
-  (js->clj o :keywordize-keys true))
+  "In node, equivalent to `(js->clj o :keywordize-keys true),
+  but successfully passes Clojure Records through. This allows JS API
+  callers to round-trip values they receive from ClojureScript APIs."
+  [x]
+  (if (record? x)
```
ncalexan: It really looks like this is identical to https://crossclj.info/fun/cljs.core.cljs/js-%3Eclj.html, save for the `record?` check.  Can you comment on exactly the difference, and explain that the repetition is needed to get things to work recursively?

rnewman: It's very similar, with the differences being no options and the ability to hook our record check in for recursing down into objects. I'll add a comment.

src/common/datomish/js.cljs
rnewman: It's simply convenient to do so. When we're happy with how this works (and there's no need to document them), we can delete them.

PR #71, opened 2016-10-07
Implement all four find specs. Fixes #38.

### LINKED ISSUE #38: Additional find specifications
We currently support relational find specifications:

```
:find ?foo ?bar
```

There are three others:

- Collection: `:find [foo ...]`
- Single tuple: `:find [?foo ?bar]`
- Single value: `:find ?title .`

These should be relatively straightforward to implement.

rnewman:


src/common/datomish/db.cljc
```diff
@@ -832,9 +809,39 @@
...
+         ;; We turn each row into either an array of values or an unadorned
+         ;; value. The row-pair-transducer does this work.
+         ;; The only thing to do to handle the full suite of find specs
+         ;; is to decide if we're then returning an array of transduced rows
+         ;; or just the first result.
+         row-pair-transducer (projection/row-pair-transducer context)
+         sql (query/context->sql-string context inputs)
+
+         first-only (context/scalar-or-tuple-query? context)
+         chan (chan (if first-only 1 50) row-pair-transducer)]
```
ncalexan: Extract the constant 50.

src/common/datomish/db.cljc
```diff
@@ -832,9 +809,39 @@
...
+
+         ;; We turn each row into either an array of values or an unadorned
+         ;; value. The row-pair-transducer does this work.
+         ;; The only thing to do to handle the full suite of find specs
+         ;; is to decide if we're then returning an array of transduced rows
+         ;; or just the first result.
+         row-pair-transducer (projection/row-pair-transducer context)
+         sql (query/context->sql-string context inputs)
+
+         first-only (context/scalar-or-tuple-query? context)
```
ncalexan: Are we clever enough to `LIMIT 1` yet?  (I will try to answer this question as I review.)

src/common/datomish/db.cljc
```diff
@@ -832,9 +809,39 @@
...
+         chan (chan (if first-only 1 50) row-pair-transducer)]
+
+     ;; Fill the channel.
+     (s/<?all-rows (.-sqlite-connection db) sql chan)
+
+     ;; If we only want the first result, great!
```
ncalexan: `<?all-rows` puts `[result err]` pairs on to `chan`, so it would "look like" `([first-result nil])` in the `first-only` case, right?  That seems correct.

src/common/datomish/query.cljc
```diff
@@ -63,10 +65,15 @@
   (let [inner-projection (projection/sql-projection-for-relation context)
         inner
         (merge
-          ;; Always SELECT DISTINCT, because Datalog is set-based.
+          ;; Always SELECT DISTINCT, because Datalog is set-based… unless
```
ncalexan: Consider putting `Because ..., we always SELECT DISTINCT -- unless...` just to keep the always/unless together.

src/common/datomish/query.cljc
```diff
@@ -142,15 +162,37 @@
...
-             :has-aggregates? (not (nil? group-by-vars))
-             :cc (clauses/patterns->cc (:default-source context) where known-types external-bindings)))))
+      (util/assoc-if
+        (assoc context
+               :find-spec find
+               :elements elements
+               :group-by-vars group-by-vars
+               :has-aggregates? (not (nil? group-by-vars))
+               :cc (clauses/patterns->cc (:default-source context) where known-types external-bindings))
+        :limit (find-spec->limit find)))))
```
ncalexan: Will this "win" if somebody supplies `:limit` and `:order-by`?  Should we bar `:limit` for certain find specs?

src/common/datomish/query/context.cljc
```diff
@@ -4,18 +4,30 @@
...
    ])
 
+(defn scalar-or-tuple-query? [context]
+  (when-let [find-spec (:find-spec context)]
+    (or (instance? FindScalar find-spec)
+        (instance? FindTuple find-spec))))
+
 (defn make-context [source]
-  (->Context source nil false nil nil nil nil))
+  (->Context source nil nil false nil nil nil nil))
```
ncalexan: Prefer `map->Context`.  Future, if you choose...

src/common/datomish/query/projection.cljc
```diff
@@ -251,18 +269,31 @@
...
-      (fn [[row err]]
-        (if err
-          [row err]
-          [(map (fn [projector] (projector row)) projectors) nil])))))
+      (if single-column-find-spec?
+        ;; We're only grabbing one result from each row.
+        (let [projector (first projectors)]
+          (fn [[row err]]
+            (if err
+              [row err]
```
ncalexan: Consider `[nil err]` for clarity.

src/common/datomish/query/projection.cljc
```diff
@@ -251,18 +269,31 @@
...
+            (instance? FindColl find-spec))]
 
     (map
-      (fn [[row err]]
-        (if err
-          [row err]
-          [(map (fn [projector] (projector row)) projectors) nil])))))
+      (if single-column-find-spec?
+        ;; We're only grabbing one result from each row.
+        (let [projector (first projectors)]
```
ncalexan: I'm missing something here.  How could you have multiple projectors for when `single-column-find-spec?` is true?  Can you make this stronger, asserting that there's only one?

test/datomish/test/query.cljc
```diff
@@ -684,3 +684,95 @@
...
+                   conn)))
+
+    ;; Tuple. We expect only one result, and indeed we only take one.
+    ;; No need for :distinct in this case!
+    (is (= {:select (list [:fulltext_datoms0.v :title])
+            :modifiers []
+            :limit 1
+            :from (list [:fulltext_datoms 'fulltext_datoms0])
+            :where (list :and [:= :fulltext_datoms0.a (:save/title attrs)])}
+           (expand [:find '[?title]
```
ncalexan: Can I get a test tuple with multiple columns, please?  This is a case that I thought through with the "single column find spec" above (which doesn't apply to that case).

test/datomish/test/query.cljc
```diff
@@ -684,3 +684,95 @@
                                    "something")
                              '[[?save]]]]}
                   conn)))))
+
+(deftest-db test-find-specs-expansion conn
+  (let [attrs (<? (<initialize-with-schema conn save-schema))]
```
ncalexan: Consider using `are` for these tests.  It's quite nice :)

test/datomish/test/query.cljc
```diff
@@ -684,3 +684,95 @@
...
+(deftest-db test-find-specs-result-shape conn
+  (let [attrs (<? (<initialize-with-schema conn save-schema))]
+    ;; Add some data.
+    (<? (d/<transact! conn
+                      [{:db/id (d/id-literal :db.part/user -1)
+                        :save/title "Some page title"}
+                       {:db/id (d/id-literal :db.part/user -2)
+                        :save/title "A different page"}]))
+
+    ;; Relation.
```
ncalexan: Why not just make the expected result a set?

test/datomish/test/query.cljc
```diff
@@ -684,3 +684,95 @@
...
+    ;; No need for :distinct in this case!
+    (let [result (<? (d/<q (d/db conn)
+                           [:find '[?title]
+                            :in '$
+                            :where '[?save :save/title ?title]]))]
+      (is (or (= ["Some page title"] result)
+              (= ["A different page"] result))))
+
+    ;; Scalar. As with the tuple form, we expect only one result.
+    (let [result (<? (d/<q (d/db conn)
```
ncalexan: Consider forcing the `:order-by` to get the two cases as well.

test/datomish/test/query.cljc
```diff
@@ -735,6 +735,34 @@
   [a b]
   (= (set a) (set b)))
 
+(deftest-db test-find-specs-empty-results conn
```
ncalexan: Kinda think we need a `deftest-schema` that opens the conn and transacts the initial schema.  Sometime later.

src/common/datomish/db.cljc
rnewman: Yes: when we're fetching a scalar or tuple, we switch from `DISTINCT` to `LIMIT 1`.

src/common/datomish/query.cljc
rnewman: If we decide to use `LIMIT 1` because of the find spec, then yes, this will win — `find-into-context` happens after `options-into-context` in `<?q`.

Probably no point being needlessly whips-and-chains about specifying a limit being wrong just because we decide we can use a lower one.

src/common/datomish/query/projection.cljc
rnewman: We should never have more than one, because the elements list should only contain one item in this case.

src/common/datomish/query/projection.cljc
rnewman: Done.

src/common/datomish/query/projection.cljc
rnewman: Assertion added:

```
          (when (second projectors)
            (raise "Single-column find spec used, but multiple projectors present."
                   {:elements elements
                    :projectors projectors
                    :find-spec find-spec}))
```

src/common/datomish/query.cljc
rnewman: Reworded entirely.

src/common/datomish/db.cljc
rnewman: Yes, exactly.

test/datomish/test/query.cljc
rnewman: If you mean changing our return value into a set: Datomic doesn't do it, and it's typically needless expense. If you mean something else, could you clarify?

rnewman:
Landed: e89544b

test/datomish/test/query.cljc
ncalexan: I meant, why not make your expected test result a set?  Sets compare without order to lists, IIRC.  So `(is (= (#{["Some page title"]["A different page"]} ...)` or similar.

rnewman: Nah mate:

```
user=> (= #{1 2 3} [3 2 1])
false
```

rnewman: You could do `(empty? (clojure.set/difference #{1 2 3} [3 2 1]))`, but that's a little obscure. Anyway, I fixed it by making the query ordered.

PR #81, opened 2016-10-11
Transact critical section

ncalexan:
Over to you, @rnewman.


src/common/datomish/util/deque.clj
```diff
@@ -0,0 +1,24 @@
+;; This Source Code Form is subject to the terms of the Mozilla Public
+;; License, v. 2.0. If a copy of the MPL was not distributed with this
+;; file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+(ns datomish.util.deque)
+
+(defn deque
+  "Return a new, empty deque."
+  []
+  (ref []))
```
rnewman: I think you want an `atom` here. There's no reason to use a `ref` unless you need to transactionally coordinate multiple `ref`s.

rnewman: … which makes this the same as the `cljs` implementation, so it can die?

rnewman: Oh, I see what's going on here. You're trying to atomically `pop`.

src/common/datomish/util/deque.cljs
```diff
@@ -0,0 +1,24 @@
...
+  [deque value]
+  (swap! deque conj value))
+
+(defn dequeue!
+  "Pop a value from the head of the deque.  Returns nil when deque is empty, like first."
+  [deque]
+  ;; There's only a single thread of control in a JS environment, so there's no possibility that
+  ;; this is interrupted and therefore no race between deref (@) and swap!.
+  (when-first [head @deque]
+    (swap! deque subvec 1)
```
rnewman: Nah, use `compare-and-set!` to build your own logic:


```
(defn dequeue! [deque]
  (loop []
    (let [vals @deque
          f (first vals)]
      (when-not (empty? vals)
        (if (compare-and-set! deque vals (pop vals))
          f
          (recur))))))
```

src/common/datomish/util/deque.clj
rnewman: See `compare-and-set!` note below.

ncalexan:
This has landed: https://github.com/mozilla/datomish/commit/ed545d4a11312e6e3a243d127c49b6c01855a8fb.

Issue #80, opened 2016-10-11
Make `<transact!` run in a critical section


ncalexan:
This is https://github.com/mozilla/tofino/issues/1321.  What's happening is that we're not "throttling" `<transact!` to be in a critical section.

rnewman:
This is the fanciest mutex I ever did see.

ncalexan:
I know.  Happy to see a less fancy one!  If we had an unbounded core.async
queue, we'd be golden.  The complexity is all about waking in order on the
JS side of the house.

ncalexan:
https://github.com/mozilla/datomish/commit/ed545d4a11312e6e3a243d127c49b6c01855a8fb

Issue #61, opened 2016-09-28
Support connection listeners


ncalexan:
Datomic and DataScript show a simple API for listening to transactions.  We'll want something similar, in particular to allow the UAS to listen for changes and propagate diffs out to clients.

rnewman:
jkerim was also mentioning how AS is (or wants to be) structured like a modern app service — lots of worker tasks pushed off to be run in the background in response to streams of data. The closer we can get to that model, the more composable this system will be.

rnewman:
@fluffyemily is this done?

fluffyemily:
@rnewman I'd say so.

PR #86, opened 2016-10-12
Critical sections and listeners, updated

ncalexan:
@rnewman -- this is an updated version of #80 and #61.  Listeners depends heavily on the exact form of transact, so here we are.

I ended up going with an unlimited buffer rather than a deque.  It's more likely we'll want an unlimited buffer in the future, tbh.  I removed the combinator (rather than fixing it) because we want a per-connection transaction critical section anyway, which was awkward to arrange; and so that we can explicitly close the transaction channel when we close the connection.  I also explicitly test the scenario where a listener calls `<transact!`, since that's so high value.  (Thinking aloud: it's so easy to make a mistake that we might want to try to catch recursive listeners.  Hmm...)

I simplified the listener interface, too, removing the explicit key and using the `mult`'s internal map instead.  Removes an atom and possible race conditions.

See what you think -- sorry it's a new PR.

### DIFF
--- a/src/common/datomish/transact.cljc
+++ b/src/common/datomish/transact.cljc
@@ -586,7 +586,7 @@
     (go
       (>! token-chan (gensym "transactor-token"))
       (loop []
-        (let [token (<! token-chan)]
+        (when-let [token (<! token-chan)]
           (when-let [[tx-data result close?] (<! (:transact-chan conn))]
             (let [pair
                   (<! (go-pair ;; Catch exceptions, return the pair.
@@ -594,10 +594,14 @@
                               report (<? (db/in-transaction!
                                            db
                                            #(-> (<with db tx-data))))]
-                          ;; We only get here if the transaction is committed.
-                          (reset! (:current-db conn) (:db-after report))
-                          (>! (:listener-source conn) report)
+                          (when report
+                            ;; <with returns non-nil or throws, but we still check report just in
+                            ;; case.  Here, in-transaction! function completed and returned non-nil,
+                            ;; so the transaction has committed.
+                            (reset! (:current-db conn) (:db-after report))
+                            (>! (:listener-source conn) report))
                           report)))]
+              ;; Even when report is nil (transaction not committed), pair is non-nil.
               (>! result pair))
             (>! token-chan token)
             (when close?
--- a/src/common/datomish/transact.cljc
+++ b/src/common/datomish/transact.cljc
@@ -6,7 +6,7 @@
   #?(:cljs
      (:require-macros
       [datomish.pair-chan :refer [go-pair <?]]
-      [cljs.core.async.macros :refer [go]]))
+      [cljs.core.async.macros :refer [go go-loop]]))
   (:require
    [datomish.query.context :as context]
    [datomish.query.projection :as projection]
@@ -25,7 +25,7 @@
    [taoensso.tufte :as tufte
     #?(:cljs :refer-macros :clj :refer) [defnp p profiled profile]]
    #?@(:clj [[datomish.pair-chan :refer [go-pair <?]]
-             [clojure.core.async :as a :refer [chan go <! >!]]])
+             [clojure.core.async :as a :refer [chan go go-loop <! >!]]])
    #?@(:cljs [[datomish.pair-chan]
               [cljs.core.async :as a :refer [chan <! >!]]]))
   #?(:clj
@@ -45,9 +45,11 @@
     [conn]
     "Get the full transaction history DB associated with this connection."))
 
-(defrecord Connection [current-db]
+(defrecord Connection [current-db transact-chan]
   IConnection
-  (close [conn] (db/close-db @(:current-db conn)))
+  (close [conn]
+    (a/close! (:transact-chan conn))
+    (db/close-db @(:current-db conn)))
 
   (db [conn] @(:current-db conn))
 
@@ -98,12 +100,15 @@
 ;; #?(:cljs
 ;;    (doseq [[tag cb] data-readers] (cljs.reader/register-tag-parser! tag cb)))
 
-;; TODO: implement support for DB parts?
+(declare start-transactor)
 
 (defn connection-with-db [db]
-  (map->Connection {:current-db (atom db)}))
-
-;; ;; TODO: persist max-tx and max-eid in SQLite.
+  (let [connection
+        (map->Connection {:current-db       (atom db)
+                          :transact-chan    (a/chan (util/unlimited-buffer))
+                          })]
+    (start-transactor connection)
+    connection))
 
 (defn maybe-datom->entity [entity]
   (cond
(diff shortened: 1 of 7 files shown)


rnewman:
I'll give this a solid block of time this evening; sorry for not getting to it in the morning!

ncalexan:
s'okay -- it was trickier than I expected to get right, and it's worth you
making sure you believe it's right :)

src/common/datomish/transact.cljc
```diff
@@ -552,12 +567,75 @@
...
+                        (let [db (db conn)
+                              report (<? (db/in-transaction!
+                                           db
+                                           #(-> (<with db tx-data))))]
+                          ;; We only get here if the transaction is committed.
+                          (reset! (:current-db conn) (:db-after report))
+                          (>! (:listener-source conn) report)
+                          report)))]
+              (>! result pair))
+            (a/close! result)
```
rnewman: Can this ever fail and cause the transactor loop to break?

src/common/datomish/transact.cljc
```diff
@@ -552,12 +567,75 @@
...
   [conn tx-data]
   {:pre [(conn? conn)]}
-  (let [db (db conn)] ;; TODO: be careful with swapping atoms.
-    (db/in-transaction!
-      db
-      #(go-pair
-         (let [report (<? (<with db tx-data))]
-           (reset! (:current-db conn) (:db-after report))
-           report)))))
+  (let [result (a/chan 1)]
```
rnewman: Oh, later I see you close it after use. Never mind. Think about it if you think it might be useful.

src/common/datomish/transact.cljc
```diff
@@ -45,9 +45,11 @@
     [conn]
     "Get the full transaction history DB associated with this connection."))
 
-(defrecord Connection [current-db]
+(defrecord Connection [current-db transact-chan]
   IConnection
-  (close [conn] (db/close-db @(:current-db conn)))
+  (close [conn]
+    (a/close! (:transact-chan conn))
+    (db/close-db @(:current-db conn)))
```
rnewman: The transact chan might still have items, and those will be processed… and will try to use the DB.

I think you should:

- Optionally: empty `transact-chan`, putting a "database closed before you ran, sorry" error into each task's result chan. If you don't, existing queued transactions will run before we close.
- Then, either:
  - Immediately close the DB, hoping that the current transact will throw and report interruption, or
  - Push a "close the DB" sentinel into `transact-chan` just before closing it. Closing the DB will then occur inside the transactor loop, between transacts. (When you close the DB, check again for tasks waiting behind, and tell them 'sorry', just in case one snuck in.)

rnewman: You probably want to break out of the transactor loop too, no?

src/common/datomish/transact.cljc
ncalexan: Closing a closed channel is a no-op.  I suppose we could do this in a `try` block, but let's see that be a problem in the wild first.

src/common/datomish/transact.cljc
```diff
@@ -552,12 +567,75 @@
...
+  {:pre [(conn? conn)]}
+  (when-not (util/unblocking-chan? listener-sink)
+    (raise "Listener sinks must be channels backed by unblocking buffers"
+           {:error :transact/bad-listener :listener-sink listener-sink}))
+  ;; Tapping an already registered sink is a no-op.
+  (a/tap (:listener-mult conn) listener-sink)
+  listener-sink)
+
+(defn- -listen-chan [f]
+  (let [c (a/chan (a/sliding-buffer 10))]
```
ncalexan: Sure.

src/common/datomish/util.cljc
```diff
@@ -101,3 +108,49 @@
...
+   (deftype UnlimitedBuffer [buf]
+     cljs.core.async.impl.protocols/UnblockingBuffer
+
+     cljs.core.async.impl.protocols/Buffer
+     (full? [this]
+       false)
+     (remove! [this]
+       (.pop buf))
+     (add!* [this itm]
+       (.unshift buf itm))
```
ncalexan: Yeah.  There are lots of ways to make this better; a persistent Clojure(Script) deque might even do better than index-munging.  For now, it's not important.

src/common/datomish/transact.cljc
ncalexan: Consumers who care about such things should use  `listen-chan!`.  I considered removing `listen!` altogether, since these choices and responsibilities are sublimated.  But it's going to be awkward for JS consumers to make `listen-chan!` work, so I'll at least expose the `n`.

src/common/datomish/transact.cljc
```diff
@@ -552,12 +580,99 @@
...
+         (raise "Connection is closed" {:error :transact/connection-closed})
+         (<? result))))))
+
+(defn- start-transactor [conn]
+  (let [token-chan (a/chan 1)]
+    (go
+      (>! token-chan (gensym "transactor-token"))
+      (loop []
+        (when-let [token (<! token-chan)]
+          (when-let [[sentinel tx-data result close?] (<! (:transact-chan conn))]
```
rnewman: Indeed, one might phrase this as:

```
(defn- put-and-close! [chan close? value]
  (>! chan value)
  (when close?
    (a/close! chan)))

(defn consume-as-closed [chan]
  (let [the-error (make-the-error)]
    (go-loop [[_ _ result close?] (<! chan)]
      (when result
        (put-and-close! result close? [nil the-error])
        (recur)))))

…
(loop []
  (when-let [token …]
    (if @(:closed? conn)
      ;; Either we closed correctly, and we should make sure our input
      ;; channel is empty, or someone screwed up.
      (if (closed? (:transact-chan conn))
        (consume-as-closed (:transact-chan conn))
        (raise "DB closed under our transact chan!"))

      ;; Otherwise, consume one item, return the token, and loop.
      (do
        (let [pair …]
          …
          ;; Don't bother checking for chan closed.
          (put-and-close! result close? pair))
        (>! token-chan token)
        (recur)))))
```

src/common/datomish/transact.cljc
```diff
@@ -45,9 +47,21 @@
...
 
-(defrecord Connection [current-db]
+(defrecord Connection [closed? current-db transact-chan]
   IConnection
-  (close [conn] (db/close-db @(:current-db conn)))
+  (close [conn]
+    (go-pair ;; Always want to return a pair-chan.
+      (when (compare-and-set! (:closed? conn) false true)
+        (let [result (a/chan 1)]
+          ;; Ask for the underlying database to be closed while draining the queue.  Invariant: we
```
ncalexan: It's really while, 'cuz closing can race with enqueing new transactions.

src/common/datomish/transact.cljc
ncalexan: The big advantage of treating all operations the same, and letting the operation handle closed, is that a new operation (perhaps we want to run queries through this?) is parallel to the existing work.

rnewman: Counter-point to myself: nothing should be able to be added to the `transact-chan` after the `:sentinel-close`, right? In that case, the close part can safely abort the loop…

src/common/datomish/transact.cljc
```diff
@@ -552,12 +580,99 @@
...
+                                ;; so the transaction has committed.
+                                (reset! (:current-db conn) (:db-after report))
+                                (>! (:listener-source conn) report))
+                              report)))))]
+              ;; Even when report is nil (transaction not committed), pair is non-nil.
+              (>! result pair))
+            (>! token-chan token)
+            (when close?
+              (a/close! result))
+            (recur)))))))
```
ncalexan: The `(recur)` is inside the `(when-let [[sentinel tx-data result close?] (<! (:transact-chan conn))] ...)` block, so as soon as the transaction channel is closed and drained, the `go` is done.

rnewman: You're right, but all that means is that when we park on the channel, we stay there forever, because the channel's closed!

We'll see the last item, process it, recur, get back to line 605, and park forever. We need a check somewhere to decide to terminate, right?

ncalexan: When the channel is closed and all items are drained, the take (`<!`) returns `nil`.  That is the check for draining done.

The token remains taken and we just "fall out of the loop", terminating the `go` block.

rnewman: Oh, you're right — I forgot that `take!` special-cases the closed check. Carry on!

src/common/datomish/transact.cljc
ncalexan: Further to our conversation, I think we will want to rewrite this approach to be much more careful with our DB threads (in Java) and handles (in JavaScript), so I'm going to leave this as is for now.

ncalexan:
This has landed: https://github.com/mozilla/datomish/commit/ed545d4a11312e6e3a243d127c49b6c01855a8fb.

PR #87, opened 2016-10-12
Correct type extraction from fulltext expressions, fix UNIONs.

rnewman:
Note that this relies on an upstream fix in honeysql (https://github.com/jkk/honeysql/pull/142). If that doesn't land soon, we can fork.

### DIFF
--- a/src/common/datomish/query/clauses.cljc
+++ b/src/common/datomish/query/clauses.cljc
@@ -469,12 +469,15 @@
           bindings (into {} (map (fn [var]
                                    (let [sym (:symbol var)]
                                      [sym [(sql/qualify alias (util/var->sql-var sym))]]))
-                                 free-vars))]
+                                 free-vars))
+
+          known-types
+          (reduce cc/combine-known-types {} (map :known-types ccs))]
 
       (cc/map->ConjoiningClauses
         {:source source
          :from [[subqueries alias]]
-         :known-types (apply merge (map :known-types ccs))
+         :known-types known-types
          :extracted-types (apply merge (map :extracted-types ccs))
          :external-bindings {}       ; No need: caller will merge.
          :bindings bindings
--- a/test/datomish/test/query.cljc
+++ b/test/datomish/test/query.cljc
@@ -824,3 +824,72 @@
                             '[?save :save/excerpt ?excerpt]]))]
       (is (or (= ["Some page title" "Some page excerpt"] result)
               (= ["A different page" "A different excerpt"] result))))))
+
+(deftest-db test-or-join-real-world conn
+  (let [attrs (<? (<initialize-with-schema
+                    conn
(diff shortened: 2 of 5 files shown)


src/common/datomish/query/clauses.cljc
```diff
@@ -469,12 +469,15 @@
           bindings (into {} (map (fn [var]
                                    (let [sym (:symbol var)]
                                      [sym [(sql/qualify alias (util/var->sql-var sym))]]))
-                                 free-vars))]
+                                 free-vars))
+
+          known-types
+          (reduce cc/combine-known-types {} (map :known-types ccs))]
```
ncalexan: This could be `transduce`, I expect, to save the lazy map.

test/datomish/test/query.cljc
```diff
@@ -824,3 +824,72 @@
...
+            (or-join [?page]
+              [(fulltext $ #{:page/url :page/title} ?str) [[?page]]]
+              (and
+                [(fulltext $ #{:save/title :save/excerpt :save/content} ?str) [[?save]]]
+                [?page :page/save ?save]))
+            [?page :page/url ?url]
+            [(get-else $ ?page :page/title "") ?title]]
+          conn)))))
+
+(deftest test-honeysql-union
```
ncalexan: Include a comment describing, or a link to, the issue, so this is less random.

test/datomish/test/query.cljc
```diff
@@ -824,3 +824,72 @@
                             '[?save :save/excerpt ?excerpt]]))]
       (is (or (= ["Some page title" "Some page excerpt"] result)
               (= ["A different page" "A different excerpt"] result))))))
+
+(deftest-db test-or-join-real-world conn
```
ncalexan: Consider a name that shows what you're testing, and perhaps a `testing` block to describe it.

src/common/datomish/query/clauses.cljc
rnewman: I gave this a shot. Expressed with `transduce`, this is:

```
(transduce (map :known-types) (completing cc/combine-known-types) {} ccs)
```

… which creates an intermediate transducer instead of a lazy sequence, an intermediate function to get a one-argument version of `combine-known-types`, and is slightly longer. Let me know if you prefer that.

(Still, it's nice to know I can figure out `transduce`!)

rnewman:
I have this rebased locally, but I won't land it until there's a working `honeysql` to point to.

ncalexan:
> I have this rebased locally, but I won't land it until there's a working honeysql to point to.

Roger that.  Land whenever.

rnewman:
Landed without the honeysql fixes, which I'll land later. feebfd0

rnewman:
I pushed a reference to my fork: 8bb0dcfa2ed5d8f84ebb34b3c20253cda9e0438a

PR #107, opened 2016-11-08
Schema management

rnewman:
This PR implements [my proposal](https://github.com/mozilla/datomish/wiki/Proposal:-application-schema-coordination-and-versioning).

Two schema management attributes are introduced: `:schema/version` and `:schema/attribute`.

Schema fragment entities are versioned, and 'contain' attributes in the store.

A _managed schema fragment_ list is passed in to `datomish.schema-management/<apply-schema-alteration`, along with optional migration code.

If a fragment is earlier than the contents of the store, an error is thrown. (Applications can continue on in a degraded state.)

If a fragment is inconsistent with the store — _e.g._, trying to claim ownership of an attribute that already belongs to another fragment — an error is thrown.

If a fragment is already at the right version in the store, no work is done for that fragment.

Otherwise, the fragment and the store are compared. Any pre/post functions for the collection of fragments and for each fragment individually are queued to run, and automated migration of the fragment is applied. All of this work happens in the same database transaction.

Most migrations won't need pre/post functions. No attempt has yet been made to expose these to JS, because that's a pain in the ass.

N.B., not all changes can be automated, and this implementation very likely does not allow a 'pre' function to impose unsupported migrations. That's easy to fix in the future, and doesn't affect the use of the current implementation.

There are tests.


src/common/datomish/pair_chan.cljc
```diff
@@ -71,6 +71,13 @@
      (consume-pair (cljs.core.async/<! ~pc-chan))
      (consume-pair (clojure.core.async/<! ~pc-chan))))
 
+(defmacro <??
+  "Takes from the channel if it's non-nil."
+  [pc-chan]
+  `(let [c# ~pc-chan]
+     (when c#
+       (datomish.pair-chan/<? c#))))
```
grigoryk: And if `par-chan` is nil, this isn't evaluated at all?

src/common/datomish/transact/bootstrap.cljc
```diff
@@ -38,6 +38,15 @@
                           :db/cardinality :db.cardinality/one}
    :db.alter/attribute   {:db/valueType   :db.type/ref
                           :db/cardinality :db.cardinality/many}
+
+   :db.schema/version    {:db/valueType   :db.type/long
+                          :db/cardinality :db.cardinality/one}
+
+   ;; unique-value because an attribute can only belong to a single
+   ;; schema fragment.
+   :db.schema/attribute  {:db/valueType   :db.type/ref
```
grigoryk: This says that attributes might have attributes, whose values must be unique. What somewhat confuses me is that there is no description/limitation of what an attribute actually is?

src/common/datomish/pair_chan.cljc
rnewman: That code _expands_. Assume the input bound to `pc-chan` is an expression, say `(foo 5)`, that returns `nil` when eventually run.

It expands at compile time to:

```
(let [c123 (foo 5)]
  (when c123
    (<? c123)))
```

When this code runs, inside of a `go-pair` block, it does the right thing: that is, if `(foo 5)` returns a value (presumably a channel), take from it; if it returns `nil` or `false`, do nothing.

src/common/datomish/transact/bootstrap.cljc
rnewman: This actually says that _schema fragments_ are related to attributes by the `:db.schema/attribute` relation.

Both schema fragments and attributes themselves are modeled as _entities_. An entity is a thing named by an `entid`. In this case, both area also identified by `ident`.

You can split up the world of Datomish into _values_ (5, "foo", 20161001T13:54:00Z), _entities_, and attributes that relate them. Attributes themselves are modeled as entities.

rnewman: `:db.type/ref` means that the range of the property is the space of all entities. The domain of every property is the space of all entities.

This is quite loosely modeled; there's no way in Datomish or Datomic to say "this property only applies to People, and can only point to Houses" — that's the realm of description logics.

Datomi{c,sh} are essentially RDF, which is a much weaker formalism.

src/common/datomish/transact.cljc
```diff
@@ -791,26 +791,30 @@
...
   Returns a pair-chan resolving to `[result error]`."
   ([conn tx-data]
    (<transact! conn tx-data (a/chan 1) true))
-  ([conn tx-data result close?]
+  ([conn tx-data-or-fn result close?]
    {:pre [(conn? conn)]}
    ;; Any race to put! is a real race between callers of <transact!.  We can't just park on put!,
    ;; because the parked putter that is woken is non-deterministic.
-   (let [closed? (not (a/put! (:transact-chan conn) [:sentinel-transact tx-data result close?]))]
+   (let [op (if (fn? tx-data-or-fn)
```
grigoryk: Here we determine what needs to be transacted - data or a function, asynchronously put a bundle of data (type of operation, fn/data,..) onto `:transact-chan` port, and finally raise if port was closed.

src/common/datomish/transact.cljc
```diff
@@ -823,30 +827,50 @@
...
+                               ;; The function must return a TxReport.
+                               ;; The function must not itself call
+                               ;; `in-transaction!` or `<transact!`.
+                               (partial tx-data-or-fn db <transact-tx-data-in-transaction!)
+
+                               :sentinel-transact
+                               ;; This is data. Apply it with `<transact-tx-data-in-transaction!`.
+                               (partial <transact-tx-data-in-transaction! db tx-data-or-fn))
+
+                               report (<? (db/in-transaction! db in-transaction-fn))]
```
grigoryk: This seems to be the meat of it. Based on type of the operation we need to perform, we construct a partial ("bound") function (`in-transaction-fn`), which is either our own `<transact-tx-data-in-transaction!` with bound data, or supplied function `tx-data-or-fn` bound to a `db` and a function which is able to perform data transactions (again, `<transact-tx-data-in-transaction!`).
This partial function is then asynchronously (and transactionally) executed by `s/in-transaction!`, and results are assigned to `report`, which in turn is offered up to `go-pair`.

src/common/datomish/pair_chan.cljc
grigoryk: I think I got it, thanks! This is neat stuff.

If `c#` will be bound to a channel once expanded, shouldn't it be called something like `<c#`? Or is that convention only for functions that return channels? In which case, shouldn't it be `[<pc-chan]`?

src/common/datomish/transact.cljc
rnewman: Exactly.

You've probably noticed that another way to do this is to push just a function into the queue, turning this loop into an unconditional "run this function with the DB" cycle.

We/I opted not to do that for a few reasons: firstly, it avoids creating thunks early, perhaps accidentally capturing scope; secondly, it makes it clear that this loop does several things; thirdly, we needed a sentinel for closing anyway, so expanding the set from two to three wasn't a big deal.

src/common/datomish/pair_chan.cljc
rnewman: Because this is a macro, `pc-chan` is a _form_ — it's a piece of syntax, not a channel. And similarly, `c#` is a gensym; not yet a thing.

The naming convention applies to what the programmer passes to `<??`, not here inside this very simple expansion, and indeed it only applies to functions — we usually call variables that _are_ channels `chan` or `c` or `pair-chan`.

So I'd use this like `(<?? (<my-channel-function 5))`, or `(let [c (<my-channel-function 5)] (println (<?? c)))`.

You can play around with this in a REPL (`lein repl`):

```
user=> (defmacro <??
  #_=>   "Takes from the channel if it's non-nil."
  #_=>   [pc-chan]
  #_=>   `(let [c# ~pc-chan]
  #_=>      (when c#
  #_=>        (datomish.pair-chan/<? c#))))
#'user/<??
user=> (macroexpand '(<?? (<my-fn 5)))
(let*
 [c__14311__auto__ (<my-fn 5)]
 (clojure.core/when c__14311__auto__ (datomish.pair-chan/<? c__14311__auto__)))
```

rnewman:
Landed; thanks for the review, @grigoryk. I'm going to file a follow-up for something else I think we'll need.

PR #116, opened 2016-11-18
Add a :none migration for schema management. Fixes #113.

### LINKED ISSUE #113: Schema management: allow migration from 0
Tofino's first release has a schema, but that schema isn't managed.

And it'll need some changes.

The simplest way to do that is to allow for an optional migration function from zero.

rnewman:
This allows for code to run before and after a schema fragment is added for the first time.

The anticipated use for this is twofold:

1. To do initial setup, e.g., defining global entities.
2. To 'adopt' unmanaged attributes already defined in the store.

This 'pre' would manually alter or retract attributes so that the transact of the new schema datoms can complete.

For example, if properties `:foo/bar` and `:foo/baz` will be unchanged, but `:noo/zob` needs to change from a string to an integer, the `:none` pre-function can alter the ident, the body of the fragment will transact the three properties up to the current version, and the `:none` post-function can migrate the renamed `:noo/zob` and clean up.

@grigoryk (and @jsantell, if you're interested), take a look? This is how I plan to impose schema management on the UAS's existing schema, which needs to be able to change.

### DIFF
--- a/src/common/datomish/schema_management.cljc
+++ b/src/common/datomish/schema_management.cljc
@@ -177,29 +177,36 @@
     [body
      (mapcat
        (fn [{:keys [name version attributes] :as fragment}]
-         (if-let [existing-version (get schema-fragment-versions name)]
-           ;; It's a change.
-           ;; Spit out any pre/post for this fragment, with a
-           ;; transact of the datoms to effect the change and
-           ;; bump the schema fragment version in the middle.
+         (let [existing-version (get schema-fragment-versions name)
+
+               datoms
+               [[:transact
+                 (if existing-version
+                   ;; It's a change.
+                   ;; Transact the datoms to effect the change and
+                   ;; bump the schema fragment version.
+                   (changed-schema-fragment->datoms
+                     (d/entid db name)
+                     symbolic-schema
+                     name
+                     attributes
+                     version)
+
+                   ;; It's new! Just do it.
+                   (managed-schema-fragment->datoms fragment))]]]
+
+           ;; We optionally allow you to provide a `:none` migration here, which
+           ;; is useful in the case where a vocabulary might have been added
+           ;; outside of the schema management system.
            (concat
              (when-let [fragment-pre-for-this
-                        (get-in fragment-pre [name existing-version])]
+                        (get-in fragment-pre [name (or existing-version :none)])]
                [[:call fragment-pre-for-this]])
-             [[:transact
-               (changed-schema-fragment->datoms (d/entid db name)
-                                                symbolic-schema
-                                                name
-                                                attributes
-                                                version)]]
+             datoms
              (when-let [fragment-post-for-this
-                        (get-in fragment-post [name existing-version])]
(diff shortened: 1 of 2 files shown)


grigoryk:
Just saw the notification - I need to start paying attention to those - will take a look in the morning.

grigoryk:
In case of already defined unmanaged attributes, it seems that the desired effect could also be achieved in at least two other ways:
1) ensure that the attributes are of the same shape as what initial fragment schema defines, so that no "adopting" is needed. I believe this is what many other schema management tools assume for the "initial" migrations on top of pre-existing schema
2) add a schema fragment in two steps, with v1 matching what's already present, and v2 migration making necessary changes

While (2) seems clumsy, (1) seems like a reasonable approach at first glance. Am I understanding this correctly, @rnewman? Could you describe why you've settled on a "none migration" approach?

rnewman:
(1) is actually what's done here: after we run the pre-hook, we'll go ahead and transact the schema fragment as if it didn't exist (which will fill in non-existent attributes). The point of the none-migration is: what do we do if the properties in the store _don't_ match the schema fragment at its current version?

If we do nothing, and just blindly try to transact the schema fragment, we will get an error telling us that we have to use `alter` instead. Even if we switch the schema management code to alter existing attributes in a fine-grained way, that might be wrong (if the meaning of the attribute has changed) or fail, or we might need to do additional work first.

That work needs to _transactionally_ clean up the store to match what we're about to do, and leave it untouched if we can't. That means a pre/post hook.

As you put it: the pre hook "ensures that the attributes are of the same shape".

(2) isn't feasible in the general case: you'd need to transactionally check for your schema version, then choose how to adopt the existing schema, then build a diff to get it up to date. At that point you need to run it in the scope of the migration, which is… a pre hook.

grigoryk:
> The point of the none-migration is: what do we do if the properties in the store don't match the schema fragment at its current version?

An alternative to this would be to say that any necessary changes must be made outside of the migration flow entirely, correct? I think that's the comparison I'm looking for - what are the pros/cons of allowing to include these changes as part of the :none migration, vs forcing the user to perform them beforehand, and keeping the first migration a very simple one. "Describe the current state of the world", as opposed to "Describe the current state of the world, but mutate it to match our description if necessary".

The "none" hook becomes a special proto-migration of sorts.

Is the goal here to essentially improve developer ergonomics of transitioning to managed schema? Or am I missing something, and this unblocks some flow which isn't possible at this point?

rnewman:
Essentially, this is the supported way to query for whether your schema fragment exists, and if it doesn't, then do arbitrary prep work. The big  improvement is that this all happens in a transaction, in a well-documented order. A failure will not corrupt the store, as a non-transactional approach would. 
Not all callers will need it, and you could build it yourself in CLJS, but that's true for this entire module.

rnewman:
Thanks for the thorough review!

rnewman:
Merged for 0.3.3: 103a86f

PR #127, opened 2016-11-30
Cast to lower-case all UUID strings in cljs. Fixes #45.

### LINKED ISSUE #45: Downcase/normalize UUIDs for storage
Until #44, we should at least make sure that two UUIDs that differ only in string case end up the same.

jsantell:


src/common/datomish/util.cljc
```diff
@@ -199,3 +199,12 @@
                  (clojure.set/union seen ks)
                  (rest key-sets)))
         overlapping))))
+
+#?(:clj
+     (defn uuidish? [x]
+       (instance? java.util.UUID x)))
+#?(:cljs
+     (let [uuid-re (js/RegExp. "[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}" "i")]
```
jsantell: As this would run on every string write, probably should look into more performant checks in JS world

src/common/datomish/sqlite_schema.cljc
```diff
@@ -203,7 +203,7 @@
 
       :cljs
       [string
-       (->SQLite [x] x)
+       (->SQLite [x] (if (util/uuidish? x) (clojure.string/lower-case x) x))
```
jsantell: Where/how could this be tested?

PR #134, opened 2016-12-16
Update test boilerplate for running on travis

bgrins:
This should be running an in-code and external test in CI once landed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 .nrepl-port
 /.lein-*
 /.nrepl-port
+Cargo.lock
 /checkouts/
 /classes/
 /node_modules/
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,6 +1,13 @@
+pub fn add_two(a: i32) -> i32 {
+    a + 2
+}
+
 #[cfg(test)]
 mod tests {
+    use super::*;
+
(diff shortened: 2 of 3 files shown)


bgrins:
Looks like it worked: https://travis-ci.org/mozilla/datomish/builds/184625808

Cargo.lock
```diff
@@ -0,0 +1,4 @@
+[root]
```
rnewman: We shouldn't check in this file.

http://doc.crates.io/guide.html#cargotoml-vs-cargolock

> If you’re building a library that other projects will depend on, put Cargo.lock in your .gitignore. If you’re building an executable like a command-line tool or an application, check Cargo.lock into git. If you're curious about why that is, see "Why do binaries have Cargo.lock in version control, but not libraries?" in the FAQ.

bgrins: Good catch, i'll remove it

bgrins:
This uses the `tests` directory to run tests out of band: ahttps://doc.rust-lang.org/book/testing.html#the-tests-directory.  Apparently subfolders don't get run automatically so we could move the js tests into this folder if we don't want to have both `test` and `tests` top level dirs

bgrins:
Pushed up the changes

PR #135, opened 2016-12-16
Include instructions for building and testing with cargo

bgrins:


ncalexan:
Is there a reason to do this in the same repository?

bgrins:
> Is there a reason to do this in the same repository?

The rust rewrite in general?  The consensus seemed that the main pro was that we could keep the existing issues in place by doing it in the same repo.  And that eventually this would become the master branch once it was deemed ready.

ncalexan:
> Is there a reason to do this in the same repository?
> 
> The rust rewrite in general? The consensus seemed that the main pro was that we could keep the existing issues in place by doing it in the same repo. And that eventually this would become the master branch once it was deemed ready.

OK, as long as there was discussion.

PR #136, opened 2016-12-16
Stub out cli crate

bgrins:
This adds a cli/ subfolder that defines a crate that builds a binary that uses an exposed variable from datomish. It's just stubbed out and only is fetching a string from the core lib.

Then the main Cargo.toml references it as a dev-dependency so it can be run in ci.  The idea is that we would do the same thing for additional binaries (like a bulk importer).  This is building both and passing: https://travis-ci.org/mozilla/datomish/builds/184661695

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,7 @@ Cargo.lock
 /classes/
 /node_modules/
 /out/
-/target/
+target/
 pom.xml
 pom.xml.asc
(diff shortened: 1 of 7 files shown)


rnewman:
(N.B., flatten commits to something like "Add a stub CLI tool and run tests on it. Fixes #136. r=rnewman" when landing, please!)

PR #138, opened 2016-12-17
(optionally merge) Create a sample crate for the parser

bgrins:
Just wanted to put this up for future reference - feel free to merge or close the PR.

I think for the short term at least, it'll be easier to manage the components within the project as crates within a single repo as opposed to separate repos deployed through crates.io.  This branch is a sample of what this could look like.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,5 +2,5 @@ language: rust
 script:
   - cargo build --verbose
   - cargo test --verbose
-  - cargo build --verbose -p datomish-cli
+  - cargo test --verbose -p datomish-parser
   - cargo test --verbose -p datomish-cli
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,5 +2,5 @@ language: rust
 script:
   - cargo build --verbose
   - cargo test --verbose
-  - cargo test --verbose -p datomish-parser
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -2,5 +2,5 @@ language: rust
 script:
   - cargo build --verbose
   - cargo test --verbose
-  - cargo build --verbose -p datomish-cli
```
bgrins: I realized that we didn't need to call `cargo build --verbose -p datomish-cli` in .travis.yml since a build error in datomish-cli will fail a build on datomish

ncalexan:
On Fri, Dec 16, 2016 at 5:49 PM, Richard Newman <notifications@github.com>
wrote:

> *@rnewman* approved this pull request.
>
> I think this is valuable, but I'd either change to datomish-query-parser
> or datomish-example-crate, 'cos we'll have more than one parser!
>
Yeah -- I was envisioning an edn parsing crate, consumed by a Datalog
parsing crate and a transaction syntax parsing crate and a pull syntax
parsing crate, etc, etc.

bgrins:
Cool, I'll rename to `datomish-query-parser` before landing

PR #147, opened 2017-01-04
Initial database explorer rust server

jsantell:
in progress

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,9 @@ authors = ["Richard Newman <rnewman@twinql.com>", "Nicholas Alexander <nalexande
 [dev-dependencies.datomish-cli]
   path = "cli"
 
+[dev-dependencies.datomish-explorer]
+  path = "explorer"
+
 [[bin]]
 name = "datomish-cli"
 path = "cli/src/main.rs"
--- a/explorer/Cargo.toml
+++ b/explorer/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "datomish-explorer"
+version = "0.0.1"
+
+[dependencies]
+iron = "0.4.x"
+toml = "0.2.1"
+clap = "2.19.3"
+
+[dependencies.datomish]
+  path = "../"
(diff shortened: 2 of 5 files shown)


explorer/src/main.rs
```diff
@@ -0,0 +1,27 @@
...
+use iron::status;
+
+fn main() {
+    let args = get_args();
+    fn hello_world(_: &mut Request) -> IronResult<Response> {
+        Ok(Response::with((status::Ok, "Hello World!")))
+    }
+
+    println!("Running Datomish Explorer at localhost:{} for database at {}", args.port, args.database);
+    let _server = Iron::new(hello_world).http(("localhost", args.port)).unwrap();
```
bgrins: I'm assuming there will also be an easy way to deliver static files here?

explorer/README.md
```diff
@@ -0,0 +1,3 @@
+# datomish-explorer
+
+An in-browser IDE for exploring a Datomish database.
```
bgrins: Can you add the command needed to start it here?

explorer/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "datomish-explorer"
+version = "0.0.1"
+
+[dependencies]
+iron = "0.4.x"
```
bgrins: any reason to use 0.4.x here but an explicit version for the others?  not really sure best practice here

Cargo.toml
```diff
@@ -11,6 +11,9 @@ authors = ["Richard Newman <rnewman@twinql.com>", "Nicholas Alexander <nalexande
 [dev-dependencies.datomish-cli]
   path = "cli"
 
+[dev-dependencies.datomish-explorer]
+  path = "explorer"
+
 [[bin]]
 name = "datomish-cli"
```
bgrins: Might need to add it here to be able to run with `cargo run`.  Assuming the command would be `cargo run --bin datomish-explorer`.

Alternatively, maybe we could fold the cli portion of the explorer into the normal cli.  So `cli` would depend on `explorer` and explorer would be just the web server code.  Since the current cli does nothing, starting the explorer could be the default behavior

bgrins: We could also make this the default bin, since the current cli doesn't do anything.  Then it'd just be `cargo run`

bgrins:
Closing in favor of #159, as discussed

PR #148, opened 2017-01-05
Take on rusqlite dependency

bgrins:
Here's some more info about the various sqlite libraries: https://users.rust-lang.org/t/sqlite-libraries-potential-for-unification/8702.

They seem to take similar approaches, chose rusqlite based on activity and available docs to populate a simple test case.   Also, there's some discussion about unifying on it at https://github.com/dckc/rust-sqlite3/issues/46

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,8 @@ version = "0.4.0"
 authors = ["Richard Newman <rnewman@twinql.com>", "Nicholas Alexander <nalexander@mozilla.com>"]
 
 [dependencies]
+rusqlite = "0.8.0"
+
 [dependencies.datomish-query-parser]
   path = "query-parser"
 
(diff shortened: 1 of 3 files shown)


bgrins:
@ncalexan @rnewman seems like there may be other plans for this in https://bugzilla.mozilla.org/show_bug.cgi?id=1328670?  Should we be looking to use sqlite3-sys instead?

rnewman:
I think the idea is that those higher-level libraries should themselves use the *-sys variant. Indeed, rusqlite does so, only offering a bundled version if you need it.

John Gallagher (the author) has chimed in on that bug, and rusqlite looks good to me (hooray for `functions`!), so I'm happy to take this path.

bgrins:
> rusqlite looks good to me (hooray for functions!), so I'm happy to take this path.

Nice!  Going to request review for this initial landing then

bgrins:
I split up the todo list for sqlite in https://github.com/mozilla/datomish/issues/133 into multiple parts that could be landed separately.  Was hoping to get the basics in place here to be able to wire up to some of the explorer work @jsantell is doing

PR #149, opened 2017-01-06
Add a basic EDN parser

joewalker:
The parser mostly works and has a decent test suite. It parses all the
queries issued by tofino-user-agent with some caveats. Known flaws:

* No support for tagged elements, comments, discarded elements or "'"
* Incomplete support for escaped characters in strings and the range of
  characters that are allowed in keywords and symbols
* Possible whitespace handling problems
* Possibly poor memory handling

### DIFF
--- a/edn/Cargo.toml
+++ b/edn/Cargo.toml
@@ -1,3 +1,17 @@
 [package]
 name = "edn"
-version = "0.0.1"
+version = "0.1.0"
+authors = ["Joe Walker <jwalker@mozilla.com>"]
+
+license = "Apache-2.0"
+repository = "https://github.com/mozilla/datomish"
+description = "EDN Parser for Datomish"
+build = "build.rs"
+readme = "./README.md"
+
+[dependencies]
+num = "0.1.35"
+ordered-float = "0.3.0"
+
+[build-dependencies]
+peg = "0.4"
--- a/edn/README.md
+++ b/edn/README.md
@@ -1,0 +1,2 @@
+# barnardsstar
+An experimental EDN parser for Datomish
--- a/edn/build.rs
+++ b/edn/build.rs
@@ -8,17 +8,8 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
-#![allow(dead_code)]
-
-extern crate ordered_float;
-extern crate num;
-
-pub mod types;
-
-pub mod parse {
-    include!(concat!(env!("OUT_DIR"), "/edn.rs"));
-}
+extern crate peg;
 
 fn main() {
-    println!("Use cargo test");
+    peg::cargo_build("src/edn.rustpeg");
 }
--- a/edn/src/edn.rustpeg
+++ b/edn/src/edn.rustpeg
@@ -1,0 +1,124 @@
+// Copyright 2016 Mozilla
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+use std::collections::{BTreeSet, BTreeMap, LinkedList};
+use std::iter::FromIterator;
+use num::BigInt;
+use types::Value;
+use ordered_float::OrderedFloat;
+
+// Goal: Be able to parse https://github.com/edn-format/edn
+// Also extensible to help parse http://docs.datomic.com/query.html
+
+// Debugging hint: test using `cargo test --features peg/trace -- --nocapture`
+// to trace where the parser is failing
+
+// TODO: Support tagged elements
+// TODO: Support comments
+// TODO: Support discard
+
+#[export]
+nil -> Value = "nil" {
+    Value::Nil
+}
+
+#[export]
+boolean -> Value =
+    "true" { Value::Boolean(true) } /
+    "false" { Value::Boolean(false) }
+
+digit = [0-9]
+sign = "-" / "+"
+
+#[export]
+bigint -> Value = b:$( sign? digit+ ) "N" {
+    Value::BigInteger(b.parse::<BigInt>().unwrap())
+}
+
+#[export]
+integer -> Value = i:$( sign? digit+ ) {
+    Value::Integer(i.parse::<i64>().unwrap())
+}
+
+frac =     sign? digit+ "." digit+
+exp =      sign? digit+            ("e" / "E") sign? digit+
+frac_exp = sign? digit+ "." digit+ ("e" / "E") sign? digit+
+
+// The order here is important - frac_exp must come before (exp / frac) or the
+// parser assumes exp or frac when the float is really a frac_exp and fails
+#[export]
+float -> Value = f:$( frac_exp / exp / frac ) {
+    Value::Float(OrderedFloat(f.parse::<f64>().unwrap()))
+}
+
+// TODO: \newline, \return, \space and \tab
+special_char = quote / tab
+quote = "\\\""
+tab = "\\tab"
+char = [^"] / special_char
+
+#[export]
+text -> Value = "\"" t:$(  char* ) "\"" {
+    Value::Text(t.to_string())
+}
+
+// TODO: Be more picky here
+symbol_char_initial = [a-z] / [A-Z] / [0-9] / [*!_?$%&=<>/.]
+symbol_char_subsequent = [a-z] / [A-Z] / [0-9] / [*!_?$%&=<>/.] / "-"
+
+#[export]
+symbol -> Value = s:$( symbol_char_initial symbol_char_subsequent* ) {
+    Value::Symbol(s.to_string())
+}
+
+keyword_char_initial = ":"
+// TODO: More chars here?
+keyword_char_subsequent = [a-z] / [A-Z] / [0-9] / "/"
+
+#[export]
+keyword -> Value = k:$( keyword_char_initial keyword_char_subsequent+ ) {
+    Value::Keyword(k.to_string())
+}
+
+#[export]
+list -> Value = "(" __ v:(__ value)* __ ")" {
+    Value::List(LinkedList::from_iter(v))
+}
+
+#[export]
+vector -> Value = "[" __ v:(__ value)* __ "]" {
+    Value::Vector(v)
+}
+
+#[export]
+set -> Value = "#{" __ v:(__ value)* __ "}" {
+    Value::Set(BTreeSet::from_iter(v))
+}
+
+pair -> (Value, Value) = k:(value) " " v:(value) ", "? {
+    (k, v)
+}
+
+#[export]
+map -> Value = "{" __ v:(pair)* __ "}" {
+    Value::Map(BTreeMap::from_iter(v))
+}
+
+// It's important that float comes before integer or the parser assumes that
+// floats are integers and fails to parse
+#[export]
+value -> Value
+    = nil / boolean / float / bigint / integer / text /
+      keyword / symbol /
+      list / vector / map / set
+
+whitespace = (" " / "\r" / "\n" / "\t")
+
+__ = whitespace*
--- a/edn/src/types.rs
+++ b/edn/src/types.rs
@@ -1,0 +1,85 @@
+// Copyright 2016 Mozilla
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software distributed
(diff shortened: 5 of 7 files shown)


edn/src/edn.rustpeg
```diff
@@ -0,0 +1,106 @@
+
```
rnewman: License blocks in all files, please.

edn/build.rs
```diff
@@ -0,0 +1,6 @@
+
+extern crate peg;
+
+fn main() {
+    peg::cargo_build("src/edn.rustpeg");
```
rnewman: Could you add `.gitattributes` and a modeline to have the `.rustpeg` file marked as Rust?

Details:

https://github.com/github/linguist/blob/master/README.md#using-gitattributes

edn/src/types.rs
```diff
@@ -0,0 +1,73 @@
+
+use std::collections::{BTreeSet, BTreeMap, LinkedList};
+use std::cmp::{Ordering, Ord, PartialOrd};
+use ordered_float::OrderedFloat;
+
+/// We're using BTree{Set, Map} rather than Hash{Set, Map} because the BTree variants implement Hash
```
rnewman: For `///`, add a little bit of explanatory documentation about what this enum is.

edn/src/types.rs
```diff
@@ -0,0 +1,73 @@
...
+            Keyword(ref ks) => match *other { Keyword(ref ko) => ko.cmp(&ks), _ => ord_order },
+            Vector(ref vs)  => match *other { Vector(ref vo)  => vo.cmp(&vs), _ => ord_order },
+            List(ref ls)    => match *other { List(ref lo)    => lo.cmp(&ls), _ => ord_order },
+            Set(ref ss)     => match *other { Set(ref so)     => so.cmp(&ss), _ => ord_order },
+            Map(ref ms)     => match *other { Map(ref mo)     => mo.cmp(&ms), _ => ord_order },
+        }
+    }
+}
+
+// TODO: There has to be a better way to do `as i32` for Value
```
rnewman: Not after `FromPrimitive` was removed. You can unsafely transmute, or you can use e.g., the `num` crate's automatically derived version, but it basically turns into this.

edn/src/types.rs
```diff
@@ -0,0 +1,73 @@
+
+use std::collections::{BTreeSet, BTreeMap, LinkedList};
+use std::cmp::{Ordering, Ord, PartialOrd};
+use ordered_float::OrderedFloat;
+
+/// We're using BTree{Set, Map} rather than Hash{Set, Map} because the BTree variants implement Hash
+/// (unlike the Hash variants which don't in order to preserve O(n) hashing time which is hard given
+/// recurrsive data structures)
+/// See https://internals.rust-lang.org/t/implementing-hash-for-hashset-hashmap/3817/1
+/// TODO: We should probably Box the collection types
```
rnewman: Why? Collection types are effectively boxes already, no? (That is, a `Vec` is a relatively small copyable structure…)

Do you have some interesting heterogeneous trait object issue in mind?

edn/src/tests.rs
```diff
@@ -0,0 +1,770 @@
+// TODO: Can't we do this just for tests?
```
rnewman: Why isn't this file in `edn/tests/`?

edn/src/edn.rustpeg
```diff
@@ -0,0 +1,106 @@
...
+boolean -> Value =
+    "true" { Value::Boolean(true) } /
+    "false" { Value::Boolean(false) }
+
+digit = [0-9]
+sign = "-" / "+"
+
+#[export]
+integer -> Value = i:$( sign? digit+ ) {
+    Value::Integer(i.parse::<i32>().unwrap())
```
rnewman: I think `i32` is probably wrong. Clojure automatically walks the numeric tower:

```clojure
user=> {:foo 12345423857458454}
{:foo 12345423857458454}
user=> {:foo 123454238574584543434343434343}
{:foo 123454238574584543434343434343N}
```

but at the very least we will want to support microsecond timestamps, which are bigger than a 32-bit signed int.

edn/src/tests.rs
```diff
@@ -0,0 +1,770 @@
...
+fn test_text() {
+    assert_eq!(text("\"hello world\"").unwrap(), Text("hello world".to_string()));
+    assert_eq!(text("\"\"").unwrap(), Text("".to_string()));
+
+    assert!(text("\"").is_err());
+    assert!(text("nil").is_err());
+}
+
+#[test]
+fn test_symbol() {
```
rnewman: Please also test the symbols `.` and `$`.

edn/src/edn.rustpeg
```diff
@@ -0,0 +1,106 @@
...
+symbol_char_subsequent = [a-z] / [A-Z] / [0-9] / [*!_?$%&=<>/.] / "-"
+
+#[export]
+symbol -> Value = s:$( symbol_char_initial symbol_char_subsequent* ) {
+    Value::Symbol(s.to_string())
+}
+
+keyword_char_initial = ":"
+// TODO: More chars here?
+keyword_char_subsequent = [a-z] / [A-Z] / [0-9] / "/"
```
rnewman: For future correction: both keywords and symbols can contain `/`, but only once, and result in a namespaced keyword. That is, `:foo/bar` has namespace `"foo"` and name `"bar"`.

(There are similar rules around `.`, which divides up the namespace into a hierarchy, but can also appear as part of the name.)

edn/.gitignore
```diff
@@ -0,0 +1,7 @@
+# Generated by Cargo
+# will have compiled files and executables
+/target/
```
bgrins: I think this file shouldn't be needed - our toplevel gitignore should handle these paths (or be updated to handle them)

ncalexan:
I've been using https://github.com/Marwes/combine to parse an EDN AST into a yet more specialized transaction AST.  One of the nice things about `combine` is that it makes providing decent error messages possible (if not easy).

While thinking about how to provide pleasing error messages that include input positions, I realized that I want my EDN input to includes _its_ input positions as well.  (Either that or I want to parse my own text, so that I can provide meaningful error messages that include input positions and ranges.)

This may not be something we want to handle at this time, but how hard is it to include input ranges using `peg`?  At least that way I could join `peg` ranges to provide some hints as to which part of your input failed.

edn/src/edn.rustpeg
joewalker: I'm not good enough at Rust enough to be able to make more than a stab at the right answer here.

The trivial thing would be s/i32/i64/g "because 64 bits should be enough for anyone". But then because it wouldn't, s/i64/[i128](https://github.com/rust-lang/rust/pull/37900)/g or s/i64/[BigInt](https://rust-num.github.io/num/num/struct.BigInt.html)/g which suddenly gets weird and or slow.

We could have the parser auto select the smallest type that fits. Except that it seems awkward if a parse of a well defined data based on a known schema results in different types depending on the data.

My currently thought is that i64 is safe, predictable, easily converted and good enough for 99.9% of cases, and that we should fail to parse for numbers bigger than i64. And that when this becomes a problem we allow global configuration of the number type to use, and when this becomes cumbersome we allow some fancy case-by-case configuration.

Or Rust might do something fancy to help us with this problem that I'm not aware of?

edn/src/edn.rustpeg
joewalker: #154

edn/build.rs
joewalker: It's fairly explicitly not rust though. It happens to have a `use` keyword that works like Rust's, along with vaguely Rust like code blocks (except that the braces must match, even in strings), and comments (except that they're 'C' style not Rust style, i.e. multi-line comments don't nest)

edn/.gitignore
joewalker: Good point. Will fix

edn/src/types.rs
joewalker: Well TBH I'm copying @jimblandy who uses `Object(Box<HashMap<String, Json>>)` in his JSON example (see enum chapter). Except that I note [the reality](https://github.com/serde-rs/json/blob/a32de277b50cbbbfd786d0a93f566dd30d71a537/json/src/value.rs#L97) is different.

edn/src/types.rs
joewalker: I agree (and I've made it better) That said, we don't really have a defined API yet, so I'm OK with better docs being lower priority.

edn/src/edn.rustpeg
rnewman: I think `i64` is a good default, and we might consider adding a `BigInt` alongside it — even if we only stub out the enum case at this point, and don't parse it.

edn/build.rs
rnewman: Hm. I suppose I was thinking that Rust syntax highlighting, indenting, etc. would be better than nothing? Up to you.

There are some syntax highlighting plugins for it:

https://github.com/treycordova/rustpeg.vim

with filetype `set filetype=rust.rustpeg`, so perhaps just add that as a Vim modeline and hope Linguist figures it out?

joewalker:
In answer to ncalexan's question about retaining input positions, I think the answer is to include some sort of line+col / offset structure alongside each [Value](https://github.com/joewalker/mentat/blob/ca2269154ef788003c7040cf1a82a4d7e35cbe44/edn/src/types.rs#L17). I don't have a concrete plan, but I believe it should work.

edn/src/edn.rustpeg
palango: The [edn format docs](https://github.com/edn-format/edn#integers) say:

> 64-bit (signed integer) precision is expected. An integer can have the suffix N to indicate that arbitrary precision is desired

joewalker: Thanks @palango 
i64 fix is in https://github.com/mozilla/mentat/pull/149/commits/29bc8081a1983e9e59b1b0450bf1307a2db053b6 and BigInt addition is in https://github.com/mozilla/mentat/pull/149/commits/41f59977f67c6cfd53e88e51b1bb2975510f82be

joewalker:
It's worth preserving a link to https://github.com/mozilla/mentat/pull/149/commits/29bc8081a1983e9e59b1b0450bf1307a2db053b6 and https://github.com/mozilla/mentat/pull/149/commits/41f59977f67c6cfd53e88e51b1bb2975510f82be as I'm about to rebase and squash and @rnewman probably wants to see those before any final r+.

edn/Cargo.toml
```diff
@@ -1,3 +1,17 @@
 [package]
 name = "edn"
-version = "0.0.1"
+version = "0.1.0"
+authors = ["Joe Walker <jwalker@mozilla.com>"]
+
+license = "Apache-2.0"
+repository = "https://github.com/mozilla/datomish"
+description = "EDN Parser for Datomish"
```
rnewman: Two 'Datomish' to replace.

edn/README.md
```diff
@@ -0,0 +1,2 @@
+# barnardsstar
+An experimental EDN parser for Datomish
```
rnewman: Same.

edn/src/edn.rustpeg
```diff
@@ -0,0 +1,124 @@
...
+sign = "-" / "+"
+
+#[export]
+bigint -> Value = b:$( sign? digit+ ) "N" {
+    Value::BigInteger(b.parse::<BigInt>().unwrap())
+}
+
+#[export]
+integer -> Value = i:$( sign? digit+ ) {
+    Value::Integer(i.parse::<i64>().unwrap())
```
rnewman: It occurs to me that `unwrap` here might be a bad thing — if I provide a query like

```
[:find ?foo :in $ :where [_ _ 12345678901234567890123456789012345678901234567890]]
```

— forgetting the 'N' — the parser will panic. Now, we could build a strategy that always handles panics in the parser, allowing us to avoid error handling, but can we instead signal a failure to parse at this point?

edn/src/lib.rs
```diff
@@ -8,4 +8,17 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
-pub mod keyword;
```
rnewman: You've unrooted the existing `Keyword` module with this change…

edn/src/lib.rs
```diff
@@ -8,4 +8,17 @@
...
+extern crate ordered_float;
+extern crate num;
+
+pub mod types;
+
+pub mod parse {
+    include!(concat!(env!("OUT_DIR"), "/edn.rs"));
+}
+
+fn main() {
```
rnewman: This shouldn't be in a `lib.rs`.

edn/src/types.rs
```diff
@@ -0,0 +1,85 @@
...
+pub enum Value {
+    Nil,
+    Boolean(bool),
+    Integer(i64),
+    BigInteger(BigInt),
+    // https://users.rust-lang.org/t/hashmap-key-cant-be-float-number-type-why/7892
+    Float(OrderedFloat<f64>),
+    Text(String),
+    Symbol(String),
+    Keyword(String),
```
rnewman: Probably this should be `Keyword(Keyword)`, with the second being a reference to the richer namespace/name-containing `Keyword` type defined in `keyword.rs`.

rnewman: … but that's #154. So roll on for now.

rnewman:
Landed in c473511, with the fixes from f12fbf23889903bb67ccc01ce1e2047b8743533a squashed in.

PR #150, opened 2017-01-06
Move the bin to src and take on clap dependency for command line arg parsing

bgrins:
I realized that keeping the cli in another crate wasn't helping and actually caused build issues once we had remote dependencies (cargo wouldn't build that sub-crate since it wasn't a dependency of datomish and just a bin reference).

We can easily keep the binary in the main src/ directory so that's what this does.  There's initial support for subcommands in place (so we can do `cargo run serve`, `cargo run whatever`.

This doesn't yet wire up the web server from #147


bgrins:
Needs a squash before landing, but changes should be ready for review.  Can handle the server in a follow-up

src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+extern crate clap;
+
+use clap::{App, Arg, SubCommand};
+
+fn main() {
+    let matches = App::new("Mentat")
+                    .subcommand(SubCommand::with_name("serve")
```
rnewman: This deep indenting doesn't excite me a ton, but oh well.

src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+                                            .about("Starts a server")
+                                            .arg(Arg::with_name("debug")
+                                                .long("debug")
+                                                .help("Print debugging info"))
+                                            .arg(Arg::with_name("database")
+                                                 .short("d")
+                                                 .long("database")
+                                                 .value_name("FILE")
+                                                 .help("Path to the Mentat database to serve")
+                                                 .default_value("temp.db")
```
rnewman: An empty string here will behave as an in-memory database, which is probably a good default.

src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+                    .get_matches();
+
+    if let Some(ref matches) = matches.subcommand_matches("serve") {
+        let debug = matches.is_present("debug");
+        println!("This doesn't work yet, but it will eventually serve the following database: {} on port: {}.  Debugging={}",
+                    matches.value_of("database").unwrap(),
+                    matches.value_of("port").unwrap(),
+                    debug);
+    } else {
+        println!("Sorry, the cli currently only supports the serve subcommand.  Try `cargo run serve -- --help`");
```
rnewman: Can you print the help here yourself?

bgrins: Found a way to handle this automatically with: https://kbknapp.github.io/clap-rs/clap/enum.AppSettings.html#variant.ArgRequiredElseHelp

src/main.rs
bgrins: Alright, switched

src/main.rs
bgrins: Yeah, ran rustfmt on the code which reduced indentation a bit

PR #152, opened 2017-01-07
Rename everything to Project Mentat

rnewman:
This is only on the Rust fork; I didn't have time to also fix d-u-a-s yet.


rnewman:
@bgrins: I added a commit to name our crates/packages with underscores; hyphens aren't always safe. Take a look?

bgrins:
> @bgrins: I added a commit to name our crates/packages with underscores; hyphens aren't always safe. Take a look?

AFAICT there aren't any issues with them in our project currently (we can still reference `extern crate datomish_query_parser` even when it's name is `datomish-query-parser`).  Is there a specific case where it isn't safe?

Either way, renaming to underscores works for me.

Issue #133, opened 2016-12-16
Rust backlog


rnewman:
We're starting on a Rust rewrite of the existing Clojure codebase, using the `rust` branch.

With an eye to parallelism, the following will get us to parity and a bit further, at which point we'll start on the list of more speculative work. 

Large portions of this will be easier than the first time around: the ClojureScript implementation required lots of deep thinking that won't need to be repeated. 

**Starter Projects**: 

- [ ] Learning Rust. This is a totally valid use of time. 
- [ ] Build an EDN parser. This doesn't seem to be a completely solved problem in Rust. 
- [ ] Build a Rust node native module. End goal: call a Rust function from JS. 
- [ ] Build a Rust library for Android. Call it via JNI. 
- [ ] Build a Rust library for iOS. Sign it. Call it from Swift. 
- [ ] Get Rust builds and tests working in CI. 
- [ ] Command-line tool. Right now it wouldn't do much, but we could put the skeleton in place and do something like print user_version for a DB path… 
- [ ] [@jsantell] Begin SQLite wrapper: dependency, open a DB at a path and run the default  pragmas, exec/query abstractions. 
  - [ ] How do we update and deliver SQLite? 
  - [ ] Survey of Rust SQLite libraries. 
- [ ] Figure out an out-of-band test solution for Rust. In-code tests are great, but they're going to get really verbose when we have piles of SQL generation tests. 

**Moving on**: 

- [ ] Core (the big blocker for much of the rest): 
  - [ ] Core type work: type codes and translation, value conversion. 
  - [ ] Core schema work. 
  - [ ] SQLite wrapper: opener, part reading, etc. 
  - [ ] Bootstrapper. 
  - [ ] Conn + DB. 
- [ ] Querying: 
  - [ ] Datalog query parser: port from DataScript. 
  - [ ] Parsed query (FindRel) -> abstract query (ConjoiningClauses). 
  - [ ] Abstract query -> SQL. 
  - [ ] Abstract query -> projectors. 
  - [ ] Querying history and logs. 
- [ ] Transactor: 
  - [ ] Transactor expansion: schema + shorthand -> datoms. 
  - [ ] Transactor writer. 
  - [ ] Transactor loop. 
  - [ ] Prepared statements/statement caching. 
  - [ ] Excision. 
  - [ ] noHistory. 
- [ ] Schema management module (porting). 
- [ ] Reader pool (query executor). 
  - [ ] Query interruption via closing connection. 
[...]

PR #158, opened 2017-01-10
Implement basic logging infrastructure

palango:
A take on the logging mentioned in #133. I used the mentioned slog crate and created just a terminal logger for now, with the granularity bound on the debug command line option. See the two examples below for example output.

The change in `lib.rs` is just to show the usage in a library, I can remove it if you want.

```
> cargo run serve
Finished debug [unoptimized + debuginfo] target(s) in 0.0 secs
Running `target/debug/mentat serve`
This doesn't work yet, but it will eventually serve the following database: temp.db on port: 3333.
Jan 10 22:38:12.487 ERRO Calling a function: mentat, version: 0.4.0
```

```
> cargo run serve -- --debug
Finished debug [unoptimized + debuginfo] target(s) in 0.0 secs
Running `target/debug/mentat serve --debug`
This doesn't work yet, but it will eventually serve the following database: temp.db on port: 3333.
Jan 10 22:37:51.396 INFO Serving database, version: 0.4.0, database: temp.db, port: 3333, debug mode: true
Jan 10 22:37:51.397 INFO Called into mentat library, version: 0.4.0, fn: get_name
Jan 10 22:37:51.397 ERRO Calling a function: mentat, version: 0.4.0
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,17 +1,20 @@
 [package]
+authors = ["Richard Newman <rnewman@twinql.com>", "Nicholas Alexander <nalexander@mozilla.com>"]
 name = "mentat"
 version = "0.4.0"
-authors = ["Richard Newman <rnewman@twinql.com>", "Nicholas Alexander <nalexander@mozilla.com>"]
 
 [dependencies]
-rusqlite = "0.8.0"
 clap = "2.19.3"
+rusqlite = "0.8.0"
+slog = "1.4.0"
+slog-scope = "0.2.2"
+slog-term = "1.3.4"
 
 [dependencies.edn]
-  path = "edn"
+path = "edn"
 
 [dependencies.mentat_query]
-  path = "query"
+path = "query"
 
 [dependencies.mentat_query_parser]
-  path = "query-parser"
+path = "query-parser"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,6 +8,11 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
+#[macro_use]
+extern crate slog;
+#[macro_use]
+extern crate slog_scope;
+
 extern crate edn;
 extern crate mentat_query;
 extern crate mentat_query_parser;
@@ -18,6 +23,7 @@ use rusqlite::Connection;
 pub mod ident;
 
 pub fn get_name() -> String {
+    info!("Called into mentat library"; "fn" => "get_name");
     return String::from("mentat");
 }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -10,7 +10,16 @@
 
 extern crate clap;
 
+#[macro_use]
+extern crate slog;
+#[macro_use]
+extern crate slog_scope;
+extern crate slog_term;
+
+extern crate mentat;
+
 use clap::{App, Arg, SubCommand, AppSettings};
+use slog::DrainExt;
 
 fn main() {
     let app = App::new("Mentat").setting(AppSettings::ArgRequiredElseHelp);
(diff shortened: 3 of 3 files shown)


src/main.rs
```diff
@@ -37,9 +46,25 @@ fn main() {
...
         let debug = matches.is_present("debug");
         println!("This doesn't work yet, but it will eventually serve the following database: {} \
-                  on port: {}.  Debugging={}",
+                  on port: {}.",
                  matches.value_of("database").unwrap(),
-                 matches.value_of("port").unwrap(),
-                 debug);
+                 matches.value_of("port").unwrap());
+
+        // set up logging
```
ncalexan: nit: full sentence comments, please.

src/lib.rs
```diff
@@ -8,6 +8,11 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
+#[macro_use]
```
ncalexan: Oh, interesting, so you get global logging without an explicit reference to a logger object.  Does this work across crates?  I.e., if we added `info!` logging to the `query` crate, would this automatically pick up the logger you configure in `main.rs` below?

ncalexan: I wonder if we want to work out how to make "dummy logging" for crates that we use in Mentat but also want to publish for third-party consumption, like (e.g. `edn`).  Maybe we can use a feature flag to control "no logging"  and replace the `info!` macros with dummy macros in that case?  I'd be curious to know what the Rust community patterns for that stuff are.

Cargo.toml
```diff
@@ -1,17 +1,20 @@
...
+authors = ["Richard Newman <rnewman@twinql.com>", "Nicholas Alexander <nalexander@mozilla.com>"]
 name = "mentat"
 version = "0.4.0"
-authors = ["Richard Newman <rnewman@twinql.com>", "Nicholas Alexander <nalexander@mozilla.com>"]
 
 [dependencies]
-rusqlite = "0.8.0"
 clap = "2.19.3"
+rusqlite = "0.8.0"
+slog = "1.4.0"
```
ncalexan: For my future self: `slog` is licensed MPL2.  I don't know if APL can depend on MPL2!  (I know that MPL2 can depend on APL.) @rnewman, @joewalker do you know off the top?

src/lib.rs
ncalexan: Ooh, I see "using slog in library does not force users of the library to use slog (but gives benefits); see crates/example-lib" on the `slog` documentation page.  That's awesome sauce!

src/main.rs
```diff
@@ -37,9 +46,25 @@ fn main() {
...
+        let log_level = if debug {
```
ncalexan: Is this formatted by `rustfmt`?  I wonder if this should be all one line, but I'm happy to roll with your preference.

Cargo.toml
rnewman: I believe so.

http://softwareengineering.stackexchange.com/questions/321565/can-i-use-libraries-published-under-mpl-2-0-and-mit-in-a-project-that-uses-apl-2

src/lib.rs
palango: It works without an explicit reference, but this makes it impossible to use special loggers as well.

I just tested to use a `slog`based lib with an `env_logger` based executable, and that doesn't work. Seems like it relies on this [line:](https://github.com/slog-rs/slog/blob/1925efe85ef3d7d64d3cd7fe9c6544195bd75ede/crates/example-lib/lib.rs#L27)
```
slog::Logger::root(slog_stdlog::StdLog.fuse(), o!())
```
That would mean we have to take care to instanciate a logger for every standalone library, an idea I don't really like.
Assuming this it seems like the better choice to use the standard `log` interface for the libraries. What do you think?

src/main.rs
palango: Yeah, it was rustfmt. I have no hard opinions either.

palango:
See #174 for a better approach. Closing.

ncalexan:
Landed in #205.  Thanks, @palango!  We'll try to be more responsive in future.

PR #159, opened 2017-01-10
Add beginning of web server for the serve subcommand

bgrins:


bgrins:
@jsantell what do you think?

src/main.rs
```diff
@@ -24,7 +30,7 @@ fn main() {
                 .long("database")
                 .value_name("FILE")
                 .help("Path to the Mentat database to serve")
-                .default_value("temp.db")
+                .default_value("")
```
rnewman: That was an earlier review comment of mine. Using `""` here means "default to an in-memory database", which I think is sensible.

src/main.rs
```diff
@@ -9,9 +9,15 @@
...
 
 extern crate clap;
+#[macro_use] extern crate nickel;
+
+use nickel::{Nickel, HttpRouter};
 
 use clap::{App, Arg, SubCommand, AppSettings};
 
+use std::u16;
+use std::str::FromStr;
```
rnewman: https://aturon.github.io/style/imports.html

Prefer

```use std::foo::{Bar, Baz};```

over

```use std::foo::*;```

because wildcard imports can change over time, introducing new name collisions.

Another approach that I'm starting to favor is:

```
use foo::edn;
…

fn bar(keyword: edn::Keyword) -> () {
```

Using explicit qualification can introduce clarity and avoids the decision about precisely what you need to `use`.

Comment (victorporof) [inline src/main.rs]: I also prefer the third approach, feels explicit while also keeping `use` statements readable and easier to maintain over time.

Comment (bgrins) [inline src/main.rs]: OK I will change it to use the third

File: src/main.rs [hunk 3]
```diff
@@ -9,9 +9,15 @@
 // specific language governing permissions and limitations under the License.
 
 extern crate clap;
+#[macro_use] extern crate nickel;
```
bgrins: This lets us use the `middleware` macro which expands out to something like this during compilation: https://github.com/nickel-org/nickel.rs/issues/346#issuecomment-225222662.

src/main.rs
bgrins: Actually after double checking I'm not sure how to make it work in this case.  Without `use std::str::FromStr;` the call to `u16::from_str` fails.  FromStr isn't referenced directly in this file but is needed for that call to compile, so the third option doesn't seem doable here.

PR #163, opened 2017-01-11
Have the EDN parser store keywords and symbols as rich types. Fixes #154.

### LINKED ISSUE #154: Make the EDN parser support the correct tokens for Keyword and Symbol
Useful links:
https://github.com/mozilla/mentat/pull/149#discussion_r94966318
https://github.com/joewalker/mentat/blob/f609f4f60c3845811a74c40288872d66611b42c6/edn/src/edn.rustpeg#L54
https://github.com/edn-format/edn

rnewman:
This demonstrates some unpleasantness with name reuse between enums and structs; the verbosity encourages the use of a test helper function.

Rust also doesn't infer types as much as I'd like, so you can't, say:

```
let v = keyword(":foo/bar");     // Parser function: returns a Value, specifically Value::Keyword
assert_eq!(v.unwrap().0, keyword::Keyword::namespaced("foo", "bar"));
```

because Rust doesn't know which enum value `v` will be.

I *could* 'inline' `keyword::Keyword` into `Value`… but then you can't define functions that work only on keywords; they'd all have to accept a `Value` and handle the other cases.


edn/src/edn.rustpeg
```diff
@@ -80,14 +82,22 @@ symbol -> Value = s:$( symbol_char_initial symbol_char_subsequent* ) {
     Value::Symbol(s.to_string())
 }
 
+namespace_divider = "."
+namespace_separator = "/"
 keyword_char_initial = ":"
+
 // TODO: More chars here?
```
ncalexan: Is this comment still relevant?  `_`, `*`, and `-` seem like good name{space} chars...

edn/src/edn.rustpeg
```diff
@@ -80,14 +82,22 @@ symbol -> Value = s:$( symbol_char_initial symbol_char_subsequent* ) {
...
+keyword_name_char = [a-z] / [A-Z] / [0-9] / "."
+keyword_name = keyword_name_char+
 
 #[export]
-keyword -> Value = k:$( keyword_char_initial keyword_char_subsequent+ ) {
-    Value::Keyword(k.to_string())
-}
+keyword -> Value
+    = keyword_char_initial ns:( kns:$(keyword_namespace) namespace_separator { kns })? n:$(keyword_name) {
+        Value::Keyword(keyword::Keyword::perhaps_namespaced(ns, n))
```
ncalexan: Is this a place where `From`/`Into` can help?

edn/src/types.rs
```diff
@@ -24,7 +26,7 @@ pub enum Value {
     Float(OrderedFloat<f64>),
     Text(String),
     Symbol(String),
-    Keyword(String),
+    Keyword(keyword::Keyword),
```
ncalexan: Yeah, I really think we want `impl From<Into<String>> for keyword::Keyword ...` if that's legal, or `From<String>` if not.  Then we can still construct `Value::Keyword` the easy way.

edn/tests/tests.rs
```diff
@@ -17,9 +17,22 @@ use std::iter::FromIterator;
 use num::bigint::ToBigInt;
 use num::traits::{Zero, One};
 use ordered_float::OrderedFloat;
+use edn::keyword;
+use edn::types::Value;
 use edn::types::Value::*;
 use edn::parse::*;
 
+// Helper for making wrapped keywords with a namespace.
+fn k_ns(ns: &str, name: &str) -> Value {
```
ncalexan: Let's spend some time to learn `Into` and `From`; these are solved problems in Rust-land.

ncalexan: Maybe not worth doing the namespace bit in the parser, but instead doing it in the constructor.  Since you have to check for empty strings anyway, might as well have one place that encodes the rules.

rnewman: We can't use `From`: those conversions cannot fail, and not all strings can be turned into a keyword.

Indeed, even `TryFrom` returns a `Result`, and thus complicates matters.

This is the root of parsing, and why I picked this division: the struct and its constructors aren't failable (they include panicking validation, but that's another matter), whereas the parser itself will refuse to parse a malformed keyword.

edn/src/types.rs
rnewman: Two reasons why not:

- Not all strings are valid keywords.
- Most constructions are namespaced, which makes that approach confusing.

edn/src/edn.rustpeg
rnewman: I decided not to attack that yet. Start narrow and broaden later.

edn/src/types.rs
ncalexan: The latter "namespaced" refers to Rust's namespacing, not EDN's, right?  I think you could make `impl TryFrom<Into<String>> ...` and `Keyword(":foo/bar")` work.

edn/tests/tests.rs
ncalexan: OK, this is fine for now.

edn/src/types.rs
rnewman: No, `namespaced` is the specific constructor function for `:foo/bar` as opposed to `:bar`.

Keywords are essentially in two disjoint sets: those with namespaces and those without.

I modeled this as an `Option<String>` for the namespace, because I didn't want yet another layer of enums. I modeled this as two constructor functions with one helper (`perhaps_namespaced`) to avoid passing `None` in places.

This *is* a spot where you can use some trait tricks to allow coercion from `&str` to `Some<&str>` to `Some<String>` within the same signature, but that would make the constructors relatively more complicated.

edn/src/edn.rustpeg
```diff
@@ -12,9 +12,11 @@
 
 use std::collections::{BTreeSet, BTreeMap, LinkedList};
 use std::iter::FromIterator;
+
```
joewalker: Something I miss from Java was a rough standard of individual imports, sorted alphabetically, inserted by some refactoring and automatically folded by the editor. Minimal code churn which the developer didn't need to think about.
I suspect that rust tooling will catch up shortly. So nothing to do now, mostly thinking aloud.

rnewman: Yeah, I decided to do much what we've been doing in Swift and Java: split std out, sort all alphabetically within that division.

rnewman:
a152e60

PR #164, opened 2017-01-12
Add tx{-parser} crates; start parsing transactions.

ncalexan:
This depends on edn and uses the combine parser combinator library.

@rnewman, you can see my progress here.  There's an abstraction around parsing the internals of vector to be worked out, but it's actually quite nice to work with (once you've absorbed some `combine` magic).  I think some of the constructors mappings might be able to be shortened, but it's not worth waiting for.

### DIFF
--- a/tx-parser/src/lib.rs
+++ b/tx-parser/src/lib.rs
@@ -1,0 +1,378 @@
+// Copyright 2016 Mozilla
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+#![allow(dead_code)]
+
+extern crate edn;
+extern crate combine;
+extern crate mentat_tx;
+
+use combine::{any, eof, many, optional, parser, satisfy_map, token, Parser, ParseResult, Stream};
+use combine::combinator::{Expected, FnParser};
+// TODO: understand why this is self::edn rather than just edn.
+use self::edn::types::Value;
+use mentat_tx::entities::*;
+
+// TODO: implement combine::Positioner on Value.  We can't do this
+// right now because Value is defined in edn and the trait is defined
+// in combine.
+
+// #[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
+// pub struct ValuePosition {
+//     pub position: usize,
+// }
+
+// impl fmt::Display for ValuePosition {
+//     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+//         write!(f, "value position: {}", self.position)
+//     }
+// }
+
+// impl combine::primitives::Positioner for Value {
+//     type Position = ValuePosition;
+//     fn start() -> ValuePosition {
+//         ValuePosition { position: 1 }
+//     }
+//     fn update(&self, position: &mut ValuePosition) {
+//         position.position += 1;
+//     }
+// }
+
+struct Tx<I>(::std::marker::PhantomData<fn(I) -> I>);
+
+type TxParser<O, I> = Expected<FnParser<I, fn(I) -> ParseResult<O, I>>>;
+
+fn fn_parser<O, I>(f: fn(I) -> ParseResult<O, I>, err: &'static str) -> TxParser<O, I>
+    where I: Stream<Item = Value>
+{
+    parser(f).expected(err)
+}
+
+impl<I> Tx<I>
+    where I: Stream<Item = Value>
+{
+    fn integer() -> TxParser<i64, I> {
+        fn_parser(Tx::<I>::integer_, "integer")
+    }
+
+    fn integer_(input: I) -> ParseResult<i64, I> {
+        return satisfy_map(|x: Value| if let Value::Integer(y) = x {
+                Some(y)
+            } else {
+                None
+            })
+            .parse_stream(input);
+    }
+
+    fn keyword() -> TxParser<String, I> {
+        fn_parser(Tx::<I>::keyword_, "keyword")
+    }
+
+    fn keyword_(input: I) -> ParseResult<String, I> {
+        return satisfy_map(|x: Value| if let Value::Keyword(y) = x {
+                Some(y)
+            } else {
+                None
+            })
+            .parse_stream(input);
+    }
+
+    fn entid() -> TxParser<EntId, I> {
+        fn_parser(Tx::<I>::entid_, "entid")
+    }
+
+    fn entid_(input: I) -> ParseResult<EntId, I> {
+        let p = Tx::<I>::integer()
+            .map(|x| EntId::EntId(x))
+            .or(Tx::<I>::keyword().map(|x| EntId::Ident(x)))
+            .parse_lazy(input)
+            .into();
+        return p;
+    }
+
+    fn lookup_ref() -> TxParser<LookupRef, I> {
+        fn_parser(Tx::<I>::lookup_ref_, "lookup-ref")
+    }
+
+    fn lookup_ref_(input: I) -> ParseResult<LookupRef, I> {
+        return satisfy_map(|x: Value| if let Value::Vector(y) = x {
+                let mut p = (Tx::<&[Value]>::entid(), any(), eof())
+                    .map(|(a, v, _)| LookupRef { a: a, v: v });
+                let r = p.parse_lazy(&y[..]).into();
+                match r {
+                    Ok((r, _)) => Some(r),
+                    _ => None,
+                }
+            } else {
+                None
+            })
+            .parse_stream(input);
+    }
+
+    fn entid_or_lookup_ref() -> TxParser<EntIdOrLookupRef, I> {
+        fn_parser(Tx::<I>::entid_or_lookup_ref_, "entid|lookup-ref")
+    }
+
+    fn entid_or_lookup_ref_(input: I) -> ParseResult<EntIdOrLookupRef, I> {
+        let p = Tx::<I>::entid()
+            .map(|x| EntIdOrLookupRef::EntId(x))
+            .or(Tx::<I>::lookup_ref().map(|x| EntIdOrLookupRef::LookupRef(x)))
+            .parse_lazy(input)
+            .into();
+        return p;
+    }
+
+    // TODO: abstract the "match Vector, parse internal stream" pattern to remove this boilerplate.
+    fn add_(input: I) -> ParseResult<Entity, I> {
+        return satisfy_map(|x: Value| -> Option<Entity> {
+                if let Value::Vector(y) = x {
+                    let mut p = (token(Value::Keyword("db/add".into())),
+                                 Tx::<&[Value]>::entid_or_lookup_ref(),
+                                 Tx::<&[Value]>::entid(),
+                                 // TODO: handle lookup-ref.
+                                 any(),
+                                 // TODO: entid or special keyword :db/tx?
+                                 optional(Tx::<&[Value]>::entid()),
+                                 eof())
+                        .map(|(_, e, a, v, tx, _)| {
+                            Entity::Add {
+                                e: e,
+                                a: a,
+                                v: ValueOrLookupRef::Value(v),
+                                tx: tx,
+                            }
+                        });
+                    // TODO: use ok() with a type annotation rather than explicit match.
+                    match p.parse_lazy(&y[..]).into() {
+                        Ok((r, _)) => Some(r),
+                        _ => None,
+                    }
+                } else {
+                    None
+                }
+            })
+            .parse_stream(input);
+    }
+
+    fn add() -> TxParser<Entity, I> {
+        fn_parser(Tx::<I>::add_, "[:db/add e a v tx?]")
+    }
+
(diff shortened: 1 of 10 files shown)


ncalexan:
I don't see these errors with
```
⋊> ~/M/mentat on parse-tx  rustc --version                                                                                                                    16:53:55
rustc 1.16.0-nightly (468227129 2017-01-03)
```
Will try with stable.

rnewman:
Builds for me locally with stable.

rnewman:
Oh, no it doesn't. `mentat_tx` is fine; `mentat_tx_parser` fails.

rnewman:
Fixed branch: https://github.com/mozilla/mentat/tree/rnewman/parse-tx

tx-parser/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+    }
+
+    fn lookup_ref_(input: I) -> ParseResult<LookupRef, I> {
+        return satisfy_map(|x: Value| if let Value::Vector(y) = x {
+                let mut p = (Tx::<&[Value]>::entid(), any(), eof())
+                    .map(|(a, v, _)| LookupRef { a: a, v: v });
+                let r = p.parse_lazy(&y[..]).into();
+                match r {
+                    Ok((r, _)) => Some(r),
+                    _ => None,
```
rnewman: You use this pattern elsewhere. At the risk of sending you down half an hour of wrestling with generics and lifetimes, is there a phrasing of this as:

```
fn result_to_option<T, E>(r: result<T, E>) -> Option<T> {
    match r {
        OK((r, _)) => Some(r),
        _ => None,
    }
}
```

that you can reuse?

tx-parser/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+    }
+
+    fn retract_entity_(input: I) -> ParseResult<Entity, I> {
+        return satisfy_map(|x: Value| -> Option<Entity> {
+                if let Value::Vector(y) = x {
+                    let mut p = (token(Value::Keyword("db/retractEntity".into())),
+                                 Tx::<&[Value]>::entid_or_lookup_ref(),
+                                 eof())
+                        .map(|(_, e, _)| Entity::RetractEntity { e: e });
+                    // TODO: use ok() with a type annotation rather than explicit match.
```
rnewman: This might render my earlier point moot.

tx-parser/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+            .parse_stream(input);
+    }
+
+    fn retract_entity() -> TxParser<Entity, I> {
+        fn_parser(Tx::<I>::retract_entity_, "[:db/retractEntity e]")
+    }
+
+    fn entity_(input: I) -> ParseResult<Entity, I> {
+        let mut p = Tx::<I>::add()
+            .or(Tx::<I>::retract())
```
rnewman: This is really nice.

tx/src/entities.rs
```diff
@@ -0,0 +1,58 @@
...
+
+///! This module defines core types that support the transaction processor.
+
+extern crate edn;
+
+// TODO: understand why this is self::edn rather than just edn.
+use self::edn::types::Value;
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum EntId {
```
rnewman: Could you steal/clean up the `EntId` stuff in the top-level repo?

tx/src/entities.rs
```diff
@@ -0,0 +1,58 @@
...
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+///! This module defines core types that support the transaction processor.
+
+extern crate edn;
+
+// TODO: understand why this is self::edn rather than just edn.
```
rnewman: See earlier note.

Also try `::edn::types::Value`, see how that differs…

tx/src/entities.rs
```diff
@@ -0,0 +1,58 @@
...
+#[derive(Clone, Debug, PartialEq)]
+pub enum EntId {
+    EntId(i64),
+    Ident(String),
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct LookupRef {
+    pub a: EntId,
+    // TODO: consider boxing to allow recursive lookup refs.
```
rnewman: You crazy.

tx/src/entities.rs
```diff
@@ -0,0 +1,58 @@
...
+
+extern crate edn;
+
+// TODO: understand why this is self::edn rather than just edn.
+use self::edn::types::Value;
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum EntId {
+    EntId(i64),
+    Ident(String),
```
rnewman: This'll eventually be `Ident(NamespacedOrPlainKeyword)`, I expect.

tx/src/entities.rs
```diff
@@ -0,0 +1,58 @@
...
+    Value(Value),
+    LookupRef(LookupRef),
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum Entity {
+    Add {
+        e: EntIdOrLookupRef,
+        a: EntId,
+        v: ValueOrLookupRef,
```
rnewman: Strongly typed cat approves of your strongly typed format.

tx-parser/src/lib.rs
palango: [An identical function exists](https://doc.rust-lang.org/std/result/enum.Result.html#method.ok) in the std lib.

ncalexan:
> An identical function exists in the std lib.

Yes, I eventually found it.  Sadly, it requires a type annotation in this case, which I couldn't quite make work.  (There are TODO's later in the commit).  If you can work this out, please do!  Thanks!

tx-parser/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+        return p;
+    }
+
+    fn lookup_ref() -> TxParser<LookupRef, I> {
+        fn_parser(Tx::<I>::lookup_ref_, "lookup-ref")
+    }
+
+    fn lookup_ref_(input: I) -> ParseResult<LookupRef, I> {
+        return satisfy_map(|x: Value| if let Value::Vector(y) = x {
+                let mut p = (Tx::<&[Value]>::entid(), any(), eof())
```
ncalexan: No, and there's a subtle thing happening with consuming `Parser` instances that I don't yet understand.  You'll note all these parsers are mutable; that's because they're consumed by `parse` and friends.  So you _can't_ avoid the invocations and allocations.  It's unclear how this impacts performance overall.

tx-parser/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+                    .map(|(a, v, _)| LookupRef { a: a, v: v });
```
ncalexan: Correct.  A tuple implements `Parser` based on the types of its elements, and the output of the tuple parser is the tupled output of its contained parsers.

tx-parser/src/lib.rs
ncalexan: I eventually figured out how to extract the whole in-a-vector pattern, but it's really hard to express as function composition.  The `combine` pattern (and, more generally, the Rust pattern -- I think) is to extract a `struct` that implements the `Parser` trait.  I know how to do this but it's finicky.  It'll be good to add such a function for `query-parser`, however, so I'll get to it eventually.

tx-parser/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+
+impl<I> Tx<I>
+    where I: Stream<Item = Value>
+{
+    fn integer() -> TxParser<i64, I> {
+        fn_parser(Tx::<I>::integer_, "integer")
+    }
+
+    fn integer_(input: I) -> ParseResult<i64, I> {
+        return satisfy_map(|x: Value| if let Value::Integer(y) = x {
```
ncalexan: `combinator.rs` is tricky; it took me three days of digging to really understand what is happening :)

tx-parser/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+    }
+
+    fn entid() -> TxParser<EntId, I> {
+        fn_parser(Tx::<I>::entid_, "entid")
+    }
+
+    fn entid_(input: I) -> ParseResult<EntId, I> {
+        let p = Tx::<I>::integer()
+            .map(|x| EntId::EntId(x))
+            .or(Tx::<I>::keyword().map(|x| EntId::Ident(x)))
```
ncalexan: Yes, I am unwrapping, although there's no real reason for this.

tx/src/entities.rs
ncalexan: Might just ban `PlainKeyword`, since it's bad form in general.

tx/src/entities.rs
ncalexan: We need a strongly typed cat meme generator, stat!

tx/src/entities.rs
rnewman: Yeah, fair. That's one reason I split, actually: the 'syntax' of the query parser is the main place I expect `Keyword`s (`:find`, `:in`), with all of the 'data' being namespaced.

tx-parser/src/lib.rs
rnewman: Because `FnParser`s (and probably other parsers, too) are implicitly stateful — they wrap a function, and there's no concept in Rust of a pure function?

(Indeed, with 'try' for lookahead, one expects LL(n) parser implementations to be stateful.)

ncalexan:
Worth calling out that the `combine` dependency should be safe: `combine` is MIT and its dependency `ascii` is MIT/Apache and it's dev dependency `byteorder` is Unlicense/MIT.  (As of `combine` version 2.1.1, from https://crates.io/crates/combine and GitHub.)

ncalexan:
@rnewman final stamp before I fold?  I'm going to punt on the entid simplification until we have a motivation to fix it.  There are some decisions about crate organization to be made that don't block this landing.

tx-parser/src/lib.rs
Marwes: > Does the parser library memoize the construction of these parsers? Should I be worried about multiple invocations of fn_parser?

Author of `combine` here. Constructing parsers should be free/very cheap since `Parser` follow the same model as `Iterator` so its all stack allocations and since most parsers are either zero-sized or just a few bytes for the function or parameters they take to construct them. As long as LLVM inlines properly there should be zero overhead.

> (Indeed, with 'try' for lookahead, one expects LL(n) parser implementations to be stateful.)

That is almost true but the `try` parser has no state itself, that is all contained in `Parser::Input` ([try](https://github.com/Marwes/combine/blob/0b263c84c23b953c05259bba1ad886dc84b586fe/src/combinator.rs#L1463), the only thing it contains is the parser it wraps).

tx-parser/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+            .map(|x| EntidOrLookupRef::Entid(x))
+            .or(Tx::<I>::lookup_ref().map(|x| EntidOrLookupRef::LookupRef(x)))
+            .parse_lazy(input)
+            .into();
+        return p;
+    }
+
+    // TODO: abstract the "match Vector, parse internal stream" pattern to remove this boilerplate.
+    fn add_(input: I) -> ParseResult<Entity, I> {
+        return satisfy_map(|x: Value| -> Option<Entity> {
```
Marwes: If I may, I believe you can change this parser into this to preserve errors from parsing the inner value.

```rust
satisfy_map(|x: Value| -> Option<Entity> {
 if let Value::Vector(y) = x { Some(y) } else { None }
}).flat_map(|y| {
    let  mut p = ...;
    p.parse(&y[..]).map(|t| t.0)
})
```
https://docs.rs/combine/2.1.1/combine/trait.Parser.html#method.flat_map

ncalexan:
> Wonder if it's worth consting these defined keyword tokens at the top level.

I think it is, but it's not possible!  You can only construct enums and structs, and the `NamespacedKeyword` members are `String` instances, which you can't construct at compile time.  Frustrating!

ncalexan:
> I think it is, but it's not possible! You can only construct enums and structs, and the NamespacedKeyword members are String instances, which you can't construct at compile time. Frustrating!

@rnewman `static` has the same restriction, so that's out too.

rnewman:
We are blessed with the `lazy_struct` crate…

tx-parser/src/lib.rs
rnewman: If I try this, I get errors like:

```
error[E0271]: type mismatch resolving `<parse::combine::combinator::SatisfyMap<&[error::edn::Value], [closure@query-parser/src/parse.rs:95:20: 95:94]> as parse::combine::Parser>::Input == I`
   --> query-parser/src/parse.rs:101:9
    |
101 |        .parse_stream(input)
    |         ^^^^^^^^^^^^ expected reference, found type parameter
    |
    = note: expected type `&[error::edn::Value]`
    = note:    found type `I`
    = note: required because of the requirements on the impl of `parse::combine::Parser` for `parse::combine::combinator::FlatMap<parse::combine::combinator::SatisfyMap<&[error::edn::Value], [closure@query-parser/src/parse.rs:95:20: 95:94]>, [closure@query-parser/src/parse.rs:96:22: 100:9]>`
```

so this isn't a trivial change.

ncalexan: Yes.  In order to do this, I needed to either use `SliceStream` or work with the `::Range` type, so that the lifetime was preserved.  It can be done, but it's awkward. See also the discussion in https://github.com/Marwes/combine/issues/74#issuecomment-270799586.

rnewman: /me gives up, runs `git checkout -- .`

Issue #170, opened 2017-01-13
Implement v1 and v2 SQL schema in Rust


ncalexan:
This ticket tracks "porting" the existing `sqlite_schema.cljc` code and definitions to Rust.  In particular, handling the SQLite user version and schema upgrade paths to ensure the underlying SQL store is coherent.

This may require defining types for the part table, symbolic schema, ident<->entid maps, but we'll see as we go.

/cc @rnewman @joewalker

rnewman:
Done enough, I think.

PR #171, opened 2017-01-13
Start installing SQL schema.

ncalexan:
WIP in progress branch for #170.


db/src/db.rs
```diff
@@ -0,0 +1,254 @@
...
+   r#"CREATE TABLE idents (ident TEXT NOT NULL PRIMARY KEY, entid INTEGER UNIQUE NOT NULL)"#,
+   r#"CREATE TABLE schema (ident TEXT NOT NULL, attr TEXT NOT NULL, value BLOB NOT NULL, value_type_tag SMALLINT NOT NULL,
+        FOREIGN KEY (ident) REFERENCES idents (ident))"#,
+   r#"CREATE INDEX idx_schema_unique ON schema (ident, attr, value, value_type_tag)"#,
+   r#"CREATE TABLE parts (part TEXT NOT NULL PRIMARY KEY, start INTEGER NOT NULL, idx INTEGER NOT NULL)"#,
+   ];
+
+const V2_STATEMENTS: [&'static str; 19] = V1_STATEMENTS;
+
+fn set_user_version(conn: &rusqlite::Connection, version: i32) -> Result<i32> {
```
rnewman: Unsigned int?

db/src/db.rs
```diff
@@ -0,0 +1,254 @@
...
+    conn.execute(&format!("PRAGMA user_version = {}", version), &[])
+}
+
+fn get_user_version(conn: &rusqlite::Connection) -> Result<i32> {
```
rnewman: Same.

ncalexan:
@rnewman, @joewalker perhaps you two could be the official reviewers for this patch bomb.

@victorporof, @jsantell if one or both of you would care to review this, please steal.

Sorry for the giant code bomb, but it's not worth slicing and dicing this further.  There's lots of TODOs and unimplemented pieces, but I want to get the types and database reading code out there, and circulate some patterns around using `error_chain` and `rusqlite`.  We can fill things in and improve things as we iterate.

rnewman:
I'll take a look tomorrow and Monday. I'm keen to get bits on disk!

db/src/db.rs
```diff
@@ -0,0 +1,541 @@
...
+
+    for statement in &V1_STATEMENTS {
+        tx.execute(statement, &[])?;
+    }
+
+    for statement in &V2_STATEMENTS {
+        try!(tx.execute(statement, &[]));
+    }
+
+    let bootstrap_partition_map = bootstrap::bootstrap_partition_map();
```
rnewman: Indeed, this is the v2 partition map.

db/src/db.rs
```diff
@@ -0,0 +1,541 @@
...
+        let tx = 1;
+        let r: Vec<Result<()>> = entities.into_iter().map(|entity: &Entity| -> Result<()> {
+            match *entity {
+                Entity::Add {
+                    e: entmod::EntidOrLookupRef::Entid(entmod::Entid::Ident(ref e_)),
+                    a: entmod::Entid::Ident(ref a_),
+                    v: entmod::ValueOrLookupRef::Value(ref v_),
+                    tx: _ } => {
+
+                    let mut stmt: rusqlite::Statement = try!(conn.prepare("INSERT INTO datoms(e, a, v, tx, value_type_tag, index_avet, index_vaet, index_fulltext, unique_value) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)"));
```
rnewman: TODO: do this `prepare` in an enclosing scope, not inside the `map` closure.

db/src/schema.rs
```diff
@@ -0,0 +1,148 @@
...
+            },
+            _ => Err(ErrorKind::BadSchemaAssertion(format!("Expected [...], got: {:?}", assertions)))
+        }?;
+
+        let mut schema_map = SchemaMap::new();
+        for (ident, attr, value) in triples {
+            let entid: &i64 = ident_map.get(&ident).ok_or(ErrorKind::BadSchemaAssertion(format!("Could not get ")))?;
+            let attributes = schema_map.entry(*entid).or_insert(Attribute::default());
+
+            // Yes, this is pretty bonkers.  Suggestions appreciated.
```
rnewman: If the value is an integer (which it is), then you can look it up in a fixed-size array with bounds checking. That is:

```
ident                           entid
------------------------------  ----------
:db.type/ref                    23
:db.type/keyword                24
:db.type/long                   25
:db.type/double                 26
:db.type/string                 27
:db.type/boolean                28
:db.type/instant                29
:db.type/bytes                  30
```

so an array:

```
[values::DB_TYPE_REF, values::DB_TYPE_KEYWORD, …]
```

which you dereference with `(*value - 23)` should do the trick…

db/src/schema.rs
```diff
@@ -0,0 +1,148 @@
...
+                    } else if *value == *values::DB_TYPE_URI {
+                        attributes.value_type = ValueType::URI;
+                    } else if *value == *values::DB_TYPE_KEYWORD {
+                        attributes.value_type = ValueType::Keyword;
+                    } else {
+                        bail!(ErrorKind::BadSchemaAssertion(format!("Expected [... :db/valueType :db.type/*] but got [... :db/valueType {:?}]", value)))
+                    }
+                },
+                ":db/cardinality" => {
+                    if *value == *values::DB_CARDINALITY_MANY {
```
rnewman: Same here.

```
ident                           entid
------------------------------  ----------
:db.cardinality/one             31
:db.cardinality/many            32
```

so two-element array, subtract 31 from the value.

db/src/types.rs
```diff
@@ -0,0 +1,251 @@
...
+// specific language governing permissions and limitations under the License.
+
+#![allow(dead_code)]
+
+use std::collections::{BTreeMap};
+
+/// Core types defining a Mentat knowledge base.
+///
+/// At its core, Mentat maintains a set of assertions of the form entity-attribute-value (EAV).  The
+/// assertions conform to a schema where-by the given attribute constrains the associated value/set
```
rnewman: s/where-by/whereby

db/src/types.rs
```diff
@@ -0,0 +1,251 @@
...
+/// of associated values.
+///
+/// ## Assertions
+///
+/// Mentat assertions are represented as rows in the `datoms` SQLite table, and each Mentat row
+/// representing an assertion are tagged with a numeric representation of :db/valueType.
```
rnewman: s/are/is

db/src/types.rs
```diff
@@ -0,0 +1,251 @@
...
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+#![allow(dead_code)]
+
+use std::collections::{BTreeMap};
+
+/// Core types defining a Mentat knowledge base.
```
rnewman: Much as I love in-tree docs, this might do nicely as a sibling of [Querying](https://github.com/mozilla/mentat/wiki/Querying), with a pointer left here. More discoverable, more ability to add tables and images…

db/src/types.rs
```diff
@@ -0,0 +1,251 @@
...
+/// order to ensure that certain index walks of related entities are efficient.  Mentat follows
+/// suit, partitioning into the following partitions:
+/// * `:db.part/db`, for core knowledge base entities;
+/// * `:db.part/user`, for user-defined entities;
+/// * `:db.part/tx`, for transaction entities.
+/// You almost certainly want to add new entities in the `:db.part/user` partition.
+///
+/// The entid sequence in a given partition is monotonically increasing, although not necessarily
+/// contiguous.  That is, it is possible for a specific entid to have never been present in the
+/// system, even though its predecessor and successor is present.
```
rnewman: s/is/are

db/src/types.rs
```diff
@@ -0,0 +1,251 @@
...
+// }
+
+/// Represents one entid in the entid space.
+///
+/// Per https://www.sqlite.org/datatype3.html (see also http://stackoverflow.com/a/8499544), SQLite
+/// stores signed integers up to 64 bits in size.  Since u32 is not appropriate for our use case, we
+/// use i64 rather than manually truncating u64 to u63 and casting to i64 throughout the codebase.
+pub type Entid = i64;
+
+/// The attribute of each Mentat assertion has an :db/valueType constraining the value to a
```
rnewman: s/an/a

db/src/types.rs
```diff
@@ -0,0 +1,251 @@
...
+/// particular domain.  Mentat recognizes the following :db/valueType values.
```
rnewman: I wouldn't use 'domain' here: attributes have a domain (always ref) and a range (which is what valueType defines). Perhaps "value space" instead.

db/src/types.rs
```diff
@@ -0,0 +1,251 @@
...
+            ValueType::Keyword => 13,
+        }
+    }
+
+    pub fn from_value_type_tag(value_type_tag: &i32) -> Option<ValueType> {
+        match *value_type_tag {
+            0 => Some(ValueType::Ref),
+            1 => Some(ValueType::Boolean),
+            4 => Some(ValueType::Instant),
+            5 => Some(ValueType::Long), // SQLite distinguishes integral from decimal types, allowing long and double to share a tag.
```
rnewman: You might consider avoiding this unidirectional and impossible conversion, which requires lying here and checking elsewhere.

That is: you can turn `(SQLite value, tag)` into a `ValueType`; you can turn a `ValueType` into a `tag`; you can turn a `tag` into a *set* of `ValueType`s; you cannot turn a `tag` into a single `ValueType`.

You could define a predicate, `uniquely_identifies_value_type`, if you wish.

This becomes more important when we get to querying.

db/src/types.rs
```diff
@@ -0,0 +1,251 @@
...
+            _ => None
+        }
+    }
+}
+
+/// Represents one partition of the entid space.
+#[derive(Clone,Debug,Eq,Hash,Ord,PartialOrd,PartialEq)]
+pub struct Partition {
+    /// The first entid in the partition.
+    pub start: i64,
```
rnewman: Maybe use a tuple type here. `pub struct Partition(i64, i64)`. Shorter syntax.

db/src/types.rs
```diff
@@ -0,0 +1,251 @@
...
+    /// `true` if this attribute is multi-valued, i.e., it is `:db/cardinality
+    /// :db.cardinality/many`.  `false` if this attribute is single-valued (the default), i.e., it
+    /// is `:db/cardinality :db.cardinality/one`.
+    pub multival: bool,
+    /// `true` if this attribute is unique-value, i.e., it is `:db/unique :db.unique/value`.
+    ///
+    /// *Unique-value* means that there is at most one assertion with the attribute and a
+    /// particular value in the datom store.
+    pub unique_value: bool,
+    /// `true` if this attribute is unique-identity, i.e., it is `:db/unique :db.unique/identity`.
```
rnewman: I'd prefer newlines between code and doc blocks.

db/src/types.rs
```diff
@@ -0,0 +1,251 @@
...
+/// TODO: consider packing this into a bitfield or similar.
+#[derive(Clone,Debug,Eq,Hash,Ord,PartialOrd,PartialEq)]
+pub struct Attribute {
+    /// The associated value type, i.e., `:db/valueType`?
+    pub value_type: ValueType,
+    /// `true` if this attribute is multi-valued, i.e., it is `:db/cardinality
+    /// :db.cardinality/many`.  `false` if this attribute is single-valued (the default), i.e., it
+    /// is `:db/cardinality :db.cardinality/one`.
+    pub multival: bool,
+    /// `true` if this attribute is unique-value, i.e., it is `:db/unique :db.unique/value`.
```
rnewman: Can you add an `is_valid` method to `Attribute`?

```
assert!(!self.multival && self.unique_value));
if self.unique_identity {
  assert!(self.value_type == ValueType::Ref);
  …
```

etc.

(Code won't compile, but you get my drift.)

rnewman: This is what John Regehr calls `checkRep` in

http://blog.regehr.org/archives/1091

rnewman: Then you can call `check_rep` from inside the struct `::new` method I suggest you define, ensuring that only valid attributes can be represented without the system panicking.

db/src/types.rs
```diff
@@ -0,0 +1,251 @@
...
+            multival: false,
+            unique_value: false,
+            unique_identity: false,
+            component: false,
+        }
+    }
+}
+
+/// Map `String` idents (`:db/ident`) to positive integer entids (`1`).
+pub type IdentMap = BTreeMap<String, Entid>;
```
rnewman: Do we care about these being ordered?

If not, I think you should use `HashMap`, per <https://doc.rust-lang.org/std/collections/index.html#when-should-you-use-which-collection>

db/src/types.rs
```diff
@@ -0,0 +1,251 @@
...
+
+/// Map positive integer entids (`1`) to `String` idents (`:db/ident`).
```
rnewman: There's a better representation for this, too — a sparse array map — but that's fine to leave as a TODO!

db/src/types.rs
```diff
@@ -0,0 +1,251 @@
...
+        Schema {
+            ident_map: ident_map,
+            entid_map: entid_map,
+            schema_map: schema_map,
+        }
+    }
+}
+
+/// Represents the metadata required to query from, or apply transactions to, a Mentat store.
+#[derive(Clone,Debug,Default,Eq,Hash,Ord,PartialOrd,PartialEq)]
```
rnewman: Perhaps leave a pointer to <https://github.com/mozilla/mentat/wiki/Thoughts:-modeling-db-conn-in-Rust> here, because eventually people will wonder why `DB` looks the way it does…

rnewman:
Three things I really care about:

- Whether we're doing the wrong thing with having a linearly increasing number of clones of simple things hanging around in memory while we process input. This doesn't seem right.
- V2 vs V1. I'd be fine seeing only V2 and the very straightforward upgrade from V1 to V2, which you can copy out of `master`. Simpler.
- Seeing if we can get the valueType conversion slightly more correct from the start. I'm a week or four away from starting to hit this, and I can see myself getting mired down if I have to juggle more than just my own code. (I'm going very slowly in Rust so far.)

Excellent work, and lots of painful edges apparently hit! The more knowledge we can pull out and share from this, the better.

rnewman:
After a little reading and thinking, I think I can summarize my position on lots of cloning.

We have five main scopes:

1. `static`. We have all of our default keywords here as `lazy_static`.
2. Within a call to `q`.
3. Within a call to `transact`.
4. The scope that receives query results. This is larger than 2.
5. The scope that receives a `tx-report`. This is larger than 3.

Now, we typically parse our query input from strings (creating new instances with new `String`s), or we're given programmatic input from scope 4 or scope 5.

That makes me think that within 2 and 3 — that is, when transacting or querying — we should be able to exclusively work with refs to keywords and `str`s. The enclosing scope, or the parsed input, provides the real data, and we only have to work with references all the way down to the database.

(Even if we didn't create new `String`s, we *could* require that the string input sticks around so we could refer to bits of it, even within the parser… but that doesn't work so well with a work queue and arbitrary threads and prepared statements. So let's say that the *query* or the transact input stick around until we're done pushing them into the database. Parsing results in a new, owned structure.)

The way back out — query results and the tx-report — might be different. The values will come from the database, and we don't want to impose on callers in 4 and 5 the requirement to clone everything when some of the values will be new and originally owned.

To fix that we probably need some kind of interning/shared data/Rc. Perhaps a query result or a `tx-report` is automatically interning, using `Rc` for all keywords…

Thoughts?

rnewman:
OBTW, I found that <http://xion.io/post/code/rust-borrowchk-tricks.html> clarified my understanding of this set of tradeoffs still further.

db/src/db.rs
ncalexan: This is a good idea, so I've filed https://github.com/mozilla/mentat/issues/190 to track it.

victorporof:
I missed the comment ping, I'll go through this as well, but mostly hoping to learn stuff rather than offer any valuable criticism.

db/src/bootstrap.rs
```diff
@@ -0,0 +1,210 @@
...
+ :db/isComponent       {:db/valueType   :db.type/boolean
+                        :db/cardinality :db.cardinality/one}
+ :db/index             {:db/valueType   :db.type/boolean
+                        :db/cardinality :db.cardinality/one}
+ :db/fulltext          {:db/valueType   :db.type/boolean
+                        :db/cardinality :db.cardinality/one}
+ :db/noHistory         {:db/valueType   :db.type/boolean
+                        :db/cardinality :db.cardinality/one}}"#;
+        edn::parse::value(s)
+            .map_err(|_| ErrorKind::BadBootstrapDefinition("Unable to parse V1_SYMBOLIC_SCHEMA".into()))
```
ncalexan: I imagine it must have both a runtime and a code-size penalty, since there are instructions involved.

However, Rust in general (and `unwrap` in particular) provides almost no backtrace help, so I think it's worth including until we prove otherwise.

db/src/bootstrap.rs
```diff
@@ -0,0 +1,210 @@
...
+            }
+        },
+        _ => bail!(ErrorKind::BadBootstrapDefinition("Expected {...}".into()))
+    }
+    Ok(assertions)
+}
+
+pub fn bootstrap_partition_map() -> PartitionMap {
+    V1_PARTS[..].into_iter()
+        .chain(V2_PARTS[..].into_iter())
```
ncalexan: Sure, I'll duplicate this.  I had it that way originally, and did this to be similar to what happened elsewhere; I'll instead decomplect `V1` from `V2` in the code, and make the variables encode the extensions.  (Assuming I can figure out the correct expression of that idea.)

db/src/db.rs
```diff
@@ -0,0 +1,541 @@
...
+   // Materialized views of the schema.
+   r#"CREATE TABLE idents (ident TEXT NOT NULL PRIMARY KEY, entid INTEGER UNIQUE NOT NULL)"#,
+   r#"CREATE TABLE schema (ident TEXT NOT NULL, attr TEXT NOT NULL, value BLOB NOT NULL, value_type_tag SMALLINT NOT NULL,
+        FOREIGN KEY (ident) REFERENCES idents (ident))"#,
+   r#"CREATE INDEX idx_schema_unique ON schema (ident, attr, value, value_type_tag)"#,
+   r#"CREATE TABLE parts (part TEXT NOT NULL PRIMARY KEY, start INTEGER NOT NULL, idx INTEGER NOT NULL)"#,
+   ];
+
+/// Additional SQL statements to be executed, in order, to create the Mentat SQL schema (version 2).
+/// We assume that the `V1_STATEMENTS` have been successfully executed (in order) before these are
```
ncalexan: I've done this.  I haven't (yet) implemented upgrading a V1 to a V2 DB.

db/src/db.rs
```diff
@@ -0,0 +1,541 @@
...
+    })
+        .chain_err(|| "Could not get_user_version")
+}
+
+// TODO: rename "SQL" functions to align with "datoms" functions.
+pub fn create_current_version(conn: &mut rusqlite::Connection) -> Result<i32> {
+    let tx = conn.transaction()?;
+
+    for statement in &V1_STATEMENTS {
+        tx.execute(statement, &[])?;
```
ncalexan: I learnt about `?` part way through writing this, and missed some `try!` invocations.  I'll use `?` consistently.

db/src/db.rs
ncalexan: Aye: added comment, filed https://github.com/mozilla/mentat/issues/191.

db/src/schema.rs
ncalexan: This seems like a level of optimization that I'm not eager to do.

db/src/types.rs
ncalexan: Mmm, I was thinking of "domain of discourse".  Will reword.

db/src/types.rs
ncalexan: I started with tuples, but I find the names helpful.  I'm not really against them.

db/src/types.rs
ncalexan: Me too, but Rust standard seems split on this.  I've added the newlines.

db/src/types.rs
ncalexan: Done.

db/src/types.rs
ncalexan: I decided to use `BTreeMap` simply because we're using it elsewhere (in `edn`).  Let's handle this in follow-up: https://github.com/mozilla/mentat/issues/192

ncalexan:
> Three things I really care about:
> 
>     Whether we're doing the wrong thing with having a linearly increasing number of clones of simple things hanging around in memory while we process input. This doesn't seem right.

The paths that clone are only in the bootstrap code path, which is both bounded size and infrequently hit.  The hot-path through the transactor isn't really present in this patch (and will require additional work to not clone).  We'll handle this as I bone out the approach.

There's definitely a good ticket here to not go via EDN for bootstrapping at all -- I did that simply 'cuz I got tired of trying to define static structures in Rust's verbose language.  I've filed https://github.com/mozilla/mentat/issues/193 to do better.

>     V2 vs V1. I'd be fine seeing only V2 and the very straightforward upgrade from V1 to V2, which you can copy out of master. Simpler.

I've dropped V1 support entirely, and filed https://github.com/mozilla/mentat/issues/194 to fill it in.

>     Seeing if we can get the valueType conversion slightly more correct from the start. I'm a week or four away from starting to hit this, and I can see myself getting mired down if I have to juggle more than just my own code. (I'm going very slowly in Rust so far.)

Your point is well taken.  We have an injection `(ValueType, edn::Value) -> (i32, rusqlite::Value)`, but we cannot factor further in the first component.  Let me see what the encoding feel of that feels like.

> Excellent work, and lots of painful edges apparently hit! The more knowledge we can pull out and share from this, the better.

Thank you.

db/src/types.rs
ncalexan: There needs to be a lazy approach here, since one can specify `:db/fulltext true` _before_ `:db/valueType :db.type/string`.  I can think of lots of ways of doing this, but few that are better than manually doing `check_rep` when we create the `Schema` from the `SchemaMap`.  One possibility would be to encapsulate all the changes to the `Attribute` into a builder pattern, sort the builders before application, and then do incremental changes.  It's a lot of work for a little pay off.

For this ticket, I'll type-check on `Schema` construction, and if we need to do more later we can do it.

ncalexan:
> Your point is well taken. We have an injection (ValueType, edn::Value) -> (i32, rusqlite::Value), but we cannot factor further in the first component. Let me see what the encoding feel of that feels like.

I started to dig further into this, and have opened https://github.com/jgallagher/rusqlite/issues/211 to try to understand our situation more clearly.  I know how to address our case without changing `rusqlite`, but it may just be a library oversight that we should try to upstream, in which case I'll put the effort there.

db/src/db.rs
```diff
@@ -0,0 +1,541 @@
...
+        .map(|_| ())
+}
+
+/// Get the SQLite user version.
+///
+/// Mentat manages its own SQL schema version using the user version.  See the [SQLite
+/// documentation](https://www.sqlite.org/pragma.html#pragma_user_version).
+fn get_user_version(conn: &rusqlite::Connection) -> Result<i32> {
+    conn.query_row("PRAGMA user_version", &[], |row| {
+        row.get(0)
```
jgallagher: FWIW I _just_ fixed this in rusqlite ([PR](https://github.com/jgallagher/rusqlite/pull/210)); the `get` here will now panic instead of silently truncating. `get_checked` is also available if you want a `Result` instead.

db/src/bootstrap.rs
jgallagher: Would [expect](https://doc.rust-lang.org/std/result/enum.Result.html#method.expect) be the right thing here? It's basically "unwrap with a better error message".

db/src/db.rs
```diff
@@ -0,0 +1,532 @@
...
+        tx.execute("INSERT INTO parts VALUES (?, ?, ?)", &[part, &partition.start, &partition.index])?;
+    }
+
+    let bootstrap_db = DB::new(bootstrap_partition_map, bootstrap::bootstrap_schema());
+    bootstrap_db.transact_internal(&tx, &bootstrap::bootstrap_entities()[..])?;
+
+    set_user_version(&tx, CURRENT_VERSION)?;
+    let user_version = get_user_version(&tx)?;
+
+    // TODO: use the drop semantics to do this automagically?
```
jgallagher: I wouldn't recommend this (i.e., I'd keep it explicit as written). By default dropping `tx` would attempt to perform a rollback. You could change that with `set_drop_behavior`, but then you don't get to see if the commit succeeded or failed.

db/src/schema.rs
jgallagher: Could you rewrite this as another `match`? Untested, but something like

```rust
match attr.as_str() {
    ":db/valueType" => {
        attributes.value_type = match *value {
            *values::DB_TYPE_REF => ValueType::Ref,
            *values::DB_TYPE_BOOLEAN => ValueType::Boolean,
            ...
        };
    },
```

It's not substantively different but is nicer to look at.

db/src/db.rs
ncalexan: Thanks for the guidance.  I'll drop the TODO instead :)

db/src/bootstrap.rs
ncalexan: I think so, particularly in the case of `Option<T> -> Result<T>`, which I do (awkwardly) a lot.  Much appreciated.

ncalexan:
> FWIW I just fixed this in rusqlite (PR); the get here will now panic instead of silently truncating. get_checked is also available if you want a Result instead.

In general, I have been preferring `get_checked(...)?` to avoid the "bare panic" (but not specifically to avoid truncating).  I read the code and must have mis-interpreted; I thought they would panic rather than truncate.

ncalexan:
@rnewman OK, this is ready for another pass.  I think the introduction of `TypedValue` really improved the code.

I realize now that I didn't move the documentation notes to the Wiki; I'll do that either before or just after landing.

@victorporof @jsantell  Let me know what y'all think!  There's lots of tests waiting to be written for this work; I'll file a few tickets to track areas that need love later today.

db/src/db.rs
```diff
@@ -285,60 +283,62 @@ pub fn ensure_current_version(conn: &mut rusqlite::Connection) -> Result<i32> {
...
-         })
-}
+impl TypedValue {
+    /// Given a SQLite `value` and a `value_type_tag`, return the corresponding `TypedValue`.
+    pub fn from_sql_value_pair(value: &rusqlite::types::Value, value_type_tag: &i32) -> Result<TypedValue> {
+        match (*value_type_tag, value) {
+            (0, &rusqlite::types::Value::Integer(ref x)) => Ok(TypedValue::Ref(*x)),
+            (1, &rusqlite::types::Value::Integer(ref x)) => Ok(TypedValue::Boolean(0 != *x)),
+            // SQLite distinguishes integral from decimal types, allowing long and double to
+            // share a tag.
```
rnewman: TODO: 4 = date, 11 = UUID.

db/src/db.rs
```diff
@@ -285,60 +283,62 @@ pub fn ensure_current_version(conn: &mut rusqlite::Connection) -> Result<i32> {
...
+    /// EDN values which are not Mentat typed values.
+    ///
+    /// This function is deterministic.
+    pub fn from_edn_value(value: &Value) -> Option<TypedValue> {
+        match value {
+            &Value::Boolean(x) => Some(TypedValue::Boolean(x)),
+            &Value::Integer(x) => Some(TypedValue::Long(x)),
+            &Value::Float(ref x) => Some(TypedValue::Double(x.clone())),
+            &Value::Text(ref x) => Some(TypedValue::String(x.clone())),
+            &Value::NamespacedKeyword(ref x) => Some(TypedValue::Keyword(x.to_string())),
```
rnewman: I think `TypedValue::Keyword` should contain a `Keyword`, preserving the namespace/name distinction and avoiding consing up a new string. This might mean you need to extract a `keyword` crate out of `edn` for reuse. Please file an issue.

db/src/db.rs
```diff
@@ -285,60 +283,62 @@ pub fn ensure_current_version(conn: &mut rusqlite::Connection) -> Result<i32> {
...
+    /// Return the corresponding SQLite `value` and `value_type_tag` pair.
+    pub fn to_sql_value_pair<'a>(&'a self) -> (ToSqlOutput<'a>, i32) {
+        match self {
+            &TypedValue::Ref(x) => (rusqlite::types::Value::Integer(x).into(), 0),
+            &TypedValue::Boolean(x) => (rusqlite::types::Value::Integer(if x { 1 } else { 0 }).into(), 1),
+            // SQLite distinguishes integral from decimal types, allowing long and double to share a tag.
+            &TypedValue::Long(x) => (rusqlite::types::Value::Integer(x).into(), 5),
+            &TypedValue::Double(x) => (rusqlite::types::Value::Real(x.into_inner()).into(), 5),
+            &TypedValue::String(ref x) => (rusqlite::types::ValueRef::Text(x.as_str()).into(), 10),
+            &TypedValue::Keyword(ref x) => (rusqlite::types::ValueRef::Text(x.as_str()).into(), 13),
```
rnewman: This is the point at which we'd produce a `str` directly from a `Keyword`.

db/src/entids.rs
```diff
@@ -0,0 +1,58 @@
...
+pub const DB_INDEX: Entid = 11;
+pub const DB_FULLTEXT: Entid = 12;
+pub const DB_NO_HISTORY: Entid = 13;
+pub const DB_ADD: Entid = 14;
+pub const DB_RETRACT: Entid = 15;
+pub const DB_PART_USER: Entid = 16;
+pub const DB_PART_TX: Entid = 17;
+pub const DB_EXCISE: Entid = 18;
+pub const DB_EXCISE_ATTRS: Entid = 19;
+pub const DB_EXCISE_BEFORET: Entid = 20;
```
rnewman: I think this should be `DB_EXCISE_BEFORE_T` for clarity.

db/src/db.rs
ncalexan: https://github.com/mozilla/mentat/issues/201

db/src/db.rs
```diff
@@ -285,60 +283,62 @@ pub fn ensure_current_version(conn: &mut rusqlite::Connection) -> Result<i32> {
...
+    /// Given a SQLite `value` and a `value_type_tag`, return the corresponding `TypedValue`.
+    pub fn from_sql_value_pair(value: &rusqlite::types::Value, value_type_tag: &i32) -> Result<TypedValue> {
+        match (*value_type_tag, value) {
+            (0, &rusqlite::types::Value::Integer(ref x)) => Ok(TypedValue::Ref(*x)),
+            (1, &rusqlite::types::Value::Integer(ref x)) => Ok(TypedValue::Boolean(0 != *x)),
+            // SQLite distinguishes integral from decimal types, allowing long and double to
+            // share a tag.
+            (5, &rusqlite::types::Value::Integer(ref x)) => Ok(TypedValue::Long(*x)),
+            (5, &rusqlite::types::Value::Real(ref x)) => Ok(TypedValue::Double((*x).into())),
+            (10, &rusqlite::types::Value::Text(ref x)) => Ok(TypedValue::String(x.clone())),
```
ncalexan: Maybe.  I thought about this and I'm okay with `into_sql_value_pair`; let me try it on for size.

ncalexan: Oh, wait, `into` is the wrong direction.

db/Cargo.toml
```diff
@@ -3,6 +3,12 @@ name = "mentat_db"
 version = "0.0.1"
 
 [dependencies]
+error-chain = "0.8.0"
+lazy_static = "0.2.2"
+# TODO: don't depend on num and ordered-float; expose helpers in edn abstracting necessary constructors.
```
ncalexan: Yep, I filed #198 while reviewing your code to track this.

db/src/entids.rs
ncalexan: Agreed, this was an oversight.

db/src/db.rs
ncalexan: I've done this.

db/src/db.rs
ncalexan: By keeping a `String` in the `Keyword`, this doesn't copy data until it hits SQLite.  If we keep a `Keyword` (really a `NamespacedKeyword`, since I'm being strict in my parser for now) then we have to cons strings and can't use the `ValueRef` in this way.

db/src/bootstrap.rs
```diff
@@ -150,6 +151,57 @@ fn idents_to_assertions(idents: &[(&str, i64)]) -> Vec<Value> {
...
+                            // We have symbolic idents but the transactor handles entids.  Ad-hoc
+                            // convert right here.  This is a fundamental limitation on the
+                            // bootstrap symbolic schema format; we can't represent "real" keywords
+                            // at this time.
+                            //
+                            // TODO: remove this limitation, perhaps by including a type tag in the
+                            // bootstrap symbolic schema, or by representing the initial bootstrap
+                            // schema directly as Rust data.
+                            let typed_value = match TypedValue::from_edn_value(value) {
+                                Some(TypedValue::Keyword(ref s)) => TypedValue::Ref(*ident_map.get(s).ok_or(ErrorKind::UnrecognizedIdent(s.clone()))?),
```
ncalexan: I think it could be made neater, but I want to keep the error handling I've handled.  Next time!

db/src/db.rs
ncalexan: I think there will be a grand reckoning where we do a lot of `String` to `&str` rewriting, so I've filed https://github.com/mozilla/mentat/issues/203 to keep this around.

PR #174, opened 2017-01-13
Implement basic logging infrastructure

palango:
A replacement for #158.

I think this is a better approach to logging. Independent libraries can use the `log` crate and should be universally usable with that.

The executable provides three options to set the logging level (in order of priority): The `MENTAT_LOG` environment variable, the `RUST_LOG` environment variable and the debug flag.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,10 +1,12 @@
 [package]
+authors = ["Richard Newman <rnewman@twinql.com>", "Nicholas Alexander <nalexander@mozilla.com>"]
 name = "mentat"
 version = "0.4.0"
-authors = ["Richard Newman <rnewman@twinql.com>", "Nicholas Alexander <nalexander@mozilla.com>"]
 
 [dependencies]
 clap = "2.19.3"
+env_logger = "0.3.5"
+log = "0.3.6"
 nickel = "0.9.0"
 rusqlite = "0.8.0"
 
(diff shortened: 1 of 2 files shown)


palango:
As for the motivation: I'm mostly interested in imroving my Rust skills and learning how a database works. Datomic seems like a smart alternative approach to databases to me, so I see this as a nice way to deepen my knowledge.
Unfortunately I have no prior knowledge to database internals and my SQL is pretty basic as well.

ncalexan:
> As for the motivation: I'm mostly interested in imroving my Rust skills and learning how a database works. Datomic seems like a smart alternative approach to databases to me, so I see this as a nice way to deepen my knowledge.

Cool!  We're all learning Rust for this project, so happy to learn with you.  I try to be pretty careful to call Mentat a "knowledge base" on top of a "database", because we really don't want to do real database work: B-trees, indexes, page caches, disk optimization, failure modes... that's for SQLite to do.

> Unfortunately I have no prior knowledge to database internals and my SQL is pretty basic as well.

Lots to learn, then!  Welcome aboard!

ncalexan:
@palango sorry that this has sat for so long.  I was looking at https://github.com/slog-rs/example-lib/blob/master/lib.rs and see an explicit comparison to vanilla `log`.  Can you explain why using vanilla `log` would be better for our crate configuration than using `slog`?  I'm having a hard time understanding the differences here.  Overall it seems like `slog` is the more robust library and will work well with our setup.

ncalexan:
Landed the original `slog` version in #205. Thanks, @palango! We'll try to be more responsive in future.

PR #176, opened 2017-01-17
edn: allow comments and improve whitespace handling

### LINKED ISSUE #166: Test EDN parser against leading, trailing, and inter-element whitespace
I saw issues parsing ` [1]`, `[1] `, and `[ [1]]`.  These are simplified and from memory, so they may not all fail.  For what it's worth I was using `r#` raw strings, although I don't think that's relevant.

ncalexan:
@rnewman, @joewalker: review?

Fixes #166.

### DIFF
--- a/edn/src/edn.rustpeg
+++ b/edn/src/edn.rustpeg
@@ -25,7 +25,6 @@ use types::Value;
 // to trace where the parser is failing
 
 // TODO: Support tagged elements
-// TODO: Support comments
 // TODO: Support discard
 
 #[export]
@@ -137,5 +136,6 @@ value -> Value
       list / vector / map / set
 
 whitespace = (" " / "\r" / "\n" / "\t")
+comment = ";" [^\r\n]* ("\r" / "\n")?
 
-__ = whitespace*
+__ = (whitespace / comment)*
--- a/edn/src/edn.rustpeg
+++ b/edn/src/edn.rustpeg
@@ -104,26 +104,26 @@ keyword -> Value
     }
 
 #[export]
-list -> Value = "(" __ v:(__ value)* __ ")" {
+list -> Value = "(" v:(value)* ")" {
     Value::List(LinkedList::from_iter(v))
 }
 
 #[export]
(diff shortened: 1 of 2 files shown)


edn/src/edn.rustpeg
```diff
@@ -105,37 +104,41 @@ keyword -> Value
...
 value -> Value
-    = nil / boolean / float / bigint / integer / text /
+    = __ v:(nil / boolean / float / bigint / integer / text /
       keyword / symbol /
-      list / vector / map / set
+      list / vector / map / set) __ {
+    v
+}
+
+whitespace = (" " / "\r" / "\n" / "\t" / ",")
```
rnewman: It's worth commenting that Clojure, and thus EDN, regard commas as whitespace, and thus this is not a mistake; `[5,,,,4]` is totally legit syntax for a two-element array.

(It's also worth noting that idiomatic EDN doesn't bother with commas.)

edn/src/edn.rustpeg
```diff
@@ -105,37 +104,41 @@ keyword -> Value
...
+map -> Value = "{" v:(pair)* "}" {
     Value::Map(BTreeMap::from_iter(v))
 }
 
 // It's important that float comes before integer or the parser assumes that
 // floats are integers and fails to parse
 #[export]
 value -> Value
-    = nil / boolean / float / bigint / integer / text /
+    = __ v:(nil / boolean / float / bigint / integer / text /
```
joewalker: Yup. Like it much better than the blind alley we were obviously headed down.

edn/src/edn.rustpeg
joewalker: Except presumably for maps, right - you need commas there?

rnewman: No — `{:foo 5 :bar 3}` is totally normal. One would only add commas if they helped clarity, as in `{5 3, 3 2, 2 3}`.

(Swift does the same thing for numbers, and Rust might too — `1_234_567` is acceptable numeric syntax, as is `12_34_56_78` if you live on bizarro planet.)

PR #196, opened 2017-01-24
Implement parsing of simple :find expressions.

rnewman:
This handles `[:find ?x ?y ?z]`, `[:find ?abc .]`, `[:find [?a ?b]]`, and `[:find [?x ...]]`. Lots more to do, but it's a start.


query/src/lib.rs
```diff
@@ -8,4 +8,206 @@
...
+pub fn is_unit_limited(spec: &FindSpec) -> bool {
+    match spec {
+        &FindSpec::FindScalar(..) => true,
+        &FindSpec::FindTuple(..)  => true,
+        &FindSpec::FindRel(..)    => false,
+        &FindSpec::FindColl(..)   => false,
+    }
+}
+
+/// Returns true if the provided `FindSpec` cares about distinct results.
```
ncalexan: Can you elaborate on "cares about" here?  I think you mean that you need to use `SELECT DISTINCT` in the SQL produced, but does that have some better interpretation in terms of the find results?

query/src/lib.rs
```diff
@@ -8,4 +8,206 @@
...
+pub fn requires_distinct(spec: &FindSpec) -> bool {
+    return !is_unit_limited(spec);
+}
+
+// Note that the "implicit blank" rule applies.
+// A pattern with a reversed attribute — :foo/_bar — is reversed
```
ncalexan: Hmm, interesting.  I wonder if I should do the same in the transaction parser.

query/src/lib.rs
```diff
@@ -8,4 +8,206 @@
...
+    Or,
+    OrJoin,
+    Pred,
+    WhereFn,
+    RuleExpr,
+    */
+    Pattern,
+}
+
+pub struct Query {
```
ncalexan: It's so beautiful!

edn/src/symbols.rs
```diff
@@ -69,10 +69,11 @@ pub struct NamespacedKeyword {
 }
 
 impl PlainSymbol {
-    pub fn new(name: &str) -> Self {
```
ncalexan: Consider extending this to the `Namespaced` variants too.  I've filed https://github.com/mozilla/mentat/issues/199 for this.

query-parser/src/util.rs
```diff
@@ -31,15 +29,15 @@ pub fn value_to_variable(v: &edn::Value) -> Option<Variable> {
 
 /// If the provided slice of EDN values are all variables as
 /// defined by `value_to_variable`, return a Vec of Variables.
-/// Otherwise, return an error.
-pub fn values_to_variables(vals: &[edn::Value]) -> Result<Vec<Variable>, FindParseError> {
+/// Otherwise, return the unrecognized Value.
```
ncalexan: Eh?  This is very surprising.  I know you want `Either`, but a `Result` with a non-`Error`-deriving error is not Rust-y at all.  Why not return `Option<Vec<Variable>>` and make callers use `unwrap_or`?

query-parser/src/find.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+fn parse_find_map(map: BTreeMap<edn::Keyword, Vec<edn::Value>>) -> FindParseResult {
+    // Eagerly awaiting `const fn`.
+    let kw_find = edn::Keyword::new("find");
+    let kw_in = edn::Keyword::new("in");
+    let kw_with = edn::Keyword::new("with");
+    let kw_where = edn::Keyword::new("where");
+
+    // Oh, if only we had `guard`.
```
ncalexan: You could do different -- not necessarily better -- with `match (map.get(&kw_find), map.get(&kw_where)) ...`.

query-parser/src/find.rs
```diff
@@ -0,0 +1,134 @@
...
+    let with_vars = with.map(values_to_variables);
+
+    //
+    // :wheres is a whole datastructure.
+
+    Ok(FindSpec::FindRel(vec!()))
+}
+
+fn parse_find_map(map: BTreeMap<edn::Keyword, Vec<edn::Value>>) -> FindParseResult {
+    // Eagerly awaiting `const fn`.
```
ncalexan: Jesus yes.

query-parser/src/find.rs
```diff
@@ -0,0 +1,134 @@
...
+fn parse_find_edn_map(map: BTreeMap<edn::Value, edn::Value>) -> FindParseResult {
+    // Every key must be a Keyword. Every value must be a Vec.
+    let mut m = BTreeMap::new();
+
+    if map.is_empty() {
+        return parse_find_map(m);
+    }
+
+    for (k, v) in map {
+        if let edn::Value::Keyword(kw) = k {
```
ncalexan: Mmm, I kind of like this `edn::Value` style. I will adopt it, eventually, in the transaction stuff.  It plays well with `mentat::Value` and `rusqlite::Value`.

query-parser/src/find.rs
```diff
@@ -0,0 +1,134 @@
...
+            if let edn::Value::Vector(vec) = v {
+                m.insert(kw, vec);
+                continue;
+            } else {
+                return Err(FindParseError::InvalidInput(v));
```
ncalexan: These errors are going to be so hard to debug.  I see two alternatives:

1. Assume that initial consumers will be comfortable inspecting and debugging Mentat itself; iterate on the error messages over time as confusion arises.
2.  Parse (probably using `combine`) the data structures directly, investing in error reporting in the parser combinator layer to help people not hate their lives.

Both are investments.

query-parser/src/parse.rs
```diff
@@ -0,0 +1,272 @@
...
+
+use self::combine::{any, eof, many1, optional, parser, satisfy_map, token, Parser, ParseResult,
+                    Stream};
+use self::combine::combinator::{Expected, FnParser, choice, try};
+use self::edn::Value::PlainSymbol;
+use self::mentat_query::{Element, FindSpec, SrcVar, Variable};
+
+use super::error::{FindParseError, FindParseResult};
+
+pub struct FindSp<I>(::std::marker::PhantomData<fn(I) -> I>);
```
ncalexan: Can we s/FindSp/FindSpec/?  I guess not due to namespacing?  Maybe just a comment that `FindSp` stands for something.

query-parser/src/parse.rs
```diff
@@ -0,0 +1,272 @@
...
+        // operate independently.
+        choice::<[&mut Parser<Input = I, Output = FindSpec>; 4],
+                 _>([&mut try(FindSp::find_scalar()),
+                     &mut try(FindSp::find_coll()),
+                     &mut try(FindSp::find_tuple()),
+                     &mut try(FindSp::find_rel())])
+            .parse_stream(input)
+    }
+
+    fn gimme<T>(r: ParseResult<T, I>) -> Option<T> {
```
ncalexan: Gentle request for a more dignified name -- and a more descriptive one :)

query-parser/src/parse.rs
```diff
@@ -0,0 +1,272 @@
...
+        r.ok().map(|x| x.0)
+    }
+}
+
+macro_rules! assert_parses_to {
```
ncalexan: We should lift this to the root crate or a utility crate.  Follow-up for sure.

query-parser/src/util.rs
```diff
@@ -0,0 +1,183 @@
...
+        }
+    }
+    return None;
+}
+
+/// If the provided slice of EDN values are all variables as
+/// defined by `value_to_variable`, return a Vec of Variables.
+/// Otherwise, return an error.
+pub fn values_to_variables(vals: &[edn::Value]) -> Result<Vec<Variable>, FindParseError> {
+    let mut out: Vec<Variable> = Vec::with_capacity(vals.len());
```
ncalexan: You can do this less explicitly with `map`, `collect`, and `ok_or`.  It's worth internalizing that there is a folding/aborting `collect` implementation for `[Option<T>] -> Option<Vec<T>>` and similarly for `Result`.

query-parser/src/util.rs
```diff
@@ -0,0 +1,183 @@
...
+
+    // Turn something like
+    //
+    //   `[:foo 1 2 3 :bar 4 5 6]`
+    //
+    // into
+    //
+    //   `Some((:foo, [1 2 3]))`
+    fn step(slice: &[edn::Value]) -> Option<(edn::Keyword, Vec<edn::Value>)> {
+        // This can't be right -- we can't handle [:foo 1 2 3 :bar].
```
ncalexan: It's not clear if you mean the code is faulty or the input is incorrect.  I think you mean the latter, but could you clarify the comment?

query-parser/src/util.rs
```diff
@@ -0,0 +1,183 @@
...
+/// `edn::Value::Vector`, and turn it into a map.
+///
+/// The slice must consist of subsequences of an initial plain
+/// keyword, followed by one or more non-plain-keyword values.
+///
+/// The plain keywords are used as keys into the resulting map.
+/// The values are accumulated into vectors.
+///
+/// Invalid input causes this function to return `None`.
+pub fn vec_to_keyword_map(vec: &[edn::Value]) -> Option<BTreeMap<edn::Keyword, Vec<edn::Value>>> {
```
ncalexan: Future enhancement: accept a predicate that defines where the break points are.  It's a map-producing partition, I guess.  It's so common in Clojure we might move it to the `edn` crate.

query-parser/src/util.rs
```diff
@@ -0,0 +1,183 @@
...
+    //
+    //   `Some((:foo, [1 2 3]))`
+    fn step(slice: &[edn::Value]) -> Option<(edn::Keyword, Vec<edn::Value>)> {
+        // This can't be right -- we can't handle [:foo 1 2 3 :bar].
+        if slice.len() < 2 {
+            return None;
+        }
+
+        // The first item must be a keyword.
+        if let edn::Value::Keyword(ref k) = slice[0] {
```
ncalexan: Consider `split_first`, `split_at` rather than explicit indexing.

ncalexan:
This is looking really good.  Overall I find the code a little iterative, which I hope will fall away over time.  (I think one of the nice features of Rust is that it allows for readable functional expressions that are still efficiently compiled.)

The only thing I really want to see changed is that non-`Error` `Result` return type.  Otherwise, bombs away!

edn/src/symbols.rs
rnewman: Aye. I was lazy!

query-parser/src/find.rs
rnewman: The main reason I went this route is that `match` doesn't short-circuit, and this allows us to fail with individual messages.

query-parser/src/parse.rs
rnewman: s/gimme/to_parsed_value.

query-parser/src/parse.rs
rnewman: Filed #200.

query-parser/src/util.rs
rnewman: Done.

query-parser/src/util.rs
rnewman: Comment added.

query-parser/src/util.rs
rnewman: I did; the split functions produce iterators, which makes the code a little more confusing. We've already validated the length above.

(What I really want is slice destructuring…)

query/src/lib.rs
```diff
@@ -8,4 +8,206 @@
...
+
+#[derive(Clone,Debug,Eq,PartialEq)]
+pub enum SrcVar {
+    DefaultSrc,
+    NamedSrc(SrcVarName),
+}
+
+/// These are the scalar values representable in EDN.
+#[derive(Clone,Debug,Eq,PartialEq)]
+pub enum NonIntegerConstant {
```
rnewman: It's not there, for a reason — an integer could be only an entid in some contexts (in the attribute place, for example) or either an entid or an integer in others. Those distinctions are modeled by the choice to use `Entid` or `EntidOrInteger` in conjunction with `NonIntegerConstant`.

There are other ways to slice this up, and I'll probably end up changing this somewhat when I flesh out the rest of the parser and start consuming these things.

query/src/lib.rs
rnewman: Two of the find specs (scalar and tuple) produce only a single result. Those don't need to be run with `SELECT DISTINCT`, and that's this distinction.

There's a more broader definition of whether `DISTINCT` is needed: for example, in some kinds of accumulation or sampling projections we might not need to do it at the SQL level. This gives us the hook to add that.

I use the words "cares about" because `find` is generally defined in terms of producing distinct results at the Datalog level.

rnewman: I'll flesh out the comment.

query/src/lib.rs
rnewman: It's simpler. Match `edn::NamespacedKeyword`, call `.is_reversed()` (once we've written it!), call `.to_reversed` to turn it the right way around.

query-parser/src/util.rs
rnewman: I made this change when I split the find parser and the query parser: the `values_to_variables` function itself doesn't know whether this should return a `FindParseError` or a `QueryParseError`.

There's also precedent in the Rust docs themselves:

> Functions return `Result` whenever errors are expected and recoverable. In the `std` crate, `Result` is most prominently used for I/O.
>
> A simple function returning `Result` might be defined and used like so:
>
> ```#[derive(Debug)]
enum Version { Version1, Version2 }

fn parse_version(header: &[u8]) -> Result<Version, &'static str> {
    match header.get(0) {
        None => Err("invalid header length"),
        Some(&1) => Ok(Version::Version1),
        Some(&2) => Ok(Version::Version2),
        Some(_) => Err("invalid version"),
    }
}```

Right below it they give an example of a `Result<i32, i32>`.

I chose `Result` instead of `Option` because this is an operation that can fail (you might choose to use `try`), and it returns the reason why it failed (the value that wasn't a variable).

PR #205, opened 2017-01-26
Implement basic logging infrastructure.

ncalexan:
Signed-off-by: Paul Lange <palango@gmx.de>

This is a rebased version of @palango's original `slog` logging PR, #158.  I don't know how to take "ownership" and update such a PR, and we (myself and @victorporof) didn't want this to languish forever, so here we are in a new PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,11 +1,14 @@
 [package]
+authors = ["Richard Newman <rnewman@twinql.com>", "Nicholas Alexander <nalexander@mozilla.com>"]
 name = "mentat"
 version = "0.4.0"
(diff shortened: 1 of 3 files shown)


ncalexan:
@palango sorry for the long delay in getting this landed, especially because we deliberated on Slack and elected to go with the original `slog` version:
```
victor
[10:18 AM]  having a proper logger is useful. for one thing, we can actually store the logs and consume them later
[10:20]  env_logger writes to stdout/stderr, which i don’t think we want in the long run
[10:20]  for now it probably doesn’t matter though
[10:21]  slog seems to be the most popular logger for rust at the moment, and has some nice things like bunyan formatting (which we like)
```

Issue #178, opened 2017-01-18
tempids for nested entities not preserved in transaction report


cascadian:
Given the following schema 
``` clj
(def schema
  [{:db/id                 (d/id-literal :db.part/user)
    :db.install/_attribute :db.part/db
    :db/cardinality        :db.cardinality/one
    :db/valueType          :db.type/long
    :db/unique             :db.unique/identity
    :db/ident              :foo/id}
   {:db/id                 (d/id-literal :db.part/user)
    :db.install/_attribute :db.part/db
    :db/cardinality        :db.cardinality/one
    :db/valueType          :db.type/ref
    :db/isComponent        true
    :db/ident              :foo/component}])
```
and this transaction

``` clj
(<? (d/<transact! conn [{:db/id    (d/id-literal :db.part/user -1)
                                                     :foo/id  10
                                                     :foo/component {:db/id (d/id-literal :db.part/user -2)} :foo/id 20}]))
```

The tempids in the transaction report are 

``` clj
{#datomish.db.TempId{:part :db.part/user, :idx -1} 65545, #datomish.db.TempId{:part :db.part/user, :idx -1000066} 65546}
```

I was expecting `:idx -2` instead of `:idx -1000066`

Note that the tempids are reported as expected if I flatten the transaction vector:

``` clj
(<? (d/<transact! conn [{:db/id (d/id-literal :db.part/user -2)
                                                     :foo/id 20}
                                                    {:db/id    (d/id-literal :db.part/user -1)
                                                     :foo/id  10
                                                     :foo/component (d/id-literal :db.part/user -2)}]))
```

ncalexan:
@cascadian hey!  We don't really support `:db/isComponent` yet (we don't implement pull expressions, we don't check for dangling entities, and we don't handle deleting components recursively), but this is very surprising.  Thanks for the report!  We'll make sure to add a test for the Rust version.

cascadian:
@ncalexan Mentat is a really good fit for my app, so I'm looking forward to the Rust version :)  I didn't realize `:db/isComponent` isn't implemented. Thanks for pointing that out. This bug is present for refs that aren't components too.

ncalexan:
Now that I think further, I'm not sure I implemented support for nested maps at all!  So double thanks for reporting this; we'll have to bone out the Rust test suite to these cases, probably as part of #180.

rnewman:
Fixed by that commit.

PR #206, opened 2017-01-26
Accept :db/id in nested maps.

ncalexan:
This should address #178.  What was happening was that we always provide `:db/id` ourselves, rather than optionally providing `:db/id` as required.


ncalexan:
@palango, would you care to review this small patch?  It would be enough for you to test that your case is addressed.

ncalexan:
@rnewman looks like Paul doesn't have time or isn't interested in this.  Might as well get the test merged -- can you review?

PR #208, opened 2017-01-26
Add is_ methods to edn::Value types and add tests. Fixes #197

### LINKED ISSUE #197: Expose is_keyword, is_vector, is_map, etc on edn::Value
Over in #196, @rnewman wants some helper functions to make destructuring `edn::Value` more pleasant.  This ticket tracks adding `is_*` predicates for all the variants, and potentially `as_vector(Value) -> Option<Vec>` and `as_map(Value) -> Option<BTreeMap<Value, Value>`, etc.

We should aim to provide the exact same helper API as [serde json library](https://github.com/serde-rs/json).

jsantell:
Just a preliminary check, is this what @ncalexan @rnewman had in mind? Want to get some eyes on it before starting the `as_*`

### DIFF
--- a/edn/src/types.rs
+++ b/edn/src/types.rs
@@ -87,13 +87,62 @@ fn test_print_edn() {
                Value::Boolean(true))).to_string());
 }
 
-impl Value {
-    pub fn is_keyword(&self) -> bool {
-        match *self {
-            Keyword(_) => true,
-            _          => false,
+/// Creates `is_$TYPE` helper functions for Value, like
+/// `is_big_integer()` or `is_text()`.
+macro_rules! def_is {
+    ($name: ident, $pat: pat) => {
+        pub fn $name(&self) -> bool {
+            match *self { $pat => true, _ => false }
         }
     }
+}
+
+/// Creates `as_$TYPE` helper functions for Value, like `as_big_integer()`,
+/// which returns the underlying value representing this Value wrapped
+/// in an Option, like `<Option<&BigInt>`.
+macro_rules! def_as {
+    ($name: ident, $kind: path, $t: ty) => {
+        pub fn $name(&self) -> Option<&$t> {
+            match *self { $kind(ref v) => Some(v), _ => None }
+        }
+    }
+}
+
+impl Value {
+    def_is!(is_nil, Nil);
+    def_is!(is_boolean, Boolean(_));
+    def_is!(is_integer, Integer(_));
+    def_is!(is_big_integer, BigInteger(_));
+    def_is!(is_float, Float(_));
+    def_is!(is_text, Text(_));
+    def_is!(is_symbol, PlainSymbol(_));
+    def_is!(is_namespaced_symbol, NamespacedSymbol(_));
+    def_is!(is_keyword, Keyword(_));
+    def_is!(is_namespaced_keyword, NamespacedKeyword(_));
+    def_is!(is_vector, Vector(_));
+    def_is!(is_list, List(_));
+    def_is!(is_set, Set(_));
+    def_is!(is_map, Map(_));
+
+    /// `as_nil` does not use the macro as it does not have an underlying
+    /// value, and returns `Option<()>`.
+    pub fn as_nil(&self) -> Option<()> {
+        match *self { Nil => Some(()), _ => None }
+    }
+
+    def_as!(as_boolean, Boolean, bool);
+    def_as!(as_integer, Integer, i64);
+    def_as!(as_big_integer, BigInteger, BigInt);
+    def_as!(as_float, Float, OrderedFloat<f64>);
+    def_as!(as_text, Text, String);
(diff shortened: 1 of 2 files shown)


ncalexan:
@jsantell yeah, just so!  I expected `serde` to have an API like this but I don't see it.  In any case, looking good!

edn/src/types.rs
```diff
@@ -42,11 +42,60 @@ pub enum Value {
...
+
+    pub fn is_boolean(&self) -> bool {
+        match *self { Boolean(_) => true, _ => false }
+    }
+
+    pub fn is_integer(&self) -> bool {
+        match *self { Integer(_) => true, _ => false }
+    }
+
+    pub fn is_big_integer(&self) -> bool {
```
rnewman: I'd be interested to see if:

```
macro_rules! def_is {
    ( $name: ident, $kind: pat ) {
        pub fn $name(&self) -> bool {
            match *self { $kind => true, _ => false }
        }
    }
}

def_is!(is_nil, Nil);
def_is!(is_integer, Integer(_));
def_is!(is_big_integer, BigInteger(_));
```

works…

edn/src/types.rs
```diff
@@ -41,15 +41,74 @@ pub enum Value {
...
+    def_is!(is_symbol, PlainSymbol(_));
+    def_is!(is_namespaced_symbol, NamespacedSymbol(_));
+    def_is!(is_keyword, Keyword(_));
+    def_is!(is_namespaced_keyword, NamespacedKeyword(_));
+    def_is!(is_vector, Vector(_));
+    def_is!(is_list, List(_));
+    def_is!(is_set, Set(_));
+    def_is!(is_map, Map(_));
+
+    pub fn as_nil(&self) -> Option<Value> {
```
jsantell: Not sure what should be returned here?

jsantell: the as_* functions were attempted to be macros, but the `v` in `Some(v)` couldn't access the `v` in the passed in pattern matcher

rnewman: Probably don't need `as_nil` when we have `is_nil`, but you could do `Option<()>`, which is really just success or failure.

edn/tests/tests.rs
```diff
@@ -887,6 +893,116 @@ fn test_utils_merge() {
...
+        }
+
+        if i == 0 { assert_eq!(value.as_nil().unwrap(), Nil) }
+        else { assert_eq!(value.as_nil(), None) }
+
+        if i == 1 { assert_eq!(*value.as_boolean().unwrap(), false) }
+        else { assert_eq!(value.as_boolean(), None) }
+
+        if i == 2 { assert_eq!(*value.as_integer().unwrap(), 1i64) }
+        else { assert_eq!(value.as_integer(), None) }
```
rnewman: This should be `assert!(!value.is_some())`.

jsantell:
@rnewman now with handfuls of macros!

edn/src/types.rs
```diff
@@ -87,15 +87,57 @@ fn test_print_edn() {
...
+    def_is!(is_symbol, PlainSymbol(_));
+    def_is!(is_namespaced_symbol, NamespacedSymbol(_));
+    def_is!(is_keyword, Keyword(_));
+    def_is!(is_namespaced_keyword, NamespacedKeyword(_));
+    def_is!(is_vector, Vector(_));
+    def_is!(is_list, List(_));
+    def_is!(is_set, Set(_));
+    def_is!(is_map, Map(_));
+
+    pub fn as_nil(&self) -> Option<()> {
```
ncalexan: Maybe a little note about why `Nil` is different.

edn/tests/tests.rs
```diff
@@ -33,6 +33,12 @@ fn k_plain(name: &str) -> Value {
     return Keyword(symbols::Keyword::new(name));
 }
 
+// Helper for making wrapped symbols with a namespace
```
ncalexan: Hmm.  I feel like there's a `From<(&str, &str)>` for `symbols::NamespacedSymbol` lurking in here, but this'll do for now.

edn/src/types.rs
```diff
@@ -87,15 +87,57 @@ fn test_print_edn() {
                Value::Boolean(true))).to_string());
 }
 
-impl Value {
-    pub fn is_keyword(&self) -> bool {
-        match *self {
-            Keyword(_) => true,
-            _          => false,
+macro_rules! def_is {
```
ncalexan: In general macros are a little hard to read, so a `///` comment explaining what they try to achieve goes a long way.

Also, a nit, consider `$pat` instead of `$kind`; _kind_ [means something precise in type theory](https://en.wikipedia.org/wiki/Kind_(type_theory)) that does not agree with your usage here.

edn/tests/tests.rs
jsantell: Feel like we can clean up a lot of these tests in a follow up! This file is large.

PR #209, opened 2017-01-26
Rudimentary printing of EDN values.

rnewman:
Note that we should be implementing `Display` instead of `ToString`, so I fixed that, too.


rnewman:
Paging @jsantell.

PR #211, opened 2017-01-27
Parse basic :find and :where clauses.

rnewman:


query-parser/src/parse.rs
```diff
@@ -19,11 +19,12 @@ use self::mentat_query::{Element, FindSpec, Variable};
 
 use super::error::{FindParseError, FindParseResult};
 
-pub struct FindSp<I>(::std::marker::PhantomData<fn(I) -> I>);
+pub struct Find<I>(::std::marker::PhantomData<fn(I) -> I>);
 
-type FindSpParser<O, I> = Expected<FnParser<I, fn(I) -> ParseResult<O, I>>>;
+// Nothing about this is specific to the type of parser.
+type ResultParser<O, I> = Expected<FnParser<I, fn(I) -> ParseResult<O, I>>>;
```
ncalexan: Mmm, we might want to lift this and share it for `tx-parser` as well.  Might not be worth the effort.

query-parser/src/parse.rs
```diff
@@ -51,6 +51,25 @@ impl<I> Query<I>
     }
 }
 
+/// Generate a `satisfy_map` expression that matches a `PlainSymbol`
+/// value with the given name.
+///
+/// We do this rather than using `combine::token` so that we don't
+/// need to allocate a new `String` inside a `PlainSymbol` inside a `Value`
+/// just to match input.
+macro_rules! matches_plain_symbol {
```
ncalexan: Very nice.  There might be an equivalent on the `tx-parser` side of the house, but I'll wait until I'm in that code again to file a more specific ticket.

query/src/lib.rs
```diff
@@ -43,14 +43,32 @@ pub type SrcVarName = String;          // Do not include the required syntactic
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Variable(pub PlainSymbol);
 
-#[derive(Clone,Debug,Eq,PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq)]
```
ncalexan: This doesn't seem to be the Rust-y style, but if `rustfmt` does this, I'm in.

query-parser/src/parse.rs
```diff
@@ -12,14 +12,25 @@ extern crate combine;
 extern crate edn;
 extern crate mentat_query;
 
-use self::combine::{eof, many1, parser, satisfy_map, Parser, ParseResult, Stream};
+use self::combine::{eof, many1, optional, parser, satisfy_map, token, Parser, ParseResult, Stream};
 use self::combine::combinator::{Expected, FnParser, choice, try};
 use self::edn::Value::PlainSymbol;
-use self::mentat_query::{Element, FindSpec, Variable};
+use self::mentat_query::{
```
ncalexan: Does `rustfmt` respect this?  (I hope it does.)

query-parser/src/parse.rs
```diff
@@ -70,6 +69,130 @@ macro_rules! matches_plain_symbol {
     }
 }
 
+trait FromValue<T> {
```
ncalexan: Hmm.  Is there a reason not to use `TryFrom`, or to implement `T : From<edn::Value>`?

There are many ways to skin this cat in Rust, so I'm not too concerned about getting this just right at this moment, but I'd like to know if you have done some of this thinking.

query-parser/src/parse.rs
```diff
@@ -70,6 +69,130 @@ macro_rules! matches_plain_symbol {
...
+    fn from_value(edn::Value) -> Option<T>;
+}
+
+impl FromValue<PatternNonValuePlace> for PatternNonValuePlace {
+    fn from_value(v: edn::Value) -> Option<PatternNonValuePlace> {
+        match v {
+            edn::Value::Integer(x) => if x >= 0 {
+                Some(PatternNonValuePlace::Entid(x as u64))
```
ncalexan: Does this panic if `x = 2^63 + 2^62`?

query-parser/src/parse.rs
```diff
@@ -70,6 +69,130 @@ macro_rules! matches_plain_symbol {
...
+            } else {
+                None
+            },
+            edn::Value::PlainSymbol(x) => if x.0.as_str() == "_" {
```
ncalexan: I find this indentation quite challenging to parse -- consider `=> { ... }` in each block to indent more clearly.

query-parser/src/parse.rs
```diff
@@ -70,6 +69,130 @@ macro_rules! matches_plain_symbol {
...
+                None
+            },
+            edn::Value::NamespacedKeyword(x) =>
+                Some(PatternNonValuePlace::Ident(x)),
+            _ => None,
+        }
+    }
+}
+
+impl FromValue<PatternValuePlace> for PatternValuePlace {
```
ncalexan: So, the `serde` way of doing these conversions is fancy: I think they generate serializing/deserializing code statically, strong typing these things.  Given how frequent `edn::Value` <-> domain object is going to be, maybe which should learn their secrets and make this first class?

query-parser/src/parse.rs
```diff
@@ -70,6 +69,130 @@ macro_rules! matches_plain_symbol {
...
+            edn::Value::Text(x) => Some(PatternValuePlace::Constant(NonIntegerConstant::Text(x))),
+            _ => None,
+        }
+    }
+}
+
+impl<I> Query<I>
+    where I: Stream<Item = edn::Value>
+{
+    fn source_var() -> ResultParser<SrcVar, I> {
```
ncalexan: I filed https://github.com/mozilla/mentat/issues/218 to extract this `fn_parser` pattern into our macro library.

query-parser/src/parse.rs
```diff
@@ -70,6 +69,130 @@ macro_rules! matches_plain_symbol {
...
+        satisfy_map(|x: edn::Value| PatternNonValuePlace::from_value(x)).parse_stream(input)
+    }
+
+    fn pattern() -> ResultParser<Pattern, I> {
+        fn_parser(Where::<I>::pattern_, "pattern")
+    }
+
+    fn pattern_(input: I) -> ParseResult<Pattern, I> {
+        satisfy_unwrap!(edn::Value::Vector, y, {
+            // While *technically* Datomic allows you to have a query like:
```
ncalexan: The grammar of this sentence is strained.

query/src/lib.rs
rnewman: It doesn't, and for no good reason I have a preference for the no-spaces style, but Jim's book uses spaces.

query-parser/src/parse.rs
rnewman: Yeah, I want to do a polishing pass.

query-parser/src/parse.rs
rnewman: This is outdated: the current version uses `i64` throughout.

query-parser/src/parse.rs
rnewman: `TryFrom` is unstable, no?

But I opted for `Option` because that's also the pattern `satisfy_map` uses. There isn't really a useful error class we could include: we just want to encode that this value wasn't a suitable `PatternNonValuePlace`. That's also why I kept this trait private: it's almost a stylistic utility for the parser, turning a collection of functions like `value_to_pattern_non_value_place` into a common method name.

query-parser/src/parse.rs
rnewman: This is static and strongly typed: one calls `PatternValuePlace::from_value(v)`, and it can only return an `Option<PatternValuePlace>`. Quite possibly it'll be inlined.

rnewman:
Depends on #218. Ready for review, @ncalexan.

query-parser/src/find.rs
```diff
@@ -38,10 +38,30 @@ extern crate mentat_query;
...
+    SrcVar,
+    Variable,
+};
+
+use super::parse::{NotAVariableError, QueryParseError, QueryParseResult};
+use super::util::vec_to_keyword_map;
+
+/// If the provided slice of EDN values are all variables as
+/// defined by `value_to_variable`, return a Vec of Variables.
+/// Otherwise, return the unrecognized Value.
```
ncalexan: nit: an `Error` wrapping the unrecognized `Value`?  Or don't wrap in `Err` at all (since you have set the precedent).

query-parser/src/find.rs
```diff
@@ -136,3 +156,49 @@ pub fn parse_find(expr: edn::Value) -> QueryParseResult {
...
+    use self::edn::types::{to_keyword, to_symbol};
+    use super::mentat_query::{
+        Element,
+        FindSpec,
+        SrcVar,
+        Variable,
+    };
+    use super::*;
+
+    // TODO: when #224 lands, fix to_keyword to be variadic.
```
ncalexan: #224 has now landed, but I don't understand how you intend to make `to_keyword` variadic.  Even with `Into<>` tricks, you still need `None`...

query-parser/src/parse.rs
```diff
@@ -10,156 +10,191 @@
...
+pub enum QueryParseError {
+    InvalidInput(edn::Value),
+    EdnParseError(edn::parse::ParseError),
+    MissingField(edn::Keyword),
+    FindParseError(FindParseError),
+}
 
-    fn find_rel() -> FindSpParser<FindSpec, I> {
-        fn_parser(FindSp::<I>::find_rel_, "find_rel")
+impl From<edn::parse::ParseError> for QueryParseError {
```
ncalexan: This kind of chaining is where `error-chain` is supposed to shine, but I'm happy with this approach until we have a pattern around `error-chain`.

query-parser/src/parse.rs
```diff
@@ -10,156 +10,191 @@
...
 
+impl<I> Query<I>
+    where I: Stream<Item = edn::Value>
+{
     fn to_parsed_value<T>(r: ParseResult<T, I>) -> Option<T> {
         r.ok().map(|x| x.0)
     }
 }
 
+def_value_satisfy_parser_fn!(Query, variable, Variable, Variable::from_value);
```
ncalexan: It's so frustrating that this is required, when all we really want is to have `Variable::from_value` be a parser!  At least it's easy to write now.

query-parser/src/parse.rs
```diff
@@ -10,156 +10,191 @@
...
+            // While *technically* Datomic allows you to have a query like:
+            // [:find … :where [[?x]]]
+            // We don't -- we require at list e, a.
+            let mut p =
+                (optional(Query::source_var()),                 // src
+                 Where::pattern_non_value_place(),              // e
+                 Where::pattern_non_value_place(),              // a
+                 optional(Where::pattern_value_place()),        // v
+                 optional(Where::pattern_non_value_place()),    // tx
+                 eof())
```
ncalexan: I wonder what it would take to make `(p, q)` always expect to parse the complete stream.  Probably another wrapper layer :(

query-parser/src/parse.rs
```diff
@@ -10,156 +10,191 @@
...
+                .map(|(src, e, a, v, tx, _)| {
+                    let v = v.unwrap_or(PatternValuePlace::Placeholder);
+                    let tx = tx.unwrap_or(PatternNonValuePlace::Placeholder);
+
+                    // Pattern::new takes care of reversal.
```
ncalexan: nit: include "attribute" -- maybe "takes care of any attributes that are reversed".

query-parser/src/parse.rs
```diff
@@ -10,156 +10,191 @@
...
+def_value_satisfy_parser_fn!(Query, source_var, SrcVar, SrcVar::from_value);
+
+pub struct Where<I>(::std::marker::PhantomData<fn(I) -> I>);
+
+def_value_satisfy_parser_fn!(Where, pattern_value_place, PatternValuePlace, PatternValuePlace::from_value);
+def_value_satisfy_parser_fn!(Where, pattern_non_value_place, PatternNonValuePlace, PatternNonValuePlace::from_value);
+
+def_value_parser_fn!(Where, pattern, Pattern, input, {
+    satisfy_map(|x: edn::Value| {
+        if let edn::Value::Vector(y) = x {
```
ncalexan: God, Rust really needs to just grow the rest of Swift's vocabulary: `if not let`, `guard`, `!`, `??`.

query-parser/src/parse.rs
```diff
@@ -10,156 +10,191 @@
...
+            let v: Option<Option<Pattern>> = r.ok().map(|x| x.0);
+            v.unwrap_or(None)
+        } else {
+            None
+        }
+    }).parse_stream(input)
+});
+
+def_value_parser_fn!(Where, clauses, Vec<WhereClause>, input, {
+    // Right now we only support patterns.
```
ncalexan: Maybe drop the TODO for those following at home?  Later on we'll include `:in` and the rest?

query-parser/src/parse.rs
```diff
@@ -10,156 +10,191 @@
...
+    satisfy_unwrap!(edn::Value::Vector, y, {
+        let r: ParseResult<FindSpec, _> =
+            Find::elements().map(FindSpec::FindTuple).parse_lazy(&y[..]).into();
+        Query::to_parsed_value(r)
+    })
+    .parse_stream(input)
+});
+
+def_value_parser_fn!(Find, find, FindSpec, input, {
+    // Any one of the four specs might apply, so we combine them with `choice`.
```
ncalexan: Beautiful.

query-parser/src/parse.rs
```diff
@@ -10,156 +10,191 @@
...
+    // Our parsers consume input, so we need to wrap them in `try` so that they
+    // operate independently.
+    choice::<[&mut Parser<Input = I, Output = FindSpec>; 4],
```
ncalexan: Consider not breaking the type across lines.

query-parser/src/parse.rs
```diff
@@ -172,37 +207,142 @@ impl<I> FindSp<I>
...
+        PatternValuePlace,
+        SrcVar,
+        Variable,
+    };
 
     use super::*;
 
     #[test]
+    fn test_pattern_mixed() {
+        let e = edn::PlainSymbol::new("_");
```
ncalexan: I think it's worth parsing EDN from strings for these tests.  It's damn hard to make sense of the input.  Next patch?

query-parser/src/parse.rs
```diff
@@ -172,37 +207,142 @@ impl<I> FindSp<I>
...
+        let tx = edn::PlainSymbol::new("?tx");
+        let input = [edn::Value::Vector(
+            vec!(edn::Value::PlainSymbol(e.clone()),
+                 edn::Value::NamespacedKeyword(a.clone()),
+                 edn::Value::Float(v.clone()),
+                 edn::Value::PlainSymbol(tx.clone())))];
+
+        let mut par = Where::pattern();
+        let result = par.parse(&input[..]);
+        if let Err(_) = result {
```
ncalexan: There's a `matches` crate that allows to use `match` as an expression, and then `assert!(matches!(...))`.  Might be worth using.

query-parser/src/find.rs
rnewman: I really mean overloading. `fn to_keyword(s: Into<&str>) -> Keyword`, `fn to_keyword(ns: Into<Option<&str>>, n: Into<&str>)`.

query-parser/src/parse.rs
rnewman: Yeah, it's pretty verbose to get to where parsing code in most other languages ends up, which is:

```
(or (parse-var in)
    (parse-constant in)
    (error "Expected var or constant"))
```

query-parser/src/parse.rs
rnewman: They're really down on `guard`, because it's syntax that affects subsequent parts of the enclosing scope. But yes :)

query-parser/src/parse.rs
rnewman: `:in` is elsewhere — this is about parsing `or`, `not`, predicate functions, and all the rest. But yes, I'll file the issues and leave a TODO.

query-parser/src/parse.rs
rnewman: I'll probably do both, out of an abundance of caution, but I'll do so in integration tests — that is, the public API for `query-parser` is to take input and produce the structures in `query`, ready for consumption by `query-executor`. I'll exercise that public API.

query-parser/src/parse.rs
rnewman: I think so — that's essentially stream lookahead.

query-parser/src/find.rs
rnewman: Oh good, there's no function overloading. Never mind.

query-parser/src/parse.rs
rnewman: Filed #239, note added.

query-parser/src/parse.rs
rnewman: *shakes fist at `rustfmt`*

query-parser/src/parse.rs
rnewman: Nice. I'll do that. (Also useful in non-test code.)

PR #212, opened 2017-01-28
Add NamespacedKeyword::is_reverse and NamespacedKeyword::to_reverse f…

### LINKED ISSUE #202: Expose is_reversed, to_reversed on NamespacedKeyword
A keyword naming an attribute — `:person/name`, `:person/friend` — can be prefixed with `_` to denote that it runs backwards. For example, the following two snippets are identical:

```
[?y :person/friend ?x]
[?x :person/hired ?y]
```
```
[?y :person/friend ?x]
[?y :person/_hired ?x]
```

In the query and transaction parsers it'll be useful to define checker and transformation methods on `NamespacedKeyword` to detect and switch between these cases.

jsantell:
…or query and transaction parsers. Fixes #202

Notes: Should we change all of the symbols to be able to handle taking a `String` instead of just a `&str`? Is there a less ugly concat way to handle `to_reverse`?  Should `to_reverse` flip ":foo/_baz" back to ":foo/baz", and if so, maybe a name change to `invert`?

### DIFF
--- a/edn/src/symbols.rs
+++ b/edn/src/symbols.rs
@@ -101,6 +101,15 @@ impl Keyword {
 }
 
 impl NamespacedKeyword {
+    /// Creates a new `NamespacedKeyword`.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use edn::symbols::NamespacedKeyword;
+    /// let keyword = NamespacedKeyword::new("foo", "bar");
+    /// assert_eq!(keyword.to_string(), ":foo/bar");
+    /// ```
     pub fn new<T>(namespace: T, name: T) -> Self where T: Into<String> {
         let n = name.into();
         let ns = namespace.into();
@@ -108,7 +117,73 @@ impl NamespacedKeyword {
         assert!(!ns.is_empty(), "Keywords cannot have an empty non-null namespace.");
 
         // TODO: debug asserts to ensure that neither field matches [ :/].
-        NamespacedKeyword { name: n, namespace: ns }
+        NamespacedKeyword {
+            name: n,
+            namespace: ns,
+        }
+    }
+
+    /// Whether this NamespacedKeyword should be interpreted in reverse order. For example,
+    /// the two following snippets are identical:
+    /// [?y :person/friend ?x]
+    /// [?x :person/hired ?y]
+    ///
+    /// [?y :person/friend ?x]
+    /// [?y :person/_hired ?x]
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use edn::symbols::NamespacedKeyword;
+    /// assert!(!NamespacedKeyword::new("foo", "bar").is_backward());
+    /// assert!(NamespacedKeyword::new("foo", "_bar").is_backward());
+    /// ```
+    #[inline]
+    pub fn is_backward(&self) -> bool {
+        self.name.starts_with("_")
+    }
+
+    /// Whether this NamespacedKeyword should be interpreted in forward order.
+    /// See `symbols::NamespacedKeyword::is_backwards`.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use edn::symbols::NamespacedKeyword;
+    /// assert!(NamespacedKeyword::new("foo", "bar").is_forward());
+    /// assert!(!NamespacedKeyword::new("foo", "_bar").is_forward());
+    /// ```
+    #[inline]
+    pub fn is_forward(&self) -> bool {
+        !self.is_backward()
+    }
+
+    /// Returns an Option<NamespacedKeyword> with the same name and namespace, but
+    /// considered reversed. See `symbols::NamespacedKeyword::is_backward`.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use edn::symbols::NamespacedKeyword;
+    /// let nsk = NamespacedKeyword::new("foo", "bar");
+    /// assert!(!nsk.is_backward());
+    /// assert_eq!(":foo/bar", nsk.to_string());
+    ///
+    /// let reversed = nsk.to_reverse();
+    /// assert!(reversed.is_backward());
+    /// assert_eq!(":foo/_bar", reversed.to_string());
+    /// ```
+    pub fn to_reverse(&self) -> NamespacedKeyword {
+        let name = if self.is_backward() {
+            self.name[1..].to_string()
+        } else {
+            format!("{}{}", "_", self.name)
+        };
+
+        NamespacedKeyword {
+            name: name,
+            namespace: self.namespace.clone(),
+        }
     }
 }
 
@@ -166,6 +241,8 @@ impl Display for NamespacedKeyword {
     /// ```rust
     /// # use edn::symbols::NamespacedKeyword;
     /// assert_eq!(":bar/baz", NamespacedKeyword::new("bar", "baz").to_string());
+    /// assert_eq!(":bar/_baz", NamespacedKeyword::new("bar", "baz").to_reverse().to_string());
+    /// assert_eq!(":bar/baz", NamespacedKeyword::new("bar", "baz").to_reverse().to_reverse().to_string());
     /// ```
     fn fmt(&self, f: &mut Formatter) -> ::std::fmt::Result {
         write!(f, ":{}/{}", self.namespace, self.name)


rnewman:
That is: a symbol's name never starts with '_' for our purposes. I'm not sure if that's true or not in edn as a whole, so pushback is acceptable. Perhaps we want a wrapper struct…

jsantell:
@rnewman makes way more sense! and in this case, to_reverse seems like it should always be able to reverse an already 'reversed' keyword?

jsantell:
Made changes suggested -- should it be `[to|is]_reverse` or `[to|is]_reversed`? Can land this on top in `fmt` in #209

jsantell:
Rebased on #209

edn/src/symbols.rs
```diff
@@ -98,12 +99,73 @@ impl Keyword {
 }
 
 impl NamespacedKeyword {
+    /// Creates a new `NamespacedKeyword`. When `name` starts with a `_`, it is considered
```
ncalexan: nit: double newline after the first sentence.

Consider repeating the `is_reversed` notes, with an example, here.

edn/src/symbols.rs
```diff
@@ -98,12 +99,73 @@ impl Keyword {
...
+    /// # use edn::symbols::NamespacedKeyword;
+    /// let nsk = NamespacedKeyword::new("foo", "bar");
+    /// assert!(!nsk.is_reversed());
+    /// assert_eq!(":foo/bar", nsk.to_string());
+    ///
+    /// let reversed = nsk.to_reversed();
+    /// assert!(reversed.is_reversed());
+    /// assert_eq!(":foo/_bar", reversed.to_string());
+    /// ```
+    pub fn to_reversed(&self) -> NamespacedKeyword {
```
ncalexan: Maybe `reverse`?

edn/src/symbols.rs
ncalexan: DataScript and the Clojure implementation refer to "not reversed" as _straight_.  Are we happy using this terminology here?  (I could imagine picking something that doesn't have any connection to sexual orientation.)  If we're happy, include _straight_ as the opposite of reversed in the comment, and add `is_straight` and `straight`/`to_straight`/`straighten`.

The fact that _straight_ and _reverse_ conjugate so differently argues for changing _straight_ but I don't have a second suggestion to hand.

edn/src/symbols.rs
```diff
@@ -98,12 +99,73 @@ impl Keyword {
...
+        NamespacedKeyword {
```
ncalexan: We are really making things hard to pattern match against, aren't we?  Oh well, don't have a better suggestion right now.

rnewman:
> I think this interface will prove awkward in practice, and doesn't help folks who need to consider reversed attributes (the transactor) actually ensure they handled the reversed cases, but I don't have a better approach to hand.

I was expecting something like:

```
let attribute = …;    // Parser output
let reversed = attribute.is_reversed();
return Pattern {
  entity: if reversed { value } else { entity },
  attribute: if reversed { attribute.reverse() } else { attribute },
  value: if reversed { entity } else { value },
  tx: tx,
};
```

which is code I'm pretty much writing now.

edn/src/symbols.rs
```diff
@@ -98,12 +99,73 @@ impl Keyword {
...
         assert!(!namespace.is_empty(), "Keywords cannot have an empty non-null namespace.");
+        let reversed = match name.chars().nth(0) {
+            Some(v) => v.to_string() == "_",
+            None => false,
+        };
 
         // TODO: debug asserts to ensure that neither field matches [ :/].
-        NamespacedKeyword { name: name.to_string(), namespace: namespace.to_string() }
+        NamespacedKeyword {
+            name: if reversed { name[1..].to_string() } else { name.to_string() },
```
rnewman: This needs to be rebased on top of #213.

edn/src/symbols.rs
```diff
@@ -98,12 +99,73 @@ impl Keyword {
...
+    ///
+    /// ```rust
+    /// # use edn::symbols::NamespacedKeyword;
+    /// let keyword = NamespacedKeyword::new("foo", "bar");
+    /// assert_eq!(keyword.to_string(), ":foo/bar");
+    /// ```
     pub fn new(namespace: &str, name: &str) -> Self {
         assert!(!name.is_empty(), "Keywords cannot be unnamed.");
         assert!(!namespace.is_empty(), "Keywords cannot have an empty non-null namespace.");
+        let reversed = match name.chars().nth(0) {
```
rnewman: Use `name.starts_with('_')`.

edn/src/symbols.rs
```diff
@@ -98,12 +99,73 @@ impl Keyword {
...
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use edn::symbols::NamespacedKeyword;
+    /// assert!(!NamespacedKeyword::new("foo", "bar").is_reversed());
+    /// assert!(NamespacedKeyword::new("foo", "_bar").is_reversed());
+    /// assert_eq!(NamespacedKeyword::new("foo", "_bar").name, "bar");
+    /// ```
+    pub fn is_reversed(&self) -> bool {
```
rnewman: `#[inline]`

edn/src/symbols.rs
rnewman: I think `reverse` would imply mutability. And while I'd normally be happy with mutability in this case, the keywords we're working with will probably be borrowed references inside values.

edn/src/symbols.rs
```diff
@@ -98,12 +99,73 @@ impl Keyword {
...
+    /// assert_eq!(":foo/bar", nsk.to_string());
+    ///
+    /// let reversed = nsk.to_reversed();
+    /// assert!(reversed.is_reversed());
+    /// assert_eq!(":foo/_bar", reversed.to_string());
+    /// ```
+    pub fn to_reversed(&self) -> NamespacedKeyword {
+        NamespacedKeyword {
+            name: self.name.clone(),
+            namespace: self.namespace.clone(),
```
rnewman: There's 'spread' syntax for this:

```
NamespacedKeyword { reversed: !self.reversed, .. self }
```

edn/src/symbols.rs
```diff
@@ -161,8 +223,9 @@ impl Display for NamespacedKeyword {
     /// ```rust
     /// # use edn::symbols::NamespacedKeyword;
     /// assert_eq!(":bar/baz", NamespacedKeyword::new("bar", "baz").to_string());
+    /// assert_eq!(":bar/_baz", NamespacedKeyword::new("bar", "_baz").to_string());
```
rnewman: Also add a `.to_reversed()` test here. You could make this new assertion pass without changing any of the other code in the file!

edn/src/symbols.rs
rnewman: I'm not super happy with 'straight'. Quite apart from the exclusionary language, this isn't a thing that bends; it's a directional flag where the two directions are opposite. So `reversed`, or `direction: forwards/backwards`, perhaps even `inverse`, but not `straight`.

edn/src/symbols.rs
jsantell: Lovely.

edn/src/symbols.rs
```diff
@@ -101,14 +101,89 @@ impl Keyword {
...
+    /// assert_eq!(":foo/_bar", reversed.to_string());
+    /// ```
+    pub fn to_reverse(&self) -> NamespacedKeyword {
+        let name = if self.is_backward() {
+            self.name[1..].to_string()
+        } else {
+            format!("{}{}", "_", self.name)
+        };
+
+        NamespacedKeyword {
```
jsantell: Unfortunately, spread properties don't seem to work due to needing to clone the props

edn/src/symbols.rs
```diff
@@ -101,14 +101,89 @@ impl Keyword {
...
+    ///
+    /// let reversed = nsk.to_reverse();
+    /// assert!(reversed.is_backward());
+    /// assert_eq!(":foo/_bar", reversed.to_string());
+    /// ```
+    pub fn to_reverse(&self) -> NamespacedKeyword {
+        let name = if self.is_backward() {
+            self.name[1..].to_string()
+        } else {
+            format!("{}{}", "_", self.name)
```
jsantell: Better way?

ncalexan: `concat`, `+`?  I don't much care, this is clear.

edn/src/symbols.rs
```diff
@@ -166,6 +241,7 @@ impl Display for NamespacedKeyword {
     /// ```rust
     /// # use edn::symbols::NamespacedKeyword;
     /// assert_eq!(":bar/baz", NamespacedKeyword::new("bar", "baz").to_string());
+    /// assert_eq!(":bar/_baz", NamespacedKeyword::new("bar", "baz").to_reverse().to_string());
```
ncalexan: Can you verify that `to_reverse().to_reverse()` is the identity?

jsantell: :+1:

PR #213, opened 2017-01-29
Add Into<String> to symbol::* constructors. Fixes #199

### LINKED ISSUE #199: Add Into<String> to symbol::* constructors
Over in #196, @rnewman is replacing `String` with `T where T: Into<String>` in a few constructors.  This ticket tracks generalizing the rest of the constructors, in particular `NamespacedSymbol` and `NamespacedKeyword`, to have similar pleasant interfaces.

jsantell:


edn/src/symbols.rs
```diff
@@ -98,12 +101,14 @@ impl Keyword {
 }
 
 impl NamespacedKeyword {
-    pub fn new(namespace: &str, name: &str) -> Self {
-        assert!(!name.is_empty(), "Keywords cannot be unnamed.");
-        assert!(!namespace.is_empty(), "Keywords cannot have an empty non-null namespace.");
+    pub fn new<T>(namespace: T, name: T) -> Self where T: Into<String> {
```
ncalexan: Given the reversed patch, is `_` a valid `name`?

rnewman: I think it would be in general-purpose EDN; Clojure certainly accepts it. I don't think it's useful for our purposes.

This kind of question is why I mentioned perhaps layering reversal on top of the EDN crate — EDN doesn't have a notion of attributes, let alone reversal! — rather than doing what's convenient for us. If we continue to have reversal as part of the EDN crate, then we should reject `:foo/_`. If we layer it on top, then not all valid EDN keywords are valid input to Mentat.

ncalexan:
> This kind of question is why I mentioned perhaps layering reversal on top of the EDN crate — EDN doesn't have a notion of attributes, let alone reversal! — rather than doing what's convenient for us.

After reviewing, as I was feeding Charlie, I thought about this layering too, and conclude the same: this isn't the right level of the stack to handle reversed attributes.  If we care to, we can add these helpers; but it seems that we should explicitly handle `ReversedKeyword` or whatever in the transactor and the query engine.

edn/src/symbols.rs
jsantell: This patch isn't related to reversing, but handling reverse in #212 should be handled differently because of the above comments.

Issue #190, opened 2017-01-23
Wrap positive integer entids in an Entid newtype


ncalexan:
This is follow-up to https://github.com/mozilla/mentat/pull/171#discussion_r96086384 and #170.  Right now, I'm representing positive integer entids as `i64`, since SQLite doesn't have an unsigned `u64` type.  It's an interesting challenge to really represent entids as an `Entid` newtype that is really `u63`.  When I started, I didn't understand the `rusqlite` traits `FromSql` and `ToSql`; now that I do, this should be straight-forward.  I'm filing this as a _[good next bug]_, I guess, since it's not trivial.  Use `OrderedFloat` for inspiration and remember that it never makes sense to do arithmetic on entids.

As a side benefit/hindrance while testing: this might allow Mentat to _control the introduction of entids_.  The Clojure version accepted any integer as an entid, which was handy for testing but did not agree with Datomic.  We might be able to make the only way to allocate an entid be via a tempid, which has good correctness properties: it'll be harder to make errors in transactions, and it'll be harder to pollute a datom store with disconnected components.

rnewman:
Also raises the modeling concept of a *known entid* vs an *unknown entid* vs a *potential entid*. Queries can, of course, refer to entids that aren't in the store.

rnewman:
#447 partly implemented this ticket: _within the scope of a transaction_ we use a newtype, `KnownEntid`, to enforce that all mentioned entids are within the range of the store's current partition maps.

Issue #188, opened 2017-01-21
[tx] Convert Clojure tests into Rust tests


ncalexan:
The Clojure implementation has an extensive correctness test suite.  Most of the tests look like:
#. install test schema;
#. transact a few assertions;
#. assert that datoms, transactions, or fulltext values table has following contents/new rows.
Since our SQL schema is identical to the Clojure implementation, it should be easy to move the input *and* output data into `.edn` files and write a generic test harness to do those steps.

It might be finicky to get the transaction IDs, timestamps, and other ephemeral data correct.  I wonder if we can rewrite `:db/tx` and `:db/txInstant` in controlled locations to get what we want?  (This could require reader syntax like `#value [:db/tx]` or similar.)

ncalexan:
I have a good start on the test framework locally, so please co-ordinate with me before doing anything in this area.

PR #214, opened 2017-01-30
[tx] Start implementing bulk SQL insertion algorithms

ncalexan:
This will eventually grow to implement https://github.com/mozilla/mentat/wiki/Transacting:-entity-to-SQL-translation.

### DIFF
--- a/db/Cargo.toml
+++ b/db/Cargo.toml
@@ -24,3 +24,7 @@ path = "../tx"
 
 [dependencies.mentat_tx_parser]
 path = "../tx-parser"
+
+# Should be dev-dependencies.
+[dependencies.tabwriter]
+version = "1.0.3"
--- a/db/Cargo.toml
+++ b/db/Cargo.toml
@@ -7,7 +7,7 @@ error-chain = "0.8.0"
 lazy_static = "0.2.2"
 
 [dependencies.rusqlite]
-version = "0.9.3"
+version = "0.9.5"
 # System sqlite might be very old.
 features = ["bundled"]
 
--- a/db/Cargo.toml
+++ b/db/Cargo.toml
@@ -4,6 +4,7 @@ version = "0.0.1"
 
 [dependencies]
 error-chain = "0.8.0"
+itertools = "0.5.9"
 lazy_static = "0.2.2"
 ordered-float = "0.4.0"
 
--- a/db/Cargo.toml
+++ b/db/Cargo.toml
@@ -7,6 +7,7 @@ error-chain = "0.8.0"
 itertools = "0.5.9"
 lazy_static = "0.2.2"
 ordered-float = "0.4.0"
+time = "0.1.35"
 
 [dependencies.rusqlite]
 version = "0.9.5"
--- a/db/Cargo.toml
+++ b/db/Cargo.toml
@@ -5,6 +5,7 @@ version = "0.0.1"
 [dependencies]
 error-chain = "0.8.0"
 lazy_static = "0.2.2"
+ordered-float = "0.4.0"
 
 [dependencies.rusqlite]
 version = "0.9.5"
--- a/db/src/bootstrap.rs
+++ b/db/src/bootstrap.rs
@@ -10,7 +10,7 @@
 
 #![allow(dead_code)]
 
-use {to_namespaced_keyword};
+use ::{to_namespaced_keyword};
 use edn;
 use edn::types::Value;
 use entids;
--- a/db/src/bootstrap.rs
+++ b/db/src/bootstrap.rs
@@ -20,6 +20,11 @@ use mentat_tx_parser;
 use types::{IdentMap, Partition, PartitionMap, Schema, TypedValue};
 use values;
 
+/// The first transaction ID applied to the knowledge base.
+///
+/// This is the start of the :db.part/tx partition.
+pub const TX0: i64 = 0x10000000;
+
 lazy_static! {
     static ref V1_IDENTS: Vec<(&'static str, i64)> = {
         vec![(":db/ident",             entids::DB_IDENT),
@@ -70,14 +75,14 @@ lazy_static! {
     static ref V1_PARTS: Vec<(&'static str, i64, i64)> = {
         vec![(":db.part/db", 0, (1 + V1_IDENTS.len()) as i64),
              (":db.part/user", 0x10000, 0x10000),
-             (":db.part/tx", 0x10000000, 0x10000000),
+             (":db.part/tx", TX0, TX0),
         ]
     };
 
     static ref V2_PARTS: Vec<(&'static str, i64, i64)> = {
         vec![(":db.part/db", 0, (1 + V2_IDENTS.len()) as i64),
              (":db.part/user", 0x10000, 0x10000),
-             (":db.part/tx", 0x10000000, 0x10000000),
+             (":db.part/tx", TX0, TX0),
         ]
     };
 
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -806,6 +806,9 @@ mod tests {
     use super::*;
     use bootstrap;
     use debug;
+    use edn;
+    use edn::symbols;
+    use mentat_tx_parser;
     use rusqlite;
     use types::*;
 
@@ -814,46 +817,110 @@ mod tests {
         // TODO: figure out how to reference the fixtures directory for real.  For now, assume we're
         // executing `cargo test` in `db/`.
         let conn = rusqlite::Connection::open("../fixtures/v2empty.db").unwrap();
-        // assert_eq!(ensure_current_version(&mut conn).unwrap(), CURRENT_VERSION);
-
-        // TODO: write :db/txInstant, bump :db.part/tx.
-        // let partition_map = read_partition_map(&conn).unwrap();
-        // assert_eq!(partition_map, bootstrap::bootstrap_partition_map());
 
         let ident_map = read_ident_map(&conn).unwrap();
         assert_eq!(ident_map, bootstrap::bootstrap_ident_map());
 
         let schema = read_schema(&conn, &ident_map).unwrap();
-        assert_eq!(schema, bootstrap::bootstrap_schema()); // Schema::default());
+        assert_eq!(schema, bootstrap::bootstrap_schema());
 
         let db = read_db(&conn).unwrap();
 
+        // Does not include :db/txInstant.
         let datoms = debug::datoms_after(&conn, &db, 0).unwrap();
-        assert_eq!(datoms.0.len(), 89); // The 89th is the :db/txInstant value.
+        assert_eq!(datoms.0.len(), 88);
 
-        // // TODO: fewer magic numbers!
-        // assert_eq!(debug::datoms_after(&conn, &db, &0x10000001).unwrap(), vec![]);
+        // Includes :db/txInstant.
+        let transactions = debug::transactions_after(&conn, &db, 0).unwrap();
+        assert_eq!(transactions.0.len(), 1);
+        assert_eq!(transactions.0[0].0.len(), 89);
+    }
+
+    /// Assert that a sequence of transactions meets expectations.
+    ///
+    /// The transactions, expectations, and optional labels, are given in a simple EDN format; see
+    /// https://github.com/mozilla/mentat/wiki/Transacting:-EDN-test-format.
+    ///
+    /// There is some magic here about transaction numbering that I don't want to commit to or
+    /// document just yet.  The end state might be much more general pattern matching syntax, rather
+    /// than the targeted transaction ID and timestamp replacement we have right now.
+    fn assert_transactions(conn: &rusqlite::Connection, db: &DB, transactions: &Vec<edn::Value>) {
+        for (index, transaction) in transactions.into_iter().enumerate() {
+            let index = index as i64;
+            let transaction = transaction.as_map().unwrap();
+            let label: edn::Value = transaction.get(&edn::Value::NamespacedKeyword(symbols::NamespacedKeyword::new("test", "label"))).unwrap().clone();
+            let assertions: edn::Value = transaction.get(&edn::Value::NamespacedKeyword(symbols::NamespacedKeyword::new("test", "assertions"))).unwrap().clone();
+            // TODO: use hyphenated keywords, like :test/expected-transaction -- when the EDN parser
+            // supports them!
+            let expected_transaction: Option<&edn::Value> = transaction.get(&edn::Value::NamespacedKeyword(symbols::NamespacedKeyword::new("test", "expectedtransaction")));
+            let expected_datoms: Option<&edn::Value> = transaction.get(&edn::Value::NamespacedKeyword(symbols::NamespacedKeyword::new("test", "expecteddatoms")));
+
+            let entities: Vec<_> = mentat_tx_parser::Tx::parse(&[assertions][..]).unwrap();
+            db.transact_internal(&conn, &entities[..], bootstrap::TX0 + index + 1).unwrap();
+
+            if let Some(expected_transaction) = expected_transaction {
+                let transactions = debug::transactions_after(&conn, &db, bootstrap::TX0 + index).unwrap();
+                assert_eq!(transactions.0[0].into_edn(),
+                           *expected_transaction,
+                           "\n{} - expected transaction:\n{}\n{}", label, transactions.0[0].into_edn(), *expected_transaction);
+            }
+
+            if let Some(expected_datoms) = expected_datoms {
+                let datoms = debug::datoms_after(&conn, &db, bootstrap::TX0).unwrap();
+                assert_eq!(datoms.into_edn(),
+                           *expected_datoms,
+                           "\n{} - expected datoms:\n{}\n{}", label, datoms.into_edn(), *expected_datoms);
+            }
+
+            // Don't allow empty tests.  This will need to change if we allow transacting schema
+            // fragments in a preamble, but for now it might catch malformed tests.
+            assert_ne!((expected_transaction, expected_datoms), (None, None),
+                       "Transaction test must include at least one of :test/expectedtransaction or :test/expecteddatoms");
+        }
     }
 
     #[test]
-    fn test_create_current_version() {
-        // // assert_eq!(bootstrap_schema().unwrap(), Schema::default());
-
-        // // Ignore result.
-        // use std::fs;
-        // let _ = fs::remove_file("/Users/nalexander/Mozilla/mentat/test.db");
-        // let mut conn = rusqlite::Connection::open("file:///Users/nalexander/Mozilla/mentat/test.db").unwrap();
-
+    fn test_add() {
         let mut conn = new_connection();
-
         assert_eq!(ensure_current_version(&mut conn).unwrap(), CURRENT_VERSION);
 
         let bootstrap_db = DB::new(bootstrap::bootstrap_partition_map(), bootstrap::bootstrap_schema());
-        // TODO: write materialized view of bootstrapped schema to SQL store.
-        // let db = read_db(&conn).unwrap();
-        // assert_eq!(db, bootstrap_db);
 
-        let datoms = debug::datoms_after(&conn, &db, 0).unwrap();
-        assert_eq!(datoms.0.len(), 89); // The 89th is the :db/txInstant value.
+        // Does not include :db/txInstant.
+        let datoms = debug::datoms_after(&conn, &bootstrap_db, 0).unwrap();
+        assert_eq!(datoms.0.len(), 88);
+
+        // Includes :db/txInstant.
+        let transactions = debug::transactions_after(&conn, &bootstrap_db, 0).unwrap();
+        assert_eq!(transactions.0.len(), 1);
+        assert_eq!(transactions.0[0].0.len(), 89);
+
+        // TODO: extract a test macro simplifying this boilerplate yet further.
+        let value = edn::parse::value(include_str!("../../tx/fixtures/test_add.edn")).unwrap();
+
+        let transactions = value.as_vector().unwrap();
+        assert_transactions(&conn, &bootstrap_db, transactions);
+    }
+
+    #[test]
+    fn test_retract() {
+        let mut conn = new_connection();
+        assert_eq!(ensure_current_version(&mut conn).unwrap(), CURRENT_VERSION);
+
+        let bootstrap_db = DB::new(bootstrap::bootstrap_partition_map(), bootstrap::bootstrap_schema());
+
+        // Does not include :db/txInstant.
+        let datoms = debug::datoms_after(&conn, &bootstrap_db, 0).unwrap();
+        assert_eq!(datoms.0.len(), 88);
+
+        // Includes :db/txInstant.
+        let transactions = debug::transactions_after(&conn, &bootstrap_db, 0).unwrap();
+        assert_eq!(transactions.0.len(), 1);
+        assert_eq!(transactions.0[0].0.len(), 89);
+
+        let value = edn::parse::value(include_str!("../../tx/fixtures/test_retract.edn")).unwrap();
+
+        let transactions = value.as_vector().unwrap();
+        assert_transactions(&conn, &bootstrap_db, transactions);
     }
 }
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -13,6 +13,7 @@
 use rusqlite;
 use rusqlite::types::{ToSql, ToSqlOutput};
 
+use {to_namespaced_keyword};
 use bootstrap;
 use edn::types::Value;
 use errors::*;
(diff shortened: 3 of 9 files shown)


ncalexan:
This is ready for review.  As written it doesn't quite match the Wiki, but the spirit is identical.  I'll try to expand on the approach and patch the Wiki as I go along.

ncalexan:
I've flagged the world for review, so lay it on me, folks.

db/Cargo.toml
```diff
@@ -4,13 +4,15 @@ version = "0.0.1"
 
 [dependencies]
 error-chain = "0.8.0"
+itertools = "0.5.9"
 lazy_static = "0.2.2"
 # TODO: don't depend on num and ordered-float; expose helpers in edn abstracting necessary constructors.
 num = "0.1.35"
 ordered-float = "0.3.0"
+time = "0.1.35"
```
jsantell: Is there anyway to keep these deps (and other modules) in sync with the top-level dependencies?

db/src/db.rs
```diff
@@ -390,6 +393,24 @@ pub fn read_db(conn: &rusqlite::Connection) -> Result<DB> {
     Ok(DB::new(partition_map, schema))
 }
 
+use itertools;
```
jsantell: Is it common to have these `use` and `type` statements in the middle of a file? Also, in general, this is getting pretty long, maybe break it out into a new file?

db/src/db.rs
```diff
@@ -416,45 +437,438 @@ impl DB {
...
+                once(e as &ToSql)
+                    .chain(once(a as &ToSql)
+                           .chain(once(value as &ToSql)
+                                  .chain(once(value_type_tag as &ToSql)
+                                         .chain(once(&tx as &ToSql)
+                                                .chain(once(to_bool_ref(added) as &ToSql)
+                                                       .chain(once(to_bool_ref(index_avet) as &ToSql)
+                                                              .chain(once(to_bool_ref(index_vaet) as &ToSql)
+                                                                     .chain(once(to_bool_ref(index_fulltext) as &ToSql)
+                                                                            .chain(once(to_bool_ref(unique_value) as &ToSql))))))))))
```
jsantell: * good
  * golly
    * miss
      * molly

db/src/db.rs
```diff
@@ -416,45 +437,438 @@ impl DB {
...
+        results.map(|_| ())
+    }
+
+    /// Take search rows and complete `temp.search_results`.
+    ///
+    /// See https://github.com/mozilla/mentat/wiki/Transacting:-entity-to-SQL-translation.
+    fn search(&self, conn: &rusqlite::Connection) -> Result<()> {
+        // First is fast, only one table walk: lookup by exact eav.
+        // Second is slower, but still only one table walk: lookup old value by ea.
+        let s = r#"
```
jsantell: Would it help with clarity/size to pull out these SQL strings into it's own module as well?

db/src/bootstrap.rs
```diff
@@ -70,14 +75,14 @@ lazy_static! {
...
-             (":db.part/tx", 0x10000000, 0x10000000),
+             (":db.part/tx", TX0, TX0),
         ]
     };
 
     static ref V2_PARTS: Vec<(&'static str, i64, i64)> = {
         vec![(":db.part/db", 0, (1 + V2_IDENTS.len()) as i64),
-             (":db.part/user", 0x10000, 0x10000),
-             (":db.part/tx", 0x10000000, 0x10000000),
+             (":db.part/user", TX0, TX0),
```
ncalexan: Oops!  `:db.part/user` should remain untouched.  Fixed locally.

db/Cargo.toml
rnewman: I'm having this trouble, too.

In theory you can do:

```
[dependencies.ordered-float]
```

and it'll use whichever one is already in use. Except that just broke for me on Travis, so instead I have to pin very, very carefully. We just need to rely on compiler errors if we get this wrong.

db/src/db.rs
rnewman: `type` more so than `use`.

db/src/db.rs
```diff
@@ -12,9 +12,12 @@
 
 use rusqlite;
 use rusqlite::types::{ToSql, ToSqlOutput};
+use time;
 
+use {repeat_values, to_namespaced_keyword};
```
rnewman: What does this `use` syntax mean?!

db/src/db.rs
```diff
@@ -416,45 +437,438 @@ impl DB {
...
+
+          SELECT t.e0, t.a0, t.v0, t.value_type_tag0, t.tx0, t.added0, t.index_avet0, t.index_vaet0, t.index_fulltext0, t.unique_value0, ':db.cardinality/one', d.rowid, d.v
+          FROM temp.inexact_searches AS t
+          LEFT JOIN datoms AS d
+          ON t.e0 = d.e AND
+             t.a0 = d.a"#;
+
+        let mut stmt = conn.prepare_cached(s)?;
+        stmt.execute(&[])
+            .map(|_c| ())
```
rnewman: Yeah, we really want a `.succeed()` method!

db/src/db.rs
```diff
@@ -416,45 +437,438 @@ impl DB {
...
+          WHERE rid IS NOT NULL AND
+                ((added0 IS 0) OR
+                 (added0 IS 1 AND search_type IS ':db.cardinality/one' AND v0 IS NOT v))"#;
+
+        let mut stmt = conn.prepare_cached(s)?;
+        stmt.execute(&[&tx])
+            .map(|_c| ())
+            .chain_err(|| "Could not insert transaction: failed to retract datoms already present")?;
+
+        Ok(())
```
rnewman: We don't need this — just kill the `?;` right above, no?

db/src/db.rs
```diff
@@ -416,45 +437,438 @@ impl DB {
...
+
+        // Insert datoms that were added and not already present.
+        let s = r#"
+          INSERT INTO datoms (e, a, v, tx, value_type_tag, index_avet, index_vaet, index_fulltext, unique_value)
+          SELECT e0, a0, v0, ?, value_type_tag0,
+                 index_avet0, index_vaet0, index_fulltext0, unique_value0
+          FROM temp.search_results
+          WHERE added0 IS 1 AND ((rid IS NULL) OR ((rid IS NOT NULL) AND (v0 IS NOT v)))"#;
+
+        let mut stmt = conn.prepare_cached(s)?;
```
rnewman: rusqlite is really nice.

db/src/db.rs
```diff
@@ -416,45 +437,438 @@ impl DB {
...
+        /// Assertions that are :db.cardinality/one and not :db.fulltext.
+        let mut non_fts_one: Vec<ReducedEntity> = vec![];
+
+        /// Assertions that are :db.cardinality/many and not :db.fulltext.
+        let mut non_fts_many: Vec<ReducedEntity> = vec![];
+
+        // Transact [:db/add :db/txInstant NOW :db/tx].
+        // TODO: allow this to be present in the transaction data.
+        let now = time::get_time();
+        let tx_instant = (now.sec as i64 * 1_000) + (now.nsec as i64 / (1_000_000));
```
rnewman: Please extract a `now_millis()` utility function. We'll need this elsewhere, I imagine, and it'd be nice to test it.

db/src/db.rs
```diff
@@ -416,45 +437,438 @@ impl DB {
...
+                    };
+
+                    let attribute: &Attribute = self.schema.require_attribute_for_entid(&a)?;
+                    if attribute.fulltext {
+                        bail!(ErrorKind::NotYetImplemented(format!("Transacting :db/fulltext entities is not yet implemented: {:?}", entity)))
+                    }
+
+                    // This is our chance to do schema-aware typechecking: to either assert that the
+                    // given value is in the attribute's value set, or (in limited cases) to coerce
+                    // the value into the attribute's value set.
```
rnewman: Add the word TODO? Or do you mean that `to_typed_value` does this to your satisfaction? Clarify the comment, please.

db/src/db.rs
```diff
@@ -416,45 +437,438 @@ impl DB {
...
+                    } else {
+                        non_fts_one.push((e, a, typed_value, added));
+                    }
+                    Ok(())
+                },
+
+                Entity::Retract {
+                    e: entmod::EntidOrLookupRef::Entid(ref e_),
+                    a: ref a_,
+                    v: entmod::ValueOrLookupRef::Value(ref v_) } => {
```
rnewman: This indenting is a little confusing.

db/src/db.rs
```diff
@@ -416,45 +437,438 @@ impl DB {
...
+                        non_fts_many.push((e, a, typed_value, added));
+                    } else {
+                        non_fts_one.push((e, a, typed_value, added));
+                    }
                     Ok(())
                 },
-                // TODO: find a better error type for this.
-                _ => panic!(format!("Transacting entity not yet supported: {:?}", entity))
+
+                _ => bail!(ErrorKind::NotYetImplemented(format!("Transacting this entity is not yet implemented: {:?}", entity)))
```
rnewman: Just use `unimplemented!("Tra…");`

db/src/db.rs
```diff
@@ -416,45 +437,438 @@ impl DB {
...
+                    };
+
+                    let a: i64 = match a_ {
+                        &entmod::Entid::Entid(ref a__) => *a__,
+                        &entmod::Entid::Ident(ref a__) => *self.schema.require_entid(&a__.to_string())?,
+                    };
+
                     let attribute: &Attribute = self.schema.require_attribute_for_entid(&a)?;
+                    if attribute.fulltext {
+                        bail!(ErrorKind::NotYetImplemented(format!("Transacting :db/fulltext entities is not yet implemented: {:?}", entity)))
```
rnewman: `unimplemented!`

db/src/db.rs
```diff
@@ -416,45 +437,438 @@ impl DB {
...
+                    };
+
+                    let a: i64 = match a_ {
+                        &entmod::Entid::Entid(ref a__) => *a__,
+                        &entmod::Entid::Ident(ref a__) => *self.schema.require_entid(&a__.to_string())?,
+                    };
+
+                    let attribute: &Attribute = self.schema.require_attribute_for_entid(&a)?;
+                    if attribute.fulltext {
+                        bail!(ErrorKind::NotYetImplemented(format!("Transacting :db/fulltext entities is not yet implemented: {:?}", entity)))
```
rnewman: `unimplemented!`

db/src/db.rs
```diff
@@ -416,45 +437,438 @@ impl DB {
...
-                _ => panic!(format!("Transacting entity not yet supported: {:?}", entity))
+
+                _ => bail!(ErrorKind::NotYetImplemented(format!("Transacting this entity is not yet implemented: {:?}", entity)))
             }
         }).collect();
 
-        let x: Result<Vec<()>> = r.into_iter().collect();
-        x.map(|_| ())
+        let r: Result<Vec<()>> = r.into_iter().collect();
+        r?;
```
rnewman: Simplify?

db/src/debug.rs
```diff
@@ -12,55 +12,203 @@
...
+pub fn datoms(conn: &rusqlite::Connection, db: &DB) -> Result<Datoms> {
+    datoms_after(conn, db, bootstrap::TX0 - 1)
+}
+
+/// Return the set of datoms in the store with transaction ID strictly greater than the given `tx`,
+/// ordered by (e, a, v, tx).
+///
+/// The datom set returned does not include any datoms of the form [... :db/txInstant ...].
+pub fn datoms_after(conn: &rusqlite::Connection, db: &DB, tx: i64) -> Result<Datoms> {
+    let mut stmt: rusqlite::Statement = conn.prepare("SELECT e, a, v, value_type_tag, tx FROM datoms WHERE tx > ? ORDER BY e ASC, a ASC, v ASC, tx ASC")?;
```
rnewman: `WHERE tx > ? AND a IS NOT ?…", &[&tx, entids::DB_TX_INSTANT]`? Might as well avoid all of those rows…

db/src/debug.rs
```diff
@@ -12,55 +12,203 @@
...
+
+    let r: Result<Vec<_>> = stmt.query_and_then(&[&tx], |row| {
+        let e: i64 = row.get_checked(0)?;
+        let a: i64 = row.get_checked(1)?;
+
+        if a == entids::DB_TX_INSTANT {
+            return Ok(None);
```
rnewman: … so you don't need to do this.

db/src/debug.rs
```diff
@@ -12,55 +12,203 @@
...
+
+        let typed_value = TypedValue::from_sql_value_pair(v, value_type_tag)?;
+        let (value, _) = typed_value.to_edn_value_pair();
+
+        let tx: i64 = row.get_checked(4)?;
+
+        Ok(Some(Datom {
+            e: to_entid(db, e),
+            a: to_entid(db, a),
+            v: value,
```
rnewman: I wonder if it's worth keeping the `TypedValue` wrapper here for `v`. The compiler should ensure that it's efficient, and without it we're strictly losing some convenient info…

db/src/debug.rs
```diff
@@ -12,55 +12,203 @@
...
+            tx: tx,
+            added: None,
+        }))
+    })?.collect();
+
+    Ok(Datoms(r?.into_iter().filter_map(|x| x).collect()))
```
rnewman: I feel there must be a simpler alternative to `filter_map(|x| x)` — perhaps don't collect the failures? Use `filter_map` earlier? Dunno.

db/src/debug.rs
rnewman: It also occurs to me: we had better make sure that users never assign a new entity ID to any of the builtins!

db/src/debug.rs
```diff
@@ -12,55 +12,203 @@
...
 
+        let tx: i64 = row.get_checked(4)?;
+        let added: bool = row.get_checked(5)?;
+
         Ok(Datom {
-            e: to_entid(e),
-            a: to_entid(a),
+            e: to_entid(db, e),
+            a: to_entid(db, a),
             v: value,
```
rnewman: Same comment about `TypedValue`.

db/src/lib.rs
```diff
@@ -35,6 +42,8 @@ mod values;
 
 use edn::symbols;
 
+pub const SQLITE_MAX_VARIABLE_NUMBER: usize = 999;
```
rnewman: This is a compile-time option in SQLite. We should upstream a patch to rusqlite.

rnewman: Note that `SQLITE_MAX_VARIABLE_NUMBER` is 500,000 on Mac OS!

We can get this value at runtime by calling `sqlite3_limit(db, 9, -1)`, where `9` is `SQLITE_LIMIT_VARIABLE_NUMBER`. I don't see that LIMIT constant exported by rusqlite.

This is the correct thing to do. Using `S_M_V_N` is expedient but won't detect a lower runtime limit imposed by `sqlite3_limit`.

Filed https://github.com/jgallagher/rusqlite/issues/220 to expose `S_M_V_N`.

Alternatively, we could consider looping over values and firing off a prepared statement over and over. I don't know which is slower: constructing the concatenated string and running one query with 999 variables, or running a small prepared statement 250 times within a transaction!

db/src/lib.rs
```diff
@@ -43,3 +52,26 @@ pub fn to_namespaced_keyword(s: &str) -> Option<symbols::NamespacedKeyword> {
...
+/// assert_eq!(repeat_values(2, 2), "(?, ?), (?, ?)".to_string());
+/// ```
+pub fn repeat_values(values_per_tuple: usize, tuples: usize) -> String {
+    assert!(values_per_tuple >= 1);
+    assert!(tuples >= 1);
+    assert!(values_per_tuple * tuples < SQLITE_MAX_VARIABLE_NUMBER, "Too many values: {} * {} >= {}", values_per_tuple, tuples, SQLITE_MAX_VARIABLE_NUMBER);
+    // Like "(?, ?, ?)".
+    let inner = format!("({})", repeat("?").take(values_per_tuple).join(", "));
+    // Like "(?, ?, ?), (?, ?, ?)".
+    let values: String = repeat(inner).take(tuples).join(", ");
```
rnewman: I know it's a huge pain in the ass, but I'm really interested to see the tradeoffs between using a fixed static string and a single prepared statement called a few times, perhaps with a mutable values array — i.e., no allocations for each write — versus doing all of this allocation in `repeat_values` and doing a single SQLite library call.

Can you informally measure?

db/src/db.rs
ncalexan: I've filed https://github.com/mozilla/mentat/issues/261 to do better.  Please comment there, take, or mentor that ticket!

db/src/db.rs
ncalexan: Generally I have found not, since it adds a level of indirection.  But Firefox does this for some modules, including Places IIRC.  I'd like to leave them more or unless inline until they're more concrete, or truly unmanageable.

db/src/db.rs
ncalexan: It's equivalent to `use ::{SYMBOL}`, which I have replaced it with.

db/src/db.rs
rnewman: Most of Firefox on all platforms keeps queries inline in code. Reasons:

- They're almost never reused. If they are, it makes the code hard to change later — the "beginners copy and paste; professionals refactor; veterans copy and paste" thing.
- Keeping them elsewhere makes it harder to understand the code, losing context.
- It makes it harder to delete unused queries if they're divorced from their calling code.

db/src/db.rs
```diff
@@ -416,45 +437,438 @@ impl DB {
...
+            let params: Vec<&ToSql> = block.iter().flat_map(|&(ref searchid, ref a, ref value, ref value_type_tag)| {
+                // Avoid inner heap allocation.
+                once(searchid as &ToSql)
+                    .chain(once(a as &ToSql)
+                           .chain(once(value as &ToSql)
+                                  .chain(once(value_type_tag as &ToSql))))
+            }).collect();
+
+            // TODO: cache these statements for selected values of `count`.
+            let values: String = repeat_values(values_per_statement, count);
```
ncalexan: This is asserted in `repeat_values`, so it's in place for all consumers.

db/src/db.rs
```diff
@@ -416,45 +437,438 @@ impl DB {
...
+    /// It is assumed that the attribute `a` in each lookup-ref is `:db/unique`, so that at most one
+    /// matching [e a v] triple exists.  (If this is not true, some matching entid `e` will be
+    /// chosen non-deterministically, if one exists.)
+    ///
+    /// Returns a map &(a, v) -> e, to avoid cloning potentially large values.  The keys of the map
+    /// are exactly those (a, v) pairs that have an assertion [e a v] in the datom store.
+    pub fn resolve_avs<'a>(&self, conn: &rusqlite::Connection, avs: &'a [AVPair]) -> Result<AVMap<'a>> {
+        // Start search_id's at some identifiable number.
+        let initial_search_id = 2000;
+        let values_per_statement = 4;
```
ncalexan: Done.

db/src/db.rs
```diff
@@ -416,45 +437,438 @@ impl DB {
...
+                    .chain(once(a as &ToSql)
+                           .chain(once(value as &ToSql)
+                                  .chain(once(value_type_tag as &ToSql))))
+            }).collect();
+
+            // TODO: cache these statements for selected values of `count`.
+            let values: String = repeat_values(values_per_statement, count);
+            let s: String = format!("WITH t(search_id, a, v, value_type_tag) AS (VALUES {}) SELECT t.search_id, d.e \
+                                     FROM t, all_datoms AS d \
+                                     WHERE d.index_avet IS NOT 0 AND d.a = t.a AND d.value_type_tag = t.value_type_tag AND d.v = t.v", values);
```
ncalexan: Done both.

db/src/db.rs
```diff
@@ -416,45 +437,438 @@ impl DB {
...
+                Ok((row.get_checked(0)?, row.get_checked(1)?))
+            })?.collect();
+            m
+        }).collect::<Result<Vec<Vec<(i64, Entid)>>>>();
+
+        // Flatten.
+        let results: Vec<(i64, Entid)> = results?.as_slice().concat();
+
+        // Create map [a v] -> e.
+        let m: HashMap<&'a AVPair, Entid> = results.into_iter().map(|(search_id, entid)| {
```
ncalexan: Yeah, there's a bunch of ways to phrase this.  I want it this way to demonstrate the chunking technique with higher order functions (partly for myself!).  I'm going to keep it as it is for now, but I've filed https://github.com/mozilla/mentat/issues/262 to do better eventually.

db/src/db.rs
```diff
@@ -416,45 +437,438 @@ impl DB {
...
+            // opaque.)
+            //
+            // N.b.: temp goes on index name, not table name.  See http://stackoverflow.com/a/22308016.
+            r#"CREATE UNIQUE INDEX IF NOT EXISTS temp.search_results_unique ON search_results (e0, a0, v0, value_type_tag0)"#,
+        ];
+
+        for statement in &statements {
+            let mut stmt = conn.prepare_cached(statement)?;
+            stmt.execute(&[])
+                .map(|_c| ())
```
ncalexan: Yeah, that or `.success()`.  I don't really care to define this right now; it could be a good first bug.

db/src/db.rs
```diff
@@ -416,45 +437,438 @@ impl DB {
...
+                    attribute.value_type == ValueType::Ref,
+                    attribute.fulltext,
+                    attribute.unique_value))
+            }).collect();
+            let block = block?;
+
+            // `params` reference computed values in `block`.
+            let params: Vec<&ToSql> = block.iter().flat_map(|&(ref e, ref a, ref value, ref value_type_tag, added, index_avet, index_vaet, index_fulltext, unique_value)| {
+                // Avoid inner heap allocation.
+                // TODO: extract some finite length iterator to make this less indented!
```
ncalexan: I filed https://github.com/mozilla/mentat/issues/261 to do better in general.

db/src/db.rs
ncalexan: Yeah, could do.

db/src/db.rs
```diff
@@ -416,45 +437,438 @@ impl DB {
...
+
+        Ok(())
+    }
+
+    /// Insert search rows into temporary search tables.
+    ///
+    /// Eventually, the details of this approach will be captured in
+    /// https://github.com/mozilla/mentat/wiki/Transacting:-entity-to-SQL-translation.
+    fn insert_non_fts_searches<'a>(&self, conn: &rusqlite::Connection, entities: &'a [ReducedEntity], tx: Entid, search_type: SearchType) -> Result<()> {
+        let values_per_statement = 10;
```
ncalexan: Done.

db/src/db.rs
ncalexan: Done in follow-up; I'm just going to leave it for now.  There's lots of churn in this area coming.

db/src/db.rs
ncalexan: Sadly `unimplemented!` doesn't include an error message, and I've been consistent with `NotYetImplemented` and a description, so I'm going to keep it this way.  Eventually, these will all go away!

db/src/db.rs
ncalexan: Sadly, the "simplification" is the turbofish (`::<>`) which is even harder to read.

db/src/debug.rs
ncalexan: I thought about this, and decided to do it as post-processing.  There's other rewriting happening and I _think_ the final form of these test functions might look a lot like a pattern matcher on `edn::Values`, in which case I don't want to filter at all.

This is really just a matter of taste, and I want to be as close to the actual DB contents in memory as I can be -- for now.

db/src/debug.rs
ncalexan: Yes, although if you footgun yourself you footgun yourself :)  I think the route to only letting the transactor allocate entids is via #190.

db/src/debug.rs
ncalexan: There might be a simpler way, but I don't know it.  Something has to `collect()` the inner `Result` instances into an outer `Result` before we can `filter_map`.  It might be that `for` loops are better in some situations -- but I reach for the higher order functions first.

db/src/lib.rs
ncalexan: There's a ton involved in doing this, and it's not time yet.  I've filed https://github.com/mozilla/mentat/issues/263 to track this for real.  This really shows when you try to import a Places database, which will be ... a while.

db/src/debug.rs
ncalexan: This is true, but the real goal of these debugging routines is to compare to EDN, c.f. #188.  With that in mind, I'm not going to expose `TypedValue` just yet.  (Now, there's `TypedValue::to_edn_value_pair` already, which could be made `Into<edn::Value>` pretty easily, so this might not be much work, but let's see if there's frustration mapping `a` to `Attribute` before we do that work.)

db/src/db.rs
ncalexan: I moved the `type` definitions to types and lifted the `use` statements to the top of the file.  These were just oversights.

ncalexan: Also, re: file length -- yes, this is long.  I have Grand Plans for a split between the DB layer and the transaction processing layer but I don't want to make the split until more of the transactor is in place.  So eventually I expect two files of similar size, but right now it's just one big file.  Sorry!

PR #217, opened 2017-01-31
Minor EDN cleanup.

rnewman:
I noticed a misnamed entry in a docstring, and `to_reverse` should really be named `to_reversed`.

This also includes my `PlainSymbol` additions — makes my next PR shorter.

### DIFF
--- a/edn/src/symbols.rs
+++ b/edn/src/symbols.rs
@@ -77,6 +77,32 @@ impl PlainSymbol {
 
         PlainSymbol(n)
     }
+
+    /// Return the name of the symbol without any leading '?' or '$'.
+    ///
+    /// ```rust
+    /// # use edn::symbols::PlainSymbol;
+    /// assert_eq!("foo", PlainSymbol::new("?foo").plain_name());
+    /// assert_eq!("foo", PlainSymbol::new("$foo").plain_name());
+    /// assert_eq!("!foo", PlainSymbol::new("!foo").plain_name());
+    /// ```
+    pub fn plain_name(&self) -> &str {
+        if self.is_src_symbol() || self.is_var_symbol() {
+            &self.0[1..]
+        } else {
+            &self.0
+        }
+    }
+
+    #[inline]
+    pub fn is_var_symbol(&self) -> bool {
+        self.0.starts_with('?')
+    }
+
+    #[inline]
+    pub fn is_src_symbol(&self) -> bool {
+        self.0.starts_with('$')
+    }
 }
 
 impl NamespacedSymbol {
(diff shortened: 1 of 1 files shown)


edn/src/symbols.rs
```diff
@@ -77,6 +77,32 @@ impl PlainSymbol {
...
+
+    /// Return the name of the symbol without any leading '?' or '$'.
+    ///
+    /// ```rust
+    /// # use edn::symbols::PlainSymbol;
+    /// assert_eq!("foo", PlainSymbol::new("?foo").plain_name());
+    /// assert_eq!("foo", PlainSymbol::new("$foo").plain_name());
+    /// assert_eq!("!foo", PlainSymbol::new("!foo").plain_name());
+    /// ```
+    pub fn plain_name(&self) -> &str {
```
jsantell: Should `NamespacedKeyword` also have this? `if self.is_backward() { self.name[1..] } else { self.name }`?

edn/src/symbols.rs
```diff
@@ -77,6 +77,32 @@ impl PlainSymbol {
...
+        if self.is_src_symbol() || self.is_var_symbol() {
+            &self.0[1..]
+        } else {
+            &self.0
+        }
+    }
+
+    #[inline]
+    pub fn is_var_symbol(&self) -> bool {
+        self.0.starts_with('?')
```
jsantell: nit: I think we use double quotes everywhere I've seen for &str primitives, but not sure if that's even the majority of cases outside of the chunks I'm familiar with

edn/src/symbols.rs
jsantell: Now I think it maybe should be `is_backwards` rather than `is_backward`

rnewman: The UK tends to say forwards/backwards, and the US to say forward/backward, I think. I'm happy either way.

edn/src/symbols.rs
rnewman: Single quotes are for characters. Double quotes are for strings.

edn/src/symbols.rs
jsantell: I think `forward` and `backwards` seems more natural to me, to complicate things, so let's do nothing :+1:

edn/src/symbols.rs
jsantell: As in `char` and `str`, or an idiom? Either way TIL, should change `.is_backward` in that case to be consistent!

rnewman: As in `char`, just like C. I'm not sure if it's Frommed into a `&str`, or if there are two code paths, but I figured the `char` approach was clearest.

PR #222, opened 2017-02-01
Add helper functions constructing OrderedFloat and BigInt to edn crate

### LINKED ISSUE #198: Add helper functions constructing OrderedFloat and BigInt to edn crate
Right now, real consumers of `edn::Value` need to import `OrderedFloat` and `BigInt` in order to pattern match `Value` completely.  That means to `extern crate` declarations, which is awkward.  This ticket tracks exposing simple constructors in the `edn` crate.  For example, maybe
```rust
impl Value {
    from_float(x: f64) -> Value { ... }
    from_bigint(x: &str) -> Value { ... }
}
```

victorporof:
Fixes #198

Notes:
* I don't know how to silence the ToBigInt import otherwise.
* I've opted to using a more idiomatic `From<>` implementation for floats, but having an explicit `from_bigint` for strings, as we also have a `Value::Text`.
* Seemed like a good idea to introduce a `NewValueError` and return a Result instead of an Option in the above case.

### DIFF
--- a/edn/src/types.rs
+++ b/edn/src/types.rs
@@ -8,12 +8,14 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
+#![allow(unused_imports)]
+
 use std::collections::{BTreeSet, BTreeMap, LinkedList};
 use std::cmp::{Ordering, Ord, PartialOrd};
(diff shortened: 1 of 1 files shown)


victorporof:
Updated as per comments in #223

rnewman:
85da91a

PR #223, opened 2017-02-01
Don't depend on num and ordered-float in the db and query crates

victorporof:
Depends on https://github.com/mozilla/mentat/pull/222

I've had to remove some derived traits from various data types when switching from OrderedFloat<f64> to the f64 primitive. I don't know what that implies, hoping for guidance.

### DIFF
--- a/db/src/types.rs
+++ b/db/src/types.rs
@@ -11,8 +11,7 @@
 #![allow(dead_code)]
 
 use std::collections::{BTreeMap};
-
-use ordered_float::{OrderedFloat};
+use edn::OrderedFloat;
 
 extern crate mentat_core;
 
--- a/query/src/lib.rs
+++ b/query/src/lib.rs
@@ -31,12 +31,8 @@
 ///! a tradeoff against well-typed function signatures and other such boundaries.
 
 extern crate edn;
-extern crate num;
-extern crate ordered_float;
 
-use num::BigInt;
-use ordered_float::OrderedFloat;
-use edn::{NamespacedKeyword, PlainSymbol};
+use edn::{BigInt, OrderedFloat, NamespacedKeyword, PlainSymbol};
 
 pub type SrcVarName = String;          // Do not include the required syntactic '$'.
 
--- a/db/Cargo.toml
+++ b/db/Cargo.toml
@@ -5,9 +5,6 @@ version = "0.0.1"
 [dependencies]
 error-chain = "0.8.0"
 lazy_static = "0.2.2"
-# TODO: don't depend on num and ordered-float; expose helpers in edn abstracting necessary constructors.
-num = "0.1.35"
-ordered-float = "0.4.0"
 
 [dependencies.rusqlite]
 version = "0.9.3"
--- a/db/src/lib.rs
+++ b/db/src/lib.rs
@@ -12,8 +12,6 @@
 extern crate error_chain;
 #[macro_use]
 extern crate lazy_static;
-extern crate num;
-extern crate ordered_float;
 extern crate rusqlite;
 
 extern crate edn;
--- a/edn/src/lib.rs
+++ b/edn/src/lib.rs
@@ -21,5 +21,7 @@ pub mod parse {
     include!(concat!(env!("OUT_DIR"), "/edn.rs"));
 }
 
-pub use self::types::Value;
-pub use self::symbols::{Keyword, NamespacedKeyword, PlainSymbol, NamespacedSymbol};
+pub use ordered_float::OrderedFloat;
+pub use num::BigInt;
+pub use types::Value;
+pub use symbols::{Keyword, NamespacedKeyword, PlainSymbol, NamespacedSymbol};
--- a/query-parser/Cargo.toml
+++ b/query-parser/Cargo.toml
@@ -5,7 +5,6 @@ version = "0.0.1"
 [dependencies]
 combine = "2.1.1"
 matches = "0.1"
-ordered-float = "0.4.0"
 
 [dependencies.edn]
   path = "../edn"
--- a/query-parser/src/parse.rs
+++ b/query-parser/src/parse.rs
@@ -244,10 +244,9 @@ mod test {
     extern crate combine;
     extern crate edn;
     extern crate mentat_query;
-    extern crate ordered_float;
 
     use self::combine::Parser;
-    use self::ordered_float::OrderedFloat;
+    use self::edn::OrderedFloat;
     use self::mentat_query::{
         Element,
         FindSpec,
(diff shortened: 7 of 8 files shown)


db/src/types.rs
```diff
@@ -38,12 +36,13 @@ pub enum ValueType {
 
 /// Represents a Mentat value in a particular value set.
 // TODO: expand to include :db.type/{instant,url,uuid}.
-#[derive(Clone,Debug,Eq,Hash,Ord,PartialOrd,PartialEq)]
+// Floats and doubles can't implement Eq and Hash.
+#[derive(Clone,Debug,PartialOrd,PartialEq)]
```
rnewman: We needed `Hash` so we're able to put `edn::Value` instances into a map. That's why we use `OrderedFloat` in the first place. I'm not sure if we'll need to use `TypedValue`s as keys in a map. What do you think, @ncalexan?

query/src/lib.rs
```diff
@@ -50,11 +46,12 @@ pub enum SrcVar {
...
 
 /// These are the scalar values representable in EDN.
-#[derive(Clone,Debug,Eq,PartialEq)]
+// Floats and doubles only implement PartialEq, but not Eq.
+#[derive(Clone,Debug,PartialEq)]
 pub enum NonIntegerConstant {
     Boolean(bool),
-    BigInteger(BigInt),
-    Float(OrderedFloat<f64>),
+    BigInteger(String),
```
rnewman: I'll need this as numeric, I think; `query` (and `query_parser`, which hasn't merged yet) need to pass along `BigInt`. I don't think there's a Rust/Cargo way of doing that neatly.

edn/src/types.rs
```diff
@@ -143,6 +162,16 @@ impl Value {
...
     def_as!(as_map, Map, BTreeMap<Value, Value>);
+
+    pub fn from_bigint(src: &str) -> Result<Value, NewValueError> {
+        Ok(Value::BigInteger(src.parse::<BigInt>()?))
+    }
+}
+
+impl From<f64> for Value {
+    fn from(src: f64) -> Value {
+        Value::Float(OrderedFloat::from(src))
```
rnewman: Nicely idiomatic!

query/src/lib.rs
victorporof: So you want to have u/i64 as source for BigInteger? How should the sign be handled?

victorporof:
This PR is for removing num and ordered-float dependencies. Adding helper functions is handled in #222. The PRs are split since I expect this one (#223) to take longer to finish based on feedback.

query/src/lib.rs
rnewman: No, I mean `query` should depend on `BigInt`. `query` represents parsed queries, and those can contain big integers.

`OrderedFloat` is an implementation detail: `Value` only needs that so that `Value` instances can be put into data structures that require hashing. `BigInt` is an intrinsic part of the data being represented; the only improvements we can make there are to provide useful factories so that *some kinds* of calling code can avoid building their own `BigInt` instances.

victorporof: Got it. I'll add BigInt back in this particular case.

db/src/types.rs
ncalexan: I'm inclined to expect we will not.  I don't feel strongly about having raw `f64` here.

edn/src/types.rs
```diff
@@ -143,6 +162,16 @@ impl Value {
     def_as!(as_list, List, LinkedList<Value>);
     def_as!(as_set, Set, BTreeSet<Value>);
     def_as!(as_map, Map, BTreeMap<Value, Value>);
+
+    pub fn from_bigint(src: &str) -> Result<Value, NewValueError> {
```
ncalexan: This might be simpler if we return `Option<Value>`.  Might call it `edn::Value::bigint_from(src: Into<str>)` as well.

victorporof: AFAIK Result is always preferable to Option when errors *can* be specified. However, if they *can* be, doesn't mean that we necessarily have to. Do you see the potential NewValueError sum type to grow in the future? Is it worth allowing consumers to know exactly why creating an edn::Value fails?

victorporof: (I know none of this is *really* that important, I'm just trying to better understand things)

rnewman: In my experience, detailed error reporting is often a waste of effort: it's only worthwhile when the caller's behavior needs to change based on what went wrong.

In this case, the caller should almost always have a valid string — it parsed it from user input using a parser that includes bigint syntax — and the only error state we need to communicate is "we couldn't turn this into a bigint for some reason". The caller already has the string and can use that for error reporting if it chooses. The user will probably be able to see what's wrong without more help.

The other scenario to consider is that often it's *not* failure *per se*; it might just have been a speculative attempt that wasn't the right branch. Parsing is often phrased in terms of "is it an X, a Y, or a Z?". With `satisfy_map`, the parsing returns an `Option<T>` to indicate that this branch of the parse didn't succeed. The error message presented to the user would then be something like "expected X, Y, or Z; got <input>". A detailed error from `parse_x` is of no use, and would be discarded immediately.

A final thought: one reason for `Result` is that it requires callers to check the value. That implies it's used for side-effecting things like filesystem writes; to me it feels more like something where the error is important, even primary.

victorporof:
I've updated the helper functions to return Result instead of Option. Landed them in https://github.com/mozilla/mentat/pull/222

I'm having a little trouble understanding how the num and ordered-float dependencies can be removed from the db and query crates if BigInt and OrderedFloat types are going to be used anyway there. At least that's what I'm understanding from https://github.com/mozilla/mentat/pull/223#pullrequestreview-19571757 and https://github.com/mozilla/mentat/pull/223#pullrequestreview-19572301

rnewman:
We can't remove `BigInt` from `query` — it needs to represent those values. We might be able to elide the existence of `OrderedFloat`, at least in the external API, by using `as_float` (which would return an `Option<f64>`, not an `Option<OrderedFloat<f64>>`).

If the calling code never needs to explicitly refer to types defined in one of these two libraries, it doesn't need to refer to the crate, and all is good. Similarly, if we have re-exported factory functions, we don't need to construct instances of those types by reference to the external crates, and all is good.

The other approach here is to `pub use ordered_float::OrderedFloat` and such, which allows reference to `edn::OrderedFloat`. I think. Give thiat a shot?

victorporof:
> If the calling code never needs to explicitly refer to types defined in one of these two libraries, it doesn't need to refer to the crate, and all is good.

How can it not refer to these types (and I assume you mean `OrderedFloat`, `BigInt`) and not refer to their respective library crates, without exporting them via `pub use` in the edn crate?

victorporof:
This is ready for re-review. @ncalexan or @rnewman ping.

victorporof:
@ncalexan or @rnewman gentle review ping.

PR #224, opened 2017-02-01
Several small idiomatic changes to the edn crate

victorporof:


edn/src/types.rs
```diff
@@ -199,15 +199,13 @@ fn to_ord(value: &Value) -> i32 {
...
 }
 
 pub fn to_keyword(namespace: Option<&str>, name: &str) -> Value {
-    if let Some(ns) = namespace {
-        return Value::NamespacedKeyword(symbols::NamespacedKeyword::new(ns, name));
-    }
-    return Value::Keyword(symbols::Keyword::new(name));
+    namespace.map_or_else(
+        || Value::Keyword(symbols::Keyword::new(name)),
+        |ns| Value::NamespacedKeyword(symbols::NamespacedKeyword::new(ns, name)))
```
rnewman: This is a great spot to add a short doc test.

edn/tests/tests.rs
```diff
@@ -946,8 +944,8 @@ fn test_is_and_as_type_helper_functions() {
             value.is_map(),
         ];
 
-        for j in 0..values.len() {
-            assert_eq!(j == i, is_result[j]);
+        for (j, result) in is_result.iter().enumerate() {
```
rnewman: If you're going to do this, you should also `assert_eq!(values.len(), is_result.len())` outside this inner loop. Previously the array indexing bounds check would panic if these arrays were mis-sized.

edn/src/types.rs
victorporof: Will add.

edn/tests/tests.rs
victorporof: Good idea.

victorporof:
Addressed comments and rebased.

PR #228, opened 2017-02-02
Add utilities for defining parsers. (#218)

### LINKED ISSUE #218: Extract macro that builds combine fn_parser instances
We have about a billion definitions like the following at https://github.com/mozilla/mentat/blob/rust/tx-parser/src/lib.rs#L36:
```rust
    fn integer() -> TxParser<i64, I> {
        fn_parser(Tx::<I>::integer_, "integer")
    }

    fn integer_(input: I) -> ParseResult<i64, I> {
        return satisfy_map(|x: Value| if let Value::Integer(y) = x {
                Some(y)
            } else {
                None
            })
            .parse_stream(input);
    }
```

They all look the same: define a function named `foo_` that produces a `ParseResult`, and a function named `foo` that turns it into an `fn_parser`; annotate the `fn_parser` with a string based on the name of the function.  This could be captured as a macro.

This is not a _[good first bug]_ unless you're strong in Rust.

rnewman:
Fixes #218.

Note that this includes `m_p_s`, which was already reviewed — it's just been moved out.

I also moved out `ResultParser`, which is nice and general.

### DIFF
--- a/parser-utils/Cargo.toml
+++ b/parser-utils/Cargo.toml
@@ -1,6 +1,7 @@
 [package]
 name = "mentat_parser_utils"
 version = "0.1.0"
-authors = ["Victor Porof <vporof@mozilla.com>"]
+authors = ["Victor Porof <vporof@mozilla.com>", "Richard Newman <rnewman@mozilla.com>"]
 
 [dependencies]
+combine = "2.1.1"
--- a/parser-utils/src/lib.rs
+++ b/parser-utils/src/lib.rs
@@ -8,6 +8,14 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
+extern crate combine;
+
+use self::combine::ParseResult;
+use self::combine::combinator::{Expected, FnParser};
+
+// Nothing about this is specific to the type of parser.
+pub type ResultParser<O, I> = Expected<FnParser<I, fn(I) -> ParseResult<O, I>>>;
+
 /// `assert_parses_to!` simplifies some of the boilerplate around running a
 /// parser function against input and expecting a certain result.
 #[macro_export]
@@ -17,4 +25,13 @@ macro_rules! assert_parses_to {
         let result = par.parse(&$input[..]);
         assert_eq!(result, Ok(($expected, &[][..])));
     }}
+}
+
+/// `satisfy_unwrap!` makes it a little easier to implement a `satisfy_map`
+/// body that matches a particular `Value` enum case, otherwise returning `None`.
+#[macro_export]
+macro_rules! satisfy_unwrap {
+    ( $cas: path, $var: ident, $body: block ) => {
+        satisfy_map(|x: edn::Value| if let $cas($var) = x $body else { None })
+    }
 }
--- a/parser-utils/src/lib.rs
+++ b/parser-utils/src/lib.rs
@@ -10,8 +10,8 @@
 
 extern crate combine;
 
-use self::combine::ParseResult;
-use self::combine::combinator::{Expected, FnParser};
+use combine::ParseResult;
+use combine::combinator::{Expected, FnParser};
 
 // Nothing about this is specific to the type of parser.
 pub type ResultParser<O, I> = Expected<FnParser<I, fn(I) -> ParseResult<O, I>>>;
--- a/query-parser/src/parser_util.rs
+++ b/query-parser/src/parser_util.rs
@@ -1,0 +1,88 @@
+// Copyright 2016 Mozilla
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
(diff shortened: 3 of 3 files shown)


parser-utils/Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [package]
 name = "mentat_parser_utils"
 version = "0.1.0"
-authors = ["Victor Porof <vporof@mozilla.com>"]
+authors = ["Victor Porof <vporof@mozilla.com>", "Richard Newman <rnewman@mozilla.com>"]
```
rnewman: I'm assuming that these will be useful for people asking for review on changes. Is that a valid assumption?

parser-utils/src/lib.rs
```diff
@@ -8,6 +8,14 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
+extern crate combine;
+
+use self::combine::ParseResult;
+use self::combine::combinator::{Expected, FnParser};
```
victorporof: Is `self` necessary here? I'm having a hard time understanding why.

parser-utils/src/lib.rs
```diff
@@ -8,6 +8,14 @@
...
+
+// Nothing about this is specific to the type of parser.
+pub type ResultParser<O, I> = Expected<FnParser<I, fn(I) -> ParseResult<O, I>>>;
```
victorporof: What is the intended use for this? Worth adding more documentation here? This is also copied in `parser_util.rs`.

parser-utils/src/lib.rs
```diff
@@ -17,4 +25,13 @@ macro_rules! assert_parses_to {
...
         assert_eq!(result, Ok(($expected, &[][..])));
     }}
-}
\ No newline at end of file
+}
+
+/// `satisfy_unwrap!` makes it a little easier to implement a `satisfy_map`
+/// body that matches a particular `Value` enum case, otherwise returning `None`.
+#[macro_export]
+macro_rules! satisfy_unwrap {
```
victorporof: It seems that `satisfy_unwrap` is both in `lib.rs` and in `parser_utils.rs`. I assume that's not intentional?

query-parser/src/parser_util.rs
```diff
@@ -0,0 +1,88 @@
...
+/// In practice this allows you to simply pass a function that accepts an `&edn::Value` and
+/// returns an `Option<$result_type>`: if a suitable value is at the front of the stream,
+/// it will be converted and returned by the parser; otherwise, the parse will fail.
+macro_rules! def_value_satisfy_parser_fn {
+    ( $parser: ident, $name: ident, $result_type: ty, $transformer: path ) => {
+        def_value_parser_fn!($parser, $name, $result_type, input, {
+            satisfy_map(|x: edn::Value| $transformer(&x)).parse_stream(input)
+        });
+    }
+}
```
victorporof: Where are all of these macro rules used? Are some deletions missing from somewhere else?

parser-utils/src/lib.rs
victorporof: Note: that's because I don't know rust, not because I think the code is bad.

parser-utils/Cargo.toml
victorporof: Probably?

parser-utils/src/lib.rs
rnewman: Force of habit! This code came from a module. This — that is, a `lib.rs` file — is the *only* place in Rust you can get away with not using `self::` to qualify an import to the crate you just imported.

`use` expressions are all absolute paths, so if you're in module `foo`, `use thing::bar` means `use ::thing::bar`, not `use foo::thing::bar`.

I'll clean this up.

parser-utils/src/lib.rs
rnewman: You'll see the use of this in #211.

parser-utils/src/lib.rs
rnewman: Deletion was lost in a rebase :/ Fixed.

query-parser/src/parser_util.rs
rnewman: #211.

parser-utils/src/lib.rs
```diff
@@ -8,6 +8,14 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
+extern crate combine;
+
+use combine::ParseResult;
+use combine::combinator::{Expected, FnParser};
+
+// Nothing about this is specific to the type of parser.
+pub type ResultParser<O, I> = Expected<FnParser<I, fn(I) -> ParseResult<O, I>>>;
```
victorporof: Some kind of explanation about what this is intended for would still be nice (at least for people like me).

rnewman:
Landed in 5d74f1ee9455cf4084ec2b96a0103503a0a64ae0 with an explanator comment for `ResultParser`.

PR #238, opened 2017-02-02
Parse and display EDN values for NaN, +Infinity and -Infinity. Fixes …

jsantell:
…#232

@victorporof can you elaborate on what EDN escaping you're talking about in the TODO? And is there any other Float formatting issues you had in mind when writing that TODO?

While we're here, is `null` a special EDN type? From what I gather, this should be printed as `nil` (that's what our parser looks for, anyway)

### DIFF
--- a/edn/src/types.rs
+++ b/edn/src/types.rs
@@ -11,6 +11,7 @@
 use std::collections::{BTreeSet, BTreeMap, LinkedList};
 use std::cmp::{Ordering, Ord, PartialOrd};
 use std::fmt::{Display, Formatter};
+use std::f64;
 
 use symbols;
 use num::BigInt;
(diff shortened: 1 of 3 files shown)


edn/src/types.rs
```diff
@@ -54,7 +55,19 @@ impl Display for Value {
             Boolean(v) => write!(f, "{}", v),
             Integer(v) => write!(f, "{}", v),
             BigInteger(ref v) => write!(f, "{}N", v),
-            Float(OrderedFloat(v)) => write!(f, "{}", v),
+            Float(ref v) => {
```
jsantell: [Matching on constants have been deprecated](https://github.com/rust-lang/rfcs/pull/1445), is there a more idiomatic way than if/elses?

victorporof:
@jsantell Should make sure that escaped characters in strings are printed/parsed in an isomorphic manner. See https://github.com/edn-format/edn#strings

Nil is indeed an EDN type: https://github.com/edn-format/edn#nil Should be printed as `nil`, not `null` though, https://github.com/mozilla/mentat/blob/rust/edn/src/edn.rustpeg#L31

edn/src/types.rs
victorporof: This is fine.

victorporof:
Also filed https://github.com/mozilla/mentat/issues/240

victorporof:
Sorry you'll have to rebase this on top of https://github.com/mozilla/mentat/commit/c038c110173e6e476a93f72720ad67fd1bb4da26

Please make sure the formatting respects the other rules in the peg file.

Issue #240, opened 2017-02-03
Make sure edn printing/parsing is isomorphic


victorporof:
A test would be nice.

PR #241, opened 2017-02-03
Ensure printing/parsing edn strings is isomorphic

victorporof:
Fixes https://github.com/mozilla/mentat/issues/240

### DIFF
--- a/edn/src/edn.rustpeg
+++ b/edn/src/edn.rustpeg
@@ -32,10 +32,10 @@ pub nil -> Value =
     "nil" { Value::Nil }
 
 pub nan -> Value =
-    "#f NaN" { Value::Float(OrderedFloat(NAN)) }
+    "#f" __ "NaN" { Value::Float(OrderedFloat(NAN)) }
 
 pub infinity -> Value =
-    "#f" s:$(sign) "Infinity" {
+    "#f" __ s:$(sign) "Infinity" {
         Value::Float(OrderedFloat(if s == "+" { INFINITY } else { NEG_INFINITY }))
     }
 
--- a/edn/src/edn.rustpeg
+++ b/edn/src/edn.rustpeg
@@ -107,17 +107,17 @@ pub keyword -> Value =
     }
 
 pub list -> Value =
-    "(" v:(value)* ")" {
+    "(" __ v:(value)* __ ")" {
         Value::List(LinkedList::from_iter(v))
     }
 
 pub vector -> Value =
-    "[" v:(value)* "]" {
+    "[" __ v:(value)* __ "]" {
         Value::Vector(v)
     }
 
 pub set -> Value =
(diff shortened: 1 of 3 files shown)


edn/src/edn.rustpeg
```diff
@@ -98,17 +98,17 @@ pub keyword -> Value =
     }
 
 pub list -> Value =
-    "(" v:(value)* ")" {
+    "(" __ v:(value)* __ ")" {
```
ncalexan: Oh, interesting, I thought I'd covered this with https://github.com/mozilla/mentat/blob/90c1c59881b1bbef1394961e86add1d95c363617/edn/src/edn.rustpeg#L128 but if there are no values, then you need the whitespace.  Good catch!

edn/src/types.rs
```diff
@@ -83,8 +83,8 @@ impl Display for Value {
             }
             Map(ref v) => {
                 write!(f, "{{")?;
-                for (key, val) in v {
-                    write!(f, " :{} {}", key, val)?;
+                for (index, (key, val)) in v.iter().enumerate() {
```
ncalexan: The leading colon is definitely wrong, so good fix there, but generally EDN does not include commas -- they're whitespace.  Why would we include commas in `Map` but not the other container types?  Because of the pairing?  I'd prefer not to include them, which agrees with Clojure{Script?, although I could be argued down.  @rnewman, do you have a strong opinion?

ncalexan: Also, consider using `join` here, which is probably more efficient.

victorporof: I'll remove the commas

victorporof:
Addressed comments.

edn/tests/tests.rs
```diff
@@ -53,6 +53,7 @@ fn test_nil() {
 
 #[test]
 fn test_nan() {
+    assert_eq!(nan("#fNaN").unwrap(), Float(OrderedFloat(f64::NAN)));
```
rnewman: I don't believe this is correct.

`#fNaN` is itself a tag, so `#fNaN` should not match the `nan` production. Whitespace is required: `#f NaN`. This test should instead assert that `#fNaN` *does not* match `nan`, right?

@victorporof, am I misreading this change?

jsantell: good point. In that case, we should explicitly force the /required/ white space.

victorporof: @rnewman, makes sense. If that's the case, whitespace should be enforced. Is that the case for + and -Infinity as well?

victorporof: (the reason I made the change in the first place was because # +Infinity and # -Infinity wasn't parseable).

victorporof: I've filed https://github.com/mozilla/mentat/issues/244, let's move the conversation there?

PR #242, opened 2017-02-03
Testing bitflag expansion

jsantell:
Based off of @ncalexan's #214, relevant commit:

https://github.com/mozilla/mentat/pull/242/commits/a021735aefb65a6155fbc1628c2527e1cacd71fc

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -81,6 +81,17 @@ fn test_typed_value() {
     assert!(TypedValue::String("foo".to_string()).is_congruent_with(None));
 }
 
+/// Bit flags used in `flags0` column in temporary tables created during search,
+/// such as the `search_results`, `inexact_searches` and `exact_searches` tables.
+/// When moving to a more concrete table, such as `datoms`, they are expanded out
+/// via these flags and put into their own column rather than a bit field.
+pub enum AttributeBitFlags {
+    IndexAVET     = 1 << 0,
+    IndexVAET     = 1 << 1,
+    IndexFulltext = 1 << 2,
+    UniqueValue   = 1 << 3,
+}
+
 /// A Mentat schema attribute has a value type and several other flags determining how assertions
 /// with the attribute are interpreted.
 ///
@@ -126,6 +137,27 @@ pub struct Attribute {
     pub component: bool,
 }
 
+impl Attribute {
+    /// Combine several attribute flags into a bitfield used in temporary search tables.
+    pub fn flags(&self) -> u8 {
+        let mut flags: u8 = 0;
+
+        if self.index {
+            flags |= AttributeBitFlags::IndexAVET as u8;
+        }
+        if self.value_type == ValueType::Ref {
+            flags |= AttributeBitFlags::IndexVAET as u8;
+        }
+        if self.fulltext {
+            flags |= AttributeBitFlags::IndexFulltext as u8;
+        }
+        if self.unique_value {
+            flags |= AttributeBitFlags::UniqueValue as u8;
+        }
+        flags
+    }
+}
+
 impl Default for Attribute {
     fn default() -> Attribute {
         Attribute {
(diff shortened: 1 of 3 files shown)


db/src/db.rs
```diff
@@ -731,14 +732,19 @@ impl DB {
             .chain_err(|| "Could not update datoms: failed to retract datoms already present")?;
 
         // Insert datoms that were added and not already present.
-        let s = r#"
+        let s = format!(r#"
```
jsantell: This is performance benchmarking, so someone should have an eye on this query other than me

db/src/db.rs
```diff
@@ -37,6 +37,11 @@ pub const CURRENT_VERSION: i32 = 2;
 const TRUE: &'static bool = &true;
 const FALSE: &'static bool = &false;
 
+const INDEX_AVET     : u8 = 1 << 0;
```
ncalexan: nit: for the final patch, comment on what these are and how the scheme works.

db/src/db.rs
```diff
@@ -591,27 +587,35 @@ impl DB {
...
-                                   /* index_fulltext0 */ bool,
-                                   /* unique_value0 */ bool)>> = chunk.map(|&(e, a, ref typed_value, added)| {
+                                   /* flags */ u8)>> = chunk.map(|&(e, a, ref typed_value, added)| {
                 count += 1;
                 let attribute: &Attribute = self.schema.require_attribute_for_entid(&a)?;
 
                 // Now we can represent the typed value as an SQL value.
                 let (value, value_type_tag): (ToSqlOutput, i32) = typed_value.to_sql_value_pair();
 
+                let mut flags: u8 = 0;
```
ncalexan: Consider implementing `Attribute.flags()` or similar.

db/src/db.rs
```diff
@@ -620,18 +624,15 @@ impl DB {
                                   .chain(once(value_type_tag as &ToSql)
                                          .chain(once(&tx as &ToSql)
                                                 .chain(once(to_bool_ref(added) as &ToSql)
-                                                       .chain(once(to_bool_ref(index_avet) as &ToSql)
-                                                              .chain(once(to_bool_ref(index_vaet) as &ToSql)
-                                                                     .chain(once(to_bool_ref(index_fulltext) as &ToSql)
-                                                                            .chain(once(to_bool_ref(unique_value) as &ToSql))))))))))
+                                                       .chain(once(flags as &ToSql)))))))
```
ncalexan: \o/

db/src/db.rs
ncalexan: Yeah, this is early days.  We're still deep in the "make it correct" phase, but I was hoping to get some idea of how much this saves.  We'll be better able to benchmark as the transactor fleshes out.

db/src/db.rs
```diff
@@ -518,10 +523,7 @@ impl DB {
                value_type_tag0 SMALLINT NOT NULL,
```
ncalexan: These tables definitely deserve a comment about the flag compression.  And flags should still be a `TINYINT`, although this doesn't make much difference in SQLite, IIUC.

jsantell:
@ncalexan rebased on that big PR. For implementing `Attribute.flags()`, to clarify, are you talking about:

* Moving BitFlags type to core/src/lib and importing that into db/src/db
* Provide an `Attribute.flags()` function that returns a bitfield containing all of the flags (index_vaet, etc) that we want in the bitfield?

ncalexan:
> Provide an Attribute.flags() function that returns a bitfield containing all of the flags (index_vaet, etc) that we want in the bitfield?

@jsantell, I was thinking of this.

jsantell:
@ncalexan  Ok, made the type changes and `flags()` helper -- good for review

core/src/lib.rs
```diff
@@ -81,6 +81,17 @@ fn test_typed_value() {
...
 }
 
+/// Bit flags used in `flags0` column in temporary tables created during search,
+/// such as the `search_results`, `inexact_searches` and `exact_searches` tables.
+/// When moving to a more concrete table, such as `datoms`, they are expanded out
+/// via these flags and put into their own column rather than a bit field.
+pub enum AttributeBitFlags {
+    IndexAVET     = 1 << 0,
+    IndexVAET     = 1 << 1,
+    IndexFullText = 1 << 2,
```
ncalexan: nit: I think we're considering "fulltext" one word, but I'm not militant about this.

db/src/db.rs
```diff
@@ -56,6 +56,8 @@ const TRUE: &'static bool = &true;
 const FALSE: &'static bool = &false;
 
 /// Turn an owned bool into a static reference to a bool.
```
ncalexan: Spurious change?

db/src/db.rs
```diff
@@ -537,31 +531,27 @@ impl DB {
...
 
+                let flags = attribute.flags();
+
                 Ok((e, a, value, value_type_tag,
                     added,
-                    attribute.index,
-                    attribute.value_type == ValueType::Ref,
-                    attribute.fulltext,
-                    attribute.unique_value))
+                    flags))
```
ncalexan: Maybe just inline this call?  Your choice, your style.

ncalexan:
Also, check the tests of all the crates -- Travis is angry :)

core/src/lib.rs
jsantell: Good call

jsantell: And looks like this is the best way to do this according to #rust

PR #243, opened 2017-02-04
Query algebrizer initial work

rnewman:


rnewman:
Ready for review, @ncalexan. @victorporof, if you have time that would also be great.

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+
+    /// Ensure that the given place can be an entity, and is congruent with existing types.
+    /// This is used for `entity` and `attribute` places in a pattern.
+    fn mark_as_ref(&mut self, value: &PatternNonValuePlace) {
+        // If it's a variable, record that it has the right type.
+        // Ident or attribute resolution errors (the only other check we need to do) will be done
+        // by the caller.
+        if let &PatternNonValuePlace::Variable(ref v) = value {
+            if !self.constrain_var_to_type(v.clone(), ValueType::Ref) {
+                self.mark_known_empty("Couldn't constrain var to Ref.");
```
ncalexan: nit: Can you inject some identification of `v` into the error message?

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+                entry.insert(this_type);
+                true
+            }
+        }
+    }
+
+    /// Ensure that the given place has the correct types to be a tx-id.
+    /// Right now this is mostly unimplemented: we fail hard if anything but a placeholder is
+    /// present.
+    fn mark_as_tx(&mut self, tx: &PatternNonValuePlace) {
```
ncalexan: Reading these functions, I don't quite see what _mark_ means in this context.  Consider elaborating on "Ensure", here and below.

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+    fn mark_known_empty(&mut self, why: &str) {
+        self.is_known_empty = true;
+        println!("{}", why);                   // TODO: proper logging.
+    }
+
+    fn entid_for_ident<'s, 'a>(&self, schema: &'s Schema, ident: &'a NamespacedKeyword) -> Option<i64> {
+        schema.get_entid(&ident.to_string())      // TODO: #291.
+    }
+
+    fn attribute_for_ident<'s, 'a>(&self, schema: &'s Schema, ident: &'a NamespacedKeyword) -> Option<&'s Attribute> {
```
ncalexan: This should be in `Schema`, I think.  I intend to pass `&Attribute` refs around in the DB layer eventually too.

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+        self.entid_for_ident(schema, ident).and_then(|x| schema.attribute_for_entid(x))
+    }
+
+    fn table_for_attribute_and_value<'s, 'a>(&self, attribute: &'s Attribute, value: &'a PatternValuePlace) -> Option<DatomsTable> {
```
ncalexan: You've thought through the two lifetimes here, which I agree are necessary for the generic helper, but I'm not sure are required for the actual functions in this file.  That is, you're electing to pass a `Schema` reference with a different lifetime from `&self` around; could you avoid a lot of this management by having an inner CC-building structure that owns a reference to the schema that has the same lifetime as itself, and then extracting the built CC?

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+                    // field.
+                    None
+                }
+            }
+        } else {
+            Some(DatomsTable::Datoms)
+        }
+    }
+
+    fn table_for_places<'s, 'a>(&self, schema: &'s Schema, attribute: &'a PatternNonValuePlace, value: &'a PatternValuePlace) -> Option<DatomsTable> {
```
ncalexan: What's a _place_?  It's a "slot" in `[e a v tx added]`, right?

Why is this "places", plural?  Is this `table_for_attribute_place_and_value_place`?  If that's the case, then I'm fine with this name.

ncalexan: Throughout this implementation, little snippets of EDN would make this easier to understand.  Things like, "If we have `[... _ v]`, then ... but if we have `[... ?binding v]` then...".  It can be tricky to build those things out of the code and descriptions.

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+        self.get_attribute(schema, pattern).map(|x| x.value_type)
+    }
+}
+
+/// Expansions.
+impl ConjoiningClauses {
+
+    /// Take the contents of `bindings` and generate inter-constraints for the appropriate
+    /// columns into `wheres`.
+    /// For example, a bindings map like
```
ncalexan: Heh.  Best to bump comments; they're good for me (and easy for you) but not valuable for others.

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+    ///
+    /// produces additional constraints:
+    ///
+    /// ```clojure
+    /// [[:= :datoms12.e :datoms13.v] [:= :datoms12.e :datoms14.e]]
+    /// ```
+    pub fn expand_bindings(&mut self) {
+        for cols in self.bindings.values() {
+            if cols.len() > 1 {
+                let ref primary = cols[0];
```
ncalexan: Huh, I did not know this `let ref` syntax.  I think I'd prefer to be consistent about `= &`; do you have strong feelings?

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+                let secondaries = cols.iter().skip(1);
+                for secondary in secondaries {
+                    // TODO: if both primary and secondary are .v, should we make sure
```
ncalexan: Is it possible to witness `a.v = b.v` and have different type tags?  In general, I think throwing _more_ constraints at the SQL optimizer is "good"; it's supposed to be able to winnow efficiently, right?

Why wouldn't requiring two values of different type tags to be equal yield an empty set?  Presumably there's some `UNION` or `JOIN` happening, and we don't yet optimize empty sub-queries away just yet?

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+    /// ```clojure
+    ///   {?foo [:datoms12.e :datoms13.v :datoms14.e]}
+    /// ```
+    ///
+    /// produces additional constraints:
+    ///
+    /// ```clojure
+    /// [[:= :datoms12.e :datoms13.v] [:= :datoms12.e :datoms14.e]]
+    /// ```
+    pub fn expand_bindings(&mut self) {
```
ncalexan: There's an expression of this using `slice::windows(2)` that might be nicer still.  Even if you keep this one, you should know about that one :)

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+        for cols in self.bindings.values() {
+            if cols.len() > 1 {
+                let ref primary = cols[0];
+                let secondaries = cols.iter().skip(1);
+                for secondary in secondaries {
+                    // TODO: if both primary and secondary are .v, should we make sure
+                    // the type tag columns also match?
+                    // We don't do so in the ClojureScript version.
+                    self.wheres.push(ColumnConstraint::EqualsColumn(primary.clone(), secondary.clone()));
+                }
```
ncalexan: Should this clear `self.bindings` or move it to some finalized location for future use?  That is, should we implement this to track the flow of data through the pipeline explicitly?

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+    ///
+    /// ```edn
+    /// [:find ?x :where
+    ///  [?x ?y true]
+    ///  [?z ?y ?x]]
+    /// ```
+    ///
+    /// where `?y` must simultaneously be a ref-typed attribute and a boolean-typed attribute. This
+    /// function deduces that and calls `self.mark_known_empty`. #293.
+    pub fn expand_type_tags(&mut self) {
```
ncalexan: Similarly, this consumes `self.known_types` and `self.extracted_types`, correct?  (It is possible that they're used again, but in a different stage.)

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+        self.mark_as_ref(&pattern.attribute);
+
+        match pattern.entity {
+            PatternNonValuePlace::Placeholder =>
+                // Placeholders don't contribute any bindings, nor do
+                // they constrain the query -- there's no need to produce
+                // IS NOT NULL, because we don't store nulls in our schema.
+                (),
+            PatternNonValuePlace::Variable(ref v) =>
+                self.bind_column_to_var(alias.1.clone(), DatomsColumn::Entity, v.clone()),
```
ncalexan: Can you pull a reference to `alias.1` out of these?  (I'm okay with an unnamed tuple globally but let's help ourselves locally.)

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+            PatternNonValuePlace::Entid(entid) =>
+                self.constrain_column_to_entity(alias.1.clone(), DatomsColumn::Entity, entid),
+            PatternNonValuePlace::Ident(ref ident) => {
+                if let Some(entid) = self.entid_for_ident(schema, ident) {
+                    self.constrain_column_to_entity(alias.1.clone(), DatomsColumn::Entity, entid)
+                } else {
+                    // A resolution failure means we're done here.
+                    self.mark_known_empty("Entity ident didn't resolve.");
```
ncalexan: Throughout: some help about what didn't resolve will be good.  Could be a good first bug, your call.

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+                }
+            }
+        }
+
+        // Determine if the pattern's value type is known.
+        // We do so by examining the value place and the attribute.
+        // At this point it's possible that the type of the value is
+        // inconsistent with the attribute; in that case this pattern
+        // cannot return results, and we short-circuit.
+        let attribute = self.get_attribute(schema, pattern);
```
ncalexan: You extracted a helper for this operation, I think.

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+                if !typed_value.is_congruent_with(value_type) {
+                    // If the attribute and its value don't match, the pattern must fail.
+                    self.mark_known_empty("Value constant not congruent with attribute type.");
+                    return;
+                }
+
+                // TODO: if we don't know the type of the attribute because we don't know the
+                // attribute, we can actually work backwards to the set of appropriate attributes
+                // from the type of the value itself! #292.
+                self.constrain_column_to_constant(alias.1.clone(), DatomsColumn::Value, c.clone().into_typed_value());
```
ncalexan: Can this latter at least be the `typed_value` you cloned earlier?

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+    fn add_attribute(schema: &mut Schema, e: Entid, a: Attribute) {
+        schema.schema_map.insert(e, a);
+    }
+
+    #[test]
+    fn test_unknown_ident() {
+        let mut cc = ConjoiningClauses::new();
+        let schema = Schema::new();
+
+        cc.apply_pattern(&schema, &Pattern {
```
ncalexan: Please use some small harness that expresses the inputs, at least, in EDN.  Trying to parse these tests is challenging.

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+            source: None,
+            entity: PatternNonValuePlace::Variable(Variable(PlainSymbol::new("?x"))),
+            attribute: PatternNonValuePlace::Ident(NamespacedKeyword::new("foo", "bar")),
+            value: PatternValuePlace::Constant(NonIntegerConstant::Boolean(true)),
+            tx: PatternNonValuePlace::Placeholder,
+        });
+
+        assert!(cc.is_known_empty);
```
ncalexan: Probably worth asserting the error message as well, at least in the early days.  There are lots of paths to the final cut that constrains the space; let's try to actually explore as many of the paths as we can.

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+
+    #[test]
+    fn test_apply_simple_pattern() {
+        let mut cc = ConjoiningClauses::new();
+        let mut schema = Schema::new();
+
+        associate_ident(&mut schema, ":foo/bar", 99);
+        add_attribute(&mut schema, 99, Attribute {
+            value_type: ValueType::Boolean,
+            ..Default::default()
```
ncalexan: Huh, interesting approach.  Cheap builder pattern.  I put this in the TIL of the notes.

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+        assert_eq!(cc.known_types.get(&x).unwrap(), &ValueType::Ref);
+
+        // ?x is bound to datoms0.e.
+        assert_eq!(cc.bindings.get(&x).unwrap(), &vec![d0_e.clone()]);
+
+        // Our 'where' clauses are two:
+        // - datoms0.a = 99
+        // - datoms0.v = true
+        // No need for a type tag constraint, because the attribute is known.
+        assert_eq!(cc.wheres, vec![
```
ncalexan: This test suggests that `wheres` should be a set, although I could imagine you want to order in some situations for the good of SQLite.  Is there another reason to not have wheres be unordered?

You might also unorder to the test(s).

query-algebrizer/src/lib.rs
```diff
@@ -9,16 +9,31 @@
...
 
+#[allow(dead_code)]
 pub fn algebrize(parsed: FindQuery) -> AlgebrizedQuery {
-    AlgebrizedQuery {}
+    AlgebrizedQuery {
+        default_source: parsed.default_source,
+        find_spec: parsed.find_spec,
+        has_aggregates: false,           // TODO: we don't parse them yet.
+        limit: None,
+        cc: cc::ConjoiningClauses::new(),
```
ncalexan: Is there a reason to not connect the knot?  Oh, right, you need to plumb a schema through, etc.  I await your plan.

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+/// !   clause, or the intersection of the vars in the two sides of the JOIN.
+/// !
+/// ! Not yet done:
+/// ! - Function clauses with bindings turn into:
+/// !   * Subqueries. Perhaps less efficient? Certainly clearer.
+/// !   * Projection expressions, if only used for output.
+/// !   * Inline expressions?
+/// !---------------------------------------------------------------------------------------
+/// !
+/// ! `from` is a list of [source alias] pairs, suitable for passing to honeysql.
```
rnewman: Yeah, this whole comment was lifted wholesale from cc.cljc.

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+/// ! `bindings` is a map from var to qualified columns.
+/// ! `known-types` is a map from var to type keyword (e.g., :db.type/ref)
+/// ! `extracted-types` is a mapping, similar to `bindings`, but used to pull
```
rnewman: Yes, exactly. If you have a query like `[:find ?v :where [_ _ ?v]]` we need to project `all_datoms.v` to get the value, and `all_datoms.value_type_tag` so we know how to treat `15` when it arrives.

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+
+    #[inline]
+    fn is_value(&self) -> bool {
+        self.1 == DatomsColumn::Value
+    }
+}
+
+/// A thing that's capable of aliasing a table name for us.
+/// This exists so that we can obtain predictable names in tests.
+pub type TableAliaser = Box<FnMut(DatomsTable) -> TableAlias>;
```
rnewman: This thing really only exists to get predictable names, and I might just replace it with a mutable counter in the CC itself.

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+    EqualsValue(QualifiedAlias, TypedValue),
+    EqualsColumn(QualifiedAlias, QualifiedAlias),
+}
+
+pub struct ConjoiningClauses {
+    is_known_empty: bool,
+    aliaser: TableAliaser,
+    from: Vec<SourceAlias>,
+    wheres: Vec<ColumnConstraint>,
+    bindings: BTreeMap<Variable, Vec<QualifiedAlias>>,
```
rnewman: I'm not at all comfortable cloning! At some point we do indeed need to have `Rc<Variable>` instead.

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+    known_types: BTreeMap<Variable, ValueType>,
+    extracted_types: BTreeMap<Variable, QualifiedAlias>,
+}
+
+impl ::std::fmt::Debug for ConjoiningClauses {
```
rnewman: For now, yes, that's true; I expect the debug representation to get clearer, though. Right now it expresses all the nested objects at full depth.

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+            aliaser: default_table_aliaser(),
+            from: vec![],
+            wheres: vec![],
+            bindings: BTreeMap::new(),
+            known_types: BTreeMap::new(),
+            extracted_types: BTreeMap::new(),
+        }
+    }
+
+    pub fn bind_column_to_var(&mut self, table: TableAlias, column: DatomsColumn, var: Variable) {
```
rnewman: Yes. The CC pipeline was always "Clojure-in-place" — that is, completely chained and consuming — with the exception of certain split-and-merge things like complex `or-join`. Not as big a jump to Rust as one might think.

query-algebrizer/src/cc.rs
```diff
@@ -0,0 +1,795 @@
...
+            !self.known_types.contains_key(&var) &&
+            !self.extracted_types.contains_key(&var);
+
+        if needs_type_extraction {
+            self.extracted_types.insert(var.clone(), alias.for_type_tag());
+        }
+        self.bindings.entry(var).or_insert(vec![]).push(alias);
+    }
+
+    pub fn constrain_column_to_entity(&mut self, table: TableAlias, column: DatomsColumn, entity: i64) {
```
rnewman: Just an accidental omission — you'll see I use `Entid` elsewhere.

query-algebrizer/src/cc.rs
rnewman: Yeah, I simply didn't want to bitrot other PRs.

query-algebrizer/src/cc.rs
rnewman: Yes, it's determining a table from the attribute and value.

query-algebrizer/src/cc.rs
rnewman: The ampersand on the right is ambiguous (in bind statements: https://github.com/rust-lang/rust-by-example/issues/390), which is why `let ref` exists. I'd opt for the unambiguous syntax by default!

query-algebrizer/src/cc.rs
rnewman: Yes, it's possible, which is why this TODO exists:

```
[_ ?a ?v]
[_ ?b ?v]
[(!= ?a ?b)]
```

This TODO is to make sure that `TypedValue::Ref(1)` and `TypedValue::Integer(1)` don't end up matching the `datoms0.v = datoms1.v` constraint.

rnewman: The type tags we're talking about here are 'live'. We already do type constraint analysis if the attributes are known, and will mark the pattern as failed if they clash.

query-algebrizer/src/cc.rs
rnewman: No: we need the bindings for other things.

query-algebrizer/src/cc.rs
rnewman: Yes, they're used again, most notably by the projector: for each projected element (e.g., `?v`) it needs to know either its known type or which additional SQL column to project to extract it.

query-algebrizer/src/cc.rs
rnewman: They're ordered because predictability is our main weapon for query optimization. The clauses are quite consistent.

query-algebrizer/src/cc.rs
rnewman: Ah, we can't just use a local counter — we want table aliases to be distinct across the whole query.

query-algebrizer/src/cc.rs
rnewman: Renamed to `constrain_to_*`.

rnewman:
I've addressed most of the comments; will land with those and then file issues or address the rest.

rnewman:
5af7082165061f530dd77f14a13196230cce5c34

query-algebrizer/src/cc.rs
rnewman: Filed #318.

query-algebrizer/src/cc.rs
rnewman: Filed #317.

query-algebrizer/src/cc.rs
rnewman: Filed #319.

query-algebrizer/src/cc.rs
rnewman: Filed #320.

query-algebrizer/src/cc.rs
ncalexan: This is not an ambiguous case, as far as I can tell -- and the discussion you link to seems to conclude that the two are equivalent: https://github.com/rust-lang/rust-by-example/issues/390#issuecomment-129167712.

query-algebrizer/src/cc.rs
ncalexan: As discussed on Slack, I do think there's a separation of this algorithm into pipeline stages that might make some of these transformations clearer.

query-algebrizer/src/cc.rs
rnewman: Indeed not — I just like to only use the unambiguous style!

Issue #195, opened 2017-01-24
Add EDN pretty printer


ncalexan:
This ticket tracks adding a pretty-printer (potentially, implementing `std::fmt::Display`) for `edn::types::Value`.

If I was to do this, I'd start by understanding how `serde` achieves this, and then whether there's a Rust implementation of Wadler's pretty-printer combinators.

ncalexan:
This is a _[good first bug]_ if you have strong development skills but not necessarily strong Rust skills.

ncalexan:
@jsantell can you bump this up your priority queue?  It's frustrating having 4k of output on one line when dumping these databases :)  Thanks!

ncalexan:
We have a lot of input and output that looks like `[[:db/add ...] [:db/add ...]]`.  It would be nice to give columnar views easily, using (for example) https://github.com/BurntSushi/tabwriter.

victorporof:
As discussed, I'll take this.

ncalexan:
I see https://github.com/freebroccolo/pretty.rs/issues/26 tracking adding some simple documentation.  The examples and tests are more than enough to complete that issue, which would be a great contribution to upstream...

Issue #220, opened 2017-02-01
[edn] Add EDN differ


ncalexan:
Similar to #195, it gets really tiresome trying to visually diff to subtly different `edn::Value` instances.  This ticket tracks implementing some structured diff algorithm.  Even just pretty printing and doing a text diff on the two outputs would be better than the madness I'm doing to compare right now!

victorporof:
I have interest in working on this as well, possibly after #195. If there's anyone else interested in taking this though, please feel free.

victorporof:
This can either be done via structural diff, or textual diff. Structural is likely more difficult to get right, one can imagine a tree traversal iterator using `itertools::diff_with`, but that likely won't offer meaningful data when diffing two sufficiently different ASTs (need to bail out early). Textual diff seems like a reasonable approach, especially after #195

ncalexan:
This is really an answer to https://github.com/mozilla/mentat/issues/247#issue-205518531, but I'm putting it here since this is the relevant ticket. 

I want pretty output during tests -- basically, I want `assert_eq!` to help me find differences in EDN (presented as `&str` slices).  This probably looks like a new macro, `assert_eq_edn!` or similar, that parses both sides into EDN (bonus points: handles `edn::Value` and anything that can be parsed into EDN) and then provides a helpful diff message as well as the string inputs if the two sides aren't equal.

victorporof:
@ncalexan It would seem that textual diffing is something sufficient for your needs (and thus some new pretty printing rules for json-like expanding of edn, which would also allow me to highlight the usefulness of some of the generalizations in #245). Is that correct?

ncalexan:
On Thu, Feb 9, 2017 at 12:07 PM, Victor Porof <notifications@github.com>
wrote:

> @ncalexan <https://github.com/ncalexan> It would seem that textual
> diffing is something sufficient for your needs (and thus some new pretty
> printing rules for json-like expanding of edn, which would also allow me to
> highlight the usefulness of some of the generalizations in #245
> <https://github.com/mozilla/mentat/pull/245>). Is that correct?
>
Textual diffing is great for test reports and helping humans figure out
where things are broken.  The EDN matching I describe is really about
letting us write tests easily, which is slightly different.  Progress on
either front would be great!

PR #245, opened 2017-02-05
Implement edn pretty printing

victorporof:
Fixes https://github.com/mozilla/mentat/issues/195

Needs some tests and documentation.

Signed-off-by: Victor Porof <vporof@mozilla.com>


victorporof:
Added tests and documentation. Ready for review.

victorporof:
Note that this printer isn't optimized for tabular-like edn data or extended json-like expansion. Different pretty printing implementations can be added for those things (e.g. for https://github.com/mozilla/mentat/issues/220)

edn/src/types_pp.rs
```diff
@@ -0,0 +1,160 @@
...
+            Value::List(ref v) => v.into(),
+            Value::Set(ref v) => v.into(),
+            Value::Map(ref v) => v.into(),
+            Value::Text(ref v) => string_leaf!(format!("\"{}\"", v)),
+            _ => string_leaf!(src.to_string())
+        }
+    }
+}
+
+impl<'a> From<&'a Vec<Value>> for PrettyPrintTree<'a> {
```
ncalexan: This seems a pretty big abstraction leak.  We want to be able to print `edn::Value`, not arbitrary vectors (or maps, or sets) of values.  I see you're using `From` throughout, but I think you should use it only for the top-level object.

edn/src/pp.rs
```diff
@@ -0,0 +1,240 @@
...
+    /// Print a newline.
+    PrintNewline,
+    /// Print a string.
+    PrintString(Cow<'a, str>),
+    /// Print a newline, the children in the tree as a continuous block
+    /// nested at `usize`, then another newline.
+    NestChildren(usize),
+}
+
+/// A tree of pretty print instructions.
```
ncalexan: Can you explain _why_ you need this intermediate representation?  That is, you're converting `edn::Value` -> `PrettyPrintTree` -> `pretty::DocBuilder` -> `String`.  What value is `PrettyPrintTree` capturing?

I expected this patch to be the real pretty printing bits (that handle `edn::Value`) that accepts an allocator and just uses it, like https://github.com/freebroccolo/pretty.rs/blob/master/examples/trees.rs#L23.  Can you explain why that approach isn't sufficient?

edn/src/pp.rs
```diff
@@ -0,0 +1,240 @@
...
+        let subdocs = match self.children {
+            Some(v) => v.into_iter().fold(alloc.nil(), |doc, child| doc.append(child.into_doc(alloc))),
+            None => alloc.nil()
+        };
+
+        match self.instr {
+            None => subdocs,
+            Some(PrintInstr::PrintNewline) => alloc.newline().append(subdocs),
+            Some(PrintInstr::PrintString(v)) => alloc.text(v).append(subdocs),
+            Some(PrintInstr::NestChildren(v)) => alloc.newline().append(subdocs).nest(v).append(alloc.newline()),
```
ncalexan: I don't understand this completely, but I expect this prefix approach (where you generate all the subdocs into one document and then insert it inline) will stop you from streaming output efficiently.  That is, in some cases you'll need to render the entire tree before pushing any bits to the writer.  Is that true?  Can it be avoided, perhaps by having the subdoc producing bit by a closure that gets invoked inline?

rnewman: My experience in writing pretty printers (which is, admittedly, fifteen years out of date) is that they are intrinsically streaming: nothing more than closed-over indents and recursion. I would also be very surprised if *any* intermediate structure were needed or desirable. Given `edn::Value::to_string()`, it should be possible to build a recursive pretty-printer by hand in about 30-40 lines of Rust, with no library at all…

edn/src/types_pp.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    #[test]
+    fn test_pp_types() {
+        let string = "[ 1 2 ( 3.14 ) #{ 4N } { foo/bar 42 :baz/boz 43 } [ ] :five :six/seven eight nine/ten true false nil #f NaN #f -Infinity #f +Infinity ]";
+
+        let data = parse::value(string).unwrap();
+        let tree = PrettyPrintTree::from(&data);
+
+        let pretty = tree.pretty_print_into_string();
+        assert_eq!(pretty.unwrap(), "[\n  1 2 (\n    3.14\n  ) #{ 4N } { foo/bar 42 :baz/boz 43 } [] :five :six/seven eight nine/ten true false nil #f NaN #f -Infinity #f +Infinity\n]");
```
victorporof: TIL!

edn/src/types_pp.rs
victorporof: I don't think there's any necessary abstraction leak here. Everything can indeed be inlined instead.

None of these subsequent implementation blocks are actually required, I've opted for implementing multiple `From` traits because it seemed more idiomatic, and gave us more pretty printable data structures for free. Splitting things into smaller functions also avoids cyclomatic complexity and is (maybe subjectively) easier to read. 

> We want to be able to print edn::Value, not arbitrary vectors (or maps, or sets) of values.

Unless there's some non-written requirement about allowing anything other than a single `edn::Value` to be pretty printable, I don't see the harm in having pretty printing implemented for free for multiple kinds of source types with this approach. They might be useful especially in tests, when having to compare different kinds of data. I don't see what is gained by not having them and inlining everything.

Like I said above, changing this is easy, but I'd like to understand why first.

edn/src/pp.rs
victorporof: The approach with having a single recursive function implementing all of the pretty printing was the first thing I did: it was sufficient and worked fine. I wanted to go a step further and build something that allows for pretty printing configuration, is type-agnostic, and abstracts away some weirdness/peculiarities of pretty.rs. A tree-like structure of instructions that can be built through macro invocations seemed like one way to go (basically, `PrettyPrintTree` is just a wrapper around pretty.rs). The advantage argument comes from ergonomics and not necessity. There are admittedly other approaches that would've been valid.

The intermediate representation also allows for easily having different pretty printing rules (e.g. for tabular data, json-like expansion etc.), that are (admittedly subjectively) easier to write in a relatively more declarative manner (which all of these macros are for). One version of this patch had more involved macros that allowed for the entire pretty printing implementation to be just a couple of lines of macro invocations in a single `From`. I've scrapped them though because they seemed a bit harder to read in the end, but maybe those would've been better to highlight my point.

edn/src/pp.rs
victorporof: @ncalexan Assuming you mean line-by-line streaming to some output, I don't understand how pretty.rs documents can be streamed when rendering needs to happen at the end, so that all the indentation instructions are already in place.

As I've said above, there's a few peculiarities with pretty.rs, where some specific sequences of documents need to be in place for printing to work, especially when handling indented subtrees (at least some of the "unwritten rules" there seemed peculiar to me).

victorporof:
@ncalexan To recap, I can indeed rewrite all of this back to my initial implementation using a single recursive function, which would work "fine". However, see my above comments and let me know how you'd like to proceed.

ncalexan:
> The approach with having a single recursive function implementing all of the pretty printing was the first thing I did: it was sufficient and worked fine.

OK, good.  Can you dig this out of your reflog and push it somewhere to see?

> I wanted to go a step further and build something that allows for pretty printing configuration, is type-agnostic, and abstracts away some weirdness/peculiarities of pretty.rs. A tree-like structure of instructions that can be built through macro invocations seemed like one way to go (basically, PrettyPrintTree is just a wrapper around pretty.rs).

A "tree-like structure of instructions" _is_ what `pretty.rs` provides.  The structure defined here doesn't appear to add value on top of `pretty.rs`, although I may be missing something subtle.  The structure `pretty.rs` defines is also "standard", in the sense that many people know and reach for Wadler's combinators whenever such a problem arises.

> The advantage argument comes from ergonomics and not necessity. There are admittedly other approaches that would've been valid.

OK.  I think we should either:
- Leverage the library and land a really simple implementation using "bare" `pretty.rs`, which I expect would be <50 lines.
- Really build something that uses the functionality you've built, whatever that might be.

Right now we're in the middle: we have a layer that seems to duplicate its library layer and little indication of the added value.  However, I think we all deserve wide latitude to do what we think is best, so if you see a future value that I do not, I trust you and will accede to your judgement.

But in any case, I'd like to see one of your earlier/simple versions (even if it doesn't quite compile, or doesn't have a full test suite, or whatever).

(Also, sorry for the delayed review -- this was in my queue yesterday but I ran out of time.)

victorporof:
@ncalexan I don't believe you're missing anything subtle. As it stands, the PrettyPrintTree doesn't add any value, and even with the implementation of additional pretty printing rules, its value might be questionable because several (albeit a little more complex) macros can be implemented on top of pretty.rs that help with writing prettification rules in a more declarative manner.

I see a total of 3 possible implementations:

1. Simple recursive function with lots of match expressions and inlined rules for how to pretty print. Inherently imperative in nature.
Advantage: barebones, easily hackable and understandable.
Disadvantage: any new pretty printing set of rules needs a completely separate rewrite; pretty.rs's API (albeit following wadler-style) instructions needs to be known and understood, with its potential quirks

2. Recursive function using macros directly on top of pretty.rs
Advantage: easy to grow new sets of pretty printing rules based on the macros
Disadvantage: macros are relatively complex

3. Recursive function using macros on top of a wrapper around pretty.rs
Advantage: same as 2, plus macro complexity is moved out in a typed tree structure which is much easier to understand than a garble of macros
Disadvantage: larger "backend" for pretty printing

I hope the above makes it clear. I started with number 1 and now have this PR implementing 3. The difference between 2 and 3 is questionable indeed, but I'm slightly more in favor of 3. I can totally understand why 1 would seem much more desirable in our current state.

I'll end by stating that this is probably the least important thing to worry about. I enjoy trying to write code that is prepared for future additions when I know beforehand what the immediate future necessities are. I do sometimes fail at this and prematurely overengineer things.

ncalexan:
> Simple recursive function with lots of match expressions and inlined rules for how to pretty print. Inherently imperative in nature.
>     Advantage: barebones, easily hackable and understandable.

I think right now, these are compelling.  Let's land something small based on `pretty.rs`, and grow it if and when we need to.

victorporof:
@ncalexan Ready for another review. I've removed the intermediary PrettyPrintTree but it quickly became apparent that simply relying on pretty.rs methods would result in lots of code duplication between edn::Value collection types. I ended up writing a few macros to dry things up, essentially pursuing implementation (2) from the above comment.

victorporof:
@ncalexan I'll add that this new code is basically doing exactly the same thing as the one relying on a `PrettyPrintTree` with two differences: there's no `From` trait implementation, and methods such as `pretty_print_to_string`, `pretty_print_to_writable` and `as_doc` are now part of `Value` instead. 

To me it felt better to separate all of these concerns outside of `Value`, while the pretty printing implementation itself on `Value` would've just handled "how the end result would look like" in a declarative manner. The intermediary representation helped with that as described in https://github.com/mozilla/mentat/pull/245#issuecomment-278715244, but alas! I'm ambivalent between all of these approaches.

ncalexan:
@victorporof this turned into a rewrite.  Have a look, see if you agree with the approach.

ncalexan:
> I'm not assigned as a reviewer, and was unassigned as the author of this PR, so I'm not sure what kind of feedback you'd like. I'll offer some high level comments.

For reasons I don't understand, I wasn't able to make GH add you as a reviewer.  Unassigning was part of me trying to get you flagged as a reviewer: I wondered if GH wouldn't let an assignee be a reviewer as well.  But no luck!
 
> This approach opts for a typed interface for bracketing, as opposed to mine which used macros. This is certainly valid, however equivalent in nature. I have no preference either way, and I'm ambivalent to these stylistic considerations.

Macros are always a superset of syntax, but you should strongly avoid macros unless there's a reason to use them.  This is true in pretty much every system that has a macro system; they always add a layer of complexity.  There's no reason to use them here; this is just about iteration.

> I have no definitive comment regarding the visual output changes for the pretty printing. To my eye the pretty printed edn looks more difficult to read now, since it's substantially more vertically expanded, but if this was the preferred output, noting it in the original issue would've been useful.

For this, I apologize.  I took as given the "standard" Clojure/EDN style, which is only standard if you've been reading and writing Clojure.

> There's no matching rules for Symbols and Keywords now when converting Values into pretty.rs documents. While this is works, my understanding is that not passing the internal strings as references would result in allocating new strings for those. My implementation carefully made sure there's no needless string allocation, since most edn queries have mostly keywords and symbols. If my understanding is correct, and performance is a consideration, this is the only thing I'd feel strongly about.

This is a good point.  Let me either reason about it or revert to your approach.  Thanks!

> Finally, not pertinent to your rewrite, I'll add that certainly to properly pretty print queries we'd need to implement this at a higher level, as edn values have no knowledge about what queries are.

Roger that.  I might just add it for fun.

victorporof:
> >There's no matching rules for Symbols and Keywords now when converting Values into pretty.rs documents. While this is works, my understanding is that not passing the internal strings as references would result in allocating new strings for those. My implementation carefully made sure there's no needless string allocation, since most edn queries have mostly keywords and symbols. If my understanding is correct, and performance is a consideration, this is the only thing I'd feel strongly about.

> This is a good point. Let me either reason about it or revert to your approach. Thanks!

I believe handling these properly can be done with your new approach as well, no reason to revert just for this. Just pass as_refed strings from symbols and keywords to automatically have them cowed when allocating text documents.

victorporof:
Rebased. Can this land soon? I want to work on edn diffing and having this in place is necessary.

ncalexan:
On Tue, Feb 21, 2017 at 1:26 AM, Victor Porof <notifications@github.com>
wrote:

> Rebased. Can this land soon? I want to work on edn diffing and having this
> in place is necessary.
>

Yes, I'll get to this right now.

PR #246, opened 2017-02-06
Properly handle whitespace for Infinity and NaN

### LINKED ISSUE #244: Properly handle whitespace for Infinity and NaN

victorporof:
Fixes #244.

Signed-off-by: Victor Porof <vporof@mozilla.com>

### DIFF
--- a/edn/src/edn.rustpeg
+++ b/edn/src/edn.rustpeg
@@ -32,10 +32,10 @@ pub nil -> Value =
     "nil" { Value::Nil }
 
 pub nan -> Value =
-    "#f" __ "NaN" { Value::Float(OrderedFloat(NAN)) }
+    "#f" whitespace+ "NaN" { Value::Float(OrderedFloat(NAN)) }
 
 pub infinity -> Value =
-    "#f" __ s:$(sign) "Infinity" {
+    "#f" whitespace+ s:$(sign) "Infinity" {
         Value::Float(OrderedFloat(if s == "+" { INFINITY } else { NEG_INFINITY }))
     }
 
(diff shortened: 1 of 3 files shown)


edn/src/edn.rustpeg
```diff
@@ -32,10 +32,10 @@ pub nil -> Value =
     "nil" { Value::Nil }
 
 pub nan -> Value =
-    "#f" __ "NaN" { Value::Float(OrderedFloat(NAN)) }
+    "#f " __ "NaN" { Value::Float(OrderedFloat(NAN)) }
 
 pub infinity -> Value =
-    "#f" __ s:$(sign) "Infinity" {
+    "#f " __ s:$(sign) "Infinity" {
```
rnewman: These are technically incorrect: EDN allows `','` as whitespace, so `#f,1.0` is valid (and you should add a test for this case). You want:

```
"#f" whitespace+ s:$(sign)
```

for this rule, and

```
"#f" whitespace+ "NaN"
```

for `nan`. That will also prevent users from putting a comment between a tag and its referent, which I think is a good thing!

edn/src/types.rs
```diff
@@ -56,7 +56,6 @@ impl Display for Value {
             Integer(v) => write!(f, "{}", v),
             BigInteger(ref v) => write!(f, "{}N", v),
             Float(ref v) => {
-                // TODO: make sure float syntax is correct.
```
rnewman: Did you check that `format`'s rules for floats match EDN's? E.g., exponent capitalization and sign?

edn/src/types.rs
```diff
@@ -56,7 +56,6 @@ impl Display for Value {
...
                 if *v == OrderedFloat(f64::INFINITY) {
                     write!(f, "#f {}", "+Infinity")
```
rnewman: You can just do

```
write!(f, "#f +Infinity")
```

for these – no need to use a format expression at all.

victorporof:
Addressed comments.

victorporof:
@rnewman or @ncalexan gentle review ping.

Issue #235, opened 2017-02-02
Use mentat-parser-utils in tx-parser


ncalexan:
@rnewman and @victorporof have pushed a bunch of shortening macros to the `mentat-parser-utils` crate in #218.  This ticket tracks using the new tools in the `tx-parser` crate (which was written first and should benefit from the new work).

This is a _[good first bug]_ if you know a little Rust.

victorporof:
Are there specific places where you think these macros need to be used, or is this more of a "let's see what can be simplified".

ncalexan:
> Are there specific places where you think these macros need to be used, or is this more of a "let's see what can be simplified"?

Little of both.  I don't know the full scope of things, but I see `def_parser_fn` instances throughout (like at https://github.com/mozilla/mentat/blob/rust/tx-parser/src/lib.rs#L36) and I see `satisfy_map` which might be shortened in a few places (like at https://github.com/mozilla/mentat/blob/rust/tx-parser/src/lib.rs#L54).

victorporof:
Making a note here that @joewalker offered to steal this from me.

PR #249, opened 2017-02-06
Use mentat-parser-utils in tx-parser. Initial stab at #235

joewalker:
Work in progress patch just to see if I'm on the right lines.


tx-parser/src/lib.rs
```diff
@@ -13,12 +13,17 @@
...
+
+#[macro_use]
+extern crate mentat_query_parser;
 
 use combine::{any, eof, many, optional, parser, satisfy_map, token, Parser, ParseResult, Stream};
 use combine::combinator::{Expected, FnParser};
 use edn::symbols::NamespacedKeyword;
 use edn::types::Value;
-use mentat_tx::entities::*;
+use mentat_tx::entities::{Entid, EntidOrLookupRef, Entity, LookupRef, ValueOrLookupRef};
```
victorporof: Any reason to import stuff explicitly instead of everything?

query-parser/src/parser_util.rs
```diff
@@ -40,6 +40,7 @@ macro_rules! matches_plain_symbol {
 /// The provided `$body` will be evaluated with `$input` bound to the input stream.
 ///
 /// `$body`, when run, should return a `ParseResult` of the appropriate result type.
+#[macro_export]
```
victorporof: Should probably go ahead and export all of these macros.

tx-parser/src/lib.rs
```diff
@@ -13,12 +13,17 @@
 extern crate edn;
 extern crate combine;
 extern crate mentat_tx;
+extern crate mentat_parser_utils;
```
victorporof: This is also likely to require a macro_use pragma if you plan on using the macros there. I'd be surprised if not.

tx-parser/src/lib.rs
```diff
@@ -30,9 +35,19 @@ fn fn_parser<O, I>(f: fn(I) -> ParseResult<O, I>, err: &'static str) -> TxParser
     parser(f).expected(err)
 }
 
+def_parser_fn!(Tx, integer, Value, i64, input, {
```
victorporof: Should probably not have to define macros here. There's another `def_parser_fn!` as well in the mentat query parser crate. Did you intend to redefine it here? I'm having a hard time understanding what this is for as well, since it's not used anywhere.

query-parser/src/parser_util.rs
rnewman: Nothing in `query-parser` should be exported; if it's directly applicable to other crates, it should be moved to `parser-utils`.

tx-parser/src/lib.rs
rnewman: It's good style. Using `*` means that a simple addition in a *different crate* can cause this one to fail to compile.

tx-parser/src/lib.rs
rnewman: `def_parser_fn!` is a macro invocation, not a macro definition.

victorporof: Ah! I misread this!

query-parser/src/parser_util.rs
victorporof: I thought not putting these in the parser_utils crate in the first place was intentional.

tx-parser/src/lib.rs
victorporof: There's a clash of styles it would seem, as @ncalexan opted for `*` in the first place. Have we settled on a single style for our codebase? My question was directed to the fact that I didn't see how it this change was related to the issue being fixed (it could've been for what I know, I still don't have a solid grasp on rust).

query-parser/src/parser_util.rs
rnewman: It was: they're very EDN-specific, and I wasn't sure how reusable they would be.

It looks like @joewalker could directly use `def_value_satisfy_parser_fn!`, though…

tx-parser/src/lib.rs
joewalker: I changed it because I misread an error message telling me it wasn't needed (it was actually the one down around line 274 in `mod tests`).

I agree with @rnewman's logic, and I'd also add that some specificity about what's required and what isn't is a good thing.

query-parser/src/parser_util.rs
joewalker: So I'm moving all (?) the parser utils from parser_utils to parser-utils, right?

victorporof: My understanding is: if it can be used in other places as well, then yes. However, indeed in this particular case `def_value_satisfy_parser_fn` could be used instead of `def_parser_fn`, so maybe don't move unless there's a really good reason to?

ncalexan:
> There's a clash of styles it would seem, as @ncalexan opted for * in the first place. Have we settled on a single style for our codebase? My question was directed to the fact that I didn't see how it this change was related to the issue being fixed (it could've been for what I know, I still don't have a solid grasp on rust).

Yeah, I think this was a mistake.  In some cases it makes sense -- for example, during development -- but in most cases it's best to just list them.  Filed https://github.com/mozilla/mentat/issues/251 to unify on the non-`use ... *` form throughout the code base.

query-parser/src/parser_util.rs
rnewman: The macros are built on each other, so you can't move and use `d_v_s_p_f` without moving the others.

victorporof:
@joewalker Is this still WIP?

joewalker:
> Is this still WIP?

Erm, maybe?

Do we have any test coverage info anywhere? All the tests pass, but can I be sure we're testing it properly?
Are there other places where we should do this?
I'm still not very confident that I did it correctly TBH.

joewalker:
Something that makes me think I'm not done yet - The [pre-PR version of entities](https://github.com/mozilla/mentat/blob/7db74953d653f3208ed004e394b5754ce8983341/tx-parser/src/lib.rs#L184) contains `"[:db/add|:db/retract ...]"`, but the [current post PR version](https://github.com/joewalker/mentat/blob/d4ac9cd97e9a4141eea74092cc3847b275d79a95/tx-parser/src/lib.rs#L181) doesn't contain this anywhere.

joewalker:
[Elsewhere](https://github.com/mozilla/mentat/pull/252#discussion_r99970983) @rnewman said (of [lib.rs:48](https://github.com/joewalker/mentat/blob/e29e5d267367ecaad926237254b843b02d7c5117/tx-parser/src/lib.rs#L38))

> This can be phrased in terms of `def_value_satisfy_parser_fn!`.

Which sounds likely, except that I'm stuck on what `$transformer` should be. I'd like to say:

    def_value_satisfy_parser_fn!(Tx, integer, i64, if let Value::Integer(y) = x {
            Some(y)
        } else {
            None
        });

But nope. Any ideas?

joewalker:
About to rebase. Head was at https://github.com/mozilla/mentat/pull/249/commits/12f17c1130c62038b86fbb0c45b43578169a8451

tx-parser/src/lib.rs
```diff
@@ -30,181 +34,150 @@ fn fn_parser<O, I>(f: fn(I) -> ParseResult<O, I>, err: &'static str) -> TxParser
...
-    fn integer_(input: I) -> ParseResult<i64, I> {
-        return satisfy_map(|x: Value| if let Value::Integer(y) = x {
-                Some(y)
-            } else {
-                None
-            })
-            .parse_stream(input);
-    }
+// The above using a def_value_satisfy_parser_fn
+//def_value_satisfy_parser_fn!(Tx, integer, i64, if let Value::Integer(y) = x { Some(y) } else { None });
```
rnewman: > In practice this allows you to simply pass a function that accepts an `&edn::Value` and
> returns an `Option<$result_type>`: if a suitable value is at the front of the stream,
> it will be converted and returned by the parser; otherwise, the parse will fail.

So if you define a function:

```
fn value_to_i64(val: &Value) -> Option<i64> {
    if let &Value::Integer(i) = val { Some(i) } else { None }
}
```

you can use it here:

```
def_value_satisfy_parser_fn!(Tx, integer, i64, value_to_i64);
```

Now it turns out we already have that function: it's `Value::as_integer`. So this should work:


```
def_value_satisfy_parser_fn!(Tx, integer, i64, Value::as_integer);
```

tx-parser/src/lib.rs
```diff
@@ -30,181 +34,150 @@ fn fn_parser<O, I>(f: fn(I) -> ParseResult<O, I>, err: &'static str) -> TxParser
...
 
-    fn keyword() -> TxParser<NamespacedKeyword, I> {
-        fn_parser(Tx::<I>::keyword_, "keyword")
-    }
+// Manually expanding the above:
+// def_value_parser_fn!(Tx, integer, i64, input, {
+//     satisfy_map(|x: edn::Value| $transformer(&x)                                      ).parse_stream(input)
+//     satisfy_map(|x: edn::Value| if let Value::Integer(y) = x { Some(y) } else { None }).parse_stream(input)
```
rnewman: The trick is that `$transformer` is a path, not a block, so you can't pass an expression to this macro.

tx-parser/src/lib.rs
```diff
@@ -30,181 +34,150 @@ fn fn_parser<O, I>(f: fn(I) -> ParseResult<O, I>, err: &'static str) -> TxParser
...
-    fn keyword_(input: I) -> ParseResult<NamespacedKeyword, I> {
-        return satisfy_map(|x: Value| if let Value::NamespacedKeyword(y) = x {
-                Some(y)
-            } else {
-                None
-            })
-            .parse_stream(input);
-    }
+def_parser_fn!(Tx, keyword, Value, NamespacedKeyword, input, {
+    return satisfy_map(|x: Value| if let Value::NamespacedKeyword(y) = x {
```
rnewman: Similarly, this would be `Value::as_namespaced_keyword`.

tx-parser/src/lib.rs
```diff
@@ -30,181 +34,150 @@ fn fn_parser<O, I>(f: fn(I) -> ParseResult<O, I>, err: &'static str) -> TxParser
...
 
-    fn lookup_ref_(input: I) -> ParseResult<LookupRef, I> {
-        return satisfy_map(|x: Value| if let Value::Vector(y) = x {
-                let mut p = (Tx::<&[Value]>::entid(), any(), eof())
-                    .map(|(a, v, _)| LookupRef { a: a, v: v });
-                let r = p.parse_lazy(&y[..]).into();
-                match r {
+// TODO: abstract the "match Vector, parse internal stream" pattern to remove this boilerplate.
+def_parser_fn!(Tx, add, Value, Entity, input, {
+    return satisfy_map(|x: Value| -> Option<Entity> {
```
rnewman: You can omit the `return` and the closing semicolon, saving yourself some space.

tx-parser/src/lib.rs
joewalker: OK, except that the signature of as_integer is `(&self) -> Option<&i64>`, and we want `-> Option<i64>`. So maybe I just include the function as stated?

rnewman: `i64` implements `Copy`, so we should fix the definition of `as_integer` (and `as_float`, `as_boolean`). No sense returning a reference type when the raw type is no larger than a machine address!

I can do that very quickly in a side PR. Momento…

rnewman: #269, @joewalker.

tx-parser/src/lib.rs
rnewman: Oh, and if this can't be done, because you get back a ref, then we perhaps want `Value::to_namespaced_keyword`, which would simply be

```
pub fn to_namespaced_keyword(&self) -> Option<NamespacedKeyword> {
    self.as_namespaced_keyword().map(|x| x.clone())
}
```

joewalker: Didn't see this comment until I'd worked it out myself, except I wrote:

    fn value_to_namespaced_keyword(val: &Value) -> Option<NamespacedKeyword> {
        if let &Value::NamespacedKeyword(ref i) = val { Some(i.clone()) } else { None }
    }

I think it makes sense to do this in Value though.

joewalker: I did this using macros: https://github.com/joewalker/mentat/commit/a4ba10c52fc14d46620fc440b16082d7634bf9ae

joewalker:
I think we can land this. OK?

tx-parser/src/lib.rs
```diff
@@ -30,181 +34,128 @@ fn fn_parser<O, I>(f: fn(I) -> ParseResult<O, I>, err: &'static str) -> TxParser
...
-    fn integer() -> TxParser<i64, I> {
-        fn_parser(Tx::<I>::integer_, "integer")
-    }
+def_value_satisfy_parser_fn!(Tx, integer, i64, Value::as_integer);
 
-    fn integer_(input: I) -> ParseResult<i64, I> {
-        return satisfy_map(|x: Value| if let Value::Integer(y) = x {
-                Some(y)
+fn value_to_namespaced_keyword(val: &Value) -> Option<NamespacedKeyword> {
+    if let &Value::NamespacedKeyword(ref i) = val { Some(i.clone()) } else { None }
```
rnewman: I think you can use `val.as_namespaced_keyword().map(|x| x.clone())` here.

joewalker: Or even:

    def_value_satisfy_parser_fn!(Tx, keyword, NamespacedKeyword, Value::to_namespaced_keyword);

joewalker: Sigh. Nope. I'll do it your way and land it, and work out the fancy thing later.

Issue #251, opened 2017-02-06
Replace `use ... *` with explicit lists


ncalexan:
Per https://github.com/mozilla/mentat/pull/249#discussion_r99631181, it's best to be explicit in our import lists.  This ticket tracks auditing any uses of `use ... *` and replacing them with explicit lists where appropriate.

joewalker:
Anyone have thoughts on `pub use ...::*;` in lib.rs?

I'm tempted to think that we should be explicit there too, but I wonder how often we'll want to not export *? Maybe the rule should be to always `pub use ...:*;` in lib.rs?

edn/src/lib.rs says:

    pub use self::types::Value;
    pub use self::symbols::{Keyword, NamespacedKeyword, PlainSymbol, NamespacedSymbol};

But that's the complete list things for export from each module.

joewalker:
I started to hack on this, but very happy to drop it if someone else wants to pick it up.

ncalexan:
> Anyone have thoughts on pub use ...::*; in lib.rs?
> 
> I'm tempted to think that we should be explicit there too, but I wonder how often we'll want to not export *? Maybe the rule should be to always pub use ...:*; in lib.rs?

Perhaps `lib.rs` is the _most_ valuable place to be explicit, since it's the master list, the central source of truth.

jsantell:
Whatever reason there is to use `use ...::*`, it will have to be more valuable than the detriment to newbie rust users and those new to contributing to Mentat, because magic variables where it's not explicit where they come from (without tooling) are awful. I'll check this out.

PR #252, opened 2017-02-07
Initial stab at #251

joewalker:
Needs rebase onto `rust`

### DIFF
--- a/parser-utils/Cargo.toml
+++ b/parser-utils/Cargo.toml
@@ -5,3 +5,9 @@ authors = ["Victor Porof <vporof@mozilla.com>", "Richard Newman <rnewman@mozilla
 
 [dependencies]
 combine = "2.1.1"
+
+[dependencies.edn]
+  path = "../edn"
+
+[dependencies.mentat_query]
+  path = "../query"
--- a/parser-utils/src/query.rs
+++ b/parser-utils/src/query.rs
@@ -11,7 +11,6 @@
 extern crate combine;
 extern crate edn;
 
-extern crate mentat_parser_utils;
 extern crate mentat_query;
 
 /// Generate a `satisfy_map` expression that matches a `PlainSymbol`
@@ -20,6 +19,7 @@ extern crate mentat_query;
 /// We do this rather than using `combine::token` so that we don't
 /// need to allocate a new `String` inside a `PlainSymbol` inside a `Value`
 /// just to match input.
+#[macro_export]
 macro_rules! matches_plain_symbol {
     ($name: expr, $input: ident) => {
         satisfy_map(|x: edn::Value| {
@@ -56,6 +56,7 @@ macro_rules! def_parser_fn {
 
 /// `def_value_parser_fn` is a short-cut to `def_parser_fn` with the input type
 /// being `edn::Value`.
+#[macro_export]
 macro_rules! def_value_parser_fn {
     ( $parser: ident, $name: ident, $result_type: ty, $input: ident, $body: block ) => {
         def_parser_fn!($parser, $name, edn::Value, $result_type, $input, $body);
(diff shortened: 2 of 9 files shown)


tx-parser/Cargo.toml
```diff
@@ -10,3 +10,9 @@ combine = "2.1.1"
 
 [dependencies.mentat_tx]
   path = "../tx"
+
+[dependencies.mentat_parser_utils]
+  path = "../parser-utils"
+
+[dependencies.mentat_query_parser]
```
ncalexan: I don't expect a dependency between parsing transactions and parsing queries.  Anything that's shared is really about parsing `edn::Value`, I expect, and should live in the parsing utilities.

tx-parser/src/lib.rs
```diff
@@ -13,12 +13,17 @@
 extern crate edn;
 extern crate combine;
 extern crate mentat_tx;
+extern crate mentat_parser_utils;
+
+#[macro_use]
+extern crate mentat_query_parser;
```
ncalexan: Further to the above, I expect the macros to come from the parsing utilities crate.

tx-parser/src/lib.rs
```diff
@@ -30,9 +35,19 @@ fn fn_parser<O, I>(f: fn(I) -> ParseResult<O, I>, err: &'static str) -> TxParser
...
+        } else {
+            None
+        })
+        .parse_stream(input);
+});
+
 impl<I> Tx<I>
     where I: Stream<Item = Value>
 {
+    /*
```
ncalexan: If this is really unused, please kill it for good.

parser-utils/src/query.rs
```diff
@@ -11,7 +11,6 @@
 extern crate combine;
```
ncalexan: These utilities aren't really query specific, they're `edn::Value` specific.  (Am I wrong?)  Can we move to `src/value_parsing.rs` or something you prefer?

parser-utils/Cargo.toml
```diff
@@ -5,3 +5,9 @@ authors = ["Victor Porof <vporof@mozilla.com>", "Richard Newman <rnewman@mozilla
 
 [dependencies]
 combine = "2.1.1"
+
+[dependencies.edn]
+  path = "../edn"
+
+[dependencies.mentat_query]
```
ncalexan: Can you explain why parsing utilities would depend on the structure of queries?  All the parsing utilities are likely to be abstract over their result type.

parser-utils/src/query.rs
rnewman: Or, indeed, just into the `lib.rs`. They're all parser utils.

tx-parser/src/lib.rs
```diff
@@ -30,9 +35,19 @@ fn fn_parser<O, I>(f: fn(I) -> ParseResult<O, I>, err: &'static str) -> TxParser
     parser(f).expected(err)
 }
 
+def_parser_fn!(Tx, integer, Value, i64, input, {
+    return satisfy_map(|x: Value| if let Value::Integer(y) = x {
```
rnewman: This can be phrased in terms of `def_value_satisfy_parser_fn!`.

joewalker:
My intent with this PR was that it address #251 (the global imports one, which Jordan has now anyway) but the PR contains the changes in the branch on which this was based. So I'll close this, and address the comments in https://github.com/mozilla/mentat/pull/249

tx-parser/Cargo.toml
joewalker: Yes, it wasn't needed.

tx-parser/src/lib.rs
joewalker: Yup. Pushed mid hack. It's gone now.

parser-utils/src/query.rs
joewalker: Moved to lib.rs, my logic being that `matches_plain_symbol` uses `satisfy_unwrap` so it's good to have it close by.

parser-utils/Cargo.toml
joewalker: They don't. Fixed.

PR #264, opened 2017-02-08
Add :db.type/{instant, uri, uuid} types to bootstrapper. Fixes #201

### LINKED ISSUE #201: [db] [tx] Support :db.type/{instant,uuid,uri}
Over in #170, I simplified things by not including placeholders for Mentat value types I wasn't going to support at the start: `:db.type/instant`, `:db.type/uuid`, and `:db.type/uri`.  (We might consider supporting JSON and a binary blob type like `:db.type/bytes` as well.)

This ticket tracks supporting these types.  Off the top, this will mean:
1. Adding new `DB_TYPE_*` values to the bootstrapper;
1. Bumping the SQL schema to accommodate the new idents;
1. Adding new `ValueType` cases;
1. Adding corresponding `TypedValue` cases;
1. Implementing the conversions to and from SQL;
1. testing the new types in the transactor and potentially in the query engine as well.

jsantell:
This is missing a v3.db to run in the `test_open_current_version` test in `db/src/db.rs`,but looking for a sanity check here. Commenting out that test, everything else passes. Unlikely I'll tackle the creation of a v3 DB in `./fixtures` and the SQL update migration for v1 -> v2 -> v3

### DIFF
--- a/db/src/bootstrap.rs
+++ b/db/src/bootstrap.rs
@@ -79,17 +79,31 @@ lazy_static! {
          ]].concat()
     };
 
+    static ref V3_IDENTS: Vec<(symbols::NamespacedKeyword, i64)> = {
+        [(*V2_IDENTS).clone(),
+         vec![(ns_keyword!("db.type", "uuid"), entids::DB_TYPE_UUID),
+              (ns_keyword!("db.type", "uri"),  entids::DB_TYPE_URI),
+         ]].concat()
+    };
+
     static ref V1_PARTS: Vec<(symbols::NamespacedKeyword, i64, i64)> = {
         vec![(ns_keyword!("db.part", "db"), 0, (1 + V1_IDENTS.len()) as i64),
              (ns_keyword!("db.part", "user"), 0x10000, 0x10000),
              (ns_keyword!("db.part", "tx"), TX0, TX0),
         ]
     };
-
+    
     static ref V2_PARTS: Vec<(symbols::NamespacedKeyword, i64, i64)> = {
         vec![(ns_keyword!("db.part", "db"), 0, (1 + V2_IDENTS.len()) as i64),
              (ns_keyword!("db.part", "user"), 0x10000, 0x10000),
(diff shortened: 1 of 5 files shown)


db/src/bootstrap.rs
```diff
@@ -138,6 +152,16 @@ lazy_static! {
             .ok_or(ErrorKind::BadBootstrapDefinition("Unable to parse V2_SYMBOLIC_SCHEMA".into()))
             .unwrap()
     };
+
+    static ref V3_SYMBOLIC_SCHEMA: Value = {
+        let s = r#"{}"#;
```
jsantell: This doesn't do anything, because the symbolic schema doesn't change, AFAIK?

db/src/bootstrap.rs
```diff
@@ -233,27 +257,27 @@ fn symbolic_schema_to_assertions(symbolic_schema: &Value) -> Result<Vec<Value>>
 }
 
 pub fn bootstrap_partition_map() -> PartitionMap {
-    V2_PARTS[..].iter()
+    V3_PARTS[..].iter()
```
jsantell: I think changing all these variables to the latest is annoying. Should just have a struct for each schema that contains parts, idents, etc., and just reference latest via `const LATEST = V3_Schema` and use that throughout.

db/src/bootstrap.rs
ncalexan: Correct, and you're welcome to just clone the `V2_SYMBOLIC_SCHEMA` or whatever is needed if you prefer.

The symbolic schema would/will change if/when we make `:db/txInstant` have `:db.valueType :db.type/inst`, like Datomic.

db/src/bootstrap.rs
ncalexan: I'm fine with this.  Do it!  Or good first bug it and mentor it.

jsantell:
Rebased, but still failing in the test that uses the v2 database @ `db/src/db.rs:884`

rnewman:
Obsoleted by #438.

PR #265, opened 2017-02-08
Consolidate Entity::{Add, Retract} to Entity::AddOrRetract.

### LINKED ISSUE #25: Lookup refs are slow
Lookup refs are resolved prior to insertion, by running a `SELECT` query.

That's fine for 'unit transacts', but for bulk insertions it's cripplingly slow: 10-50msec per query multiplied by the number of lookup refs.

It would literally be quicker to read the entire database into memory, build a lookup table, and resolve in one pass. As a general policy, we should not run a linear number of queries on insert.

The best solution is probably to run a single query with `WHERE (a = ? AND v = ?) OR (a = ? AND v = ?)` — or just `WHERE a IN (?, ?)` for large inputs — and build a lookup table.

### LINKED ISSUE #255: [tx-parser] Don't split Entity::{Add,Retract}
The transaction processor mostly treats `Entity::Add` and `Entity::Retract` in the same path, and it's easy to match on a flag if one case should be handled differently.  Keeping them separate, however, makes treating the two cases in the same manner verbose.

This ticket tracks unifying on `Entity::AddOrRetract`.  To do this, we need to:
1. Remove `tx` from `Entity::Add` entirely -- this was just a mistake;
1. Add an `Op` or `OpType` enum that is `Add` or `Retract`;
1. Combine `Add` and `Retract` into one `AddOrRetract` with an `op: OpType`;
1. Update `tx-parser` to produce `AddOrRetract` with the correct `op`;
1. Update tests to use the new type.

This ticket, or a follow-up ticket, could also make the `tx-parser` handle `:db/add` and `:db/retract` uniformly, reducing duplication in the implementation.  This isn't crucial, though.

jsantell:
Fixes #255

### DIFF
--- a/tx-parser/src/lib.rs
+++ b/tx-parser/src/lib.rs
@@ -14,11 +14,11 @@ extern crate edn;
 extern crate combine;
 extern crate mentat_tx;
 
-use combine::{any, eof, many, optional, parser, satisfy_map, token, Parser, ParseResult, Stream};
+use combine::{any, eof, many, parser, satisfy_map, token, Parser, ParseResult, Stream};
 use combine::combinator::{Expected, FnParser};
 use edn::symbols::NamespacedKeyword;
 use edn::types::Value;
-use mentat_tx::entities::*;
+use mentat_tx::entities::{Entid, EntidOrLookupRef, Entity, LookupRef, OpType, ValueOrLookupRef};
 
 pub struct Tx<I>(::std::marker::PhantomData<fn(I) -> I>);
 
@@ -114,15 +114,13 @@ impl<I> Tx<I>
                                  Tx::<&[Value]>::entid(),
                                  // TODO: handle lookup-ref.
                                  any(),
-                                 // TODO: entid or special keyword :db/tx?
-                                 optional(Tx::<&[Value]>::entid()),
                                  eof())
-                        .map(|(_, e, a, v, tx, _)| {
-                            Entity::Add {
+                        .map(|(_, e, a, v, _)| {
+                            Entity::AddOrRetract {
+                                op: OpType::Add,
                                 e: e,
                                 a: a,
                                 v: ValueOrLookupRef::Value(v),
-                                tx: tx,
                             }
                         });
                     // TODO: use ok() with a type annotation rather than explicit match.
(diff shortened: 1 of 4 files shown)


tx-parser/src/lib.rs
```diff
@@ -117,12 +117,12 @@ impl<I> Tx<I>
                                  // TODO: entid or special keyword :db/tx?
                                  optional(Tx::<&[Value]>::entid()),
                                  eof())
-                        .map(|(_, e, a, v, tx, _)| {
-                            Entity::Add {
+                        .map(|(_, e, a, v, _, _)| {
```
jsantell: What do we do with the extra `tx` here?

ncalexan:
> What do we do with the extra tx here?

Don't accept it during parsing -- it's not valid.

tx-parser/src/lib.rs
```diff
@@ -138,7 +138,7 @@ impl<I> Tx<I>
     }
 
     fn add() -> TxParser<Entity, I> {
-        fn_parser(Tx::<I>::add_, "[:db/add e a v tx?]")
+        fn_parser(Tx::<I>::add_, "[:db/add e a v]")
```
jsantell: Removed from parser

tx/src/entities.rs
```diff
@@ -41,17 +41,18 @@ pub enum ValueOrLookupRef {
...
-        a: Entid,
-        v: ValueOrLookupRef,
-        tx: Option<Entid>,
-    },
-    Retract {
+    AddOrRetract {
         e: EntidOrLookupRef,
         a: Entid,
         v: ValueOrLookupRef,
+        op: OpType,
```
ncalexan: Can we put the `op` first?

tx-parser/src/lib.rs
```diff
@@ -117,12 +117,12 @@ impl<I> Tx<I>
                                  // TODO: entid or special keyword :db/tx?
                                  optional(Tx::<&[Value]>::entid()),
```
ncalexan: Drop this, it's not legal.  Maybe you did in another commit and I didn't see it?

tx-parser/src/lib.rs
ncalexan: Ah -- this is just the string description; the actual parser declaration is at https://github.com/mozilla/mentat/pull/265/files/7105861f047775c19fabf11148b8f0bc7dd95963#diff-5a81d11a5ad2928cc7cf2547288345eaL118.  Remove the optional, that'll drop the element from the tuple.

Issue #260, opened 2017-02-08
[db] Convert Schema into edn::Value for pretty-printing, debugging, etc


ncalexan:
Mentat stores its `Schema` in an optimized representation, where keys are integer entids and attributes are represented by `Attribute` structures with bit flags.  It's not easy to read, and it's definitely not easy to transact back against an existing store, or compare to Datomic schemas.

This ticket tracks converting `Schema` into `edn::Value`.  The output will be a map that looks something like the "symbolic schema" defined in the bootstrapping code around https://github.com/mozilla/mentat/blob/c585715224c180ff84630c05aa763d0d271f350a/db/src/bootstrap.rs#L86, i.e.:
```edn
{ :db/attribute
  { :db/valueType :db.type/long
    ...
  }
  ...
}
```
I don't think this will be common enough to warrant a `From`/`Into` implementation -- just a `Schema.as_edn_value(&self) -> edn::Value` should be enough.

victorporof:
Assigning to hopefully learn something about the db crate.

ncalexan:
> Assigning to hopefully learn something about the db crate.

Sure!  @jsantell filed #268, which would be much easier with this function, so it might be worth getting feedback from him and doing that as a consumer of this.

fluffyemily:
Can I just confirm the exact output that you want here.

I read this as, given a schema with an attribute with ident `NamespacedKeyword { namespace: "foo", name: "bar" }` and attribute `{ value_type: Boolean, multival: false, unique: None, index: false, fulltext: false, component: false }` should the output be this?:

```
{ :db/attribute
    { :db/id                :foo/bar
      :db/valueType  :db.type/bool
      :db/cardinality  :db.cardinality/one}
}
```
and for ident `NamespacedKeyword { namespace: "foo", name: "baz" }` and attribute `{ value_type: Ref, multival: true, unique: Some(attribute::Unique::Value), index: true, fulltext: true, component: true }` should the output be this?:

```
{ :db/attribute
    { :db/id                   :foo/baz
      :db/valueType      :db.type/ref
      :db/cardinality      :db.cardinality/many
      :db/unique            :db.unique/identity
      :db/indexing         true
      :db/fulltext            true
      :db/isComponent true }
}
```

and, I'm sorry, but I am a little confused how to chain them together, with multiple attributes in the same schema. to give a single output for a large schema.

rnewman:
As an example, you're trying to produce something like the value of the `:schema/attributes` entry in

[...]

Issue #193, opened 2017-01-23
[tx] Don't parse EDN to construct bootstrap schema during bootstrap phase


ncalexan:
In #170, I got frustrated defining static structures in Rust and lazily parsed some EDN to represent symbolic schemas.  We then turn around and transform the parsed structures into simpler triples to iterate over.  This is all around the bootstrap `V2_SYMBOLIC_SCHEMA` code.

This ticket tracks defining a nested `[(IDENT, [(ATTRIBUTE, VALUE), ...]), ...]` structure in Rust directly, and avoiding using the EDN parser (and potentially even `Value` for the ident, attribute, and tree structure parts) during bootstrap.

This is a reasonable _[good first bug]_.

rnewman:
Somewhere I thought I filed an issue about parsing queries at compile time.

It's hard without `const fn` and perhaps some other support, but eventually there's no reason why a static EDN tree can't be parsed at compile time, all the way through to an abstract `transact` input or an abstract query.

Obviously we can also define a bunch of macros (or do it manually) to define the intermediate structures programmatically, which I guess is this ticket :)

PR #267, opened 2017-02-08
Ensure :db/index true for :db/unique _. Fixes #254. r=nalexander

### LINKED ISSUE #254: [tx] Make :db/unique _ imply :db/index true.
This was just an oversight.  From the Datomic documentation at http://docs.datomic.com/identity.html: "A unique identity attribute is always indexed by value, in order to support uniqueness checks. Specifying :db/index true is redundant and not recommended."  This is true for a unique value attribute as well.

This ticket tracks enforcing that assertion.  Concretely, we need to:
1. Encode the implication around https://github.com/mozilla/mentat/blob/rust/db/src/schema.rs#L109 and https://github.com/mozilla/mentat/blob/rust/db/src/schema.rs#L110
2. Ensure that it's always true around https://github.com/mozilla/mentat/blob/rust/db/src/schema.rs#L22

Note that at this time, Mentat takes no position on whether specifying :db/unique _ and :db/index true is "not recommended", but it doesn't really matter.

jsantell:
What's the best way to set up tests for the schema validations?

### DIFF
--- a/db/src/schema.rs
+++ b/db/src/schema.rs
@@ -19,6 +19,10 @@ fn validate_schema_map(entid_map: &EntidMap, schema_map: &SchemaMap) -> Result<(
     for (entid, attribute) in schema_map {
         let ident = entid_map.get(entid).ok_or(ErrorKind::BadSchemaAssertion(format!("Could not get ident for entid: {}", entid)))?;
 
+        if attribute.unique_value && !attribute.index {
+            bail!(ErrorKind::BadSchemaAssertion(format!(":db/unique :db/unique_value true without :db/index true for entid: {}", ident)))
+        }
+
         if attribute.unique_identity && !attribute.unique_value {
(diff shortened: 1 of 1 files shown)


ncalexan:
> What's the best way to set up tests for the schema validations?

Right now, you could use https://github.com/mozilla/mentat/blob/afafcd64a042d51ba5f8ac44e258e708b5922f6a/db/src/schema.rs#L65 to ensure a bad schema fails.

In the context of #260, The inverse of #260 -- turning an `edn::Value` symbolic schema into a `Schema` would be good for testing.  We already do this during bootstrapping, it's just spread between functions -- and #193 tracks not doing this at all, for efficiency during bootstrapping!  Perhaps follow-up to extract this to a function (mostly for testing, but maybe also for future use during schema fragment management) would be best, and then tests for this would be much easier?

Issue #256, opened 2017-02-08
[edn] Add edn::Value::into_* that consumes its argument


ncalexan:
This is similar to `edn::Value::as_*` from #197, but with signature `into_*(self)` rather than `as_*(&self)`.

This ticket tracks getting the patch at https://github.com/ncalexan/mentat/commit/8f2bbd661c3662370102789d0503b18a273eea3a rebased, reviewed, and landed.

PR #269, opened 2017-02-09
Expose EDN as_ methods that return copied values, not references.

rnewman:


ncalexan:
Man, with #256, we have absolutely everything, eh?

It seems to me like we really do want to give the flexibility to callers and accept that `edn::Value` is, in my Dad's words, a "hell bin" into which all manner of access is dropped.

This is fine by me if you're fine with #256.

edn/tests/tests.rs
```diff
@@ -1003,6 +1003,14 @@ fn test_utils_merge() {
     test(&right, &left, &expected);
 }
 
+macro_rules! def_test_as_value_type {
```
joewalker: Testing with macros feels somewhat dangerous to me, especially if we're implementing with macros. Real world usage is likely to be less hygienic and there's a risk of parallel systemic failures in the impl and test.
That said, this feels simple enough that mistakes are unlikely?

PR #275, opened 2017-02-09
Run pragmas to configure SQLite store and connection when we open a DB

### LINKED ISSUE #259: [db] Run pragmas to configure SQLite store and connection when we open a DB
This ticket tracks running SQLite pragma commands when we open a database, just like we do in the Clojure implementation.  See https://github.com/mozilla/mentat/blob/84a80f40f5c888f8452d07bd15f3b5fba49d3963/src/common/datomish/db_factory.cljc#L126.

We should make the `new_connection` function around https://github.com/mozilla/mentat/blob/c585715224c180ff84630c05aa763d0d271f350a/db/src/db.rs#L23 accept a URI (so it can handle file-backed and in-memory databases), and then run the pragmas in that function.

victorporof:
Fixes #259

### DIFF
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -11,6 +11,7 @@
 #![allow(dead_code)]
 
 use std::iter::once;
+use std::path::Path;
 
 use itertools;
 use itertools::Itertools;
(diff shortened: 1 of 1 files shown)


db/src/db.rs
```diff
@@ -27,8 +28,23 @@ use mentat_tx::entities as entmod;
...
 
-pub fn new_connection() -> rusqlite::Connection {
-    return rusqlite::Connection::open_in_memory().unwrap();
+pub fn new_connection<T>(uri: T) -> rusqlite::Result<rusqlite::Connection> where T: AsRef<Path> {
+    let res = match uri.as_ref().to_string_lossy().len() {
+        0 => rusqlite::Connection::open_in_memory(),
+        _ => rusqlite::Connection::open(uri),
+    };
+
+    if let Ok(conn) = res.as_ref() {
```
ncalexan: This would be simpler, I think, if it was like
```rust
fn (..) {
  let conn = match {
    _ => ... open(uri)?
  };
  conn.execute_batch()
}
```
The `as_ref` with a `Result` threw me a little.

victorporof:
Addressed comments.

db/src/db.rs
victorporof: Nice! Good idea.

Issue #258, opened 2017-02-08
[edn] Expose line/column/character span position information from parsed EDN streams


ncalexan:
This ticket tracks including line/column/character position information with parsed `edn::Value` trees.  It addresses https://github.com/mozilla/mentat/pull/149#issuecomment-271029255; it is a pre-requisite, I think, for #168.

Concretely, I think we want a pair type like:
```rust
struct edn::ValueAndSpan {
  value: edn::Value,
  span: edn::Span,
}
```
which encapsulates a `Span` of some description coming out of the PEG EDN parser.  Sadly, this means we'll need to rewrite or duplicate the `edn::Value` type to have `edn::ValueAndSpan` children.  I could imagine something like:

- `edn::Value<V>`, where `V` parameterizes the children (so `V = edn::Value` is what we have now)

or something like

- `edn::Value<S>`, where `S` parameterizes the span (so `S = ()` is what we have now).

We'll want to expose this new type, and probably convert to the existing type as well (just dropping all the spans).

This is pretty awkward -- better suggestions appreciated.  This could be a good first bug if the patch doesn't try to change existing code too much (i.e., just adds new types and conversions into existing types).

rnewman:
Why not just make `edn::Value` your `ValueAndSpan`?

Each fork of the enum would be `(value, span)`. Callers would need to match with `_` to discard the span, but that's OK: it forces them to think about it.

rnewman:
That is: we already have a type for just the internal value — the internal value itself.

ncalexan:
> Why not just make edn::Value your ValueAndSpan?

Two reasons:
* it churns a lot of existing code, making this not a good first bug;
* because it makes programatically generating `edn::Value` difficult.  You can introduce some `NullSpan`, I guess, but eventually you're going to want just "raw values".

PR #282, opened 2017-02-10
Expose line/column/character span position information from parsed EDN streams

victorporof:
Fixes https://github.com/mozilla/mentat/issues/258

### DIFF
--- a/edn/src/edn.rustpeg
+++ b/edn/src/edn.rustpeg
@@ -16,8 +16,8 @@ use std::f64::{NAN, INFINITY, NEG_INFINITY};
 
 use num::BigInt;
 use ordered_float::OrderedFloat;
-use types;
-use types::Value;
+
+use types::{SpannedValue, Span, ValueAndSpan};
 
 // Goal: Be able to parse https://github.com/edn-format/edn
 // Also extensible to help parse http://docs.datomic.com/query.html
@@ -28,20 +28,43 @@ use types::Value;
 // TODO: Support tagged elements
 // TODO: Support discard
 
-pub nil -> Value =
-    "nil" { Value::Nil }
-
-pub nan -> Value =
-    "#f" whitespace+ "NaN" { Value::Float(OrderedFloat(NAN)) }
-
-pub infinity -> Value =
-    "#f" whitespace+ s:$(sign) "Infinity" {
-        Value::Float(OrderedFloat(if s == "+" { INFINITY } else { NEG_INFINITY }))
+pub nil -> ValueAndSpan =
+    start:#position "nil" end:#position {
+        ValueAndSpan {
+            inner: SpannedValue::Nil,
+            span: Span(start, end)
+        }
     }
 
-pub boolean -> Value =
-    "true" { Value::Boolean(true) } /
-    "false" { Value::Boolean(false) }
+pub nan -> ValueAndSpan =
+    start:#position "#f" whitespace+ "NaN" end:#position {
+        ValueAndSpan {
+            inner: SpannedValue::Float(OrderedFloat(NAN)),
+            span: Span(start, end)
+        }
+    }
+
+pub infinity -> ValueAndSpan =
+    start:#position "#f" whitespace+ s:$(sign) "Infinity" end:#position {
+        ValueAndSpan {
+            inner: SpannedValue::Float(OrderedFloat(if s == "+" { INFINITY } else { NEG_INFINITY })),
+            span: Span(start, end)
+        }
+    }
+
+pub boolean -> ValueAndSpan =
+    start:#position "true" end:#position {
+        ValueAndSpan {
+            inner: SpannedValue::Boolean(true),
+            span: Span(start, end)
+        }
+    } /
+    start:#position "false" end:#position {
+        ValueAndSpan {
+            inner: SpannedValue::Boolean(false),
+            span: Span(start, end)
+        }
+    }
 
 digit = [0-9]
 alphanumeric = [0-9a-zA-Z]
@@ -50,30 +73,45 @@ validbase = [3][0-6] / [12][0-9] / [2-9]
 hex = [0-9a-fA-F]
 sign = "-" / "+"
 
-pub bigint -> Value =
-    b:$( sign? digit+ ) "N" {
-        Value::BigInteger(b.parse::<BigInt>().unwrap())
+pub bigint -> ValueAndSpan =
+    start:#position b:$( sign? digit+ ) "N" end:#position {
+        ValueAndSpan {
+            inner: SpannedValue::BigInteger(b.parse::<BigInt>().unwrap()),
+            span: Span(start, end)
+        }
     }
 
-pub octalinteger -> Value =
-    "0" i:$( octaldigit+ ) {
-        Value::Integer(i64::from_str_radix(i, 8).unwrap())
+pub octalinteger -> ValueAndSpan =
+    start:#position "0" i:$( octaldigit+ ) end:#position {
+        ValueAndSpan {
+            inner: SpannedValue::Integer(i64::from_str_radix(i, 8).unwrap()),
+            span: Span(start, end)
+        }
     }
 
-pub hexinteger -> Value =
-    "0x" i:$( hex+ ) {
-        Value::Integer(i64::from_str_radix(i, 16).unwrap())
+pub hexinteger -> ValueAndSpan =
+    start:#position "0x" i:$( hex+ ) end:#position {
+        ValueAndSpan {
+            inner: SpannedValue::Integer(i64::from_str_radix(i, 16).unwrap()),
+            span: Span(start, end)
+        }
     }
 
-pub basedinteger -> Value =
+pub basedinteger -> ValueAndSpan =
     // Only allow values 2-36
-    b:$( validbase ) "r" i:$( alphanumeric+ ) {
-        Value::Integer(i64::from_str_radix(i, b.parse::<u32>().unwrap()).unwrap())
+    start:#position b:$( validbase ) "r" i:$( alphanumeric+ ) end:#position {
+        ValueAndSpan {
+            inner: SpannedValue::Integer(i64::from_str_radix(i, b.parse::<u32>().unwrap()).unwrap()),
+            span: Span(start, end)
+        }
     }
 
-pub integer -> Value =
-    i:$( sign? digit+ ) {
-        Value::Integer(i.parse::<i64>().unwrap())
+pub integer -> ValueAndSpan =
+    start:#position i:$( sign? digit+ ) end:#position {
+        ValueAndSpan {
+            inner: SpannedValue::Integer(i.parse::<i64>().unwrap()),
+            span: Span(start, end)
+        }
     }
 
 frac =     sign? digit+ "." digit+
@@ -82,9 +120,12 @@ frac_exp = sign? digit+ "." digit+ ("e" / "E") sign? digit+
 
 // The order here is important - frac_exp must come before (exp / frac) or the
 // parser assumes exp or frac when the float is really a frac_exp and fails
-pub float -> Value =
-    f:$( frac_exp / exp / frac ) {
-        Value::Float(OrderedFloat(f.parse::<f64>().unwrap()))
+pub float -> ValueAndSpan =
+    start:#position f:$( frac_exp / exp / frac ) end:#position {
+        ValueAndSpan {
+            inner: SpannedValue::Float(OrderedFloat(f.parse::<f64>().unwrap())),
+            span: Span(start, end)
+        }
     }
 
 // TODO: \newline, \return, \space and \tab
@@ -93,9 +134,12 @@ quote = "\\\""
 tab = "\\tab"
 char = [^"] / special_char
 
-pub text -> Value =
-    "\"" t:$( char* ) "\"" {
-        Value::Text(t.to_string())
+pub text -> ValueAndSpan =
+    start:#position "\"" t:$( char* ) "\"" end:#position {
+        ValueAndSpan {
+            inner: SpannedValue::Text(t.to_string()),
+            span: Span(start, end)
+        }
     }
 
 namespace_divider = "."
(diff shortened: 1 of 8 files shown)


edn/src/edn.rustpeg
```diff
@@ -29,31 +30,60 @@ use types::Value;
 // TODO: Support discard
 
 pub nil -> Value =
-    "nil" { Value::Nil }
```
ncalexan: I feel like `rust-peg` probably solves this problem generically, marking up your tokens with spans automatically.  Did you investigate this?

edn/src/types.rs
```diff
@@ -17,9 +17,21 @@ use symbols;
...
 #[derive(PartialEq, Eq, Hash, Clone, Debug)]
-pub enum Value {
+pub struct Value {
+    pub span: Option<Span>,
+    pub inner: InnerValue,
+}
+
+/// Span represents the current offset into the input string.
+#[derive(PartialEq, Eq, Hash, Clone, Debug)]
+pub struct Span(usize);
```
ncalexan: Let's do what the Rust compiler and other places do, which is that a span is a start and an end (or a length, but I prefer end) into the input buffer.

edn/src/edn.rustpeg
victorporof: Need to add a #position myself. This is WIP.

edn/src/types.rs
victorporof: Rustpeg only gives positions. Compute the length myself?

victorporof:
Fine -_-

edn/src/types.rs
```diff
@@ -214,12 +264,6 @@ impl Value {
     }
 }
 
-impl From<f64> for Value {
```
ncalexan: One type can have many `From` implementations.  If anything, I want `edn::Value` to have `From` for _more_ types (when there's no ambiguity).  Can you explain why you removed this?

edn/src/types.rs
```diff
@@ -248,6 +292,18 @@ impl Ord for Value {
     }
 }
 
+impl PartialOrd for ValueAndSpan {
```
ncalexan: This surprises me.  I don't expect `ValueAndSpan` instances with _different_ spans to compare or order equal.

@rnewman, do you?

edn/src/types.rs
```diff
@@ -321,44 +377,45 @@ mod test {
...
     fn test_value_from() {
-        assert_eq!(Value::from(42f64), Value::Float(OrderedFloat::from(42f64)));
+        assert_eq!(Value::from_float(42f64), Value::Float(OrderedFloat::from(42f64)));
+        assert_eq!(Value::from_ordered_float(OrderedFloat::from(42f64)), Value::Float(OrderedFloat::from(42f64)));
         assert_eq!(Value::from_bigint("42").unwrap(), Value::BigInteger(BigInt::from(42)));
     }
 
     #[test]
     fn test_print_edn() {
         let string = "[ 1 2 ( 3.14 ) #{ 4N } { foo/bar 42 :baz/boz 43 } [ ] :five :six/seven eight nine/ten true false nil #f NaN #f -Infinity #f +Infinity ]";
```
ncalexan: You're mapping `Value` into `ValueAndSpan`, manually.  Why not map `ValueAndSpan` onto `Value`, which we need for all the other consumers?  As written, this isn't testing the span business anyway.

edn/src/types.rs
```diff
@@ -31,21 +43,33 @@ pub enum Value {
     NamespacedSymbol(symbols::NamespacedSymbol),
     Keyword(symbols::Keyword),
     NamespacedKeyword(symbols::NamespacedKeyword),
-    Vector(Vec<Value>),
+    Vector(Vec<ValueAndSpan>),
     // We're using a LinkedList here instead of a Vec or VecDeque because the
     // LinkedList is faster for appending (which we do a lot of).
     // See https://github.com/mozilla/mentat/issues/231
-    List(LinkedList<Value>),
+    List(LinkedList<ValueAndSpan>),
```
ncalexan: I think this should remain `Value`, for current consumers.

Then, we want a parallel hierarchy like you've defined here, which could have two types (`ValueAndSpan(inner: InnerValueAndSpan , span: Span)` like you have, and `InnerValueAndSpan` where the container values are `ValueAndSpan`, just like here.

The map between is clear.

edn/src/types.rs
ncalexan: Oh, I see, you made these macros.  For just a few implementations I think this trickery is overkill, but it's your call.

edn/src/edn.rustpeg
```diff
@@ -28,32 +29,61 @@ use types::Value;
 // TODO: Support tagged elements
 // TODO: Support discard
 
-pub nil -> Value =
-    "nil" { Value::Nil }
+pub nil -> ValueAndSpan =
+    start:#position "nil" end:#position {
+        ValueAndSpan {
+            inner: Nil,
+            span: Some(Span(start, end))
```
ncalexan: Ah, I see.  Nifty!

edn/src/types.rs
victorporof: I see. That's a lot of code duplication. No way around it?

edn/src/types.rs
victorporof: It's a leftover from the previous implementation, but makes this a bit more symmetrical/generic so I'll keep it.

edn/src/types.rs
victorporof: Not strictly necessary, I'll remove it if I have to.

edn/src/types.rs
victorporof: If I duplicate the hierarchy as described above, then yes, mapping ValueAndSpan onto Value would be possible here. Otherwise not (because Value::Collection was over ValueAndSpan)

victorporof:
Addressed comments and rebased.

victorporof:
@ncalexan this is ready for review.

edn/src/types.rs
```diff
@@ -214,9 +318,9 @@ impl Value {
     }
 }
 
-impl From<f64> for Value {
-    fn from(src: f64) -> Value {
-        Value::Float(OrderedFloat::from(src))
+impl ValueAndSpan {
+    pub fn reduce(self) -> Value {
```
ncalexan: This is a reduction, but Rust tends to name general reductions `fold` and special cases like this `into_inner`, to be clear that you're consuming `self`.

edn/src/types.rs
```diff
@@ -44,12 +44,96 @@ pub enum Value {
...
+    }
+}
+
+impl From<ValueAndSpan> for Value {
+    fn from(src: ValueAndSpan) -> Value {
+        Value::from(src.inner)
+    }
+}
+
+impl From<Value> for ValueAndSpan {
```
ncalexan: Hmm.  Not sure I want this to be trivial.

ncalexan:
Oh -- and do you think there might be a reason to just accept a generic type `S` for the span?  I could imagine other situations where one might want to mark up trees (with weights, scores, metadata) and the structure is there.

edn/src/types.rs
victorporof: Removed.

edn/src/types.rs
victorporof: Changed.

victorporof:
> Oh -- and do you think there might be a reason to just accept a generic type S for the span? I could imagine other situations where one might want to mark up trees (with weights, scores, metadata) and the structure is there.

I don't think there's any value in having that at this time.

edn/src/types.rs
victorporof: The reason for this was because it seemed useful to have a conversion between SpannedValue and Value, but in order to have that, we'd also need to have something from ValueAndSpan to value, because SpannedValue collections have ValueAndSpan children. Not having trivial conversion here meant that a lot of code (like all the from_foo, from_keyword, from_namespace etc. methods would've had to be duplicated. I've ultimately done it with macros to avoid any trivial conversion from Value to ValueAndSpan.

victorporof:
@ncalexan this is ready for final review. I've addressed all comments.

edn/src/edn.rustpeg
```diff
@@ -112,48 +156,69 @@ symbol_name = ( symbol_char_initial+ / "." ) ( symbol_char_subsequent* / "." )
...
+        ValueAndSpan {
+            inner: SpannedValue::Map(BTreeMap::from_iter(v)),
+            span: Span(start, end)
+        }
     }
 
 // It's important that float comes before integer or the parser assumes that
 // floats are integers and fails to parse
-pub value -> Value =
+pub value -> ValueAndSpan =
```
ncalexan: Is it not possible to do all of the `#position` stuff here, in `value`?  That is, capture the posititions before and after the `__` markers below, parse `Value` in the internal bits, and then convert to `ValueAndSpan`?  Maybe not, because of the internal whitespace that gets ignored?

Ah, as I think more about this, I think not: the recursive nature of `ValueAndSpan` denies that possibility, since the children of a container need to have the right type.  So sad.  (This might be possible with a `ValueAndSpan<S>` parameterized by an optional span type.)

edn/src/types.rs
```diff
@@ -44,12 +43,63 @@ pub enum Value {
     Map(BTreeMap<Value, Value>),
 }
 
-use self::Value::*;
+/// SpannedValue is the same as Value but used in ValueAndSpan
```
ncalexan: nit: `SpannedValue` is parallel to `Value` but container types have `ValueAndSpan` children.

edn/src/types.rs
```diff
@@ -44,12 +43,63 @@ pub enum Value {
...
+            SpannedValue::Map(v) => Value::Map(v.into_iter().map(|(x, y)| (x.fold(), y.fold())).collect()),
+        }
+    }
+}
 
 impl Display for Value {
     // TODO: Make sure float syntax is correct, handle NaN and escaping.
     // See https://github.com/mozilla/mentat/issues/232
     fn fmt(&self, f: &mut Formatter) -> ::std::fmt::Result {
+        use self::Value::*;
```
ncalexan: Nifty -- I did not know this was possible in a function scope.  TIL!

edn/src/types.rs
```diff
@@ -115,9 +187,9 @@ macro_rules! def_is {
     }
 }
 
-/// Creates `as_$TYPE` helper functions for Value, like `as_integer()`,
-/// which returns the underlying value representing this Value wrapped
-/// in an Option, like `<Option<i64>`.
+/// Creates `as_$TYPE` helper functions for Value or SpannedValue, like
+/// `as_integer()`, which returns the underlying value representing the
+/// original variable wrapped in an Option, like `<Option<i64>`.
```
ncalexan: nit: here and below, drop the outer `<>`.  So "like `Option<i64>`".

edn/src/types.rs
ncalexan: You called this `fold`, right?  This isn't a general `fold`, it's `into_inner`.

edn/tests/tests.rs
```diff
@@ -15,207 +15,424 @@ extern crate ordered_float;
...
+fn_parse_into_value!(keyword);
+fn_parse_into_value!(list);
+fn_parse_into_value!(vector);
+fn_parse_into_value!(set);
+fn_parse_into_value!(map);
+fn_parse_into_value!(value);
 
 #[test]
 fn test_nil() {
-    assert_eq!(nil("nil").unwrap(), Nil);
```
ncalexan: Throughout this whole file, you've converted a large number of tests from comparing against `T` to comparing against the `edn::Value` version of `T`.  That's a valid choice, but why did you choose it?  And could you have done this as a mechanical `Pre:` patch?  (With a commit comment explaining why you wanted this or what it makes easier later, etc, etc.)  As it stands I can't see what's really new here and what's just different.

tx-parser/tests/parser.rs
```diff
@@ -26,7 +26,7 @@ fn test_entities() {
     let input = r#"[[:db/add 101 :test/a "v"]
 [:db/retract 102 :test/b "w"]]"#;
 
-    let edn = parse::value(input).unwrap();
+    let edn = parse::value(input).unwrap().fold();
```
ncalexan: I know I suggested `into_inner`, but maybe we should be explicit and say `without_spans`?  That'll certainly read easiest...

edn/src/types.rs
```diff
@@ -248,7 +389,32 @@ impl Ord for Value {
...
 
+impl PartialOrd for SpannedValue {
+    fn partial_cmp(&self, other: &SpannedValue) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for SpannedValue {
+    fn cmp(&self, other: &SpannedValue) -> Ordering {
+        Value::from(self.clone()).cmp(&Value::from(other.clone()))
```
ncalexan: This is going to be very expensive.  Why do either of these types implement `Ord` and `PartialOrd` at all?  What is the desired ordering?  Is it a domain concept?  Is it because we're using `edn::Value` in some ordered data structure?  Maybe the inner `BTreeSet` and `BTreeMap`?  We might not want to do that.

Can we be clear on what we're doing here and why?

edn/src/edn.rustpeg
victorporof: As you've figured out, not possible due to the fact that SpannedValue has ValueAndSpan children in collections.

edn/src/types.rs
victorporof: I'll update the comment. Should be obvious though.

edn/src/types.rs
victorporof: Ok.

edn/tests/tests.rs
victorporof: To avoid even more changes to this file.

victorporof: I think I might be able to get away with a bit less changes, but the crux of the issue is that both SpannedValue and Value are imported the the tests file, and the edn parser returns ValueAndSpan now. Without those changes, we would't be checking Values anymore, but SpannedValues, so some kind of conversion needs to be done: this is via folding, or maintaining the previous intent to test Value, or something else. My approach was one of many possible ones. To make life easier for other people, I chose to drop the `*` import and make things specific.

edn/src/types.rs
victorporof: As discussed on Slack, cloning isn't necessary, just a shortcut I took in order to get something that is landable (as in: provides value that doesn't exist yet, is better than nothing etc.). Changing this implementation to avoid cloning is relatively trivial, sufficient for a good followup good first bug, but I'll do it in this PR.

tx-parser/tests/parser.rs
victorporof: Will do.

victorporof:
@ncalexan Addressed comments, avoided cloning in `Ord` implementation, and reduced test code churn as much as I could (although this seems to be strictly worse in terms of readability, because Value and Spanned value both have the same types so it's a bit hard to determine what exactly is being tested. Ready for another review.

victorporof:
@ncalexan I've lost some changes in a rebase, stand by.

victorporof:
Reflog rescued me, ready for review.

victorporof:
@ncalexan gentle review ping.

victorporof:
@ncalexan The failure is due to something landing in the meantime. Easy fix.

victorporof:
@ncalexan Fixed.

Issue #184, opened 2017-01-21
[tx] Handle id-literals in the transactor


ncalexan:
This is the hardest part of transacting, but I worked out an efficient approach in the Clojure implementation: see #88.  This depends on #169.

Similar to #183, a sophisticated implementation might have multiple regimes depending on the input characteristics.

ncalexan:
This is done: https://github.com/mozilla/mentat/commit/16e9740d8a117c4b9384c4be262b0b64fe941d9f.

PR #283, opened 2017-02-10
Implement upsert resolution algorithm.

ncalexan:
This PR will end up implementing #184.  Right now it's just a place-holder for some issues I want to file.

### DIFF
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -804,9 +804,9 @@ impl DB {
             match *entity {
                 Entity::AddOrRetract {
                     op: OpType::Add,
-                    e: entmod::EntidOrLookupRef::Entid(ref e_),
+                    e: entmod::EntidOrLookupRefOrTempId::Entid(ref e_),
                     a: ref a_,
-                    v: entmod::ValueOrLookupRef::Value(ref v_)} => {
+                    v: ref v_ } => {
 
                     let e: i64 = match e_ {
                         &entmod::Entid::Entid(ref e__) => *e__,
@@ -839,9 +839,9 @@ impl DB {
 
                 Entity::AddOrRetract {
                     op: OpType::Retract,
-                    e: entmod::EntidOrLookupRef::Entid(ref e_),
+                    e: entmod::EntidOrLookupRefOrTempId::Entid(ref e_),
                     a: ref a_,
-                    v: entmod::ValueOrLookupRef::Value(ref v_) } => {
+                    v: ref v_ } => {
 
                     let e: i64 = match e_ {
                         &entmod::Entid::Entid(ref e__) => *e__,
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -1114,25 +1114,40 @@ mod tests {
             let assertions: edn::Value = transaction.get(&edn::Value::NamespacedKeyword(symbols::NamespacedKeyword::new("test", "assertions"))).unwrap().clone();
             let expected_transaction: Option<&edn::Value> = transaction.get(&edn::Value::NamespacedKeyword(symbols::NamespacedKeyword::new("test", "expected-transaction")));
             let expected_datoms: Option<&edn::Value> = transaction.get(&edn::Value::NamespacedKeyword(symbols::NamespacedKeyword::new("test", "expected-datoms")));
+            let expected_error_message: Option<&edn::Value> = transaction.get(&edn::Value::NamespacedKeyword(symbols::NamespacedKeyword::new("test", "expected-error-message")));
 
             let entities: Vec<_> = mentat_tx_parser::Tx::parse(&[assertions][..]).unwrap();
 
-            let report = db.transact(&conn, entities).unwrap();
-            assert_eq!(report.tx_id, bootstrap::TX0 + index + 1);
+            let maybe_report = db.transact(&conn, entities);
 
             if let Some(expected_transaction) = expected_transaction {
+                if expected_transaction.is_nil() {
+                    assert!(maybe_report.is_err());
+
+                    if let Some(expected_error_message) = expected_error_message {
+                        let expected_error_message = expected_error_message.as_text();
+                        assert!(expected_error_message.is_some(), "Expected error message to be text:\n{:?}", expected_error_message);
+                        let error_message = maybe_report.unwrap_err().to_string();
+                        assert!(error_message.contains(expected_error_message.unwrap()), "Expected error message:\n{}\nto contain:\n{}", error_message, expected_error_message.unwrap());
+                    }
+                    continue
+                }
+
+                let report = maybe_report.unwrap();
+                assert_eq!(report.tx_id, bootstrap::TX0 + index + 1);
+
                 let transactions = debug::transactions_after(&conn, &db, bootstrap::TX0 + index).unwrap();
                 assert_eq!(transactions.0.len(), 1);
-                assert_eq!(transactions.0[0].into_edn(),
-                           *expected_transaction,
-                           "\n{} - expected transaction:\n{}\n{}", label, transactions.0[0].into_edn(), *expected_transaction);
+                assert_eq!(*expected_transaction,
+                           transactions.0[0].into_edn(),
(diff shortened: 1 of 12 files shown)


ncalexan:
@rnewman, @jsantell this is ready for initial review.  There are some loose ends like collecting tempids (and testing them) and several follow up tickets (supporting lookup refs, map notation).  But it's best to get some review and hopefully get this stuff landed to get other people involved.

ncalexan:
It's worth mentioning that I do have a vision for the relationship between `DB`, `Tx`, and (future) `Conn`, but I'm not close enough to the implementation to worry too much about getting it right just yet.  So `Tx` and `DB` remain inter-twingled for now; the division of responsibilities will become clearer in time.

db/src/db.rs
```diff
@@ -759,12 +763,71 @@ impl DB {
...
+        self.transact_internal(conn, entities, tx, tx_instant)
+    }
+
     /// Transact the given `entities` against the given SQLite `conn`, using the metadata in
     /// `self.DB`.
     ///
     /// This approach is explained in https://github.com/mozilla/mentat/wiki/Transacting.
     // TODO: move this to the transactor layer.
-    pub fn transact_internal(&self, conn: &rusqlite::Connection, entities: &[Entity], tx_id: Entid, tx_instant: i64) -> Result<TxReport> {
+    pub fn transact_internal(&mut self, conn: &rusqlite::Connection, entities: &[Entity], tx_id: Entid, tx_instant: i64) -> Result<TxReport> {
```
rnewman: This is indeed starting to solidify on a wonky abstraction: `transact_internal` should take a reference to a `DB` as input, and return a report that refers to a (potentially changed) `DB`, no? Or, rather, altering the current internal `DB` of the conn is a side-effect.

Its contract is that the `DB` must match the contents of the database, so it can't take just any `DB`…

db/src/tx.rs
```diff
@@ -0,0 +1,263 @@
...
+use mentat_tx::entities as entmod;
+use mentat_tx::entities::{Entity, OpType};
+use rusqlite;
+use types::*;
+use upsert_resolution::Generation;
+
+/// A transaction on its way to being applied.
+#[derive(Debug)]
+pub struct Tx<'conn> {
+    /// The metadata to use to interpret the transaction entities with.
```
rnewman: Ah, gotcha. I just didn't get this far yet.

db/src/tx.rs
```diff
@@ -0,0 +1,263 @@
...
+    pub db: &'conn mut DB,
+
+    /// The SQLite connection to apply against.  In the future, this will be a Mentat connection.
+    pub conn: &'conn rusqlite::Connection,
+
+    /// The transaction ID of the transaction.
+    pub tx_id: Entid,
+
+    /// The timestamp when the transaction began to be commited.
```
rnewman: "committed"

db/src/tx.rs
```diff
@@ -0,0 +1,263 @@
...
+    ///
+    /// This is milliseconds after the Unix epoch according to the transactor's local clock.
+    // TODO: :db.type/instant.
+    pub tx_instant: i64,
+}
+
+impl<'conn> Tx<'conn> {
+    pub fn new(db: &'conn mut DB, conn: &'conn rusqlite::Connection, tx_id: Entid, tx_instant: i64) -> Tx<'conn> {
+        Tx { db: db,
```
rnewman: Newline after `{`

db/src/tx.rs
```diff
@@ -0,0 +1,263 @@
...
+             conn: conn,
+             tx_id: tx_id,
+             tx_instant: tx_instant }
```
rnewman: Trailing comma and `}` on its own line

db/src/db.rs
```diff
@@ -689,130 +762,67 @@ impl DB {
...
-
-                    // This is our chance to do schema-aware typechecking: to either assert that the
-                    // given value is in the attribute's value set, or (in limited cases) to coerce
-                    // the value into the attribute's value set.
-                    let typed_value: TypedValue = self.to_typed_value(v_, &attribute)?;
-
-                    let added = false;
+    /// Update the current partition map materialized view.
+    // TODO: only update changed partitions.
+    pub fn update_partition_map(&self, conn: &rusqlite::Connection) -> Result<()> {
```
rnewman: Is it sensible to use `&mut rusqlite::Connection` here as a signal that this is a mutating operation?

rnewman: Indeed, only one thread should ever be using the same `Connection`, so perhaps we want `type SQLite = &mut rusqlite::Connection;`…?

db/src/db.rs
```diff
@@ -26,7 +26,7 @@ use entids;
 use mentat_tx::entities as entmod;
 use mentat_tx::entities::{Entity, OpType};
 use errors::{ErrorKind, Result, ResultExt};
-use types::{Attribute, AttributeBitFlags, DB, Entid, IdentMap, Partition, PartitionMap, Schema, TypedValue, ValueType};
+use types::{AVMap, AVPair, Attribute, AttributeBitFlags, DB, Entid, IdentMap, Partition, PartitionMap, Schema, TypedValue, TxReport, ValueType};
```
joewalker: Doesn't that just replace a "too wide" problem with a "too high" problem? Or maybe I'm taking you to literally?

    use stuff::{
        Long, List, Of, Things, To, Import,
        Split, Over, Several, Lines,
    };

?

rnewman: It leaves us in Java's import situation: a sorted line-list with one diff line per insert or removal. Not the most compact representation, but good at minimally and clearly representing change over time.

ncalexan: I'm happy to push these onto multiple lines for clean diffs.  I'll do it throughout.

db/src/db.rs
```diff
@@ -759,12 +763,71 @@ impl DB {
...
+        // supported in the Clojure implementationat all, and might not be supported in Mentat soon,
+        // so this is very low priority.
+        let mut stmt = conn.prepare_cached(s.as_str())?;
+        stmt.execute(&params[..])
+            .map(|_c| ())
+            .chain_err(|| "Could not update partition map")
+    }
+
+    /// Allocate a single fresh entid in the given `partition`.
+    fn allocate_entid(&mut self, partition: String) -> i64 {
```
ncalexan: Done.

db/src/db.rs
ncalexan: I see you commented further, but I think a duality in many Rust expressions:
- `&mut self -> ()` and caller owns coordination;
- `&self -> Self` and callee owns coordination.

I elected for the former since I expect to build a `Conn` to manage the coordination in #296.

db/src/tx.rs
ncalexan: Fixed, here and another place.

db/src/tx.rs
ncalexan: Done.

db/src/db.rs
ncalexan: I haven't really worked out the details, but I think there's an entirely different division of responsibilities in the future.  Right now, the data structures `DB` and `Tx` "own" the work of updating the SQLite store and processing an incoming transaction.  My vision is that there's a split between the "updating the SQLite gubbins" and "term rewriting, resolution, etc" on top.

Following your suggestion, we might indeed define type
```rust
type SQLiteConnection = &mut rusqlite::Connection;
```
and a trait that defines the "low level interface", like
```rust
trait ConcreteRepresentation {
  pub fn lookup_avs(...);
  pub fn insert_non_fts_one(...);
  ...
}
```
Then, we might
```rust
impl ConcreteRepresentation for SQLiteConnection {
  pub fn lookup_avs(...) { // Using SQLite-specific features, etc.
  ...
}
```

That would separate the `DB` type from the implementation details that are specific to SQLite.  You could imagine that some `ConcreteRepresentation` implementations might not support fulltext indexing at all; or that there's an implementation that uses Postgres or some other backing store.

Does that seem sensible?

PR #286, opened 2017-02-10
Add octal, hexadecimal, and arbitrary base integers to the EDN parser

jsantell:


edn/src/edn.rustpeg
```diff
@@ -44,13 +44,31 @@ pub boolean -> Value =
...
+        Value::Integer(i64::from_str_radix(i, 8).unwrap())
+    }
+
+pub hexinteger -> Value =
+    "0x" i:$( hex+ ) {
+        Value::Integer(i64::from_str_radix(i, 16).unwrap())
+    }
+
+pub basedinteger -> Value =
+    b:$(digit digit?) "r" i:$( alphanumeric+ ) {
```
jsantell: This throws when using a base that's not in `(2, 36)` -- but still gets parsed. Should this be handled differently?

jsantell: To clarify, it should fail when using a base outside of those values, but none of the other parser's RHS's have that opportunity AFAICT

rnewman: One option is to define the acceptable bases more accurately:

```
[2..9] / [12][0-9] / 3[0..6]
```

Another is to figure out how rust-peg expects rules to not apply. I believe that if you do this:

```
  b:$(…) … $(alphanumeric+) {?
      let base = b.parse::<u32>()?;
      let num = i64::from_str_radix(i, base)?
      Value::Integer(num)
  }
```

it'll do what you want.

victorporof:
@jsantell @rnewman There aren't any Display rules for those newly added integers. Is parsing/printing still isomorphic?

victorporof:
I guess these will print to regular base 10 integers, so we have isomorphism, just not for bases. I think that's fine?

rnewman:
Yes. We also don't preserve whitespace, string escaping choices, redundant plus signs, etc. etc.

PR #287, opened 2017-02-11
s/expected[datoms|transaction]/expected-*

jsantell:


jsantell:
Also updated [the wiki](https://github.com/mozilla/mentat/wiki/Transacting:-EDN-test-format) -- does anything need to be done for the cljs version?

PR #289, opened 2017-02-11
Stub out mentat::q_once.

rnewman:


query-parser/src/find.rs
```diff
@@ -158,6 +158,17 @@ fn parse_find_edn_map(map: BTreeMap<edn::Value, edn::Value>) -> QueryParseResult
     parse_find_map(m)
 }
 
+impl From<edn::parse::ParseError> for QueryParseError {
+    fn from(err: edn::parse::ParseError) -> QueryParseError {
+        QueryParseError::EdnParseError(err)
+    }
+}
+
+pub fn parse_find_string(string: &str) -> QueryParseResult {
```
ncalexan: There's an `edn::Value::From<&str>` lurking here, isn't there?

rnewman: Trait all the things!

PR #290, opened 2017-02-14
Move Schema from mentat_db to mentat_core.

rnewman:
This is another split-and-move-to-core PR: the error-producing schema-building complex bits move into a private trait for `mentat_db`, and the simple lookup-oriented type definition moves into `mentat_core` so I can use it during query algebrizing.

I also defined `SchemaMap` in terms of `Entid`, not `i64`.

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -222,6 +222,14 @@ impl Schema {
     pub fn attribute_for_entid(&self, x: &Entid) -> Option<&Attribute> {
         self.schema_map.get(x)
     }
+
+    pub fn is_attribute(&self, x: &Entid) -> bool {
+        self.schema_map.contains_key(x)
+    }
+
+    pub fn identifies_attribute(&self, x: &String) -> bool {
+        self.get_entid(x).map(|e| self.is_attribute(e)).unwrap_or(false)
+    }
 }
 
 #[cfg(test)]
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -212,20 +212,20 @@ pub struct Schema {
 }
 
 impl Schema {
-    pub fn get_ident(&self, x: &Entid) -> Option<&String> {
-        self.entid_map.get(x)
+    pub fn get_ident(&self, x: Entid) -> Option<&String> {
+        self.entid_map.get(&x)
     }
 
-    pub fn get_entid(&self, x: &String) -> Option<&Entid> {
-        self.ident_map.get(x)
+    pub fn get_entid(&self, x: &String) -> Option<Entid> {
+        self.ident_map.get(x).map(|x| *x)
     }
 
-    pub fn attribute_for_entid(&self, x: &Entid) -> Option<&Attribute> {
-        self.schema_map.get(x)
+    pub fn attribute_for_entid(&self, x: Entid) -> Option<&Attribute> {
+        self.schema_map.get(&x)
     }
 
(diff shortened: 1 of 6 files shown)


rnewman:
I stacked some helpers to this, and also got rid of the need to pass references to `Entid`s in and out. They're `Copy`, so it's trivial to just dereference them on the way out, and of course take references when needed for `BTreeMap`.

core/src/lib.rs
```diff
@@ -173,6 +175,55 @@ impl Default for Attribute {
...
+impl Schema {
+    pub fn get_ident(&self, x: &Entid) -> Option<&String> {
+        self.entid_map.get(x)
+    }
+
+    pub fn get_entid(&self, x: &String) -> Option<&Entid> {
+        self.ident_map.get(x)
+    }
+
+    pub fn attribute_for_entid(&self, x: &Entid) -> Option<&Attribute> {
```
ncalexan: Might as well stick the `require_` variants here too.

core/src/lib.rs
```diff
@@ -222,6 +222,14 @@ impl Schema {
     pub fn attribute_for_entid(&self, x: &Entid) -> Option<&Attribute> {
         self.schema_map.get(x)
     }
+
+    pub fn is_attribute(&self, x: &Entid) -> bool {
+        self.schema_map.contains_key(x)
+    }
+
+    pub fn identifies_attribute(&self, x: &String) -> bool {
```
ncalexan: I had to think about what you meant here, so include the docstrings, please!

core/src/lib.rs
rnewman: @ncalexan: for the record, I didn't move the `require_` methods because that would require schlepping all of the `db` error machinery into `core`. The error kinds include rusqlite errors like `BadSQLValuePair`, and I don't know how `error_chain!` would handle merging two sets of errors from two different crates. So… I left it, because I don't need 'em!

core/src/lib.rs
rnewman: Done in bfb6230.

ncalexan:
On Mon, Feb 13, 2017 at 7:39 PM, Richard Newman <notifications@github.com>
wrote:

> *@rnewman* commented on this pull request.
> ------------------------------
>
> In core/src/lib.rs <https://github.com/mozilla/mentat/pull/290>:
>
> > +    ///
> +    /// Invariant: key-set is the same as the key-set of `entid_map` (equivalently, the value-set of
> +    /// `ident_map`).
> +    pub schema_map: SchemaMap,
> +}
> +
> +impl Schema {
> +    pub fn get_ident(&self, x: &Entid) -> Option<&String> {
> +        self.entid_map.get(x)
> +    }
> +
> +    pub fn get_entid(&self, x: &String) -> Option<&Entid> {
> +        self.ident_map.get(x)
> +    }
> +
> +    pub fn attribute_for_entid(&self, x: &Entid) -> Option<&Attribute> {
>
> @ncalexan <https://github.com/ncalexan>: for the record, I didn't move
> the require_ methods because that would require schlepping all of the db
> error machinery into core. The error kinds include rusqlite errors like
> BadSQLValuePair, and I don't know how error_chain! would handle merging
> two sets of errors from two different crates. So… I left it, because I
> don't need 'em!
>
Roger that. Another thing we'll need to figure out eventually, but it can
be deferred and I have work to be done in that area anyway.

PR #300, opened 2017-02-15
Store Ident/Entid mappings as NamespacedSymbols

jsantell:
First round, stopped at this point because I think it might make more sense for the mappings to hold a reference to a NamespacedKeyword instead, any issues you can imagine doing that? Lots of cloning currently. Haven't dealt with much ownership/borrowing code in Rust.

### DIFF
--- a/db/src/bootstrap.rs
+++ b/db/src/bootstrap.rs
@@ -14,6 +14,7 @@ use ::{to_namespaced_keyword};
 use edn;
 use errors::{ErrorKind, Result};
 use edn::types::Value;
+use edn::symbols;
 use entids;
 use db::TypedSQLValue;
 use mentat_tx::entities::Entity;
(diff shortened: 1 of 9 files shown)


db/src/debug.rs
```diff
@@ -107,7 +107,7 @@ impl Transactions {
 
 /// Convert a numeric entid to an ident `Entid` if possible, otherwise a numeric `Entid`.
 fn to_entid(db: &DB, entid: i64) -> Entid {
-    db.schema.get_ident(entid).and_then(|ident| to_namespaced_keyword(&ident)).map_or(Entid::Entid(entid), Entid::Ident)
+    db.schema.get_ident(entid).map_or(Entid::Entid(entid), |ref ident| Entid::Ident(&ident))
```
jsantell: This doesn't compile -- ownership issues

jsantell:
Rebased again

db/src/bootstrap.rs
```diff
@@ -195,14 +197,17 @@ fn symbolic_schema_to_triples(ident_map: &IdentMap, symbolic_schema: &Value) ->
                             // bootstrap symbolic schema, or by representing the initial bootstrap
                             // schema directly as Rust data.
                             let typed_value = match TypedValue::from_edn_value(value) {
-                                Some(TypedValue::Keyword(ref s)) => TypedValue::Ref(*ident_map.get(s).ok_or(ErrorKind::UnrecognizedIdent(s.clone()))?),
+                                Some(TypedValue::Keyword(ref s)) => {
+                                    let entid = ident_map.get(&to_namespaced_keyword(s).unwrap());
```
rnewman: If you unwrap here, you should comment about why it's safe… but I think you probably want to not unwrap.

db/src/db.rs
```diff
@@ -395,8 +396,9 @@ impl TypedSQLValue for TypedValue {
 /// Read the ident map materialized view from the given SQL store.
 pub fn read_ident_map(conn: &rusqlite::Connection) -> Result<IdentMap> {
     let mut stmt: rusqlite::Statement = conn.prepare("SELECT ident, entid FROM idents")?;
-    let m = stmt.query_and_then(&[], |row| -> Result<(String, Entid)> {
-        Ok((row.get(0), row.get(1)))
+    let m = stmt.query_and_then(&[], |row| -> Result<(symbols::NamespacedKeyword, Entid)> {
+        let ident: String = row.get(0);
+        Ok((to_namespaced_keyword(&ident).unwrap(), row.get(1)))
```
rnewman: Again, it's probably safest not to unwrap — we're already returning `Result`, so you could simply have `to_namespaced_keyword` return a `Result`, or make one with `.ok_or`.

edn/src/symbols.rs
```diff
@@ -277,3 +277,30 @@ impl Display for NamespacedKeyword {
         write!(f, ":{}/{}", self.namespace, self.name)
     }
 }
+
+#[macro_export]
+macro_rules! ns_keyword {
+    ($name: expr) => {{
+        let splits = [':', '/'];
+        let mut i = $name.split(&splits[..]);
```
rnewman: This stuff happens at runtime. I suspect you want to just do `ns_keyword { ($ns: expr, $name: expr) …` and make the callers more efficient!

jsantell:
Fixed macro usage, and does not prematurely unwrap `Option` types

edn/src/symbols.rs
```diff
@@ -277,3 +277,29 @@ impl Display for NamespacedKeyword {
         write!(f, ":{}/{}", self.namespace, self.name)
     }
 }
+
+macro_rules! ns_keyword {
+    ($name: expr) => {{
+        let splits = [':', '/'];
+        let mut i = $name.split(&splits[..]);
```
ncalexan: I realized a while back that this isn't quite right -- it will accept the splits in the wrong order, like `/bad:keyword`.  File a follow-up to do this in two stages?

db/src/bootstrap.rs
```diff
@@ -33,63 +34,63 @@ use values;
...
-             (":db.type/boolean",      entids::DB_TYPE_BOOLEAN),
-             (":db.type/instant",      entids::DB_TYPE_INSTANT),
-             (":db.type/bytes",        entids::DB_TYPE_BYTES),
-             (":db.cardinality/one",   entids::DB_CARDINALITY_ONE),
-             (":db.cardinality/many",  entids::DB_CARDINALITY_MANY),
-             (":db.unique/value",      entids::DB_UNIQUE_VALUE),
-             (":db.unique/identity",   entids::DB_UNIQUE_IDENTITY),
-             (":db/doc",               entids::DB_DOC),
+    static ref V1_IDENTS: Vec<(symbols::NamespacedKeyword, i64)> = {
+        vec![(ns_keyword!("db", "ident"),             entids::DB_IDENT),
```
ncalexan: You introduced a nice macro for doing the splitting in the previous patch, and then you don't use that nice form in this patch.  I don't think you can do the string checking at compile time with a macro (you probably can with a build script), so is this by design?  If it is, get rid of the earlier patch entirely.

db/src/bootstrap.rs
```diff
@@ -195,14 +197,20 @@ fn symbolic_schema_to_triples(ident_map: &IdentMap, symbolic_schema: &Value) ->
...
+                                        .and_then(|ident| ident_map.get(&ident))
+                                        .ok_or(ErrorKind::UnrecognizedIdent(s.clone()))?;
+
+                                    TypedValue::Ref(*entid)
+                                },
                                 Some(v) => v,
                                 _ => bail!(ErrorKind::BadBootstrapDefinition(format!("Expected Mentat typed value for value but got '{:?}'", value)))
                             };
 
                             triples.push((ident.clone(),
```
ncalexan: This was long and is now short; consider putting it on one line.

db/src/bootstrap.rs
```diff
@@ -195,14 +197,20 @@ fn symbolic_schema_to_triples(ident_map: &IdentMap, symbolic_schema: &Value) ->
                             // bootstrap symbolic schema, or by representing the initial bootstrap
                             // schema directly as Rust data.
                             let typed_value = match TypedValue::from_edn_value(value) {
-                                Some(TypedValue::Keyword(ref s)) => TypedValue::Ref(*ident_map.get(s).ok_or(ErrorKind::UnrecognizedIdent(s.clone()))?),
+                                Some(TypedValue::Keyword(ref s)) => {
+                                    let entid = to_namespaced_keyword(s)
```
ncalexan: This is fine, but it would be more consistent if you further `map` all the way into the `TypedValue::Ref`.

db/src/db.rs
```diff
@@ -387,21 +388,30 @@ impl TypedSQLValue for TypedValue {
...
 }
 
 /// Read the ident map materialized view from the given SQL store.
 pub fn read_ident_map(conn: &rusqlite::Connection) -> Result<IdentMap> {
     let mut stmt: rusqlite::Statement = conn.prepare("SELECT ident, entid FROM idents")?;
-    let m = stmt.query_and_then(&[], |row| -> Result<(String, Entid)> {
-        Ok((row.get(0), row.get(1)))
+    let m = stmt.query_and_then(&[], |row| -> Result<(symbols::NamespacedKeyword, Entid)> {
+        let ident: String = row.get(0);
+        to_namespaced_keyword(&ident)
```
ncalexan: It's possible to make `let ident: NamespacedKeyword = row.get(0)` just work, by implementing `rusqlite::types::FromSql`.  However, that's awkward because the type is in the `edn` crate and the trait is in the `rusqlite` trait.  It might be worth newtyping this to be automatic, but we can do that as follow-up.  This might just get better with your next ticket, too.

db/src/debug.rs
```diff
@@ -107,7 +107,10 @@ impl Transactions {
 
 /// Convert a numeric entid to an ident `Entid` if possible, otherwise a numeric `Entid`.
 fn to_entid(db: &DB, entid: i64) -> Entid {
-    db.schema.get_ident(entid).and_then(|ident| to_namespaced_keyword(&ident)).map_or(Entid::Entid(entid), Entid::Ident)
```
ncalexan: Just for your understanding of the mapping functions, this could also be expressed:
```rust
db.schema.get_ident(entid).map_or(Entid::Entid(entid), |ident| Entid::Ident(ident.clone()))
```

edn/src/symbols.rs
```diff
@@ -278,28 +278,17 @@ impl Display for NamespacedKeyword {
     }
 }
 
+#[macro_export]
 macro_rules! ns_keyword {
```
ncalexan: Is it not possible to use the macro you introduced in the previous part?  Is there something about cross-crate macro usage happening?  Is it awkward in tests?

db/src/bootstrap.rs
jsantell: The first patch had string checking at runtime, as per @rnewman's suggestion, it now checks at compile time (although less beautiful in usage)

edn/src/symbols.rs
jsantell: I think there's a misunderstanding between the two commits; they'll be squashed

db/src/db.rs
jsantell: Filed #330

Issue #317, opened 2017-02-15
[query] Better debug representation for ConjoiningClauses


rnewman:
It currently prints like:

```
…
        Variable(
            PlainSymbol(
                "?y"
            )
        ): [
            QualifiedAlias(
                "datoms01",
                Value
            )
        ]
    },
    known_types: {
        Variable(
            PlainSymbol(
                "?x"
            )
        ): Ref,
        Variable(
            PlainSymbol(
                "?y"
            )
        ): Boolean
    },
```

which is really verbose. I'd prefer:

```
  known_types:
    ?x:     Ref
    ?y:     Boolean
```

Issue #320, opened 2017-02-16
[query] Extend CC test cases to use EDN input (and perhaps EDN output)


rnewman:
They currently painstakingly construct the input pattern and schema. It'd be nice if they were easier to read.

joewalker:
Can you give a pointer to the tests in question, and something that looks nice?

rnewman:
The tests are indeed those; they look like:

```rust
        cc.apply_pattern(&schema, &Pattern {
            source: None,
            entity: PatternNonValuePlace::Variable(Variable(PlainSymbol::new("?x"))),
            attribute: PatternNonValuePlace::Ident(NamespacedKeyword::new("foo", "bar")),
            value: PatternValuePlace::Constant(NonIntegerConstant::Boolean(true)),
            tx: PatternNonValuePlace::Placeholder,
});
```

You can imagine a version that runs through the `pattern` rule of the query parser after an EDN parse, allowing us to write:

```rust
    let pattern = Pattern::from_edn("[?x :foo/bar true _]");
    cc.apply_pattern(&schema, &pattern);
```

which is substantially clearer.

ncalexan:
@rnewman also filed #317 to make debug output of `ConjoiningClauses` nicer; I'd like to gently suggest that we turn clauses into EDN and use the EDN matcher @victorporof is working on in #271 to make sure we're getting the desired output.  It might be that the EDN matcher would need to grow some string/regex features to support these tests, but that's oaky.

This would be a nice unification of our testing patterns, and might again allow to run the same tests against the Clojure and Rust implementations, a la #188.  (For the record, I think this is both less likely and less valuable than sharing the transactor tests between the Clojure and Rust implementations.)

rnewman:
One of the problems with using textual/structural comparison is that it forces tests to be written in terms of exact comparison, rather than expressing expected arbitrary invariants. It also makes it less natural to test intermediate states.

The majority of the testing I do would not use EDN comparison, because there's always _something_ for which I want to say "there's an entry for `?x` but I don't care what it is", or "the thing in this map matches this thing in this array over here", or whatever.

Furthermore, making test cases test some output format breaks one very good piece of unit testing advice, which is that tests should be orthogonal. Each thing you alter in your code should, ideally, break one test. Template-based testing means that you break _most of your tests_, because each one redundantly checks the same things, obscuring the real breakage. Add a field to the output? Now you need to put a correct value for that field in every test result string.

[...]

PR #321, opened 2017-02-16
Initial stab at a better debug for ConjoiningClauses (i.e. #317)

joewalker:


joewalker:
Clearly, I don't think this is 'done'. I'm wondering if you were expecting this to be done inside ConjoiningClauses.fmt rather than spread out across Variable, ValueType etc.

query/src/lib.rs
```diff
@@ -68,6 +69,12 @@ impl Variable {
     }
 }
 
+impl fmt::Debug for Variable {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.0)
```
rnewman: I'm happy with this kind of change: "`Debug` should format the output in a programmer-facing, debugging context", and we know implicitly that it's a `Variable`. But I can imagine other developers _not_ being happy about `Variable` and `PlainSymbol` now printing the same!

joewalker: I wonder if we could we solve that with

    write!(f, "{} #_ (variable)", self.0)

Or something similar?
(It feels like it's more natural that `PlainSymbol` should be the default.

rnewman: There's no requirement that `Debug` output follow any particular format, so no need for EDN — we could just as easily do `write!(f, "var({})", self.0)` if we wanted to be unambiguous.

ncalexan:
Per https://github.com/mozilla/mentat/issues/320#issuecomment-280406436, I'd like to gently suggest that we convert clauses to EDN and use that output for testing.

rnewman:
I'd separate `Debug` printing (intended for humans to read) from any kind of stable programmatic string representation used for testing. And I don't think that escapes from doing white-box testing of the in-memory data structures…

joewalker:
> Per #320 (comment), I'd like to gently suggest that we convert clauses to EDN and use that output for testing.

Currently pressing on because I agree that debug printing is a different use case from testing.

joewalker:
So the output from testing was:

```
ConjoiningClauses {
    is_known_empty: false,
    from: [
        SourceAlias(
            Datoms,
            "datoms00"
        ),
        SourceAlias(
            Datoms,
            "datoms01"
        )
    ],
    wheres: [
        EqualsEntity(
            QualifiedAlias(
                "datoms00",
                Attribute
            ),
            98
        ),
        EqualsValue(
            QualifiedAlias(
                "datoms00",
                Value
            ),
            String(
                "idgoeshere"
            )
        ),
        EqualsEntity(
            QualifiedAlias(
                "datoms01",
                Attribute
            ),
            99
        ),
        EqualsColumn(
            QualifiedAlias(
                "datoms00",
                Entity
            ),
            QualifiedAlias(
                "datoms01",
                Entity
            )
        )
    ],
    bindings: {
        Variable(
            PlainSymbol(
                "?x"
            )
        ): [
            QualifiedAlias(
                "datoms00",
                Entity
            ),
            QualifiedAlias(
                "datoms01",
                Entity
            )
        ],
        Variable(
            PlainSymbol(
                "?y"
            )
        ): [
            QualifiedAlias(
                "datoms01",
                Value
            )
        ]
    },
    known_types: {
        Variable(
            PlainSymbol(
                "?x"
            )
        ): Ref,
        Variable(
            PlainSymbol(
                "?y"
            )
        ): Boolean
    },
    extracted_types: {}
}
```

joewalker:
And with this push it goes down to:

```
ConjoiningClauses {
    is_known_empty: false,
    from: [
        SourceAlias(Datoms, datoms00),
        SourceAlias(Datoms, datoms01)
    ],
    wheres: [
        cc::Entity(datoms00/Attribute = 98),
        cc::Value(datoms00/Value = String("idgoeshere")),
        cc::Entity(datoms01/Attribute = 99),
        cc::Column(datoms00/Entity = datoms01/Entity)
    ],
    bindings: {
        var(?x): [
            QualifiedAlias("datoms00", Entity),
            QualifiedAlias("datoms01", Entity)
        ],
        var(?y): [
            QualifiedAlias("datoms01", Value)
        ]
    },
    known_types: {
        var(?x): Ref,
        var(?y): Boolean
    },
    extracted_types: {}
}
```

query-algebrizer/src/cc.rs
```diff
@@ -13,6 +13,7 @@
 extern crate mentat_core;
 extern crate mentat_query;
 
+use std::fmt;
 use ::std::collections::BTreeMap;
```
rnewman: Could you make these styles match? I used `::std` out of some kind of misguided exploration of Rust style, I think; if your `std` works, let's do that.

query-algebrizer/src/cc.rs
```diff
@@ -115,13 +128,29 @@ pub fn default_table_aliaser() -> TableAliaser {
...
     EqualsEntity(QualifiedAlias, Entid),
     EqualsValue(QualifiedAlias, TypedValue),
     EqualsColumn(QualifiedAlias, QualifiedAlias),
 }
 
+impl fmt::Debug for ColumnConstraint {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            &ColumnConstraint::EqualsEntity(ref qa, ref entid) => {
+                write!(f, "cc::Entity({}/{:?} = {:?})", qa.0, qa.1, entid)
```
rnewman: I'd prefer these be:

```rust
   write!(f, "{}.{} = entity({:?})", qa.0, qa.1.as_str(), entid)
```

and so on — instead of

```
    cc::Entity(datoms00/Attribute = 98),
    cc::Column(datoms01/Value = datoms02/Entity),
```

it would be


```
    datoms00.a = entity(98),
    datoms01.v = datoms02.e,
```

which is much closer to the syntax we use in example SQL, and a little less crusted in syntax.

rnewman: If you really want these to look like standalone delimited things, wrap them in `[]`, perhaps? No preference from me.

joewalker: On the adding `[]` question, I wasn't sure that helped. I also considered removing the spaces either side of the equals.
In the end, no preference from me either so I left it as it was.

joewalker:
So the example output is now:

```
ConjoiningClauses {
    is_known_empty: false,
    from: [
        SourceAlias(Datoms, datoms00),
        SourceAlias(Datoms, datoms01)
    ],
    wheres: [
        datoms00.a = entity(98),
        datoms00.v = value(String("idgoeshere")),
        datoms01.a = entity(99),
        datoms00.e = datoms01.e
    ],
    bindings: {
        var(?x): [
            datoms00.e,
            datoms01.e
        ],
        var(?y): [
            datoms01.v
        ]
    },
    known_types: {
        var(?x): Ref,
        var(?y): Boolean
    },
    extracted_types: {}
}

```

joewalker:
Pausing before I push because I made the [requested change](https://github.com/mozilla/mentat/pull/321#discussion_r102291964) in SourceAlias instead, which changes things like the bindings.

rnewman:
Ship it!

PR #326, opened 2017-02-17
Codeify minimum rustc version in a build script so we can print a hel…

jsantell:
…pful message if a developer is using an out of date rustc. r=rnewman

### DIFF
--- a/build/version.rs
+++ b/build/version.rs
@@ -1,0 +1,26 @@
+// Copyright 2016 Mozilla
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use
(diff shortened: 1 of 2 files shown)


build/version.rs
```diff
@@ -0,0 +1,26 @@
...
+// specific language governing permissions and limitations under the License.
+
+extern crate rustc_version;
+
+use std::io::{self, Write};
+use std::process::exit;
+use rustc_version::version_matches;
+
+/// Change this when there's a new minimum version of rustc required
+/// to build the project
```
ncalexan: nit: full sentence.

PR #329, opened 2017-02-17
TypedValue::Keyword now wraps a NamespacedKeyword rather than a String. Fixes #203.

### LINKED ISSUE #203: [db] Wrap an edn::symbols::NamespacedKeyword rather than a String in db::TypedValue
Per https://github.com/mozilla/mentat/pull/171#discussion_r97866454.  Right now, we extract and clone a `String` in `db::TypedValue::Keyword`.  We'd like to keep this kind of data structured as much as possible, but I don't want to do this work for #170 before landing, so I'm tracking it here.

jsantell:
Only pay mind to 4634502, this is based off of #300 PR

### DIFF
--- a/db/src/bootstrap.rs
+++ b/db/src/bootstrap.rs
@@ -10,7 +10,6 @@
 
 #![allow(dead_code)]
 
-use ::{to_namespaced_keyword};
 use edn;
 use errors::{ErrorKind, Result};
 use edn::types::Value;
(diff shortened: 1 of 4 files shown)


db/src/bootstrap.rs
```diff
@@ -198,9 +197,8 @@ fn symbolic_schema_to_triples(ident_map: &IdentMap, symbolic_schema: &Value) ->
                             // schema directly as Rust data.
                             let typed_value = match TypedValue::from_edn_value(value) {
                                 Some(TypedValue::Keyword(ref s)) => {
-                                    let entid = to_namespaced_keyword(s)
-                                        .and_then(|ident| ident_map.get(&ident))
-                                        .ok_or(ErrorKind::UnrecognizedIdent(s.clone()))?;
+                                    let entid = ident_map.get(s)
```
ncalexan: There might be a `require` helper for `IdentityMap` lurking in here, like in `db/src/schema.rs`.  That'll give you one place for all these TODO references.

PR #333, opened 2017-02-17
Change to_namespaced_keyword(s) to return a Result

### LINKED ISSUE #331: Collapse `to_namespace_keyword(s)` error handling

jsantell:
Change to_namespaced_keyword(s) to return a Result rather than Option to reduce error handling throughout db code. Fixes #331. I think this is a good approach for reducing the to-be-implemented custom ErrorKind throughout code. You mentioned combining it into `schema.require_entid`, but that's a different usage (it checks the ident_map).

### DIFF
--- a/db/src/schema.rs
+++ b/db/src/schema.rs
@@ -11,7 +11,7 @@
 #![allow(dead_code)]
 
 use entids;
-use errors::*;
+use errors::{ErrorKind, Result};
 use edn::symbols;
 use mentat_core::{
     Attribute,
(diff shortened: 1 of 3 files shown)


db/src/schema.rs
```diff
@@ -11,7 +11,7 @@
 #![allow(dead_code)]
 
 use entids;
-use errors::*;
+use errors::{ErrorKind, Result};
```
ncalexan: Worth noting that `error-chain` explicitly suggests `use *` here; but I'd rather we're happy, so keep it.

PR #334, opened 2017-02-20
Remove most uses of `use foo::*`

joewalker:
Not expanded are nested instances of `use super::*`, but maybe we should add those too. I held off just because that would be code that breaks itself.

### DIFF
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -865,7 +865,6 @@ mod tests {
     use edn::symbols;
     use mentat_tx_parser;
     use rusqlite;
-    use types::*;
 
     #[test]
     fn test_open_current_version() {
--- a/db/src/internal_types.rs
+++ b/db/src/internal_types.rs
@@ -18,7 +18,12 @@ use std::rc::Rc;
 
 use errors;
 use errors::ErrorKind;
-use types::*;
+use types::{
+    AVMap,
+    AVPair,
+    Entid,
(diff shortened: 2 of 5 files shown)


ncalexan:
> Not expanded are nested instances of use super::*, but maybe we should add those too. I held off just because that would be code that breaks itself.

This is the standard in test code, AFAICT.

query-algebrizer/src/cc.rs
```diff
@@ -69,7 +69,7 @@ pub enum DatomsColumn {
 
 impl DatomsColumn {
     pub fn as_str(&self) -> &'static str {
-        use DatomsColumn::*;
+        use DatomsColumn::{Attribute, Entity, Tx, Value, ValueTypeTag};
```
ncalexan: I don't think we should expand this style, and will bring it up in the meeting.  The whole point is to avoid being verbose; we're trading `match _ { X::Y => ; X::Z => ; }` for `use X::{Y, Z}; match _ { X => ; Z => }`, which isn't a win to my eyes.

victorporof: I agree. I think it's fine to import * in *small* contained scopes like function bodies.

victorporof: Note that this also imports all inner types of a sum type, so there's zero change to break "unowned" code.

PR #335, opened 2017-02-20
Shell script for creating docs and running tests

joewalker:


mentat.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/sh
```
ncalexan: nit: license header?

mentat.sh
```diff
@@ -0,0 +1,45 @@
...
+
+doc() {
```
ncalexan: Could you extract a list of packages and interpolate it into the two commands?

victorporof:
It would seem that we _do_ have support for `cargo test --all`, so is this PR really necessary now?

joewalker:
I'll probably close this. Travis is failing with:

```
error: no method named `write_pretty` found for type `types::ValueAndSpan` in the current scope
  --> edn/src/pretty_print.rs:84:25
   |
84 |         assert_eq!(data.write_pretty(40, &mut Vec::new()).is_ok(), true);
   |                         ^^^^^^^^^^^^
```

Although it's passing for me.

joewalker:
Worth noting: we don't have the doc build support that this script provided, but that was kind of a nice to have anyway.

Issue #271, opened 2017-02-09
[tests] Use the EDN matcher for testing transaction output


ncalexan:
In #214, I implemented the first parts of #188, which tracks converting Clojure implementation tests into generic EDN format tests that could run against the Clojure implementation and the Rust Mentat implementation.

As I say in #188:

> It might be finicky to get the transaction IDs, timestamps, and other ephemeral data correct. I wonder if we can rewrite :db/tx and :db/txInstant in controlled locations to get what we want?

And that's what I did in #214: I translated transaction IDs and certain timestamps into `edn::Value::Symbol` instances like `?txN` and `?msN`, and then compared the resulting EDN with test EDN using the same symbols.  That is, _after_ reading from the store I post-processed the transaction assertions and datom assertions, etc, looking for particular patterns and combinations to rewrite.

This ticket tracks doing something more general.  I'd like to be able to take arbitrary EDN and _pattern match_ it against EDN, without having to postprocess.  I want to handle a few things:
- `_` matches arbitrary sub-EDN;
- `?name` matches sub-EDN, which must be identical each place `?name` appears;
- `...` matches arbitrary sibling-EDN

For example:
- `[_]` matches an arbitrary one-element vector
- `[_ _]`matches an arbitrary two-element vector
- `[?x ?x]` matches `[1 1]` and `[#{} #{}]` but not `[1 2]` or `[[] #{}]`
- `[1 ...]` matches `[1]` and `[1 1]` and `[1 2 3]`, but not `[2]` or `[2 1]` or `#{1 2}`.

Note that `...` is a valid EDN symbol.  Sibling EDN is a little tricky because it needs to handle un-ordered sets and maps, but I think it's do-able.

This is a particularly useful case of a structural EDN differ, as discussed in #220.

ncalexan:
This is a fun ticket if you know some Rust, but it's probably not a great first bug.  It's a good second bug, though!

You'll want to update:
-  the test harness around https://github.com/mozilla/mentat/blob/42580539b8a58f9d10e5891a22112f8c1b697bff/db/src/db.rs#L891 to use your new differ;
- the debug utilities that the harness uses to not post-process; see https://github.com/mozilla/mentat/blob/42580539b8a58f9d10e5891a22112f8c1b697bff/db/src/debug.rs#L130;
- the text data itself around https://github.com/mozilla/mentat/tree/42580539b8a58f9d10e5891a22112f8c1b697bff/tx/fixtures;
[...]

PR #338, opened 2017-02-21
Implement a basic edn matcher

victorporof:
Fixes part of #271

### DIFF
--- a/edn/src/matcher.rs
+++ b/edn/src/matcher.rs
@@ -1,0 +1,581 @@
+// Copyright 2016 Mozilla
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+use std::collections::HashMap;
+use std::cell::RefCell;
+use itertools::diff_with;
+
+use symbols;
+use types::Value;
+
+/// A trait defining pattern matching rules for any given pattern of type `T`.
+trait PatternMatchingRules<'a, T> {
+    /// Return true if the given pattern matches an arbitrary value.
+    fn matches_any(pattern: &T) -> bool;
+
+    /// Return the placeholder name if the given pattern matches a placeholder.
+    fn matches_placeholder(pattern: &'a T) -> Option<(&'a String)>;
+}
+
+/// A default type implementing `PatternMatchingRules` specialized on
+/// EDN values using plain symbols as patterns. These patterns are:
+/// * `_` matches arbitrary sub-EDN;
+/// * `?name` matches sub-EDN, which must be identical each place `?name` appears;
+struct DefaultPatternMatchingRules;
+
+impl<'a> PatternMatchingRules<'a, Value> for DefaultPatternMatchingRules {
+    fn matches_any(pattern: &Value) -> bool {
+        match *pattern {
+            Value::PlainSymbol(symbols::PlainSymbol(ref s)) => s.starts_with('_'),
+            _ => false
+        }
+    }
+
+    fn matches_placeholder(pattern: &'a Value) -> Option<(&'a String)> {
+        match *pattern {
+            Value::PlainSymbol(symbols::PlainSymbol(ref s)) => if s.starts_with('?') { Some(s) } else { None },
+            _ => None
+        }
+    }
+}
+
+/// Pattern matcher for EDN values utilizing specified pattern matching rules.
+/// For example, using this with `DefaultPatternMatchingRules`:
+/// * `[_]` matches an arbitrary one-element vector;
+/// * `[_ _]` matches an arbitrary two-element vector;
+/// * `[?x ?x]` matches `[1 1]` and `[#{} #{}]` but not `[1 2]` or `[[] #{}]`;
+struct Matcher<'a> {
+    placeholders: RefCell<HashMap<&'a String, &'a Value>>
+}
+
+impl<'a> Matcher<'a> {
+    /// Creates a Matcher instance.
+    fn new() -> Matcher<'a> {
+        Matcher {
+            placeholders: RefCell::default()
+        }
+    }
+
+    /// Performs pattern matching between two EDN `Value` instances (`value`
+    /// and `pattern`) utilizing a specified pattern matching ruleset `T`.
+    /// Returns true if matching succeeds.
+    pub fn match_with_rules<T>(value: &'a Value, pattern: &'a Value) -> bool
+    where T: PatternMatchingRules<'a, Value> {
+        let matcher = Matcher::new();
+        matcher.match_internal::<T>(value, pattern)
+    }
+
+    /// Recursively traverses two EDN `Value` instances (`value` and `pattern`)
+    /// performing pattern matching. Note that the internal `placeholders` cache
+    /// might not be empty on invocation.
+    fn match_internal<T>(&self, value: &'a Value, pattern: &'a Value) -> bool
+    where T: PatternMatchingRules<'a, Value> {
+        use Value::*;
+
+        if T::matches_any(pattern) {
+            true
+        } else if let Some(symbol) = T::matches_placeholder(pattern) {
+            let mut placeholders = self.placeholders.borrow_mut();
+            value == *placeholders.entry(symbol).or_insert(value)
+        } else {
+            match (value, pattern) {
+                (&Vector(ref v), &Vector(ref p)) =>
+                    diff_with(v, p, |a, b| self.match_internal::<T>(a, b)).is_none(),
+                (&List(ref v), &List(ref p)) =>
+                    diff_with(v, p, |a, b| self.match_internal::<T>(a, b)).is_none(),
+                (&Set(ref v), &Set(ref p)) =>
+                    v.len() == p.len() &&
+                    v.iter().all(|a| p.iter().any(|b| self.match_internal::<T>(a, b))) &&
+                    p.iter().all(|b| v.iter().any(|a| self.match_internal::<T>(a, b))),
+                (&Map(ref v), &Map(ref p)) =>
+                    v.len() == p.len() &&
+                    v.iter().all(|a| p.iter().any(|b| self.match_internal::<T>(a.0, b.0) && self.match_internal::<T>(a.1, b.1))) &&
+                    p.iter().all(|b| v.iter().any(|a| self.match_internal::<T>(a.0, b.0) && self.match_internal::<T>(a.1, b.1))),
+                _ => value == pattern
+            }
+        }
+    }
+}
+
+impl Value {
+    /// Performs default pattern matching between this value and some `pattern`.
+    /// Returns true if matching succeeds.
+    pub fn matches(&self, pattern: &Value) -> bool {
+        Matcher::match_with_rules::<DefaultPatternMatchingRules>(self, pattern)
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use parse;
+
+    macro_rules! assert_match {
+        ( $pattern:tt, $value:tt, $expected:expr ) => {
+            let pattern = parse::value($pattern).unwrap().without_spans();
+            let value = parse::value($value).unwrap().without_spans();
+            assert_eq!(value.matches(&pattern), $expected);
+        };
+        ( $pattern:tt =~ $value:tt ) => {
+            assert_match!($pattern, $value, true);
+        };
+        ( $pattern:tt !~ $value:tt ) => {
+            assert_match!($pattern, $value, false);
+        }
+    }
+
+    #[test]
+    fn test_match_primitives() {
+        assert_match!("nil" =~ "nil");
+        assert_match!("true" =~ "true");
+        assert_match!("true" !~ "false");
+        assert_match!("1" =~ "1");
+        assert_match!("1" !~ "2");
+        assert_match!("1N" =~ "1N");
+        assert_match!("1N" !~ "2N");
+        assert_match!("1.0" =~ "1.0");
+        assert_match!("1.0" !~ "2.0");
+        assert_match!("\"a\"" =~ "\"a\"");
+        assert_match!("\"a\"" !~ "\"b\"");
+        assert_match!("foo" =~ "foo");
+        assert_match!("foo" !~ "bar");
+        assert_match!("foo/bar" !~ "foo");
+    }
+
+    #[test]
+    fn test_match_collections_sorted() {
+        assert_match!("[nil, true, 1, \"foo\", bar, :baz]" =~ "[nil, true, 1, \"foo\", bar, :baz]");
+        assert_match!("(nil, true, 1, \"foo\", bar, :baz)" =~ "(nil, true, 1, \"foo\", bar, :baz)");
+        assert_match!("#{nil, true, 1, \"foo\", bar, :baz}" =~ "#{nil, true, 1, \"foo\", bar, :baz}");
+        assert_match!("{nil true, 1 \"foo\", bar :baz}" =~ "{nil true, 1 \"foo\", bar :baz}");
+    }
+
+    #[test]
+    fn test_match_collections_unsorted() {
+        assert_match!("[nil, true, 1, \"foo\", bar, :baz]" !~ "[1, \"foo\", nil, true, bar, :baz]");
+        assert_match!("(nil, true, 1, \"foo\", bar, :baz)" !~ "(1, \"foo\", nil, true, bar, :baz)");
+        assert_match!("#{nil, true, 1, \"foo\", bar, :baz}" =~ "#{1, \"foo\", nil, true, bar, :baz}");
+        assert_match!("{nil true, 1 \"foo\", bar :baz}" =~ "{1 \"foo\", nil true, bar :baz}");
+    }
+
+    #[test]
+    fn test_match_maps_switched_key_values() {
+        assert_match!("{1 2, 3 4}" =~ "{1 2, 3 4}");
+        assert_match!("{2 1, 3 4}" !~ "{1 2, 3 4}");
+        assert_match!("{2 1, 4 3}" !~ "{1 2, 3 4}");
+        assert_match!("{1 2, 4 3}" !~ "{1 2, 3 4}");
+    }
+
+    #[test]
+    fn test_match_maps_ordered_collection_keys_and_values() {
+        assert_match!("{[1, 2] (3, 4)}" =~ "{[1, 2] (3, 4)}");
+        assert_match!("{[2, 1] (3, 4)}" !~ "{[1, 2] (3, 4)}");
+        assert_match!("{[2, 1] (4, 3)}" !~ "{[1, 2] (3, 4)}");
+        assert_match!("{[1, 2] (4, 3)}" !~ "{[1, 2] (3, 4)}");
+
+        assert_match!("{(3, 4) [1, 2]}" !~ "{[1, 2] (3, 4)}");
+        assert_match!("{(3, 4) [2, 1]}" !~ "{[1, 2] (3, 4)}");
+        assert_match!("{(4, 3) [2, 1]}" !~ "{[1, 2] (3, 4)}");
+        assert_match!("{(4, 3) [1, 2]}" !~ "{[1, 2] (3, 4)}");
+    }
+
+    #[test]
+    fn test_match_maps_unordered_collection_keys_and_values() {
+        assert_match!("{#{1, 2} #{3, 4}}" =~ "{#{1, 2} #{3, 4}}");
+        assert_match!("{#{2, 1} #{3, 4}}" =~ "{#{1, 2} #{3, 4}}");
+        assert_match!("{#{2, 1} #{4, 3}}" =~ "{#{1, 2} #{3, 4}}");
+        assert_match!("{#{1, 2} #{4, 3}}" =~ "{#{1, 2} #{3, 4}}");
+
+        assert_match!("{#{3, 4} #{1, 2}}" !~ "{#{1, 2} #{3, 4}}");
+        assert_match!("{#{3, 4} #{2, 1}}" !~ "{#{1, 2} #{3, 4}}");
+        assert_match!("{#{4, 3} #{2, 1}}" !~ "{#{1, 2} #{3, 4}}");
+        assert_match!("{#{4, 3} #{1, 2}}" !~ "{#{1, 2} #{3, 4}}");
+    }
+
+    #[test]
+    fn test_match_any_simple() {
+        assert_match!("_" =~ "nil");
+        assert_match!("_" =~ "true");
+        assert_match!("_" =~ "1");
+        assert_match!("_" =~ "1N");
+        assert_match!("_" =~ "1.0");
+        assert_match!("_" =~ "\"a\"");
+        assert_match!("_" =~ "_");
+        assert_match!("_" =~ "symbol");
+        assert_match!("_" =~ "ns/symbol");
+        assert_match!("_" =~ ":keyword");
+        assert_match!("_" =~ ":ns/keyword");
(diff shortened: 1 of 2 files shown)


victorporof:
@ncalexan Does the `WIP` commit look as you'd expect? If not, I'd like to catch any errors in my approach early.

edn/src/matcher.rs
```diff
@@ -0,0 +1,114 @@
...
+        }
+        (&NamespacedSymbol(ref a), &NamespacedSymbol(ref b)) => a == b,
+        (&Keyword(ref a), &Keyword(ref b)) => a == b,
+        (&NamespacedKeyword(ref a), &NamespacedKeyword(ref b)) => a == b,
+        (&Vector(ref a), &Vector(ref b)) => diff_with(a, b, |a, b| match_values(a, b)).is_none(),
+        (&List(ref a), &List(ref b)) => diff_with(a, b, |a, b| match_values(a, b)).is_none(),
+        (&Set(ref a), &Set(ref b)) => a.len() == b.len() && a.iter().all(|v|
+            // TOOD
+            // Should use `match_values` for this, to allow for matching rules.
+            b.contains(v)
```
victorporof: This is incorrect, also because sets are unordered, will change.

edn/src/matcher.rs
```diff
@@ -0,0 +1,114 @@
...
+        assert_match_eq!("_" == "_", true);
+        assert_match_eq!("_" == "symbol", true);
+        assert_match_eq!("_" == "ns/symbol", true);
+        assert_match_eq!("_" == ":keyword", true);
+        assert_match_eq!("_" == ":ns/keyword", true);
+        assert_match_eq!("_" == "[nil, true, 1, \"foo\", bar, :baz]", true);
+        assert_match_eq!("_" == "(nil, true, 1, \"foo\", bar, :baz)", true);
+        assert_match_eq!("_" == "(nil, true, 1, \"foo\", bar, :baz)", true);
+        assert_match_eq!("_" == "#{nil, true, 1, \"foo\", bar, :baz}", true);
+        assert_match_eq!("_" == "{nil true, 1 \"foo\", bar :baz}", true);
```
victorporof: TODO: add much more tests.

edn/src/matcher.rs
```diff
@@ -0,0 +1,114 @@
...
+
+use std::collections::HashMap;
+use regex::Regex;
+use itertools::diff_with;
+
+use types::Value;
+
+///
+pub fn match_values<T>(pattern: T, input: T) -> bool where T: AsRef<Value> {
+    use Value::*;
```
victorporof: `AsRef` is not strictly necessary, but I've found it convenient.

edn/src/matcher.rs
```diff
@@ -0,0 +1,114 @@
...
+        (&Keyword(ref a), &Keyword(ref b)) => a == b,
+        (&NamespacedKeyword(ref a), &NamespacedKeyword(ref b)) => a == b,
+        (&Vector(ref a), &Vector(ref b)) => diff_with(a, b, |a, b| match_values(a, b)).is_none(),
+        (&List(ref a), &List(ref b)) => diff_with(a, b, |a, b| match_values(a, b)).is_none(),
+        (&Set(ref a), &Set(ref b)) => a.len() == b.len() && a.iter().all(|v|
+            // TOOD
+            // Should use `match_values` for this, to allow for matching rules.
+            b.contains(v)
+        ),
+        (&Map(ref a), &Map(ref b)) => a.len() == b.len() && a.iter().all(|(input_key, input_value)| {
```
ncalexan: The inner maps will compare correctly, unsorted or no.  And in the case of `BTree{Map,Set}`, it's even easier, since the iterator is ordered.

edn/src/matcher.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    use parse;
+
+    macro_rules! assert_match_eq {
+        ( $a:tt == $b:tt, $expected:expr ) => {
```
ncalexan: Interpolating the `==` like this doesn't agree with `assert_eq!`, which should be your guide.  It's misleading since you're _not_ testing equality (but matching, which is written `~=` in some languages, like Perl), and since you _can't_ do `!=` for the inversion.  Why not just `assert_matches!`, `assert_not_matches!`?

edn/src/types.rs
```diff
@@ -398,6 +398,12 @@ macro_rules! def_common_value_impl {
             def_common_value_methods!($t<$tchild>);
         }
 
+        impl AsRef<$t> for $t {
```
ncalexan: This really doesn't make much sense; it's equivalent to `&x`.  You say it's convenient, but can you explain why?

edn/src/matcher.rs
```diff
@@ -0,0 +1,114 @@
...
+
+///
+pub fn match_values<T>(pattern: T, input: T) -> bool where T: AsRef<Value> {
+    use Value::*;
+    use symbols::PlainSymbol as S;
+
+    let mut named_values: HashMap<&String, &Value> = HashMap::new();
+    let name_re = Regex::new(r"^\?[^?]+$").unwrap();
+
+    match (pattern.as_ref(), input.as_ref()) {
```
ncalexan: You're driving on the pattern and the input like equality, but I'd drive it from the pattern.  That is, just match the pattern, and handle the special cases (`_`, `?name`, `...`) specially.  Recurse for collection patterns and dispatch to `pattern == input` for atom patterns.  This'll be very short.

edn/src/matcher.rs
ncalexan: Do we have a type that is `AsRef<Value>` other than `Value`?  It can't be `ValueAndSpan` because the inner children are wrong...

edn/Cargo.toml
```diff
@@ -13,6 +13,8 @@ readme = "./README.md"
 [dependencies]
 num = "0.1.35"
 ordered-float = "0.4.0"
+regex = "0.2.1"
```
ncalexan: It's not critical, but checking for `.startswith("?")` should be enough; no need for regexes right away.

edn/src/matcher.rs
```diff
@@ -0,0 +1,114 @@
...
+use itertools::diff_with;
+
+use types::Value;
+
+///
+pub fn match_values<T>(pattern: T, input: T) -> bool where T: AsRef<Value> {
+    use Value::*;
+    use symbols::PlainSymbol as S;
+
+    let mut named_values: HashMap<&String, &Value> = HashMap::new();
```
victorporof: This obviously won't work, need to store the hash map outside of the recursive function. Placeholder for now.

edn/Cargo.toml
victorporof: Makes sense.

edn/src/types.rs
victorporof: I like not necessarily having to make the borrow explicit, and pattern matching should be able to work on both owned or borrowed values. So, just as pretty much any other use of AsRef, convenience, in this case just abstracting over the fact that borrowing is an implementation detail. Can remove if you don't like it.

edn/src/matcher.rs
victorporof: I agree, the way this is named is rather silly. The `==` is just sugar that looked good to my eyes, but something more explicit like `assert_matches` is probably better. However, do note that we also need something like `assert_not_matches` in that case, and I'd like to not have to duplicate macros. Will find a middle ground, I think a better name and `~=` is the way to go.

edn/src/matcher.rs
victorporof: Interesting, so we don't care about order anyway since the iterator yields values orderly. And because all of these collections are comparable directly, we can just use `==` for everything. That's nice.

victorporof:
> Do we have a type that is AsRef<Value> other than Value? It can't be ValueAndSpan because the inner children are wrong...

SpannedValue comes to mind.

edn/src/matcher.rs
victorporof: Indeed, I forgot that edn::Value itself implements `Eq` and the likes

victorporof:
@ncalexan I don't see how something like `Into<Option<PatternMatching>` can work when `PatternMatching` isn't a sized type. Both `Into` and `Option` require their types to be sized afaict, and I can't find any magical rust incantations to make this work.

See https://is.gd/cuA4SA

victorporof:
@ncalexan Care to take another look at this before I implement `...`?

edn/src/matcher.rs
```diff
@@ -0,0 +1,541 @@
...
+use types::Value;
+
+/// A trait defining pattern matching rules for any given pattern of type `T`.
+trait PatternMatchingRules<'a, T> {
+    /// Return true if the given pattern matches an arbitrary value.
+    fn matches_any(pattern: &T) -> bool;
+
+    /// Return the placeholder name if the given pattern matches a placeholder.
+    fn matches_placeholder(pattern: &'a T) -> Option<(&'a String)>;
+}
```
victorporof: Not implementing `...` yet, but should be relatively straightforward to add later.

victorporof: Since we're not consuming `pattern` here, I don't see why we'd want to return `Option<Symbol, Value>` or anything similar.

edn/src/matcher.rs
```diff
@@ -0,0 +1,541 @@
...
+    fn matches_any(pattern: &Value) -> bool {
+        match *pattern {
+            Value::PlainSymbol(symbols::PlainSymbol(ref s)) => s == "_",
+            _ => false
+        }
+    }
+
+    fn matches_placeholder(pattern: &'a Value) -> Option<(&'a String)> {
+        match *pattern {
+            Value::PlainSymbol(symbols::PlainSymbol(ref s)) => if s.starts_with('?') { Some(s) } else { None },
```
victorporof: Extracting the string directly here since it was more ergonomic to have a HashMap over strings instead of `symbols::PlainSymbol` (only because there's less to type).

edn/src/matcher.rs
```diff
@@ -0,0 +1,541 @@
...
+    }
+}
+
+/// Pattern matcher for EDN values utilizing specified pattern matching rules.
+/// For example, using this with `DefaultPatternMatchingRules`:
+/// * `[_]` matches an arbitrary one-element vector;
+/// * `[_ _]` matches an arbitrary two-element vector;
+/// * `[?x ?x]` matches `[1 1]` and `[#{} #{}]` but not `[1 2]` or `[[] #{}]`;
+struct Matcher<'a> {
+    placeholders: RefCell<HashMap<&'a String, &'a Value>>
```
victorporof: `RefCell` because interior immutability is not relevant for consumers of this type. There's no reason to expose the fact that some internal structure mutates while pattern matching happens, since the public API allows for this matcher to be used only once.

Alternatively, we could expose mutability and make sure `placeholders` is cleared before every match, but this seems to be irrelevant for the outside world.

edn/src/matcher.rs
```diff
@@ -0,0 +1,541 @@
...
+        use Value::*;
+
+        if T::matches_any(pattern) {
+            true
+        } else if let Some(symbol) = T::matches_placeholder(pattern) {
+            let mut cache = self.placeholders.borrow_mut();
+            if !cache.contains_key(symbol) {
+                cache.insert(symbol, value); true
+            } else {
+                value == *cache.get(symbol).unwrap()
```
victorporof: I couldn't find a way to minimize cache access here: it would seem that we need to have both `contains_key` and `get` calls, because a `get` on a `HashMap` borrows it as immutable, so we can't `insert` later if the `get` returned `None`.

edn/src/matcher.rs
```diff
@@ -0,0 +1,541 @@
...
+                (&List(ref a), &List(ref b)) =>
+                    diff_with(a, b, |a, b| self.match_internal::<T>(a, b)).is_none(),
+                (&Set(ref a), &Set(ref b)) =>
+                    a.len() == b.len() &&
+                    a.iter().all(|av| b.iter().any(|bv| self.match_internal::<T>(av, bv))) &&
+                    b.iter().all(|bv| a.iter().any(|av| self.match_internal::<T>(av, bv))),
+                (&Map(ref a), &Map(ref b)) =>
+                    a.len() == b.len() &&
+                    a.iter().all(|(ak, av)| b.iter().any(|(bk, bv)| self.match_internal::<T>(ak, bk) && self.match_internal::<T>(av, bv))) &&
+                    b.iter().all(|(bk, bv)| a.iter().any(|(ak, av)| self.match_internal::<T>(ak, bk) && self.match_internal::<T>(av, bv))),
```
victorporof: This ended up being just a tad bit more complicated than expected, since we can't just compare collections directly with equality since that would render the patterns useless. For example, `[_]` would never be equal to `[1]` unless we mess with the `Eq` trait implementation instead of deriving the default.

edn/src/matcher.rs
```diff
@@ -0,0 +1,541 @@
...
+        Matcher::match_with_rules::<DefaultPatternMatchingRules>(self, pattern)
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use parse;
+
+    macro_rules! assert_match {
+        ( $pattern:tt ~= $value:tt, $expected:expr ) => {
```
victorporof: I hope this form makes things more obvious. I'd like to avoid having both `assert_matches` and `assert_not_matches`, since that would seem overkill.

edn/src/matcher.rs
```diff
@@ -0,0 +1,541 @@
...
+        assert_match!("#{1 _}" ~= "#{3 4}", false);
+    }
+
+    #[test]
+    fn test_match_multiple_any_in_set_with_multiple_values() {
+        // These are false because _ is a symbol and sets guarantee
+        // uniqueness of children. So pattern matching will fail because
+        // the pattern is a set of length 2, while the matched edn is a set
+        // of length 3. If _ were unique, all of these assertions would
+        // be true. Need to better handle pattern rules.
```
victorporof: @ncalexan please pay close attention to these and let me know what you think.

edn/src/matcher.rs
```diff
@@ -0,0 +1,541 @@
...
+        assert_match!("{1 _}" ~= "{3 4}", false);
+    }
+
+    #[test]
+    fn test_match_multiple_any_in_map_with_multiple_values() {
+        // These are false because _ is a symbol and maps guarantee
+        // uniqueness of keys. So pattern matching will fail because
+        // the pattern is a map of length 2, while the matched edn is a map
+        // of length 3. If _ were unique, all of these assertions would
+        // be true. Need to better handle pattern rules.
```
victorporof: @ncalexan please pay close attention to these and let me know what you think.

edn/src/matcher.rs
```diff
@@ -0,0 +1,541 @@
...
+    fn new() -> Matcher<'a> {
+        Matcher {
+            placeholders: RefCell::default()
+        }
+    }
+
+    /// Performs pattern matching between two EDN `Value` instances (`value`
+    /// and `pattern`) utilizing a specified pattern matching ruleset `T`.
+    /// Returns true if matching succeeds.
+    pub fn match_with_rules<T>(value: &'a Value, pattern: &'a Value) -> bool
```
victorporof: Returning bool for now, will implement better matching results in a followup.

edn/src/matcher.rs
```diff
@@ -0,0 +1,576 @@
...
+                (&List(ref v), &List(ref p)) =>
+                    diff_with(v, p, |a, b| self.match_internal::<T>(a, b)).is_none(),
+                (&Set(ref v), &Set(ref p)) =>
+                    v.len() == p.len() &&
+                    v.iter().all(|a| p.iter().any(|b| self.match_internal::<T>(a, b))) &&
+                    p.iter().all(|b| v.iter().any(|a| self.match_internal::<T>(a, b))),
+                (&Map(ref v), &Map(ref p)) =>
+                    v.len() == p.len() &&
+                    v.iter().all(|a| p.iter().any(|b| self.match_internal::<T>(a.0, b.0) && self.match_internal::<T>(a.1, b.1))) &&
+                    p.iter().all(|b| v.iter().any(|a| self.match_internal::<T>(a.0, b.0) && self.match_internal::<T>(a.1, b.1))),
```
victorporof: This ended up being just a tad bit more complicated than expected, since we can't just compare collections directly with equality since that would render the patterns useless. For example, [_] would never be equal to [1] unless we mess with the Eq trait implementation instead of deriving the default.

edn/src/matcher.rs
ncalexan: I was wondering why my initial trait sketch was wrong, but of course -- you take (or compare to) the corresponding `Value` from the _input_.

edn/src/matcher.rs
ncalexan: I think this will work, but it's a subtle interaction between different parts of the code that could produce bugs.  I really think a mutable `Matcher`, consumed by `matches`, is the simplest; but do as you see fit.

edn/src/matcher.rs
```diff
@@ -0,0 +1,576 @@
...
+    /// Returns true if matching succeeds.
+    pub fn match_with_rules<T>(value: &'a Value, pattern: &'a Value) -> bool
+    where T: PatternMatchingRules<'a, Value> {
+        let matcher = Matcher::new();
+        matcher.match_internal::<T>(value, pattern)
+    }
+
+    /// Recursively traverses two EDN `Value` instances (`value` and `pattern`)
+    /// performing pattern matching. Note that the internal `placeholders` cache
+    /// may not be empty on invocation.
```
ncalexan: "may", as in "might not"?  Or "may", as in "must not"?

edn/src/matcher.rs
```diff
@@ -0,0 +1,576 @@
...
+    fn match_internal<T>(&self, value: &'a Value, pattern: &'a Value) -> bool
+    where T: PatternMatchingRules<'a, Value> {
+        use Value::*;
+
+        if T::matches_any(pattern) {
+            true
+        } else if let Some(symbol) = T::matches_placeholder(pattern) {
+            let mut cache = self.placeholders.borrow_mut();
```
ncalexan: There's a nice approach using `entry()` for handling these in one shot.  You can easily insert or get the existing value.  Alternately, you could _always_ `insert()`, and always check the returned `Option` -- although maybe it's no better, since you don't own the inserted value to test against any more.

edn/src/matcher.rs
ncalexan: Mmm, I was going to suggest that as an implementation strategy.  That is, you could make a newtype wrapper with a custom `Eq` and `Ord` implementation... it seems complicated.

edn/src/matcher.rs
ncalexan: First, sorry -- I got this backwards, Perl uses `=~`: see http://www.comp.leeds.ac.uk/Perl/matching.html.

If you're going to use `=~`, why not drop `expected` and also handle the Perl operator `!~`?

edn/src/matcher.rs
```diff
@@ -0,0 +1,576 @@
...
+            if !cache.contains_key(symbol) {
+                cache.insert(symbol, value); true
+            } else {
+                value == *cache.get(symbol).unwrap()
+            }
+        } else {
+            match (value, pattern) {
```
ncalexan: Is there a reason to not do this with `(*value, *pattern)` like you suggested in that other issue?

edn/src/matcher.rs
ncalexan: Wow, you're right, this is far more subtle than I'd thought.  As I was laying awake in the middle of the night, I realized this might need to test all orderings, or do back-tracking or something.  That is, how would you handle matching `#{?a ?b}` against `#{1 2}`?  That is, how do you know you capture the bindings in the "correct" order?

I'm inclined to look at the transactor tests and see if we could just outlaw patterns in unordered collections, at least for a while.  This seems very difficult to get right; I'll look at your tests now to see how you handle this.

edn/src/matcher.rs
```diff
@@ -0,0 +1,576 @@
...
+        assert_match!("[1 [2 [3 ?x]] 5 [?y 7]]" ~= "[1 [2 [3 4]] 5 [6 7]]", true);
+        assert_match!("[1 [2 [3 ?x]] 5 [?y 7]]" ~= "[1 [2 [3 4]] 5 [4 7]]", true);
+        assert_match!("[1 [2 [3 ?x]] 5 [?x 7]]" ~= "[1 [2 [3 4]] 5 [6 7]]", false);
+        assert_match!("[1 [2 [3 ?x]] 5 [?y 7]]" ~= "[1 [2 [3 4]] 5 [4 7]]", true);
+
+        assert_match!("[?x ?y ?x ?y]" ~= "[1 2 1 2]", true);
+        assert_match!("[?x ?y ?x ?y]" ~= "[1 2 2 1]", false);
+
+        assert_match!("[[?x ?y] [?x ?y]]" ~= "[[1 2] [1 2]]", true);
+        assert_match!("[[?x ?y] [?x ?y]]" ~= "[[1 2] [2 1]]", false);
```
ncalexan: I guess the tricky ones alluded to above would be `[#{?x ?y} ?x] ~= [#{1 2} 1]` _and_ `[#{?x ?y} ?y] ~= [#{1 2} 2]`.  It's probably not worth getting this right.

edn/src/matcher.rs
ncalexan: By "outlaw patterns in unordered collections", I think I mean that I don't want to allow patterns as the direct children of an unordered collection, so `#{ _ }` and `{_ 0}` are not legal, but `#{[?a _]}` and `{:x ?b}` are legal.  The idea would be to avoid these tricky situations around ordering and duplication, while still allowing to match on atoms and ordered collections inside unordered collections.

edn/src/matcher.rs
victorporof: If the API is consuming, that makes sense. I'll remove the `RefCell` in favor of this.

edn/src/matcher.rs
victorporof: I can't seem to figure out the magical rust incantations to make that work. Care to help me out?

edn/src/matcher.rs
victorporof: I think that actually works with the current implementation, but I'll add a test to make sure.

edn/src/matcher.rs
victorporof: "could" is a better word. It's a hint for the subtle interactions that could end up producing bugs if future changes to this code aren't aware of how it works. But I think favoring consumed mutable references to `Matcher` over using `RefCell`s is better, so this will change.

edn/src/matcher.rs
victorporof: I've tried always inserting and it doesn't make sense: you could end up overwriting old values with something different, which will make some pattern matching tests (especially around sets and maps) fail. Will try the `entry` approach though.

edn/src/matcher.rs
victorporof: Yes, currently placeholders cache values depending on the iteration order, which is ordered. So `#{?a, ?b}` matching `#{1 2}` will have `?a`=1 and `?b`=2, while `#{?b, ?a}` matching `#{1 2}` will end up with the same cached values, because of the iteration order. Similarly, `#{?a, ?b}` matching `#{2 1}` will have `?a`=2 and `?b`=1, while `#{?b, ?a}` matching `#{2 1}` will, again, end up with the same cached values as before. This is obviously not ideal, but at least if we understand this "rule" we can use it to our advantage when writing our patterns.

edn/src/matcher.rs
ncalexan: Let's just leave it, and try to understand what's going on later.  Sometimes I have the same experience: this should work, why doesn't it, screw it.

edn/src/matcher.rs
ncalexan: If you overwrite with something different, that should always fail, right?

ncalexan:
Just so that we capture this: in Slack, I suggested we generalize the placeholder’s to start with `_`, just like unused pattern bindings in Rust.  So you could match `[_1 _2]`.  I still think that handling unordered collections might be difficult.

edn/src/matcher.rs
victorporof: Not necessarily in ways you'd expect. For maps and sets we do double traversal to check that the two collections are bijective.

edn/src/matcher.rs
victorporof: These indeed pass. I've tested them explicitly.

PR #339, opened 2017-02-22
Moar SQL

rnewman:
The bulk of this PR is about accumulating values into SQL.

When faced with a Datalog query like:

```clojure
[:find … [?x _ "foo bar"]]
```

We have a choice: try to expand those literals into the query, escaping appropriately, reserving the parameter list for user-supplied parameters… or we can not solve the escaping problem and instead use bindings.

I took the latter approach for all non-numeric values. But that means we can end up with a SQL query that interleaves user parameter inputs with query literals:

```clojure
[:find ?x :in ?foo [?x :foo/baz "something"] [?x :foo/bar ?foo]]
```

turns into something like

```sql
SELECT datoms00.e AS x
FROM datoms AS datoms00, datoms AS datoms01
WHERE datoms00.a = 65591
  AND datoms00.v = ?                -- This needs to be "something"
  AND datoms01.a = 65592
  AND datoms01.v = ?                 -- This needs to be ?foo
```

Using positional arguments for `?foo` would be difficult: we might have positions 0, 1, and 3 be pre-bound and positions 2 and 4 being inputs.

The obvious solution is to use named parameters. That's the approach taken here: we use named parameters for these 'synthetic' value bindings, and also named parameters for input variables.

This handily also avoids the need to ever duplicate an argument: SQLite will take care of looking them up multiple times.

On top of this work, I've implemented rudimentary algebrizing and rudimentary translation into SQL, culminating in a simple end-to-end query example.

### DIFF
--- a/query-translator/Cargo.toml
+++ b/query-translator/Cargo.toml
@@ -4,6 +4,9 @@ version = "0.0.1"
 workspace = ".."
 
 [dependencies]
+[dependencies.mentat_core]
+path = "../core"
+
 [dependencies.mentat_sql]
 path = "../sql"
 
@@ -12,3 +15,7 @@ path = "../query"
 
 [dependencies.mentat_query_algebrizer]
 path = "../query-algebrizer"
+
+# Only for tests.
+[dev-dependencies.mentat_query_parser]
+path = "../query-parser"
--- a/query-translator/src/translate.rs
+++ b/query-translator/src/translate.rs
@@ -1,0 +1,76 @@
+// Copyright 2016 Mozilla
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+#![allow(dead_code, unused_imports)]
+
+use mentat_query_algebrizer::{
+    AlgebraicQuery,
+    ColumnConstraint,
+    ConjoiningClauses,
+    DatomsColumn,
+    DatomsTable,
+    QualifiedAlias,
+    SourceAlias,
+};
+
+use types::{
+    ColumnOrExpression,
+    Constraint,
+    FromClause,
+    Projection,
+    SelectQuery,
+    TableList,
+};
+
+trait ToConstraint {
+    fn to_constraint(self) -> Constraint;
+}
+
+trait ToColumn {
+    fn to_column(self) -> ColumnOrExpression;
+}
+
+impl ToColumn for QualifiedAlias {
+    fn to_column(self) -> ColumnOrExpression {
+        ColumnOrExpression::Column(self)
+    }
+}
+
+impl ToConstraint for ColumnConstraint {
+    fn to_constraint(self) -> Constraint {
+        use self::ColumnConstraint::*;
+        match self {
+            EqualsEntity(qa, entid) =>
+                Constraint::equal(qa.to_column(), ColumnOrExpression::Integer(entid)),
+            EqualsValue(qa, tv) =>
+                Constraint::equal(qa.to_column(), ColumnOrExpression::Value(tv)),
+            EqualsColumn(left, right) =>
+                Constraint::equal(left.to_column(), right.to_column()),
+        }
+    }
+}
+
+
+/// Consume a provided `ConjoiningClauses` to yield a new
+/// `SelectQuery`. A projection list must also be provided.
+pub fn cc_to_select(projection: Projection, cc: ConjoiningClauses) -> SelectQuery {
+    SelectQuery {
+        projection: projection,
+        from: FromClause::TableList(TableList(cc.from)),
+        constraints: cc.wheres
+                       .into_iter()
+                       .map(|c| c.to_constraint())
+                       .collect(),
+    }
+}
+
+pub fn cc_to_exists(cc: ConjoiningClauses) -> SelectQuery {
+    cc_to_select(Projection::One, cc)
+}
--- a/query-translator/src/types.rs
+++ b/query-translator/src/types.rs
@@ -10,6 +10,8 @@
 
 #![allow(dead_code, unused_imports)]
 
+use mentat_core::TypedValue;
+
 use mentat_query_algebrizer::{
     AlgebraicQuery,
     ConjoiningClauses,
@@ -25,26 +27,46 @@ use mentat_sql::{
     QueryBuilder,
     QueryFragment,
     SQLiteQueryBuilder,
+    SQLQuery,
 };
 
 //---------------------------------------------------------
 // A Mentat-focused representation of a SQL query.
 
-enum ColumnOrExpression {
+pub enum ColumnOrExpression {
     Column(QualifiedAlias),
-    Integer(i64),             // Because it's so common.
+    Integer(i64),            // Because it's so common.
+    Value(TypedValue),       // TODO: replace with a SQLite value.
 }
 
-type Name = String;
-struct Projection (ColumnOrExpression, Name);
+pub type Name = String;
 
-#[derive(Clone)]
-struct Op(String);      // TODO
-enum Constraint {
+pub struct ProjectedColumn(pub ColumnOrExpression, pub Name);
+
+pub enum Projection {
+    Columns(Vec<ProjectedColumn>),
+    Star,
+    One,
+}
+
+#[derive(Copy, Clone)]
+pub struct Op(&'static str);      // TODO
+
+pub enum Constraint {
     Infix {
         op: Op,
         left: ColumnOrExpression,
         right: ColumnOrExpression
+    }
+}
+
+impl Constraint {
+    pub fn equal(left: ColumnOrExpression, right: ColumnOrExpression) -> Constraint {
+        Constraint::Infix {
+            op: Op("="),
+            left: left,
+            right: right,
+        }
     }
 }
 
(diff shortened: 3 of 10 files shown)


sql/src/lib.rs
```diff
@@ -8,18 +8,40 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
+extern crate ordered_float;
```
ncalexan: It seems like most Rust libraries do all the `extern crate` declarations in `lib.rs`.  This avoids some shenanigans with nested directories and rooted `use` statements.  I noticed you haven't done this in some parts of the `query-*` hierarchy.  I see no guidance at https://aturon.github.io/style/imports.html.

Do you have a strong opinion?  (This particular case does not conflict, just trying to straighten this out.)

sql/src/lib.rs
```diff
@@ -8,18 +8,40 @@
...
 use std::error::Error;
 
+use ordered_float::OrderedFloat;
+
+use mentat_core::TypedValue;
+
 pub type BuildQueryError = Box<Error + Send + Sync>;
 pub type BuildQueryResult = Result<(), BuildQueryError>;
 
+pub enum BindParamError {
```
ncalexan: It would be nice to build this on top of #341, which provides a nice framework for these error types.

sql/src/lib.rs
```diff
@@ -8,18 +8,40 @@
...
+    InvalidParameterName(String),
+    BindParamCouldBeGenerated(String),
+}
+
+/// We want to accumulate values that will later be substituted into a SQL statement execution.
+/// This struct encapsulates the generated string and the _initial_ argument list.
```
ncalexan: Elaborate on the emphasis: "The _final_ argument list provided to the SQL engine might be smaller than the initial list, because duplicate arguments may be coalesced; and the final argument list may be differently ordered.

sql/src/lib.rs
```diff
@@ -8,18 +8,40 @@
...
+pub struct SQLQuery {
+    pub sql: String,
+
+    /// These will eventually perhaps be rusqlite `ToSql` instances.
+    pub args: Vec<(String, String)>,
```
ncalexan: Is this a map from name to something?  Should this be a set?  Justify this surprising choice.

And storing `ToSql` instances is trickier than I expected in other situations; you might need a lifetime.

sql/src/lib.rs
```diff
@@ -47,14 +69,36 @@ impl QueryFragment for () {
...
+    pub fn with_prefix(prefix: String) -> Self {
         SQLiteQueryBuilder {
             sql: String::new(),
+            arg_prefix: prefix,
+            arg_counter: 0,
+            args: vec![],
         }
     }
+
+    fn push_static_arg(&mut self, val: String) {
```
ncalexan: Hmm.  So `val` is always a SQL value represented as a `String`?  OK.

ncalexan: Ah yes, it's more clear in `push_typed_value`.  So much cloning!

sql/src/lib.rs
```diff
@@ -69,11 +113,70 @@ impl QueryBuilder for SQLiteQueryBuilder {
         Ok(())
     }
 
-    fn push_bind_param(&mut self) {
-        self.push_sql("?");
+    fn push_typed_value(&mut self, value: &TypedValue) -> BuildQueryResult {
+        use TypedValue::*;
+        match value {
```
ncalexan: I think clippy suggest `match *value`, but I don't quite understand why.  @victorporof and I are trying to standardize on the former, assuming that clippy "knows something"; will you join us?

query-translator/Cargo.toml
```diff
@@ -4,6 +4,9 @@ version = "0.0.1"
 workspace = ".."
 
 [dependencies]
+[dependencies.mentat_core]
```
ncalexan: README.md for this new create, please!

query-translator/src/translate.rs
```diff
@@ -0,0 +1,76 @@
...
+
+impl ToColumn for QualifiedAlias {
+    fn to_column(self) -> ColumnOrExpression {
+        ColumnOrExpression::Column(self)
+    }
+}
+
+impl ToConstraint for ColumnConstraint {
+    fn to_constraint(self) -> Constraint {
+        use self::ColumnConstraint::*;
```
ncalexan: It's not clear to me why you have `self::ColumnConstraint::*` rather than `ColumnConstraint::*` here.  Is this an interaction with test code?

query-translator/src/types.rs
```diff
@@ -88,28 +115,47 @@ impl QueryFragment for ColumnOrExpression {
...
     fn push_sql(&self, out: &mut QueryBuilder) -> BuildQueryResult {
-        self.0.push_sql(out)?;
-        out.push_sql(" AS ");
-        out.push_identifier(self.1.as_str())
+        use self::Projection::*;
+        match self {
+            &One => out.push_sql("1"),
+            &Star => out.push_sql("*"),
+            &Columns(ref cols) => {
+                let &ProjectedColumn(ref col, ref alias) = &cols[0];
```
ncalexan: There's no difference between the two blocks.  You're trying to ensure failure if `cols` is empty?  There are better ways of expressing this.

query-translator/src/types.rs
```diff
@@ -25,37 +27,57 @@ use mentat_sql::{
...
 
 //---------------------------------------------------------
 // A Mentat-focused representation of a SQL query.
 
-enum ColumnOrExpression {
+pub enum ColumnOrExpression {
     Column(QualifiedAlias),
-    Integer(i64),             // Because it's so common.
+    Integer(i64),            // Because it's so common.
+    Value(TypedValue),       // TODO: replace with a SQLite value.
```
ncalexan: A note that the intention is to have this independent of the Mentat data types (if this is true, which I think it is) would be nice.

sql/src/lib.rs
rnewman: This happens to be `lib.rs`: the simple case is easy! For many of the query libraries you're right, I didn't do so.

I tend to disagree with 'importing' everything at the top level: it makes it less clear which files depend on which crates. But this is one of Rust's warts: per-module crate dependencies are ugly, and `Cargo.toml` makes the whole thing redundant anyway…

sql/src/lib.rs
rnewman: I'll see if I can do that after I review it! I figure that the error-propagation story is far from complete, particularly for querying, and is very likely to change when we focus on developer ergonomics and an external API.

sql/src/lib.rs
rnewman: Not quite true: the final argument list will only add to this one (adding any `:in` bindings), and the order of arguments doesn't matter.

sql/src/lib.rs
rnewman: `rusqlite` uses — as does SQLite — a simple array of pairs to represent named (as opposed to positional) arguments.

I expect that if this turns into `ToSql` it'll actually be boxed `AsRef<ToSql>` or `Borrow<ToSql>` instances, or something similarly verbose.

For now, it's only keywords and strings that need this handling…

sql/src/lib.rs
rnewman: Opinions in `#rust` were divided. ("A good question for StackOverflow")

It's possible ("check the asm") that the `match *v` version could copy, or could move, depending on the match arms. The `match v` version should never be bad. If the ASM disagrees, then it becomes personal preference.

Opinion was also somewhat against Clippy: as with all linters, it doesn't always give the best advice. Quoth:

> Manishearth: clippy isn't intended to be blindly followed

So, given that I'm too lazy to disassemble Mentat to find out if using `match *foo` produces worse assembly, I'll stick with `match foo`.

query-translator/Cargo.toml
rnewman: There is one!

rnewman: It was added in:
```
commit 9ecd02ef95154b5e3610a965090487da93ce283b
Author: Richard Newman <rnewman@twinql.com>
Date:   Thu Feb 16 15:39:19 2017 -0800

    Begin serializing queries to SQL. r=nalexander
```

query-translator/src/translate.rs
rnewman:
```
error[E0432]: unresolved import `ColumnConstraint::*`
  --> query-translator/src/translate.rs:48:13
   |
48 |         use ColumnConstraint::*;
   |             ^^^^^^^^^^^^^^^^^^^^ Did you mean `self::ColumnConstraint`?
```

query-translator/src/types.rs
rnewman: I vacillated on this. As you saw earlier, I simply allowed the SQL layer to accept a `TypedValue`, which avoids _this_ layer eventually having to worry about `rusqlite::ToSql`. I should probably remove this comment. The other approach is to have this be `ToSql` (or similar), and translate `TypedValue` at the point of constructing a `SelectQuery`. Do you have a preference, @ncalexan?

query-translator/src/types.rs
rnewman: Different bug: there's supposed to be a leading accumulation of a comma within the loop. This is the usual "do the first, then loop the second onward with a separator" pattern.

rnewman: Fixed.

sql/src/lib.rs
ncalexan: Aye, or a box.

query-translator/src/types.rs
ncalexan: I don't have a strong preference.  On the transactor side of the house, I am brewing a db (concrete storage) vs. tx (Mentat transaction processing) split, and I see some value in separating the two concerns.  It's not clear to me where the "SQL data from here out" is, exactly, in the query flow, so I'm happy to leave this as TODO.

query-translator/src/types.rs
ncalexan: OK.

ncalexan: There's [`intersperse`](https://docs.rs/itertools/0.5.9/src/itertools/intersperse.rs.html#13-19) for these situations...

rnewman:
5e3cdd1

Issue #328, opened 2017-02-17
Expose errors in query parser


rnewman:
We currently completely collapse errors from `combine` in `mentat_query_parser`:

```rust
#[derive(Clone, Debug, Eq, PartialEq)]
pub enum WhereParseError {
    Err,
}

…

pub fn clause_seq_to_patterns(clauses: &[edn::Value]) -> WhereParseResult {
    Where::clauses()
        .parse(clauses)
        .map(|x| x.0)
        .map_err(|_| WhereParseError::Err)
}
```

After https://github.com/Marwes/combine/pull/87, we'll be better able to turn these errors into something that's meaningful to propagate out of the query interface.

Issue #296, opened 2017-02-14
[core] Implement Conn managing DB metadata, transacting, and SQLite threading


ncalexan:
This issue tracks implementing the start of the `Conn` abstraction.  A Mentat connection is the conceptual equivalent of a SQLite connection.  One crucial difference is that Mentat enforces a single writer and handles some isolation.  This is similar to [Datomic's Java `Connection` API](http://docs.datomic.com/javadoc/datomic/Connection.html).

@rnewman (and I) have been thinking at some length about how to represent the structure in Rust; he's written about this at https://github.com/mozilla/mentat/wiki/Thoughts:-modeling-db-conn-in-Rust and more he shared some notes at https://public.etherpad-mozilla.org/p/db-conn-thoughts.

Getting some pieces of this correct up front will help guide a lot of the abstractions we're building -- see comments like https://github.com/mozilla/mentat/pull/283#pullrequestreview-21589524 and friends.

ncalexan:
This is not a good first or second bug.

ncalexan:
I think @rnewman's review in #283 will be easier if I sketch some of this out in code.

Issue #261, opened 2017-02-08
Improve pattern for finite-length iterators in flat_map expansions


ncalexan:
There are a bunch of places where we want to turn a collection of tuples (like `Vec<(A, B, C)>`) into a flat iterator, like `A1, B1, C1, A2, B2, C2, ...`.  The pattern we're using right now is to combine `flat_map` and `std::iter::once`, like:
```rust
collection.flat_map(|(a, b)| once(a).chain(once(b)))
```

That avoids heap allocation in favour of stack allocation, but gets pretty gnarly when the tuple length increases.  For example:
- https://github.com/mozilla/mentat/pull/214/files/64ff2efe012bb4297056c5899076af0c7f4ca49d#diff-54794907269c3b2d0bebbfd6a8aeab5dR617
- https://github.com/mozilla/mentat/pull/214/files/64ff2efe012bb4297056c5899076af0c7f4ca49d#diff-54794907269c3b2d0bebbfd6a8aeab5dR476

This ticket tracks doing something better.  Perhaps a few helper functions (like `once_2` and `once_3`?), or `impl (A, B) { fn once(...) { } }`, or a macro to at least make the callsite easier to read.

This is probably not a good first bug unless you're strong in Rust, because some taste is required to determine what's _better_ than the pattern we have now.

victorporof:
Assigning to find out whether or not I have taste.

victorporof:
FWIW, the links in the first comment aren't working anymore.

PR #341, opened 2017-02-22
Db and errors

ncalexan:
This commit sequence is a bit all over the place.  It's all preparatory work for a top-level `Conn`, as tracked by #296.  That's still in progress in my tree.

The first series of  commits, up to and including https://github.com/ncalexan/mentat/commit/201ffc712cf2b1189903a12d16d3190a13293476, are about separating the SQLite bits from the `DB` type.  See the comments around the `MentatStoring` trait.

The last few commits, including and following https://github.com/ncalexan/mentat/commit/b8a7c84ea47b5e0d9b0cc27b85faab55c4ff1c00, are about moving the existing code to `error-chain`.  You can see the chaining (a little), and unifying some error types.  This also surfaces errors out of `combine`; see some of the commit messages for notes.

### DIFF
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -873,11 +873,11 @@ mod tests {
         let db = read_db(&conn).unwrap();
 
         // Does not include :db/txInstant.
-        let datoms = debug::datoms_after(&conn, &db, 0).unwrap();
+        let datoms = debug::datoms_after(&conn, &db.schema, 0).unwrap();
         assert_eq!(datoms.0.len(), 88);
 
         // Includes :db/txInstant.
-        let transactions = debug::transactions_after(&conn, &db, 0).unwrap();
+        let transactions = debug::transactions_after(&conn, &db.schema, 0).unwrap();
         assert_eq!(transactions.0.len(), 1);
         assert_eq!(transactions.0[0].0.len(), 89);
     }
@@ -920,7 +920,7 @@ mod tests {
                 let report = maybe_report.unwrap();
                 assert_eq!(report.tx_id, bootstrap::TX0 + index + 1);
 
-                let transactions = debug::transactions_after(&conn, &db, bootstrap::TX0 + index).unwrap();
+                let transactions = debug::transactions_after(&conn, &db.schema, bootstrap::TX0 + index).unwrap();
                 assert_eq!(transactions.0.len(), 1);
                 assert_eq!(*expected_transaction,
                            transactions.0[0].into_edn(),
@@ -928,7 +928,7 @@ mod tests {
             }
 
             if let Some(expected_datoms) = expected_datoms {
-                let datoms = debug::datoms_after(&conn, &db, bootstrap::TX0).unwrap();
+                let datoms = debug::datoms_after(&conn, &db.schema, bootstrap::TX0).unwrap();
                 assert_eq!(*expected_datoms,
                            datoms.into_edn(),
                            "\n{} - expected datoms:\n{}\nbut got datoms:\n{}", label, *expected_datoms, datoms.into_edn())
@@ -947,11 +947,11 @@ mod tests {
         let mut db = ensure_current_version(&mut conn).unwrap();
 
         // Does not include :db/txInstant.
-        let datoms = debug::datoms_after(&conn, &db, 0).unwrap();
+        let datoms = debug::datoms_after(&conn, &db.schema, 0).unwrap();
         assert_eq!(datoms.0.len(), 88);
 
         // Includes :db/txInstant.
-        let transactions = debug::transactions_after(&conn, &db, 0).unwrap();
+        let transactions = debug::transactions_after(&conn, &db.schema, 0).unwrap();
         assert_eq!(transactions.0.len(), 1);
         assert_eq!(transactions.0[0].0.len(), 89);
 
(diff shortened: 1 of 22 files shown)


ncalexan:
@rnewman lots of code movement in the DB layer, and then some real work to use `error-chain` throughout.  Hope it makes some sense.

db/src/debug.rs
```diff
@@ -172,8 +173,8 @@ pub fn transactions_after(conn: &rusqlite::Connection, schema: &Schema, tx: i64)
         let added: bool = row.get_checked(5)?;
 
         Ok(Datom {
-            e: to_entid(schema, e),
-            a: to_entid(schema, a),
+            e: to_entid(schema.borrow(), e),
+            a: to_entid(schema.borrow(), a),
```
rnewman: Does it make a difference if we `borrow()` only once?

db/src/debug.rs
```diff
@@ -140,8 +141,8 @@ pub fn datoms_after(conn: &rusqlite::Connection, db: &DB, tx: i64) -> Result<Dat
         let tx: i64 = row.get_checked(4)?;
 
         Ok(Some(Datom {
-            e: to_entid(db, e),
-            a: to_entid(db, a),
+            e: to_entid(schema.borrow(), e),
+            a: to_entid(schema.borrow(), a),
```
rnewman: Does it make a difference if we only borrow once here?

db/src/db.rs
```diff
@@ -457,17 +462,149 @@ pub enum SearchType {
...
-    /// are exactly those (a, v) pairs that have an assertion [e a v] in the datom store.
-    pub fn resolve_avs<'a>(&self, conn: &rusqlite::Connection, avs: &'a [&'a AVPair]) -> Result<AVMap<'a>> {
+    /// are exactly those (a, v) pairs that have an assertion [e a v] in the store.
+    fn resolve_avs<'a>(&self, avs: &'a [&'a AVPair]) -> Result<AVMap<'a>>;
+
+    /// Begin (or prepare) the underlying storage layer for a new Mentat transaction.
+    ///
+    /// Use this to create temporary tables, prepare indices, set pragmas, etc, before the initial
+    /// `insert_non_fts_searches` invocation.
+    fn begin_transaction(&self) -> Result<()>;
```
rnewman: This is a difficult thing to abstract, but yeah, I guess!

parser-utils/src/lib.rs
```diff
@@ -105,3 +105,85 @@ macro_rules! def_value_satisfy_parser_fn {
...
+    fn description(&self) -> &str {
+        "parse error parsing EDN values"
+    }
+}
+
+impl<'a> From<combine::primitives::ParseError<&'a [edn::Value]>> for ValueParseError {
+    fn from(e: combine::primitives::ParseError<&'a [edn::Value]>) -> ValueParseError {
+        ValueParseError {
+            position: e.position,
+            errors: e.errors.into_iter().map(|e| e.map_range(|r| {
```
rnewman: I confess that I'm a little surprised to have `errors` be something we can move, but I guess that helps!

I was worried that we'd have copying on the error path, where the error might be due to OOM…

query-parser/src/parse.rs
```diff
@@ -222,15 +185,17 @@ pub fn find_seq_to_find_spec(find: &[edn::Value]) -> FindParseResult {
...
+        .map_err(|e| Error::from_kind(ErrorKind::FindParseError(e)))
 }
 
 #[allow(dead_code)]
 pub fn clause_seq_to_patterns(clauses: &[edn::Value]) -> WhereParseResult {
     Where::clauses()
         .parse(clauses)
         .map(|x| x.0)
-        .map_err(|_| WhereParseError::Err)
+        .map_err::<ValueParseError, _>(|e| e.translate_position(clauses).into())
```
rnewman: This fixes #328.

rnewman:
This is fine. I kinda dislike `error_chain`; it's obtuse, but so it goes.

rnewman:
Err… now I can't mark this as approved. Well done, GitHub.

db/src/db.rs
```diff
@@ -659,17 +653,16 @@ impl DB {
                     .chain(once(a as &ToSql)
                            .chain(once(value as &ToSql)
                                   .chain(once(value_type_tag as &ToSql)
-                                         .chain(once(&tx as &ToSql)
-                                                .chain(once(to_bool_ref(added) as &ToSql)
-                                                       .chain(once(flags as &ToSql)))))))
+                                         .chain(once(to_bool_ref(added) as &ToSql)
```
ncalexan: I concur: #261.

db/src/debug.rs
ncalexan: Nope, so I've lifted a temporary.

db/src/db.rs
ncalexan: Yeah, I'm basically just punting and working in a live scratchpad until the abstractions get more clear.

parser-utils/src/lib.rs
ncalexan: I can only imagine this does badly in OOM situations, but we're not going to get `combine::ParseError` instances in those situations anyway.  I guess it's possible we get `ParseError` _into_ OOM... but that should blow up during this mapping.

query-parser/src/parse.rs
ncalexan: Will close it when I push.

ncalexan:
Gah, force pushing a replacement since I made two typoes in the commit message.

PR #342, opened 2017-02-22
Add top-level `Conn`. Fixes #296.

### LINKED ISSUE #296: [core] Implement Conn managing DB metadata, transacting, and SQLite threading

ncalexan:
See commit comment for some notes.

### DIFF
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -208,7 +208,7 @@ pub fn create_current_version(conn: &mut rusqlite::Connection) -> Result<DB> {
 
     // TODO: return to transact_internal to self-manage the encompassing SQLite transaction.
     let bootstrap_schema = bootstrap::bootstrap_schema();
-    let (_report, next_partition_map, next_schema) = transact(&tx, &bootstrap_partition_map, &bootstrap_schema, bootstrap::bootstrap_entities())?;
+    let (_report, next_partition_map, next_schema) = transact(&tx, bootstrap_partition_map, &bootstrap_schema, bootstrap::bootstrap_entities())?;
     if next_schema.is_some() {
         // TODO Use custom ErrorKind https://github.com/brson/error-chain/issues/117
         bail!(ErrorKind::NotYetImplemented(format!("Initial bootstrap transaction did not produce expected bootstrap schema")));
@@ -902,7 +902,7 @@ mod tests {
 
             let entities: Vec<_> = mentat_tx_parser::Tx::parse(&[assertions][..]).unwrap();
 
-            let maybe_report = transact(&conn, partition_map, schema, entities);
+            let maybe_report = transact(&conn, partition_map.clone(), schema, entities);
 
             if let Some(expected_transaction) = expected_transaction {
                 if !expected_transaction.is_nil() {
--- a/db/src/tx.rs
+++ b/db/src/tx.rs
@@ -46,6 +46,7 @@
 //! keep everything straight.
 
 use std;
+use std::borrow::Cow;
 use std::collections::BTreeSet;
 
 use ::{to_namespaced_keyword};
@@ -275,7 +276,7 @@ impl<'conn, 'a> Tx<'conn, 'a> {
         let temp_id_allocations: TempIdMap = unresolved_temp_ids.into_iter().zip(entids).collect();
 
         let final_populations = generation.into_final_populations(&temp_id_allocations)?;
-        {
+
         /// Assertions that are :db.cardinality/one and not :db.fulltext.
         let mut non_fts_one: Vec<db::ReducedEntity> = vec![];
 
@@ -325,10 +326,6 @@ impl<'conn, 'a> Tx<'conn, 'a> {
         }
 
         self.store.commit_transaction(self.tx_id)?;
-        }
-
-        // let mut next_schema = self.schema.to_mut();
-        // next_schema.ident_map.insert(NamespacedKeyword::new("db", "new"), 1000);
 
         // TODO: update idents and schema materialized views.
         db::update_partition_map(self.store, &self.partition_map)?;
@@ -340,25 +337,20 @@ impl<'conn, 'a> Tx<'conn, 'a> {
     }
 }
 
-use std::borrow::Cow;
-
-/// Transact the given `entities` against the given SQLite `conn`, using the metadata in
-/// `self.DB`.
+/// Transact the given `entities` against the given SQLite `conn`, using the given metadata.
 ///
 /// This approach is explained in https://github.com/mozilla/mentat/wiki/Transacting.
 // TODO: move this to the transactor layer.
-pub fn transact<'conn, 'a, I>(conn: &'conn rusqlite::Connection, partition_map: &'a PartitionMap, schema: &'a Schema, entities: I) -> Result<(TxReport, PartitionMap, Option<Schema>)> where I: IntoIterator<Item=Entity> {
+pub fn transact<'conn, 'a, I>(conn: &'conn rusqlite::Connection, mut partition_map: PartitionMap, schema: &'a Schema, entities: I) -> Result<(TxReport, PartitionMap, Option<Schema>)> where I: IntoIterator<Item=Entity> {
     // Eventually, this function will be responsible for managing a SQLite transaction.  For
     // now, it's just about the tx details.
 
     let tx_instant = ::now(); // Label the transaction with the timestamp when we first see it: leading edge.
-
-    let mut next_partition_map: PartitionMap = partition_map.clone();
-    let tx_id = next_partition_map.allocate_entid(":db.part/tx");
+    let tx_id = partition_map.allocate_entid(":db.part/tx");
 
     conn.begin_transaction()?;
 
-    let mut tx = Tx::new(conn, next_partition_map, schema, tx_id, tx_instant);
+    let mut tx = Tx::new(conn, partition_map, schema, tx_id, tx_instant);
 
     let report = tx.transact_entities(entities)?;
 
--- a/src/conn.rs
+++ b/src/conn.rs
@@ -1,0 +1,215 @@
+// Copyright 2016 Mozilla
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+#![allow(dead_code)]
+
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex};
+
+use rusqlite;
+
+use edn;
+use errors::*;
+use mentat_core::{
+    Schema,
+    TypedValue,
+};
+use mentat_db::db;
+use mentat_db::{
+    transact,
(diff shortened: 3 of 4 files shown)


ncalexan:
@joe, I ended up writing less than I expected, mostly 'cuz I want to move on to something else before revisiting this.  See if this new module makes sense to you none-the-less.

@rnewman, a little different than what we discussed.  See how you feel.

My preference is to leave significant changes until we have a concrete consumer to guide our API --  such as a CLI or HTTP interface.

ncalexan:
This builds on #341.

db/src/db.rs
```diff
@@ -758,41 +758,66 @@ impl DB {
...
+}
 
-        // TODO: only cache the latest of these statements.  Changing the set of partitions isn't
-        // supported in the Clojure implementation at all, and might not be supported in Mentat soon,
-        // so this is very low priority.
-        let mut stmt = conn.prepare_cached(s.as_str())?;
-        stmt.execute(&params[..])
-            .map(|_c| ())
-            .chain_err(|| "Could not update partition map")
+/// Update the current partition map materialized view.
```
joewalker: The diff has totally broken down at this point.
It looks like there's some moving around of functions -> Just transact() into PartitionMap?

db/src/db.rs
```diff
@@ -457,17 +462,149 @@ pub enum SearchType {
...
+    fn resolve_avs<'a>(&self, avs: &'a [&'a AVPair]) -> Result<AVMap<'a>>;
+
+    /// Begin (or prepare) the underlying storage layer for a new Mentat transaction.
+    ///
+    /// Use this to create temporary tables, prepare indices, set pragmas, etc, before the initial
+    /// `insert_non_fts_searches` invocation.
+    fn begin_transaction(&self) -> Result<()>;
+
+    // TODO: this is not a reasonable abstraction, but I don't want to really consider non-SQL storage just yet.
+    fn insert_non_fts_searches<'a>(&self, entities: &'a [ReducedEntity], search_type: SearchType) -> Result<()>;
```
joewalker: This function name bothers me in the same way that it bothers me when people talk about PIN numbers. Which is to say it's not actually annoying, but is distracting, because it's not what I want to be thinking about.

You altered a function with a similar name earlier, but then you weren't adding the function so I let you off ;-)

rnewman: Bear in mind that these are two different kinds of "search" — "full-text search" is one, and the search against existing datoms that we perform during insertion is another. This would be like "with seventeen debit cards in my wallet, the PIN numbers are too damn high!"

db/src/db.rs
```diff
@@ -209,13 +207,20 @@ pub fn create_current_version(conn: &mut rusqlite::Connection) -> Result<DB> {
     }
 
     // TODO: return to transact_internal to self-manage the encompassing SQLite transaction.
-    let mut bootstrap_db = DB::new(bootstrap_partition_map, bootstrap::bootstrap_schema());
-    bootstrap_db.transact(&tx, bootstrap::bootstrap_entities())?;
+    let bootstrap_schema = bootstrap::bootstrap_schema();
+    let (_report, next_partition_map, next_schema) = transact(&tx, &bootstrap_partition_map, &bootstrap_schema, bootstrap::bootstrap_entities())?;
+    if next_schema.is_some() {
+        // TODO Use custom ErrorKind https://github.com/brson/error-chain/issues/117
+        bail!(ErrorKind::NotYetImplemented(format!("Initial bootstrap transaction did not produce expected bootstrap schema")));
```
joewalker: We use `"...".to_string()` in most other code? Feels like the latter is less work?

db/src/db.rs
```diff
@@ -457,17 +462,149 @@ pub enum SearchType {
...
 
-impl DB {
-
+/// `MentatStoring` will be the trait that encapsulates the storage layer.  It is consumed by the
+/// transaction processing layer.
+///
+/// Right now, the only implementation of `MentatStoring` is the SQLite-specific SQL schema.  In the
+/// future, we might consider other SQL engines (perhaps with different fulltext indexing), or
+/// entirely different data stores, say ones shaped like key-value stores.
+pub trait MentatStoring {
```
joewalker: `MentatStoring` feels like a funny name to me. Why not `MentatStorage`?

ncalexan: @rnewman set the trend with `SchemaBuilding` rather than `SchemaBuilder`.  I think I gently suggested nouns, but I've continued with the verbs for now.  Maybe we should discuss this in a meeting, since I see no guidance in the [Rust Guidelines](https://aturon.github.io/features/traits/README.html).

db/src/db.rs
ncalexan: I probably just copy-paste and then removed arguments.  Happy to `to_string()`.

db/src/db.rs
ncalexan: Yeah, there's a lot of moving bits around -- now landed -- that isn't relevant to this little piece.

ncalexan:
@rnewman still waiting on your review.  (You looked at a lot of the earlier code, and perhaps this too.)

db/src/tx.rs
```diff
@@ -340,25 +341,20 @@ impl<'conn, 'a> Tx<'conn, 'a> {
...
-use std::borrow::Cow;
-
-/// Transact the given `entities` against the given SQLite `conn`, using the metadata in
-/// `self.DB`.
+/// Transact the given `entities` against the given SQLite `conn`, using the given metadata.
 ///
 /// This approach is explained in https://github.com/mozilla/mentat/wiki/Transacting.
 // TODO: move this to the transactor layer.
-pub fn transact<'conn, 'a, I>(conn: &'conn rusqlite::Connection, partition_map: &'a PartitionMap, schema: &'a Schema, entities: I) -> Result<(TxReport, PartitionMap, Option<Schema>)> where I: IntoIterator<Item=Entity> {
+pub fn transact<'conn, 'a, I>(conn: &'conn rusqlite::Connection, mut partition_map: PartitionMap, schema: &'a Schema, entities: I) -> Result<(TxReport, PartitionMap, Option<Schema>)> where I: IntoIterator<Item=Entity> {
```
rnewman: Why does `partition_map` need to be `mut`? IIRC the only reason for that here is if you assign to `partition_map` in the body of the function, and you don't, right?

src/conn.rs
```diff
@@ -0,0 +1,215 @@
...
+
+
+/// Connection metadata required to query from, or apply transactions to, a Mentat store.
+///
+/// Owned data for the volatile parts (generation and partition map), and `Arc` for the infrequently
+/// changing parts (schema) that we want to share across threads.
+///
+/// See https://github.com/mozilla/mentat/wiki/Thoughts:-modeling-db-conn-in-Rust.
+pub struct Metadata {
+    pub generation: usize,
```
rnewman: Why `usize` over `u64`?

src/conn.rs
```diff
@@ -0,0 +1,215 @@
...
+    pub partition_map: PartitionMap,
```
rnewman: I'm thinking about this. Is there a situation in which a caller will take a reference to the `partition_map` or `generation`, perhaps in order to validate work optimistically depending on the schema? Does this modeling help or hinder?

src/conn.rs
```diff
@@ -0,0 +1,215 @@
...
+
+    /// Yield the current `Schema` instance.
+    pub fn current_schema(&self) -> Arc<Schema> {
+        // We always unwrap the mutex lock: if it's poisoned, this will propogate panics to all
+        // accessing threads.  This is perhaps not reasonable; we expect the mutex to be held for
+        // very short intervals, but a panic during a critical update section is possible, since the
+        // lock encapsulates committing a SQL transaction.
+        //
+        // That being said, in the future we will provide an interface to take the mutex, providing
+        // maximum flexibility for Mentat consumers.
```
rnewman: The only thing one can do in this case is recover, probably by effectively closing and reopening. File an issue?

Poisoning should be rare.

src/conn.rs
```diff
@@ -0,0 +1,215 @@
...
+        //
+        // This approach might need to change when we support interrupting query threads (#297), and
+        // will definitely need to change if we support interrupting transactor threads.
+        self.metadata.lock().unwrap().schema.clone()
+    }
+
+    /// Query the Mentat store, using the given connection and the current metadata.
+    pub fn q_once<T>(&self,
```
rnewman: Lovely. I'll build on top of this.

src/conn.rs
```diff
@@ -0,0 +1,215 @@
...
+
+        // The transaction is processed while the mutex is not held.
+        let (report, next_partition_map, next_schema) = transact(&tx, current_partition_map, &*current_schema, entities)?;
+
+        {
+            // The mutex is taken during this block.
+            let mut metadata = self.metadata.lock().unwrap();
+
+            if current_generation != metadata.generation {
+                // Somebody else wrote!
```
rnewman: File: retry the transact automatically?

src/conn.rs
```diff
@@ -0,0 +1,215 @@
...
+
+        let tx = sqlite.transaction()?;
+
+        let (current_generation, current_partition_map, current_schema) =
+        {
+            // The mutex is taken during this block.
+            let ref current: Metadata = *self.metadata.lock().unwrap();
+            (current.generation,
+             // Expensive, but the partition map is updated after every committed transaction.
+             current.partition_map.clone(),
```
rnewman: Ah, gotcha.

db/src/db.rs
rnewman: There's conflicting precedent in Rust.

There are traits named after verbs: `Write`. After nouns: `Iterator`. After words that could be either: `Clone`.

Jim's book offers examples phrased as adjectives like `Spliceable` and `Visible` (with methods like `render`), named after their defining method`IsEmoji` (with method `is_emoji`), and named after a subtype (`Creature`).

rnewman: aturon's guide offers:

*Traits*: 	`CamelCase`; prefer transitive verbs, nouns, and then adjectives; avoid suffixes (like able)

rnewman: https://aturon.github.io/style/naming.html

src/conn.rs
ncalexan: I went back and forth on this.  `generation` is `Copy`, so that's not a big deal.  But `partition_map` is volatile (as you note below), so it deserves special care.  It's almost always small (just 3 partitions) so it's probably best just to `clone` it for consumers rather than to try to `Arc` it at all times.  If a consumer like `query_history` comes along which really wants a handle on the state of `:db.part/tx`, I'm happy to re-evaluate.  We might find that `partition_map` isn't even the right way to give that information to such a consumer!

src/conn.rs
ncalexan: No particular reason.

src/conn.rs
ncalexan: Filed https://github.com/mozilla/mentat/issues/356.

PR #349, opened 2017-02-24
Use error_chain in sql, query-parser, query-translator.

rnewman:
Note that I bumped to 0.9.0, which appears to fix some bugs. I did this while trying to solve an obscure macro expansion error (complaining about an extra `;`, when the issue was a _missing_ `;` somewhere else).

The last commit is a coal-face; you only really need to worry about the Error bits.

@ncalexan knows most about error_chain. @victorporof, please also take a look.

### DIFF
--- a/query-parser/src/parse.rs
+++ b/query-parser/src/parse.rs
@@ -29,46 +29,52 @@ use self::mentat_query::{
     WhereClause,
 };
 
-#[derive(Clone, Debug, Eq, PartialEq)]
-pub struct NotAVariableError(pub edn::Value);
+error_chain! {
+    types {
+        Error, ErrorKind, ResultExt, Result;
+    }
 
-#[allow(dead_code)]
-#[derive(Clone, Debug, Eq, PartialEq)]
-pub enum FindParseError {
-    Err,
-}
+    foreign_links {
+        EdnParseError(edn::parse::ParseError);
+    }
 
-#[allow(dead_code)]
-#[derive(Clone, Debug, Eq, PartialEq)]
-pub enum WhereParseError {
-    Err,
-}
+    errors {
+        NotAVariableError(value: edn::Value) {
+            description("not a variable")
+            display("not a variable: '{}'", value)
+        }
 
-#[derive(Clone,Debug,Eq,PartialEq)]
-pub enum QueryParseError {
-    InvalidInput(edn::Value),
-    EdnParseError(edn::parse::ParseError),
-    MissingField(edn::Keyword),
-    FindParseError(FindParseError),
-    WhereParseError(WhereParseError),
-    WithParseError(NotAVariableError),
-}
+        FindParseError {
+            description(":find parse error")
+            display(":find parse error")
+        }
 
-impl From<WhereParseError> for QueryParseError {
-    fn from(err: WhereParseError) -> QueryParseError {
-        QueryParseError::WhereParseError(err)
+        WhereParseError {
+            description(":where parse error")
+            display(":where parse error")
+        }
+
+        // Not yet used.
+        WithParseError {
+            description(":with parse error")
+            display(":with parse error")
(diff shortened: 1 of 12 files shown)


query-parser/src/parse.rs
```diff
@@ -219,18 +225,20 @@ def_value_parser_fn!(Find, find, FindSpec, input, {
 //
 #[allow(dead_code)]
 pub fn find_seq_to_find_spec(find: &[edn::Value]) -> FindParseResult {
-    Find::find()
-        .parse(find)
-        .map(|x| x.0)
-        .map_err(|_| FindParseError::Err)
```
rnewman: @ncalexan: pretty sure this is familiar to you. Combine's `ParseError` doesn't implement `Error`, so we can't use `chain_err`. `Result` isn't `Result`, and the compiler can't figure out the types, so we can't just use `map_err` here. (I even tried `.map_err(…).chain_err()`, and it complained about that, too.)

We also can't directly add a `ParseError` as an `ErrorKind`, because it has an associated lifetime and is generic up the wazoo, as before. I believe you have code to improve this, so we can align after that lands!

src/query.rs
```diff
@@ -31,30 +41,39 @@ pub enum QueryResults {
...
 #[allow(unused_variables)]
-pub fn q_once(sqlite: SQLiteConnection,
-              db: DB,
-              query: &str,
-              inputs: Option<HashMap<String, TypedValue>>) -> QueryExecutionResult {
+pub fn q_once<'sqlite, 'schema, 'query>
+(sqlite: &'sqlite rusqlite::Connection,
+ schema: &'schema Schema,
+ query: &'query str,
+ inputs: Option<HashMap<String, TypedValue>>) -> QueryExecutionResult {
```
victorporof: This formatting looks quite weird to me.

src/errors.rs
```diff
@@ -0,0 +1,33 @@
...
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+use rusqlite;
+
+use mentat_query_parser;
+use mentat_sql;
+
+error_chain! {
```
victorporof: So is the general consensus now to use error_chain everywhere we have multiple kinds of errors which need to propagate?

rnewman: We have little to no choice.

We either define a stack of enums as deep as our call stack, or we use (or reimplement) `error_chain` to flatten the stack into a broad `ErrorKind` enum at compile time and at runtime.

Notably if you have a module that uses `error_chain`, the modules _it_ calls must either use `error_chain` or return `Result<T, std::error::Error>` — if you provide non-error types as the error branch of the `Result`, `error_chain` lacks the coercions it needs, and you have to do a bunch of work at every call site to bail with the right value.

The simplest thing to do is make _everything_ use `error_chain`. I dislike its lack of specificity (there are situations in which some error kinds cannot occur, but it can't model that easily), and I dislike the proliferation of `Result` types, and I dislike that it makes error handling less visible. But the alternative is to have a pretty decent chunk of the codebase devoted to modeling different error enums and `From` conversions between them.

src/query.rs
rnewman: Yeah, it is. But there's no good indenting rules for long scope generic lists, which is probably why people use `'a, 'b, 'c`!

There are for types: `<T> (… foo: T) where T: Some<Long<Thing>>`, of course.

src/errors.rs
victorporof: Yeah, ideally we'd like to avoid Err(FooErr(BarErr(BazErr(...)))), but there's no way to do that unless we somehow declare all of our error types top level. And that's what error chain seems to do, so I'm in favor.

PR #350, opened 2017-02-24
Add edn differ

victorporof:
Fixes https://github.com/mozilla/mentat/issues/220

### DIFF
--- a/edn/src/pretty_print.rs
+++ b/edn/src/pretty_print.rs
@@ -9,24 +9,40 @@
 // specific language governing permissions and limitations under the License.
 
 use itertools::Itertools;
-use pretty;
+use pretty::{
+    BoxAllocator,
+    DocAllocator,
+    DocBuilder
+};
 
+use std::usize;
 use std::io;
 use std::borrow::Cow;
 
 use types::Value;
 
+pub struct MaxLineLength(pub usize);
+pub struct Indent(pub usize);
+
+pub enum Style {
+    Clojure(MaxLineLength),
+    Expanded(Indent)
+}
+
 impl Value {
     /// Return a pretty string representation of this `Value`.
-    pub fn to_pretty(&self, width: usize) -> Result<String, io::Error> {
+    pub fn to_pretty(&self, style: Style) -> Result<String, io::Error> {
         let mut out = Vec::new();
-        self.write_pretty(width, &mut out)?;
+        self.write_pretty(style, &mut out)?;
         Ok(String::from_utf8_lossy(&out).into_owned())
     }
 
     /// Write a pretty representation of this `Value` to the given writer.
-    pub fn write_pretty<W>(&self, width: usize, out: &mut W) -> Result<(), io::Error> where W: io::Write {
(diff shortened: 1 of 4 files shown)


victorporof:
@ncalexan ready for review.

victorporof:
@ncalexan Indeed. We had quite lengthy conversations about how to achieve this, and textual seemed like the simples approach that offered some value. Using structural diffing was my first thought too, but textual diffing gives us something useful (probably useful enough for 99% of our testing requirements) with less effort.

Adding more newlines to help the diffing with this approach is probably better.

victorporof:
@ncalexan Given my above response, and the fact that this approach is still useful for now (and using structural diffing and the existing edn differ would require a bit more work, especially around returning something nicer than `bool` when matching fails), I'd suggest landing this. What do you think?

victorporof:
@ncalexan gentle review ping.

edn/src/pretty_print.rs
```diff
@@ -37,70 +53,134 @@ impl Value {
...
-        assert_eq!(data.to_pretty(40).unwrap(), "[[] () #{} {} \"\"]");
+        assert_eq!(data.to_pretty(Style::Clojure(MaxLineLength(40))).unwrap(), "[[] () #{} {} \"\"]");
+    }
+
+    #[test]
+    fn test_pp_types_empty_expanded() {
+        let string = "[ [ ] ( ) #{ } { }, \"\" ]";
+        let data = parse::value(string).unwrap().without_spans();
+
+        // Ending whitespace with `\x20` to avoid editors from removing trailing
```
ncalexan: I saw this in your earlier patch and didn't comment explicitly; I should have.  This is a really strong indicator that something is wrong.  We shouldn't be pretty printing output that we wouldn't commit to the tree!

victorporof: We have the same issue in mozilla-central: sometimes trailing whitespace doesn't want to be removed, and the escaped whitespace is to make life easiers for people who's editors eagerly trim whitespace. I strongly disagree with the escaping here being a strong indicator of anything "wrong", or anything at all.

victorporof: Can you clarify what's exactly "wrong" with this? Have you closely looked at the reasoning behind the escaping?

victorporof:
> If you really think this is the right way to achieve this...

We had lengthy conversations about how to best achieve this, and we both arrived at the conclusion that textual diffing is better for the time being and for the intended purpose. If I didn't misunderstand, why the sudden change of heart?

> Can we at least not add this binary mode flag and just extract a separate recursive function? 

What's the objective advantage to doing that instead?

> There's no advantage to using pretty.rs for this printout. That is, why indent, group, nest, at all?

What's the advantage to NOT using pretty.rs?

> Can you see if just pretty printing with line length 1 (so everything breaks) does something close enough to what you want?

This would be USING pretty.rs. If we go this route, we'd just be replacing allocator.newline() with allocator.space(), and breaking lines at 1 instead of max, while removing some of the grouping. Would this change really be that significant to warrant a rewrite here?

PR #351, opened 2017-02-25
Initial implementation to see if we're on the right lines

joewalker:
No polish, just to see if I'm on the right lines

### DIFF
--- a/edn/src/matcher.rs
+++ b/edn/src/matcher.rs
@@ -73,33 +73,38 @@ impl<'a> Matcher<'a> {
 
     // This is the same as itertools::diff_with() except that it:
     // - returns a boolean
-    // - takes account of '...' (':ellipsis' for now)
+    // - takes account of '...'
     // - automatically calls match_internal rather than requiring an is_equal() closure
-    fn match_iter_internal<I, T>(&self, v: I, p: I) -> bool
-        where I: IntoIterator<Item=&'a Value>,
+    fn match_iter_internal<II, T>(&self, values: II, patterns: II) -> bool
+        where II: IntoIterator<Item=&'a Value>,
               T: PatternMatchingRules<'a, Value> {
-        let mut v = v.into_iter();
-        let mut p = p.into_iter();
-        while let Some(v_elem) = v.next() {
-            match p.next() {
-                None => return false,
-                Some(p_elem) => {
-                    match *p_elem {
-                        Value::Keyword(ref keyword) => {
-                            // We don't support having anything after the elipsis (yet?)
-                            match p.next() { Some(_) => unimplemented!(), None => {} }
+        let mut values = values.into_iter();
+        let mut patterns = patterns.into_iter();
 
-                            if keyword.0 == "ellipsis" { return true }
+        while let Some(value) = values.next() {
+            match patterns.next() {
+                None => { return false; },
+                Some(match_elem) => {
+                    match *match_elem {
+                        Value::PlainSymbol(ref symbol) if symbol.0 == "..." => {
+                            // We don't support having anything after the ellipsis
+                            match patterns.next() { Some(_) => unimplemented!(), None => {} }
+                            return true;
                         },
                         _ => {},
                     }
-                    if !self.match_internal::<T>(&v_elem, &p_elem) {
+                    if !self.match_internal::<T>(&value, &match_elem) {
                         return false;
                     }
                 },
             }
         }
-        match p.next() { None => true, Some(_) => false }
+
+        match patterns.next() {
+            None => true,
+            Some(&Value::PlainSymbol(ref symbol)) if symbol.0 == "..." => true,
+            _ => false,
+        }
     }
 
     /// Recursively traverses two EDN `Value` instances (`value` and `pattern`)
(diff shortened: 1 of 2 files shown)


edn/src/matcher.rs
```diff
@@ -72,6 +71,37 @@ impl<'a> Matcher<'a> {
...
+                        },
+                        _ => {},
+                    }
+                    if !self.match_internal::<T>(&v_elem, &p_elem) {
+                        return false;
+                    }
+                },
+            }
+        }
+        match p.next() { None => true, Some(_) => false }
```
victorporof: So if I'm reading this correctly , the code does the following:
* iterate the "value" and "pattern" collections in lockstep, with the "value" collection being the driver
* if the "pattern" collection is finished while the "value" collection still has some elements to iterate over, bail with `false`.
* do a bunch of special casing, equality checking and peeking when encountering something considered to be an "ellipsis" as the currently iterated element in the "pattern" collection.
* recurse matching otherwise.
* if the "value" collection is finished while the "pattern" collection still has some elements to iterate over, bail with `false`.

This approach seems overly complex for a number of reasons. For example: `[1 2 3 ...] !~ [1 2 3]`, when it should. Similarly, `[1 ... 4] != [1 2 3 4]`. Unordered collections are even more interesting: `#{3 2 ... 4}` should match `#{1 2 3 4 5}`, and since symbols have a higher precedence than uints for example, I don't see how any of these can be achieved with lockstep iteration (in this case *everything* will come after the ellipsis).

Let's step back a bit and think about we're actually trying to achieve: surjection[1] between the two groups, where the "value" collection is the domain and the "pattern" collection is the codomain: all elements in the "value" collection have a correspondent in the "pattern" collection, and vice versa but with the special rule that `...` can match anything. You could interpret `...` as sort of an reverse `_`: make sure each item in the "value" collection has a correspondent in the "pattern" collection, where "correspondents" can be something that pass `match_internal` or are equal to `...`.

I expect that it should be possible to implement this with just easy calls to `Iterator` consumers [2], like it is for the `set` and `map` collections.

[1] https://en.wikipedia.org/wiki/Surjective_function
[2] https://doc.rust-lang.org/stable/book/iterators.html#consumers

victorporof:
There might be some discussion needed revolving around whether or not `...` can match "nothing", that is, should `[...] =~ []` or not? I certainly believe `...` matching "nothing" would make things easier and more mathematically elegant, but that shouldn't be the driving force here. @ncalexan what do you think?

victorporof:
I've had a quick play with this just to make sure whether or not what I'm saying actually makes sense.

1. What I haven't anticipated, is that our edn parser can't parse `...` symbols. I've filed https://github.com/mozilla/mentat/issues/352 for this.
2. It doesn't seem absolutely necessary to reimplement `matches_with` at all. It's not incorrect to do it, but I think the resulting code is much simpler if you don't. One approach would be to just split the input collections around the `...` symbol (making sure that `...` can represent multiple values), and ascertaining whether or not the two resulting "left" and "right" subcollections match. Here's a quick sketch: https://gist.github.com/victorporof/acb72f6bd52cce16bd10e47089ac6393

edn/src/matcher.rs
```diff
@@ -72,6 +71,37 @@ impl<'a> Matcher<'a> {
...
+                            if keyword.0 == "ellipsis" { return true }
+                        },
+                        _ => {},
+                    }
+                    if !self.match_internal::<T>(&v_elem, &p_elem) {
+                        return false;
+                    }
+                },
+            }
+        }
```
ncalexan: The stanza below is `p.next().is_none()`.

ncalexan:
> There might be some discussion needed revolving around whether or not ... can match "nothing", that is, should [...] =~ [] or not? I certainly believe ... matching "nothing" would make things easier and more mathematically elegant, but that shouldn't be the driving force here. @ncalexan what do you think?

I concur, `...` is most naturally allowed to be empty.

joewalker:
@ncalexan:
> Further to @victorporof's "step back" and rethink, I'd like to try to do something simpler...

> * ordered Value::Vector and Value::List with a single ellipsis

OK. If you really need elements after the ellipsis. I was headed for the simpler, "ellipsis must be the last element" rule.

> * unordered Value::Map, where we accept a single ellipsis key (with ignored value, or mandatory value _).

I think this mostly comes down to semantics in writing patterns. What I really want to write is `{ a 1, ... }`, but maybe what better expresses that is `{ a 1, ... ... }`?

> * _possibly_ unordered Value::Set with a single ellipsis key, and all child element compared literally.

If you're saying _possibly_ and you'll be using this in writing tests, then I suggest leaving it to a follow-up.

victorporof:
> > unordered Value::Map, where we accept a single ellipsis key (with ignored value, or mandatory value _).
> I think this mostly comes down to semantics in writing patterns. What I really want to write is { a 1, ... }, but maybe what better expresses that is { a 1, ... ... }?

Since maps contain (key value) pairs, and not (key value+) tuples or (key) without value, it doesn't make any sense to me to have `...` matching values. Since `...` means "zero or many", and values in maps cannot be empty (note that "nil" isn't "empty"), nor contain more than one "value", `_` is a much better candidate.

victorporof:
Accepting a single ellipsis key certainly makes things easier, however it's potentially doable to make it possible to accept multiple ellipsis. For example, I can imagine this being useful in ordered sets, where `[1 ... 4 ... 6]` would match `[1 2 3 4 5 6]`. The suggested implementation in https://github.com/mozilla/mentat/pull/351#issuecomment-282546498 would simply have to make multiple "cuts" instead of a single one to support multiple ellipses. However, not sure if this feature is really necessary for our use case, so let's not pursue it for now unless there's strong desire.

PR #353, opened 2017-02-28
Projection and miscellany

rnewman:
This PR is work to get us to the point of running a complete query from input string to results. Most of that work will be in projection.

### DIFF
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -331,6 +331,24 @@ pub fn ensure_current_version(conn: &mut rusqlite::Connection) -> Result<DB> {
     }
 }
 
+pub trait SQLValueType {
+    fn value_type_tag(&self) -> i32;
+}
+
+impl SQLValueType for ValueType {
+    fn value_type_tag(&self) -> i32 {
+        match *self {
+            ValueType::Ref =>      0,
+            ValueType::Boolean =>  1,
+            ValueType::Instant =>  4,
+            ValueType::Long =>     5,
+            ValueType::Double =>   5,
+            ValueType::String =>  10,
+            ValueType::Keyword => 13,
+        }
+    }
+}
+
 pub trait TypedSQLValue {
     fn from_sql_value_pair(value: rusqlite::types::Value, value_type_tag: i32) -> Result<TypedValue>;
     fn to_sql_value_pair<'a>(&'a self) -> (ToSqlOutput<'a>, i32);
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -341,6 +341,7 @@ impl SQLValueType for ValueType {
             ValueType::Ref =>      0,
             ValueType::Boolean =>  1,
             ValueType::Instant =>  4,
+            // SQLite distinguishes integral from decimal types, allowing long and double to share a tag.
             ValueType::Long =>     5,
             ValueType::Double =>   5,
             ValueType::String =>  10,
--- a/query-projector/src/lib.rs
+++ b/query-projector/src/lib.rs
@@ -117,7 +117,12 @@ impl TypedIndex {
     /// - If the retrieved value can't be coerced to a rusqlite `Value`.
     /// - Either index is out of bounds.
     ///
-    /// This function will return an error if the type code is unknown, or the value is
+    /// Because we construct our SQL projection list, the code that stored the data, and this
+    /// consumer, a panic here implies that we have a bad bug — we put data of a very wrong type in
+    /// a row, and thus can't coerce to Value, we're retrieving from the wrong place, or our
+    /// generated SQL is junk.
+    ///
+    /// This function will return a runtime error if the type code is unknown, or the value is
     /// otherwise not convertible by the DB layer.
     fn lookup<'a, 'stmt>(&self, row: &Row<'a, 'stmt>) -> Result<TypedValue> {
         use TypedIndex::*;
@@ -136,7 +141,6 @@ impl TypedIndex {
     }
 }
 
-// TODO: this sucks. Can we take ownership and avoid the clone?
 fn column_name(var: &Variable) -> Name {
     let &Variable(PlainSymbol(ref s)) = var;
     s.clone()
@@ -147,15 +151,25 @@ fn value_type_tag_name(var: &Variable) -> Name {
     format!("{}_value_type_tag", s)
 }
 
-/// Walk an iterator of `Element`s, collecting projector templates and
-/// pushing columns into the provided `Vec`.
+/// Walk an iterator of `Element`s, collecting projector templates and columns.
+///
+/// Returns a pair: the SQL projection (which should always be a `Projection::Columns`)
+/// and a `Vec` of `TypedIndex` 'keys' to use when looking up values.
+///
+/// Callers must ensure that every `Element` is distinct -- a query like
+///
+/// ```edn
+/// [:find ?x ?x :where [?x _ _]]
+/// ```
+///
+/// should fail to parse. See #358.
 fn project_elements<'a, I: IntoIterator<Item = &'a Element>>(
+    count: usize,
     elements: I,
-    starting_at: i32,
-    query: &AlgebraicQuery,
-    cols: &mut Vec<ProjectedColumn>) -> Vec<TypedIndex> {
+    query: &AlgebraicQuery) -> (Projection, Vec<TypedIndex>) {
 
-    let mut i: i32 = starting_at;
+    let mut cols = Vec::with_capacity(count);
+    let mut i: i32 = 0;
     let mut templates = vec![];
 
     for e in elements {
@@ -164,7 +178,15 @@ fn project_elements<'a, I: IntoIterator<Item = &'a Element>>(
             // into the SQL projection, aliased to the name of the variable,
             // and we push an annotated index into the projector.
             &Element::Variable(ref var) => {
-                let qa = query.cc.bindings.get(var).unwrap()[0].clone();   // TODO
+                // Every variable should be bound by the top-level CC to at least
+                // one column in the query. If that constraint is violated it's a
+                // bug in our code, so it's appropriate to panic here.
+                let columns = query.cc
+                                   .bindings
+                                   .get(var)
+                                   .expect("Every variable has a binding");
+
+                let qa = columns[0].clone();
                 let name = column_name(var);
 
                 if let Some(t) = query.cc.known_types.get(var) {
@@ -187,7 +209,7 @@ fn project_elements<'a, I: IntoIterator<Item = &'a Element>>(
         }
     }
 
-    templates
+    (Projection::Columns(cols), templates)
 }
 
 pub trait Projector {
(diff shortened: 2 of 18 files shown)


db/src/db.rs
```diff
@@ -331,6 +331,24 @@ pub fn ensure_current_version(conn: &mut rusqlite::Connection) -> Result<DB> {
...
+    fn value_type_tag(&self) -> i32;
+}
+
+impl SQLValueType for ValueType {
+    fn value_type_tag(&self) -> i32 {
+        match *self {
+            ValueType::Ref =>      0,
+            ValueType::Boolean =>  1,
+            ValueType::Instant =>  4,
+            ValueType::Long =>     5,
```
ncalexan: nit: we've been fairly consistent about including the note about overlapping tags more or less every place it could be included.  Follow suit?

src/errors.rs
```diff
@@ -31,6 +32,7 @@ error_chain! {
     links {
         DbError(mentat_db::Error, mentat_db::ErrorKind);
         QueryParseError(mentat_query_parser::Error, mentat_query_parser::ErrorKind);
+        ProjectorError(mentat_query_projector::Error, mentat_query_projector::ErrorKind);
```
ncalexan: BOILERS PLATED, SIR.

src/query.rs
```diff
@@ -59,21 +56,22 @@ pub fn q_once<'sqlite, 'schema, 'query>
...
+                .map(|&(ref k, ref v)| (k.as_str(), v as &ToSql))
+                .collect();
+        statement.query_named(refs.as_slice())?
     };
-    */
-
-
 
-    Ok(QueryResults::Scalar(Some(TypedValue::Boolean(true))))
+    select.projector
```
ncalexan: Is this commit out of order?  Check before landing please, although this is often just a Github UI foible.

query-projector/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+    Name,
+    Projection,
+    ProjectedColumn,
+    /*
+    SelectQuery,
+    TableList,
+    */
+};
+
+error_chain! {
```
ncalexan: Huh, interesting, instead of `errors.rs`.  Not sure if I prefer this inline our in its own file right now.  Roll on.

query-projector/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+enum TypedIndex {
+    Known(Index, ValueTypeTag),
+    Unknown(Index, Index),
+}
+
+impl TypedIndex {
+    /// Look up this index and type(index) pair in the provided row.
+    /// This function will panic if:
+    ///
+    /// - This is an `Unknown` and the retrieved type code isn't an i32.
```
ncalexan: It looks like you're only going to panic when `Rusqlite::get` fails.  There's `Rusqlite::get_checked` which will allow you to recover.  It's almost always better to just surface your `Result` rather than panic.  Do you feel strongly otherwise?

query-projector/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+            &Unknown(value_index, type_index) => {
+                let v: rusqlite::types::Value = row.get(value_index);
+                let value_type_tag: i32 = row.get(type_index);
+                TypedValue::from_sql_value_pair(v, value_type_tag).map_err(|e| e.into())
+            },
+        }
+    }
+}
+
+// TODO: this sucks. Can we take ownership and avoid the clone?
```
ncalexan: I don't see why not, just take `var: Variable` and consume it, either by pattern destructuring or using `var.0.0` or whatever.

query-projector/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+    let mut i: i32 = starting_at;
+    let mut templates = vec![];
+
+    for e in elements {
+        match e {
+            // Each time we come across a variable, we push a SQL column
+            // into the SQL projection, aliased to the name of the variable,
+            // and we push an annotated index into the projector.
+            &Element::Variable(ref var) => {
+                let qa = query.cc.bindings.get(var).unwrap()[0].clone();   // TODO
```
ncalexan: Worth a note that a panic here is an implementation, not a runtime, error.

query-projector/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+            .iter()
+            .map(|ti| ti.lookup(&row))
+            .collect::<Result<Vec<TypedValue>>>()
+    }
+}
+
+impl Projector for RelProjector {
+    fn project<'stmt>(&self, mut rows: Rows<'stmt>) -> Result<QueryResults> {
+        let mut out: Vec<Vec<TypedValue>> = vec![];
+        while let Some(r) = rows.next() {
```
ncalexan: So, this is one way to iterate `rusqlite` cursors.  There's also `query_map` and `query_and_then`.  You chose the lowest level option, but then immediately build the higher level functionality of `query_and_then` on top (by pushing into `out`); can you explain why you expressed this in this way?

query-projector/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+            let bindings = self.collect_bindings(row)?;
+            out.push(bindings);
+        }
+        Ok(QueryResults::Rel(out))
+    }
+}
+
+/// A coll projector produces a vector of values.
+/// Each value is sourced from the same column.
+struct CollProjector {
```
ncalexan: nit: consider s/Coll/Collection/.

query-projector/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+    let &Variable(PlainSymbol(ref s)) = var;
+    s.clone()
+}
+
+fn value_type_tag_name(var: &Variable) -> Name {
+    let &Variable(PlainSymbol(ref s)) = var;
+    format!("{}_value_type_tag", s)
+}
+
+/// Walk an iterator of `Element`s, collecting projector templates and
```
ncalexan: I feel like you're making some assumptions about uniqueness of `Element` variable names, or some assumption about the SQL names they'll be mapped to.  What does the code do in the case of `:find ?x ?x :where ?x ...`?  Do you get two SQL variables, or just one?  (I expect two, given my reading of the patch.)

query-projector/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+/// pushing columns into the provided `Vec`.
+fn project_elements<'a, I: IntoIterator<Item = &'a Element>>(
+    elements: I,
+    starting_at: i32,
+    query: &AlgebraicQuery,
+    cols: &mut Vec<ProjectedColumn>) -> Vec<TypedIndex> {
```
ncalexan: I'm not against this, but I do find taking an accumulator _and_ returning something else unusual.  Especially since your usage below doesn't exploit the fact that you can append to an existing `cols` with this approach (really -- it does allocate `with_capacity`, but that's not a big enough win to bend convention for).

query-translator/tests/translate.rs
```diff
@@ -72,8 +72,8 @@ fn test_rel() {
     let input = r#"[:find ?x :where [?x :foo/bar "yyy"]]"#;
     let parsed = parse_find_string(input).expect("parse failed");
     let algebrized = algebrize(&schema, parsed);
-    let select = cc_to_exists(algebrized.cc);
-    let SQLQuery { sql, args } = select.to_sql_query().unwrap();
-    assert_eq!(sql, "SELECT 1 FROM `datoms` AS `datoms00` WHERE `datoms00`.a = 99 AND `datoms00`.v = $v0");
+    let select = query_to_select(algebrized);
```
ncalexan: Further to my earlier point, can I see `[:find ?x ?x ... ]` in these tests as well?

query-projector/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+
+/// Convenience function for grabbing a single-element projection.
+fn one_typed_index(element: &Element, query: &AlgebraicQuery, mut cols: &mut Vec<ProjectedColumn>) -> TypedIndex {
+    let mut templates = project_elements(iter::once(element), 0, query, &mut cols);
+    assert_eq!(1, templates.len());
+    templates.pop().expect("Couldn't pop a vec of len 1!")
+}
+
+/// Combines the two things you need to turn a query into SQL and turn its results into
+/// `QueryResults`.
```
ncalexan: The `README.md` has a nice description of the two things.  It's worth repeating.

nit: consider s/projector/datalog_projector/ or similar to really drive the "two sides, one coin" point.  I found the two projections/projectors a little tricky to keep track of why reading this, and had to refer to the high-level note when I read this.

tests/query.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    // Rel.
+    let start = time::PreciseTime::now();
+    let results = q_once(&c, &db.schema,
+                         "[:find ?x ?ident :where [?x :db/ident ?ident]]", None)
+        .expect("Query failed");
+    let end = time::PreciseTime::now();
+
+    // This will need to change each time we add a default ident.
+    assert_eq!(37, results.len());
```
ncalexan: Hmm.  You could export something from `bootstrap` to make this a little future proof, but that's a weird thing to leak to the global namespace.

tests/query.rs
```diff
@@ -0,0 +1,121 @@
...
+
+use mentat::{
+    NamespacedKeyword,
+    QueryResults,
+    new_connection,
+    q_once,
+};
+
+#[test]
+fn test_query() {
```
ncalexan: nit: consider splitting this test function into individual bits, so that the failure messages are clearer.

query-projector/src/lib.rs
rnewman: Because we construct our SQL projection list, the code that stored the data, and this consumer, a panic here implies that we have a bad bug — we put data of a very wrong type in a row, and thus can't coerce to `Value`, we're retrieving from the wrong place, or our generated SQL is junk.

Generally I'd rather a bug like this cause a crash with a stack rather than propagate up to application error handling, which — let's face it — is normally just "Oops, an error occurred", or showing an empty list, and muffling the problem altogether.

query-projector/src/lib.rs
rnewman: Because we're pulling this `Variable` out of the `Elements` clause, which comes from the iterator we're given, which comes from the `FindSpec`, which is part of the query itself. Owning this variable puts strong limits on what else can be done with the query after this function is called, right?

query-projector/src/lib.rs
rnewman: You'll get two entries in the SQL projection list that project the same column (indeed, to the same name). I'll make sure this works correctly.

query-projector/src/lib.rs
rnewman: This was just an awkward refactor.

query-projector/src/lib.rs
rnewman: Filed #358.

query-projector/src/lib.rs
rnewman: I switched to `expect` and added a comment.

query-projector/src/lib.rs
rnewman: Two reasons:

- Some find specs/projectors — as you can see in `ScalarProjector` — will only ever consume one row. Rather than having two code paths, we simply always pass `Rows` to the projector, which can choose how much to consume.
- Some future projections will decouple the input rows from the output results, accumulating 0–N results for each SQL row; some of those behaviors will require context (such as accumulating into a `Set`). `query_map` wouldn't be appropriate for those.

query-translator/tests/translate.rs
rnewman: Will add as part of #358.

PR #354, opened 2017-02-28
Accept symbol names and namespaces with '.'. Fixes #352.

### LINKED ISSUE #352: Make the edn parser properly handle `...` symbols
The `...` strings parses as three separate symbols at the moment, when it should be just one.

ncalexan:
I did some testing in the Clojure REPL to verify these.

### DIFF
--- a/edn/src/edn.rustpeg
+++ b/edn/src/edn.rustpeg
@@ -148,11 +148,11 @@ namespace_separator = "/"
 // TODO: Be more picky here
 // Keywords follow the rules of symbols, except they can (and must) begin with :
 // e.g. :fred or :my/fred. See https://github.com/edn-format/edn#keywords
-symbol_char_initial = [a-z] / [A-Z] / [0-9] / [*!_?$%&=<>]
-symbol_char_subsequent = [a-z] / [A-Z] / [0-9] / [-*!_?$%&=<>]
+symbol_char_initial = [a-z] / [A-Z] / [0-9] / [*!_?$%&=<>.]
+symbol_char_subsequent = [a-z] / [A-Z] / [0-9] / [-*!_?$%&=<>.]
 
-symbol_namespace = symbol_char_initial symbol_char_subsequent* (namespace_divider symbol_char_subsequent+)*
-symbol_name = ( symbol_char_initial+ / "." ) ( symbol_char_subsequent* / "." )
+symbol_namespace = symbol_char_initial symbol_char_subsequent*
+symbol_name = symbol_char_initial symbol_char_subsequent*
 
 keyword_prefix = ":"
 
--- a/edn/tests/tests.rs
+++ b/edn/tests/tests.rs
@@ -313,9 +313,13 @@ fn test_span_text() {
 fn test_symbol() {
     assert_eq!(symbol("$").unwrap(), s_plain("$"));
     assert_eq!(symbol(".").unwrap(), s_plain("."));
+    assert_eq!(symbol("...").unwrap(), s_plain("..."));
 
     assert_eq!(symbol("hello/world").unwrap(), s_ns("hello", "world"));
     assert_eq!(symbol("foo-bar/baz-boz").unwrap(), s_ns("foo-bar", "baz-boz"));
+    assert_eq!(symbol("hello/...").unwrap(), s_ns("hello", "..."));
+    assert_eq!(symbol("./hello").unwrap(), s_ns(".", "hello"));
+    assert_eq!(symbol("../..").unwrap(), s_ns("..", ".."));
 
     assert_eq!(symbol("foo-bar/baz_boz").unwrap(), s_ns("foo-bar", "baz_boz"));
     assert_eq!(symbol("foo_bar/baz-boz").unwrap(), s_ns("foo_bar", "baz-boz"));
@@ -337,6 +341,10 @@ fn test_span_symbol() {
         inner: SpannedValue::from_symbol("hello", "world"),
         span: Span(0, 11)
     });
+    assert_eq!(parse::symbol("...").unwrap(), ValueAndSpan {
+        inner: SpannedValue::from_symbol(None, "..."),
+        span: Span(0, 3)
+    });
 }
 
 #[test]
@@ -353,6 +361,9 @@ fn test_keyword() {
     assert_eq!(keyword(":foo-bar").unwrap(), k_plain("foo-bar"));
     assert_eq!(keyword(":foo_bar").unwrap(), k_plain("foo_bar"));
 
+    assert_eq!(keyword(":./world").unwrap(), k_ns(".", "world"));
+    assert_eq!(keyword(":hello/...").unwrap(), k_ns("hello", "..."));
+
     assert!(keyword(":").is_err());
     assert!(keyword(":foo/").is_err());
 }
@@ -366,6 +377,10 @@ fn test_span_keyword() {
     assert_eq!(parse::keyword(":hello/world").unwrap(), ValueAndSpan {
         inner: SpannedValue::from_keyword("hello", "world"),
         span: Span(0, 12)
+    });
+    assert_eq!(parse::keyword(":.../...").unwrap(), ValueAndSpan {
+        inner: SpannedValue::from_keyword("...", "..."),
+        span: Span(0, 8)
     });
 }
 


edn/src/edn.rustpeg
```diff
@@ -148,11 +148,11 @@ namespace_separator = "/"
 // TODO: Be more picky here
 // Keywords follow the rules of symbols, except they can (and must) begin with :
 // e.g. :fred or :my/fred. See https://github.com/edn-format/edn#keywords
-symbol_char_initial = [a-z] / [A-Z] / [0-9] / [*!_?$%&=<>]
-symbol_char_subsequent = [a-z] / [A-Z] / [0-9] / [-*!_?$%&=<>]
+symbol_char_initial = [a-z] / [A-Z] / [0-9] / [*!_?$%&=<>.]
```
rnewman: This conflicts with the Clojure reader definition:

> Symbols begin with a non-numeric character and can contain alphanumeric characters and `*`, `+`, `!`, `-`, `_`, `'`, and `?` (other characters may be allowed eventually).
>
> `/` has special meaning, it can be used once in the middle of a symbol to separate the namespace from the name, e.g. `my-namespace/foo`. `/` by itself names the division function.
>
> `.` has special meaning - it can be used one or more times in the middle of a symbol to designate a fully-qualified class name, e.g. `java.util.BitSet`, or in namespace names. Symbols beginning or ending with `.` are reserved by Clojure. Symbols containing `/` or `.` are said to be 'qualified'.

That implies to me that either (a) we should document that our interpretation of EDN is diverging from Clojure's reader, or (b) we should special-case `.` and `...` as acceptable symbols.

I seem to remember that we did so…

edn/src/edn.rustpeg
```diff
@@ -148,11 +148,11 @@ namespace_separator = "/"
...
+symbol_char_subsequent = [a-z] / [A-Z] / [0-9] / [-*!_?$%&=<>.]
 
-symbol_namespace = symbol_char_initial symbol_char_subsequent* (namespace_divider symbol_char_subsequent+)*
-symbol_name = ( symbol_char_initial+ / "." ) ( symbol_char_subsequent* / "." )
```
rnewman: Isn't it a sufficient fix to this bug to simply change the final `"."` in this production to `"."*`?

edn/src/edn.rustpeg
```diff
@@ -148,11 +148,11 @@ namespace_separator = "/"
 // TODO: Be more picky here
 // Keywords follow the rules of symbols, except they can (and must) begin with :
 // e.g. :fred or :my/fred. See https://github.com/edn-format/edn#keywords
-symbol_char_initial = [a-z] / [A-Z] / [0-9] / [*!_?$%&=<>]
-symbol_char_subsequent = [a-z] / [A-Z] / [0-9] / [-*!_?$%&=<>]
+symbol_char_initial = [a-z] / [A-Z] / [0-9] / [*!_?$%&=<>.]
+symbol_char_subsequent = [a-z] / [A-Z] / [0-9] / [-*!_?$%&=<>.]
```
rnewman: This should also allow for `+`.

edn/src/edn.rustpeg
joewalker: Or we could argue that the Clojure reader has diverged from https://github.com/edn-format/edn

joewalker:
I think it makes sense *not* to use `…` in place of `...` because it's hard to type. I `…` it because it's an option if things get complex with `...`.

edn/src/edn.rustpeg
ncalexan: Something has diverged, for sure.  I didn't read https://github.com/edn-format/edn to verify what it says because Clojure is so liberal about what it accepts: just `.` characters; prefix `.` characters; suffix `.` characters; in name and namespace, in any combination.

edn/src/edn.rustpeg
rnewman: I think the idea was related to "Symbols beginning or ending with . are reserved by Clojure.". That is: symbols are an alphanumeric character, followed by alphanum+some specials, and they can include a `java.divided.namespace`, an optional slash, and so on.

While Clojure's reader is very very lazy, and the `symbol` function doesn't do any validation, it's probably not *intentional* if either the EDN grammar or the Clojure reader allows a symbol like `...aaa......a.a..a...a.`

So we might have two pathways: 'normal' symbols, which must look like

`foo.bar_noo/b-az10`

and 'reserved' symbols, which can be things like

`...`, `.`, `.reserved/suffix-.`

That said, if Clojure reads it, and the EDN spec allows it, then I guess we have to allow variables in queries like `?..asda.as+das..aa.....`, right?

edn/src/edn.rustpeg
rnewman: It's hard to have a lot of faith in the spec, because it seems to be taking the tack of defining things imprecisely and then expanding the allowed set when it conflicts with Clojure:

https://github.com/edn-format/edn/issues/62

Probably the best source of truth is the ClojureScript reader package, but I haven't dug into that myself.

ncalexan:
> While Clojure's reader is very very lazy, and the symbol function doesn't do any validation, it's probably not intentional if either the EDN grammar or the Clojure reader allows a symbol like ...aaa......a.a..a...a.

And yet...
```clojure
user=> (type (quote ..a..a+a..a+))
#<Class@1b68b9a4 clojure.lang.Symbol>
```

I have no particular dog in this fight, other than wanting to avoid adding new `Special{Namespaced}Symbol` types to `edn::Value`.  Perhaps the best thing is to be liberal and land something like this patch, but with `+` as well as `.`?

rnewman:
> And yet...

Oh yes, I know it allows all kinds of junk!

```
user=> '...aa.a.a..a.a....
...aa.a.a..a.a....
user=> '?..asda.as+das..aa.....
?..asda.as+das..aa.....
```

What I mean by this whole line of inquiry is that if you look at the human-facing parts of both Clojure and EDN (the docs, issues, and loose specs), it's not at all clear that the implementation follows the _intention_. 

If one were to define a legal form to accept traditional Irish names, and wrote a one-line definition that they consist of letters, spaces, and apostrophes — which is how I think the Clojure implementation would do it — then if someone handed in a form with the name `"   ''''Os'''HA"`, you'd probably reply "well, that's not really what I meant".

Clojure has historically not bothered with good error messages, thorough validation, or polishing off the rough edges; it expects developers to take care to avoid getting cut.

We don't necessarily need to follow the same approach: we could quite easily allow only the special symbols `.` and `...`, otherwise requiring a sensible production like `foo-bar.baz/noo`. That might avoid pain later — remember, keywords read here can end up in the database and live forever, and the query parser needs to accept these symbols as tokens, eventually being escaped into SQL. The more narrowly we define EDN, the smaller our bug and attack surface elsewhere.

edn/src/edn.rustpeg
victorporof: Why do we even have `namespace_divider`? What are the actual rules for namespace dividers, that is: given `...` I'd expect it to be a symbol with no namespace and a name equal to `...`. Then again, I don't know that much about these EDN nuances, but seems like we're complicating our lives a bit with these distinctions.

rnewman: Correct, `...` would be a `PlainSymbol`.

`namespace_divider` is there to divide the namespace from the name. It's allowed to occur only once or never, and there are (in most interpretations) different rules for namespaces and names. E.g., Clojure expects namespaces to approximately be valid Java package names `foo.bar.baz`, so `...` _shouldn't_ be used as a namespace. (That doesn't mean the Clojure reader will reject it.)

joewalker:
I think that `..a...a.a` parses is absolutely part of the intention. Consider this from the spec: "_If -, + or . are the first character, the second character (if any) must be non-numeric._". You could almost say that another way "If it doesn't parse as a number then it's a symbol", or in other words, symbols are intentionally super broad.

DANGER DANGER - opinions about Lisp from someone who isn't a lisp programmer: Initially I thought this was silly. But from a Lisp (or any homoiconic language) POV, wouldn't it make sense to allow anything because you're extending the language from inside not from outside?

Either way, I think the right thing to do to `edn.rustpeg` is https://github.com/mozilla/mentat/pull/351/files?diff=unified#diff-85609b045eafaeaec1bfa7a81f296369

rnewman:
> Consider this from the spec: "If -, + or . are the first character, the second character (if any) must be non-numeric."

For symbol names, yes. But Clojure clear _wants_ namespace names to be sequences of parts separated by '.', yet its reader doesn't prevent that:

```
user=> 'a....aaaa..aa/a123'
a....aaaa..aa/a123'
```

If you name a namespace that way and pass it to `require`, you'll get a silly error in some other part of the code, because you're not following the rules that are baked into assumptions elsewhere (in this case, how namespaces map to file paths), yielding GIGO:

```
user=> (require 'a...a.aaaaaa)

FileNotFoundException Could not locate a///a/aaaaaa__init.class or a///a/aaaaaa.clj on classpath.  clojure.lang.RT.load (RT.java:456)
```

or similarly if the namespace is meant to identify a Java package, or look up documentation, or whatever.

The point of this diversion is just to say: I wouldn't necessarily trust Clojure's docs for individual modules, like the reader, as a good description of the inputs that Clojure the system actually sensibly consumes. That's all culture.

> You could almost say that another way "If it doesn't parse as a number then it's a symbol", or in other words, symbols are intentionally super broad.

Yes, that's exactly true for Clojure. It's basically any sequence of non-whitespace characters, optionally separated by a slash.

Perhaps the best way to write down my intuition is: I'd agree with this wholeheartedly if I were writing a Clojure parser in Rust; Clojure needs you to support symbols like `mypackage/++` and plain symbols like `..>++?`. The standard library has namespaced symbols like `<!`.

I might even agree with this if I were writing an EDN parser that absolutely must support every EDN file found in the wild; some of those will contain Clojure code!

But EDN "as she is spoke", particularly in Datalog queries, is a narrower set that I'd be happy to use to narrow down the input set that we must safely process — we already don't support the "extensible" part of EDN at all, so perhaps we can also exclude symbols that don't name functions or reasonable-looking variables? That is, can we make the entrance to our funnel match the rest of the system, so our culture doesn't diverge from our parser?

I'm happy with going broad if you (a) think that having a wide-ranging reusable parser is a noble goal, (b) think we can get the productions right, and (c) are OK with making sure we have validation in the two EDN-consuming parsers, in order to prevent users from adding attributes that might trip us up elsewhere (like `:+a../..+-`), or writing queries with variables like `???????`.

My motivation for getting involved here was to advocate for the minimal fix, which is to minimally broaden the set of parsed symbols to include `...`. It's less risky, but it yields a less reusable parser.

ncalexan:
OK, @rnewman has convinced me.  If we do the minimal thing to accept `.` and `...`, we might save ourselves some pain later.  If we find that general EDN consumers arise, we can always generalize.

@joewalker is that OK by you?

ncalexan:
This landed: 85f3b79f758f264adba0875218e1b60c7765e965

PR #359, opened 2017-03-06
Support a limited set of '.'-prefixed symbols. (#352)

rnewman:
This commit allows `.` and `...` to parse correctly as `PlainSymbol`.

Tests in edn, query-translator, and the top level have been added.

I opted for the minimal solution.


rnewman:
85f3b79

I added a little tweak to avoid allowing keywords.

PR #362, opened 2017-03-07
WIP: more algebrizing

rnewman:
This fleshes out some of the missing parts of the algebrizer:

- Non-ident keywords.
- Known-type optimization.
- Supporting arbitrary numerics: `[?x _ 15]`.

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -74,11 +74,13 @@ impl TypedValue {
             &TypedValue::Keyword(_) => ValueType::Keyword,
         }
     }
+
 }
 
 // Put this here rather than in `db` simply because it's widely needed.
 pub trait SQLValueType {
     fn value_type_tag(&self) -> i32;
+    fn accommodates_integer(&self, int: i64) -> bool;
 }
 
 impl SQLValueType for ValueType {
@@ -92,6 +94,28 @@ impl SQLValueType for ValueType {
             ValueType::Double =>   5,
             ValueType::String =>  10,
             ValueType::Keyword => 13,
+        }
+    }
+
+    /// Returns true if the provided integer is in the SQLite value space of this type. For
+    /// example, `1` is how we encode `true`.
+    ///
+    /// ```
+    /// use mentat_core::{ValueType, SQLValueType};
+    /// assert!(ValueType::Boolean.accommodates_integer(1));
+    /// assert!(!ValueType::Boolean.accommodates_integer(-1));
+    /// assert!(!ValueType::Boolean.accommodates_integer(10));
+    /// assert!(!ValueType::String.accommodates_integer(10));
+    /// ```
+    fn accommodates_integer(&self, int: i64) -> bool {
+        use ValueType::*;
+        match *self {
+            Ref               => int >= 0,
+            Boolean           => (int == 0) || (int == 1),
+            Instant | Long    => true,
+            Double            => true,
+            ValueType::String => false,
+            Keyword           => false,
         }
     }
 }
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -76,6 +76,26 @@ impl TypedValue {
     }
 }
 
+// Put this here rather than in `db` simply because it's widely needed.
+pub trait SQLValueType {
+    fn value_type_tag(&self) -> i32;
+}
+
+impl SQLValueType for ValueType {
+    fn value_type_tag(&self) -> i32 {
+        match *self {
+            ValueType::Ref =>      0,
+            ValueType::Boolean =>  1,
+            ValueType::Instant =>  4,
+            // SQLite distinguishes integral from decimal types, allowing long and double to share a tag.
+            ValueType::Long =>     5,
+            ValueType::Double =>   5,
+            ValueType::String =>  10,
+            ValueType::Keyword => 13,
+        }
+    }
+}
+
 #[test]
 fn test_typed_value() {
     assert!(TypedValue::Boolean(false).is_congruent_with(None));
--- a/query-algebrizer/src/lib.rs
+++ b/query-algebrizer/src/lib.rs
@@ -29,8 +29,24 @@ pub struct AlgebraicQuery {
     default_source: SrcVar,
     pub find_spec: FindSpec,
     has_aggregates: bool,
-    pub limit: Option<i64>,
+    pub limit: Option<u64>,
     pub cc: cc::ConjoiningClauses,
+}
+
+impl AlgebraicQuery {
+    pub fn apply_limit(&mut self, limit: Option<u64>) {
+        match self.limit {
+            None => self.limit = limit,
+            Some(existing) =>
+                match limit {
+                    None => (),
+                    Some(new) =>
+                        if new < existing {
+                            self.limit = limit;
+                        },
+                },
+        };
+    }
 }
 
 #[allow(dead_code)]
--- a/query-algebrizer/src/lib.rs
+++ b/query-algebrizer/src/lib.rs
@@ -34,6 +34,9 @@ pub struct AlgebraicQuery {
 }
 
 impl AlgebraicQuery {
+    /**
+     * Apply a new limit to this query, if one is provided and any existing limit is larger.
+     */
     pub fn apply_limit(&mut self, limit: Option<u64>) {
         match self.limit {
             None => self.limit = limit,
@@ -46,6 +49,10 @@ impl AlgebraicQuery {
                         },
                 },
         };
+    }
+
+    pub fn cannot_succeed(&self) -> bool {
+        self.cc.is_known_empty
     }
 }
 
--- a/query-projector/src/lib.rs
+++ b/query-projector/src/lib.rs
@@ -27,11 +27,11 @@ use rusqlite::{
 };
 
 use mentat_core::{
+    SQLValueType,
     TypedValue,
 };
 
 use mentat_db::{
-    SQLValueType,
     TypedSQLValue,
 };
 
--- a/query-projector/src/lib.rs
+++ b/query-projector/src/lib.rs
@@ -100,6 +100,16 @@ impl QueryResults {
             &Rel(ref v)    => v.len(),
         }
     }
+
+    pub fn empty(spec: &FindSpec) -> Box<Fn() -> QueryResults> {
+        use self::FindSpec::*;
+        match spec {
+            &FindScalar(_) => Box::new(|| QueryResults::Scalar(None)),
+            &FindTuple(_)  => Box::new(|| QueryResults::Tuple(None)),
+            &FindColl(_)   => Box::new(|| QueryResults::Coll(vec![])),
+            &FindRel(_)    => Box::new(|| QueryResults::Rel(vec![])),
+        }
+    }
 }
 
 type Index = i32;            // See rusqlite::RowIndex.
@@ -214,6 +224,22 @@ fn project_elements<'a, I: IntoIterator<Item = &'a Element>>(
 
 pub trait Projector {
     fn project<'stmt>(&self, rows: Rows<'stmt>) -> Result<QueryResults>;
+}
+
+struct NullProjector {
+    results: Box<Fn() -> QueryResults>,
+}
+
+impl NullProjector {
+    fn null(result_factory: Box<Fn() -> QueryResults>) -> NullProjector {
+        NullProjector { results: result_factory }
+    }
+}
+
+impl Projector for NullProjector {
+    fn project<'stmt>(&self, _: Rows<'stmt>) -> Result<QueryResults> {
+        Ok((self.results)())
+    }
 }
 
 struct ScalarProjector {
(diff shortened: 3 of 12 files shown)


rnewman:
@ncalexan: this isn't even close to review, but note that I moved some of the type code stuff out of `mentat_db`. It's fairly central to a bunch of stuff, and `mentat_db` is a huge crate with rusqlite dependencies. 

The type code mapping is, from one perspective, very closely coupled to our SQL schema, so it sort of belongs in `db`. On the other hand, just about _everything_ apart from the transact expander depends on the SQL schema (the translator, the projector, even parts of the algebrizer), and we don't really get any encapsulation if the query translator is importing `mentat_db` to do type code translation. As such, I don't think we gain much by keeping the simple `ValueType -> i32` mapping in `db`, and it's more convenient in core. What do you think?

ncalexan:
> @ncalexan: this isn't even close to review, but note that I moved some of the type code stuff out of mentat_db. It's fairly central to a bunch of stuff, and mentat_db is a huge crate with rusqlite dependencies.
> 
> The type code mapping is, from one perspective, very closely coupled to our SQL schema, so it sort of belongs in db. On the other hand, just about everything apart from the transact expander depends on the SQL schema (the translator, the projector, even parts of the algebrizer), and we don't really get any encapsulation if the query translator is importing mentat_db to do type code translation. As such, I don't think we gain much by keeping the simple ValueType -> i32 mapping in db, and it's more convenient in core. What do you think?

I'm fine with this: it's certainly convenient in `core`.  The large Rust projects I've seen all seem to grow some sort of `types` crate specifically to encapsulate these shared pieces; `core` is ours.  Land this when you get a chance.

rnewman:
This is now actually ready for review — I just didn't get around to pinging!

query-algebrizer/src/lib.rs
```diff
@@ -29,10 +29,26 @@ pub struct AlgebraicQuery {
     default_source: SrcVar,
     pub find_spec: FindSpec,
     has_aggregates: bool,
-    pub limit: Option<i64>,
+    pub limit: Option<u64>,
     pub cc: cc::ConjoiningClauses,
 }
 
+impl AlgebraicQuery {
+    pub fn apply_limit(&mut self, limit: Option<u64>) {
```
ncalexan: Worth commenting that `limit` is monotonically _decreasing_.

src/query.rs
```diff
@@ -53,14 +53,21 @@ pub type QueryExecutionResult = Result<QueryResults>;
 /// The caller is responsible for ensuring that the SQLite connection is in a transaction if
 /// isolation is required.
 #[allow(unused_variables)]
-pub fn q_once<'sqlite, 'schema, 'query>
+pub fn q_once<'sqlite, 'schema, 'query, T, U>
```
ncalexan: You might consider a fluent API, like:
```rust
q(sqlite, schema)
  .find("[?e ?a]")
  .where("[(> ?e :value)]")
  .with(<hashmap>)
  .limit(5)
  .results()
```
or similar.

query-sql/src/lib.rs
```diff
@@ -172,6 +172,19 @@ impl QueryFragment for Op {
     }
 }
 
+macro_rules! interpose {
```
ncalexan: Blech.  At least comment on what this does, for the next poor sod.  You already have a `QueryFragment` and I see that source aliases don't implement it 'cuz of ownership.  Consider newtyping `SourceAlias` so that can impl `QF` -- you might not need this macro.

query-translator/src/translate.rs
```diff
@@ -68,6 +73,40 @@ impl ToConstraint for ColumnConstraint {
...
+            EqualsPrimitiveLong(table, value) => {
+                let value_column = QualifiedAlias(table.clone(), DatomsColumn::Value).to_column();
+                let tag_column = QualifiedAlias(table, DatomsColumn::ValueTypeTag).to_column();
+                Constraint::And {
+                    constraints: vec![
+                        Constraint::equal(value_column, ColumnOrExpression::Value(TypedValue::Long(value))),
+                        Constraint::In {
+                            left: tag_column,
+                            list:
+                                if value < 0 {
```
ncalexan: This is not wrong, but it's the first time that we've used this information (both for refs and for booleans).  We've been consistent for booleans that `false` is 0, but we haven't used that `true` is 1 AFAIK -- we've always used that `true` is not 0.  No change needed, just noting this.

ncalexan: Consider finding a way to encode these symbolically, referencing the `ValueType` type in some way.  (Rust does not make this as pleasant as Swift.)

query-translator/src/translate.rs
```diff
@@ -68,6 +73,40 @@ impl ToConstraint for ColumnConstraint {
...
+                                         ColumnOrExpression::Integer(4),
+                                         ColumnOrExpression::Integer(5),]
+                                },
+                        },
+                    ],
+                }
+            },
+
+
+            HasType(table, value_type) => {
```
ncalexan: nit: newlines are a little irregular here.

query-translator/tests/translate.rs
```diff
@@ -87,11 +89,43 @@ fn test_limit() {
...
+    let SQLQuery { sql, args } = translate(&schema, input, 5);
     assert_eq!(sql, "SELECT `datoms00`.e AS `?x` FROM `datoms` AS `datoms00` WHERE `datoms00`.a = 99 AND `datoms00`.v = $v0 LIMIT 5");
     assert_eq!(args, vec![("$v0".to_string(), "yyy".to_string())]);
 }
+
+#[test]
+fn test_unknown_attribute_keyword_value() {
+    let schema = Schema::default();
+
+    let input = r#"[:find ?x :where [?x _ :ab/yyy]]"#;
```
ncalexan: Consider a test where you have a concrete ref attribute (say `:db/ident`) and the query includes a concrete unknown ident.

query-sql/src/lib.rs
rnewman: No, I later use `interpose` wherever I need to produce a sequence of separated things, not just in `SourceAlias` — it's used four times by the end of this PR.

query-translator/src/translate.rs
rnewman: The query input is non-symbolic: we get an integer, and now we're trying to find out which types it could match. The best I think we could do here is to push this logic into `ValueType` as a static function: "which types could be represented by this value?".

rnewman:
Re-review on that last commit, please, @ncalexan; they're fairly large changes.

core/src/lib.rs
```diff
@@ -94,6 +96,28 @@ impl SQLValueType for ValueType {
...
+    /// assert!(!ValueType::Boolean.accommodates_integer(-1));
+    /// assert!(!ValueType::Boolean.accommodates_integer(10));
+    /// assert!(!ValueType::String.accommodates_integer(10));
+    /// ```
+    fn accommodates_integer(&self, int: i64) -> bool {
+        use ValueType::*;
+        match *self {
+            Ref               => int >= 0,
+            Boolean           => (int == 0) || (int == 1),
+            Instant | Long    => true,
```
ncalexan: nit: no pipe for symmetry with `Double`.

query-algebrizer/src/lib.rs
```diff
@@ -47,6 +50,10 @@ impl AlgebraicQuery {
                 },
         };
     }
+
+    pub fn cannot_succeed(&self) -> bool {
```
ncalexan: Success is a little different than "can return results".  Consider something like `is_empty`, although I suppose that's a property of the `QueryResult` (set) rather than the query itself.  Maybe `query_result_is_empty`?

query-projector/src/lib.rs
```diff
@@ -100,6 +100,16 @@ impl QueryResults {
             &Rel(ref v)    => v.len(),
         }
     }
+
```
ncalexan: Consider implementing `is_empty` here while you're at it.

query-projector/src/lib.rs
```diff
@@ -216,6 +226,22 @@ pub trait Projector {
     fn project<'stmt>(&self, rows: Rows<'stmt>) -> Result<QueryResults>;
 }
 
+struct NullProjector {
```
ncalexan: "A projector that produces a `QueryResult` containing no data."

query-sql/src/lib.rs
```diff
@@ -127,6 +136,35 @@ fn push_column(qb: &mut QueryBuilder, col: &DatomsColumn) {
...
+///   1, 2, 3
+/// ```
+///
+/// or
+///
+/// ```sql
+/// x = 1 AND y = 2
+/// ```
+///
+/// without producing an intermediate string sequence.
```
ncalexan: You don't need to do this now, but I think you'd find a plain function taking closures easier to read: there's no "dangling ident", and each closure can take `|out|` or whatever.

query-sql/src/lib.rs
```diff
@@ -280,8 +305,15 @@ impl QueryFragment for FromClause {
     fn push_sql(&self, out: &mut QueryBuilder) -> BuildQueryResult {
         use self::FromClause::*;
         match self {
-            &TableList(ref table_list) => table_list.push_sql(out),
-            &Join(ref join) => join.push_sql(out),
+            &TableList(ref table_list) => {
+                out.push_sql(" FROM ");
```
ncalexan: You don't have to do this now, but further to my trait comment earlier, you could implement `" FROM ".push_sql(out)`, both for symmetry and potentially for easier use with `itertools`, etc.

query-sql/src/lib.rs
```diff
@@ -290,19 +322,15 @@ impl QueryFragment for SelectQuery {
     fn push_sql(&self, out: &mut QueryBuilder) -> BuildQueryResult {
         out.push_sql("SELECT ");
         self.projection.push_sql(out)?;
-
-        out.push_sql(" FROM ");
         self.from.push_sql(out)?;
 
-        if self.constraints.is_empty() {
-            return Ok(());
+        if !self.constraints.is_empty() {
```
ncalexan: Is it ever the case that `self.constraints` will be non-empty but `from` will be `None`?  If it is, you might consider trying to simplify the SQL produced from `self.projection` and `self.constraints`.  This is very much up to you, though, since I doubt you want to make the query itself represent that it is from `None`.

query-translator/src/translate.rs
```diff
@@ -69,40 +70,45 @@ impl ToConstraint for ColumnConstraint {
...
+                /// However, '1' and '0' are used to represent booleans, and some integers are also
+                /// used to represent FTS values. We don't want to accidentally match those.
+                ///
+                /// We ask `SQLValueType` whether this value is in range for how booleans are
+                /// represented in the database.
+                ///
+                /// We only hit this code path when the attribute is unknown, so we're querying
+                /// `all_datoms`. That means we don't see FTS IDs at all -- they're transparently
+                /// replaced by their strings. If that changes, then you should also exclude the
+                /// string type code (10) here.
```
ncalexan: Belt and braces: should we just do this across the board?

query-translator/src/translate.rs
```diff
@@ -117,14 +123,23 @@ pub struct CombinedSelectQuery {
...
-    SelectQuery {
-        projection: projection,
-        from: FromClause::TableList(TableList(cc.from)),
-        constraints: cc.wheres
-                       .into_iter()
-                       .map(|c| c.to_constraint())
-                       .collect(),
-        limit: limit.into(),
+    if cc.is_known_empty {
+        SelectQuery {
```
ncalexan: Ah, interesting -- further to my earlier comment, you do have the mechanism in place to statically represent a `NoneQuery`.  Consider lifting the semantic meaning of `FromClause::Nothing` up to the query level.  This would push your concrete `SelectQuery` to either be an enum, or to be a trait object in some form, and it might be that you don't want to do that for other reasons; I leave that to your discretion, since this will clearly work.

query-translator/tests/translate.rs
```diff
@@ -113,19 +141,40 @@ fn test_unknown_attribute_integer_value() {
...
+    let schema = Schema::default();
+
+    let impossible = r#"[:find ?x :where [?x :db/ident :no/exist]]"#;
+    let parsed = parse_find_string(impossible).expect("parse failed");
+    let algebrized = algebrize(&schema, parsed);
+
+    // This query cannot succeed: the ident doesn't resolve for a ref-typed attribute.
+    assert!(algebrized.cannot_succeed());
+
+    // If you insist…
```
ncalexan: Heh :)

query-sql/src/lib.rs
rnewman: It's theoretically possible: `SELECT 1 WHERE 1 < 2` is valid SQL. I'm not sure what you mean by 'simplify', though — can you elaborate?

query-translator/src/translate.rs
rnewman: It's not free to add new constraints to the query. I opted to file #363 instead — if we have a test that fails when we violate this assumption, then we should be good to go.

query-sql/src/lib.rs
ncalexan: If you know that `from` is `None`, you probably never need to process `constraints` or `projection`.  If you look later on, these are always trivial when `from` is `None`.  I'm suggesting you lift that "block representation" of `None, trivial, trivial` to `SelectQuery::None` or similar.

rnewman: I see what you mean. As I noted above: it's possible to have constraints and/or a projection when `from` is `None`. Indeed, in Datalog `ground` is `SELECT 123 AS x`, and `not` is effectively `SELECT 1 WHERE $condition`.

query-translator/src/translate.rs
rnewman: I'm working on this: it's possible for both SQL queries and Datalog queries to be known-empty, and that can indeed be represented in enums.

Similarly, it's possible for them to have known results — empty is just a special case of this — and it might be useful to flesh this out later. For example, `[:find ?x . :where [?x :db/ident :foo/bar]]` can at algebrizing time be completely computed without touching the database, if we so choose.

rnewman: I decided to push this up to the consuming point — that avoids us creating a projection list.

rnewman:
bf38105

PR #364, opened 2017-03-08
Convert EDN transaction tests to Rust code. Fixes #271.

### LINKED ISSUE #271: [tests] Use the EDN matcher for testing transaction output

ncalexan:
@victorporof I thought you might want to review a patch that depends so heavily on the work you did with the pattern matcher.  Thanks for it!


db/src/db.rs
```diff
@@ -899,135 +970,216 @@ mod tests {
...
+        // tempids in :db/retract that do upsert are retracted.  The ref given doesn't exist, so the
+        // assertion will be ignored.
+        let _report = conn.transact("[[:db/add \"t1\" :db/ident :name/Ivan]
+                                      [:db/retract \"t1\" :db.schema/attribute 103]]");
+        assert_matches!(conn.last_transaction(),
+                        "[[?tx :db/txInstant ?ms ?tx true]]");
+
+        // A multistep upsert.  The upsert algorithm will first try to resolve "t1", fail, and then
+        // allocate both "t1" and "t2".
+        let _report = conn.transact("[[:db/add \"t1\" :db/ident :name/Josef]
```
ncalexan: One thing I see and will fix throughout is that Rust thinks that `let _...` is a usage.  In this case, I forgot to `.unwrap()`; if there was no `let _...`, the compiler would tell me I forgot to use the result.

db/src/db.rs
```diff
@@ -899,135 +970,216 @@ mod tests {
...
-
-        assert_transactions(&conn, &mut db.partition_map, &mut db.schema, transactions);
+        let mut conn = TestConn::default();
+
+        // Test inserting :db.cardinality/one elements.
+        conn.transact("[[:db/add 100 :db/ident :keyword/value1]
+                        [:db/add 101 :db/ident :keyword/value2]]").unwrap();
+        assert_matches!(conn.last_transaction(),
+                        "[[100 :db/ident :keyword/value1 ?tx true]
+                          [101 :db/ident :keyword/value2 ?tx true]
```
rnewman: Does the matcher assert that `?tx` is the same throughout these patterns?

db/src/db.rs
```diff
@@ -899,135 +970,216 @@ mod tests {
...
-        assert_transactions(&conn, &mut db.partition_map, &mut db.schema, transactions);
+        let mut conn = TestConn::default();
+
+        // Insert a few :db.cardinality/one elements.
+        conn.transact("[[:db/add 100 :db/ident :keyword/value1]
+                        [:db/add 101 :db/ident :keyword/value2]]").unwrap();
+        assert_matches!(conn.last_transaction(),
+                        "[[100 :db/ident :keyword/value1 ?tx true]
+                          [101 :db/ident :keyword/value2 ?tx true]
+                          [?tx :db/txInstant ?ms ?tx true]]");
```
rnewman: Can you validate that the timestamp is expected? That is, if the `conn` exposes a `ts()` function, defined to be monotonically increasing (a simplification), then `?ms >= prev && ?ms <= now`?

db/src/db.rs
```diff
@@ -899,135 +970,216 @@ mod tests {
...
-
-        let transactions = value.as_vector().unwrap();
-        assert_transactions(&conn, &mut db.partition_map, &mut db.schema, transactions);
+        let mut conn = TestConn::default();
+
+        // Insert a few :db.cardinality/one elements.
+        conn.transact("[[:db/add 100 :db/ident :keyword/value1]
+                        [:db/add 101 :db/ident :keyword/value2]]").unwrap();
+        assert_matches!(conn.last_transaction(),
+                        "[[100 :db/ident :keyword/value1 ?tx true]
```
rnewman: Can you validate that the `?tx` is (a) numeric, (b) in the range of tx-ids, and (c) earlier than the current tx-id pointer in the connection metadata?

db/src/db.rs
ncalexan: Within _one_ pattern, yes.  Between multiple patterns, no.  This is by design.

db/src/db.rs
ncalexan: I'd like to do all of these tests as separate tests, like we do in the Clojure implementation: https://github.com/mozilla/mentat/blob/84a80f40f5c888f8452d07bd15f3b5fba49d3963/test/datomish/db_test.cljc#L426.

Filed https://github.com/mozilla/mentat/issues/368 to track this.  I'll land without it, but I'm happy to get to this fairly quickly, since it's important to get it right.

db/src/db.rs
rnewman: Sorry, by "pattern" I meant each `[100 … ?tx …]` — that is, this acts like the query engine does, and so this behaves something like

```clojure
(not-empty
  (q (last-transaction conn)
     "[:find ?tx . :in $last-tx :where [100 :db/ident :keyword/value1 ?tx] [101 :db/ident :keyword/value2 ?tx]]"))
```

ncalexan: Yes, I think your understanding is correct.  The pattern above would match
```edn
[[100 :db/ident :keyword/value1 12345 true]
 [100 :db/ident :keyword/value2 12345 true]]
```
but not
```edn
[[100 :db/ident :keyword/value1 12345 true]
 [100 :db/ident :keyword/value2 99999 true]]
```

PR #365, opened 2017-03-09
Collect tempids after upsert resolution. Fixes #299.

### LINKED ISSUE #299: [tx] Test collecting tempids after upsert resolution
In #283, I didn't quite finish the tempid collection implementation.  It's difficult to test without #294 since there aren't attributes with the correct form in the bootstrap database.

This ticket tracks actually testing the `expected-tempids` in files like https://github.com/mozilla/mentat/blob/16e9740d8a117c4b9384c4be262b0b64fe941d9f/tx/fixtures/test_upsert_vector.edn#L29.

This will get better after #271.

ncalexan:


ncalexan:
Built on top of #364.

db/src/tx.rs
```diff
@@ -263,8 +267,24 @@ impl<'conn, 'a> Tx<'conn, 'a> {
         // And evolve them forward.
         while generation.can_evolve() {
             // Evolve further.
-            let temp_id_map = self.resolve_temp_id_avs(&generation.temp_id_avs()[..])?;
+            let temp_id_map: TempIdMap = self.resolve_temp_id_avs(&generation.temp_id_avs()[..])?;
             generation = generation.evolve_one_step(&temp_id_map);
+
+            // Report each tempid that resolves via upsert.
+            for (tempid, &entid) in &temp_id_map {
```
rnewman: Why do we not take ownership of `temp_id_map`, and then take ownership of each `tempid` in the body? It's not used later.

db/src/tx.rs
```diff
@@ -263,8 +267,24 @@ impl<'conn, 'a> Tx<'conn, 'a> {
...
+                // keys of temp_id_map are unique between generations.Suppose that id->e and id->e*
+                // are two such mappings, resolved on subsequent evolutionary steps, and that `id`
+                // is a key in the intersection of the two key sets. This can't happen: if `id` maps
+                // to `e` via id->e, all instances of `id` have been evolved forward (replaced with
+                // `e`) before we try to resolve the next set of `UpsertsE`.  That is, we'll never
+                // successfully upsert the same tempid in more than one generation step.  (We might
+                // upsert the same tempid to multiple entids via distinct `[a v]` pairs in a single
+                // generation step; in this case, the transaction will fail.)
+                assert!(!tempids.contains_key(&**tempid));
+                tempids.insert((**tempid).clone(), entid);
```
rnewman: `insert` returns `None` if there was no existing value for that key. You can use that instead of the separate `contains_key` check:

```rust
let previous = tempids.insert(…);
assert!(previous.is_none());
```

db/src/tx.rs
```diff
@@ -242,16 +249,13 @@ impl<'conn, 'a> Tx<'conn, 'a> {
     // TODO: move this to the transactor layer.
     pub fn transact_entities<I>(&mut self, entities: I) -> Result<TxReport> where I: IntoIterator<Item=Entity> {
         // TODO: push these into an internal transaction report?
+        let mut tempids: BTreeMap<String, Entid> = BTreeMap::default();
```
rnewman: Move this allocation as low as it'll go; consider removing the comment above.

db/src/tx.rs
rnewman: Are you aware of the `Extend` trait?

```
use std::collections::BTreeMap;

fn main() {
    let mut m = BTreeMap::<String, i64>::default();
    let kvs = vec![("foo".to_string(), 32), ("bar".to_string(), 64)];
    m.extend(kvs.into_iter());
    println!("{:?}", m);
}
```
You don't get to verify that every key is non-present, but that's a tradeoff to explicitly consider.

db/src/tx.rs
ncalexan: In the final patch, it's used immediately after, so I think this is addressed.

db/src/tx.rs
ncalexan: I am aware of it, but it's not something I often use.  I think I'll keep the explicit check here.

db/src/tx.rs
ncalexan: I'll do this, but it's not the win I think you expect.  Each `tempid` is `Rc<String>`, so we can take ownership of the `Rc` but we can't avoid the `clone` below (safely), since we might have more consumers of any individual `Rc<String>` at this point.

db/src/tx.rs
ncalexan: Re: `insert` and `None` -- this is nice and I've done it.

PR #369, opened 2017-03-10
Support inputs in ConjoiningClauses

rnewman:
Note that VSCode is fine with comments in JSON files!

### DIFF
--- a/query-algebrizer/src/cc.rs
+++ b/query-algebrizer/src/cc.rs
@@ -18,7 +18,11 @@ use std::fmt::{
     Formatter,
     Result,
 };
-use std::collections::BTreeMap;
+use std::collections::{
+    BTreeMap,
+    BTreeSet,
+};
+
 use std::collections::btree_map::Entry;
 
 use self::mentat_core::{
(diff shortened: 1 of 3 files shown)


rnewman:
This doesn't have a consumer yet, but I expect you can imagine how it would work — when we have `:in` in the parsed query, and `:inputs` to `q_once`, we pass them in when creating the top-level CC.

(We will have kinda-nested CCs, so we might need borrows or `Rc` here eventually. This will do for now.)

query-algebrizer/src/cc.rs
```diff
@@ -242,30 +265,63 @@ impl Default for ConjoiningClauses {
...
+        cc.known_types
+          .extend(cc.known_bindings.iter()
+                    .map(|(k, v)| (k.clone(), v.value_type())));
+        cc
+    }
+}
+
+impl ConjoiningClauses {
+    fn known_binding(&self, var: &Variable) -> Option<TypedValue> {
+        self.known_bindings.get(var).cloned()
```
ncalexan: Huh, TIL about `cloned()`.

query-algebrizer/src/cc.rs
```diff
@@ -242,30 +265,63 @@ impl Default for ConjoiningClauses {
...
+            column_bindings: BTreeMap::new(),
+            known_bindings: BTreeMap::new(),
             known_types: BTreeMap::new(),
             extracted_types: BTreeMap::new(),
         }
     }
 }
 
 impl ConjoiningClauses {
+    fn with_bindings(bindings: BTreeMap<Variable, TypedValue>) -> ConjoiningClauses {
```
ncalexan: Consider `with_known_bindings` or `with_input_values`.  You just distinguished two types of bindings; let's not muddy the waters with a function name that doesn't distinguish.

query-algebrizer/src/cc.rs
```diff
@@ -209,14 +213,31 @@ pub struct ConjoiningClauses {
...
+    /// algebrized.
+    ///
+    /// Known bindings must agree with `known_types`. If you write a query like
+    /// ```edn
+    /// [:find ?x :in $ ?val :where [?x :foo/int ?val]]
+    /// ```
+    ///
+    /// and for `?val` provide `TypedValue::String("foo".to_string())`, the query will be known at
+    /// algebrizing time to be empty.
+    known_bindings: BTreeMap<Variable, TypedValue>,
```
ncalexan: You might consider calling this `input_values`.  Is there ever a time that `input_variables` will not be a superset of the keys of `known_bindings`?  (I may answer this question as I read.)

You might also consider `inputs: BTreeMap<Variable, Option<TypedValue>>` to unify where you track things.  Perhaps this doesn't make sense for data sources (i.e., for `$`).

query-algebrizer/src/cc.rs
```diff
@@ -917,4 +975,75 @@ mod testing {
...
+
+        cc.apply_pattern(&schema, &Pattern {
+            source: None,
+            entity: PatternNonValuePlace::Variable(x.clone()),
+            attribute: PatternNonValuePlace::Ident(NamespacedKeyword::new("foo", "bar")),
+            value: PatternValuePlace::Variable(y.clone()),
+            tx: PatternNonValuePlace::Placeholder,
+        });
+
+        // The type of the provided binding doesn't match the type of the attribute.
```
ncalexan: You had a mechanism, IIRC, for explaining _why_ the clauses are empty.  Can you assert something about this?  It's valuable to provide such type-checking to consumers.

query-algebrizer/src/cc.rs
rnewman: I think I agree with your naming suggestion. I might also opt for `value_bindings` (cf `column_bindings`).

From the perspective of these being used to bind inputs, yes, `input_variables` should be a superset. The cljs impl doesn't have an analogue of `input_variables`.

The only doubt I have is what to do with nested CCs (e.g., `not` clauses). For an 'inner' CC, it's not obvious what its inputs are.

In the cljs implementation we merge, as we descend, the outer CC's external bindings (mappings from var to expressions like `(sql/param :foo)`) with its column bindings to form the inner CC's external bindings. That allows us to cleanly merge on the way back up: we don't contribute external bindings to the parent.

I haven't got to that point in the Rust codebase yet.

query-algebrizer/src/cc.rs
rnewman: `with_value_bindings`.

query-algebrizer/src/cc.rs
rnewman: Yeah, `mark_known_empty`. I'll add a member to track a reason.

rnewman:
6109a63

Issue #295, opened 2017-02-14
[tx] Handle :db.alter/attribute in transactor


ncalexan:
Right now, the transactor doesn't allow to modify the schema. After #294 (which itself depends on #283), this will be ready for action. This ticket tracks:
- recognizing and acting on `:db.alter/attribute` assertions;
  This might include allowing certain "safe" data migrations.  I.e., converting numeric types; or adding uniqueness constraints; or allowing significant schema changes as long as there's no data in the store; or ...
- updating the relevant metadata structures;
- updating the idents and schema SQL materialized views;
- testing.

@rnewman implemented support for altering the schema in the Clojure implementation in #78 and filed follow-ups in #105.  We might just re-purpose at least #105!

ncalexan:
I see tests in the Clojure implementation around https://github.com/mozilla/mentat/blob/84a80f40f5c888f8452d07bd15f3b5fba49d3963/test/datomish/schema_changes_test.cljc#L212.

Issue #294, opened 2017-02-14
[tx] Handle :db.install/attribute in transactor


ncalexan:
Right now, the transactor doesn't allow to update (or modify) the schema.  After #283, this'll be ready for action.  This ticket tracks:
- recognizing and acting on `:db.install/attribute` assertions;
- updating the relevant metadata structures;
- updating the `idents` and `schema` SQL materialized views;
- testing.

ncalexan:
I'll be working on this next.

PR #370, opened 2017-03-10
Schema alteration.

ncalexan:
This is WIP towards #294 and #295 and the equivalent for `:db/ident`.  Not quite ready for review yet.


core/src/schema.rs
```diff
@@ -0,0 +1,112 @@
...
+        self.get_entid(x).map(|e| self.is_attribute(e)).unwrap_or(false)
+    }
+
+    /// Associate the given ident with given entid if neither are already recognized.
+    ///
+    /// Return `true` if an association was added, `false` otherwise.
+    pub fn associate_ident(&mut self, i: NamespacedKeyword, e: Entid) -> bool {
+        use std::collections::btree_map::Entry;
+        match (self.entid_map.entry(e), self.ident_map.entry(i.clone())) {
+            (Entry::Vacant(entid_entry), Entry::Vacant(ident_entry)) => {
```
rnewman: You can get ever so slightly better performance via:

```
if let Entry::Vacant(entid_entry) = self.entid_map.entry(e) {
    if let Entry::Vacant(…) = … {
        entid_entry.insert(i);
        ident_entry.insert(e);
        return true;
    }
}
false
```

— don't bother looking up the second if the first isn't vacant.

core/src/schema.rs
```diff
@@ -0,0 +1,112 @@
...
+            },
+            _ => false,
+        }
+    }
+
+    /// XXX
+    pub fn update_ident(&mut self, i: NamespacedKeyword, e: Entid) -> bool {
+        use std::collections::btree_map::Entry;
+        if let Entry::Occupied(mut entid_entry) = self.entid_map.entry(e) {
+            // TODO: avoid clone?
```
rnewman: To avoid the clone here you can use the return value of `remove` to decide whether to return false. You don't do anything with the second entry beyond making sure that it's occupied before removing it; `remove` returns `None` if there was nothing to remove.

db/src/metadata.rs
```diff
@@ -0,0 +1,379 @@
...
+
+impl MetadataReport {
+    pub fn is_empty(&self) -> bool {
+        self.installed.is_empty() && self.altered.is_empty()
+    }
+}
+
+lazy_static! {
+    pub static ref ATTRIBUTE_SET: String = {
+        // let mut stmt = self.prepare_cached("SELECT e, a, v, value_type_tag, added FROM transactions WHERE tx = ? AND a IN
```
rnewman: ?

db/src/metadata.rs
```diff
@@ -0,0 +1,379 @@
...
+                    TypedValue::Ref(entids::DB_CARDINALITY_MANY) => { builder.multival(true); },
+                    TypedValue::Ref(entids::DB_CARDINALITY_ONE) => { builder.multival(false); },
+                    _ => bail!(ErrorKind::BadSchemaAssertion(format!("Expected [... :db/cardinality :db.cardinality/many|:db.cardinality/one] but got [... :db/cardinality {:?}]", value)))
+                }
+            },
+
+            entids::DB_UNIQUE => {
+                match *value {
+                    // TODO: accept nil in some form.
+                    // TypedValue::Nil => {
```
rnewman: I'm not sure it should be a `TypedValue`. After all, we don't allow nulls in the DB, or in queries.

What's wrong with using `TypedValue::Boolean(false)` to mean neither?

Datomic uses retraction to model 'unset', and 'unset' means things: "You can also retract optional schema attributes." in http://docs.datomic.com/schema.html#altering-schema-attributes

db/src/metadata.rs
```diff
@@ -0,0 +1,379 @@
...
+                }
+            },
+
+            _ => {
+                bail!(ErrorKind::BadSchemaAssertion(format!("Do not recognize attribute '{}' for entid '{}'", attr, entid)))
+            }
+        }
+    };
+
+    let mut asserted: BTreeSet<Entid> = BTreeSet::default();
```
rnewman: You should also handle retraction here.

db/src/metadata.rs
```diff
@@ -0,0 +1,379 @@
...
+                if !builder.is_valid_install_attribute() {
+                    bail!(ErrorKind::BadSchemaAssertion(format!("Schema attribute for :db.install/attribute with entid '{}' does not set :db/valueType", entid)));
+                }
+                entry.insert(builder.build());
+                asserted.insert(entid);
+            },
+            Entry::Occupied(mut entry) => {
+                // Computing mutations before checking if we have a valid alter attribute looks odd,
+                // but during bootstrapping, we are given complete attributes (including
+                // :db/valueType) that are already in the schema.  This allows to ignore that case.
```
rnewman: allows us

db/src/metadata.rs
```diff
@@ -0,0 +1,379 @@
...
+        } else {
+            if let Some(asserted_value) = asserted.remove(&(e, a)) {
+                updated.insert((e, a), (typed_value, asserted_value));
+            } else {
+                retracted.insert((e, a), typed_value);
+            }
+        }
+    }
+
+    // Datomic does not allow to retract attributes or idents.  For now, Mentat follows suit.
```
rnewman: The schema doc suggests you can retract _something_ to unset things like uniqueness constraints.

db/src/schema.rs
```diff
@@ -52,6 +55,135 @@ fn validate_schema_map(entid_map: &EntidMap, schema_map: &SchemaMap) -> Result<(
...
+    multival: Option<bool>,
+    unique_value: Option<bool>,
+    unique_identity: Option<bool>,
+    index: Option<bool>,
+    fulltext: Option<bool>,
+    component: Option<bool>,
+}
+
+impl AttributeBuilder {
+    pub fn value_type<'a>(&'a mut self, value_type: ValueType) -> &'a mut Self {
```
rnewman: Why mutable references rather than ownership? Surely the model is

```
let foo = AttributeBuilder::new().value_type(ValueType::Boolean).build();
```

which can be modeled as

```
pub fn value_type(self, value_type: ValueType) -> Self {
```

no?

db/src/metadata.rs
```diff
@@ -0,0 +1,379 @@
...
+                entids::DB_INSTALL_VALUE_TYPE,
+                entids::DB_IS_COMPONENT,
+                entids::DB_UNIQUE,
+                entids::DB_VALUE_TYPE)
+    };
+}
+
+/// Return `false` if the given reduced entity will not change the metadata: recognized idents,
+/// schema, partitions in the partition map.
+pub fn might_update_metadata(reduced: &db::ReducedEntity) -> bool {
```
rnewman: I'd suggest marking this as `#[inline]`, see if it speeds up the transactor tight loop.

db/src/metadata.rs
```diff
@@ -0,0 +1,379 @@
...
+    match reduced.1 {
+        // Idents.
+        entids::DB_IDENT |
+        // Schema.
+        entids::DB_ALTER_ATTRIBUTE |
```
rnewman: Could you put these in numeric order, with the numeric value from `entids.rs` alongside?

I'd also be interested in an optimization here: hard-code the range in `entids.rs` (1–35, I think), and bail early if the value isn't in that range. No point painstakingly comparing `220` to 11 different small integers only to return `false`. The vast majority of assertions won't be in this range, so I'm keen to make this function ~free in the common case.

db/src/metadata.rs
```diff
@@ -0,0 +1,379 @@
...
+    // Group into asserted, retracted, and updated.  We assume all our attributes are
+    // :db/cardinality :db.cardinality/one (so they'll only be added or retracted at most once),
+    // and all modifications are simple changes.
+    let mut asserted: BTreeMap<(Entid, Entid), TypedValue> = BTreeMap::new();
+    let mut retracted: BTreeMap<(Entid, Entid), TypedValue> = BTreeMap::new();
+    let mut updated: BTreeMap<(Entid, Entid), (TypedValue, TypedValue)> = BTreeMap::new();
+
+    for (e, a, typed_value, added) in assertions.into_iter() {
+        let attribute = schema.require_attribute_for_entid(a)?;
+        // XXX TODO NOTE
```
rnewman: ?

db/src/tx.rs
```diff
@@ -299,10 +341,15 @@ impl<'conn, 'a> Tx<'conn, 'a> {
                     }
 
                     let added = op == OpType::Add;
+                    let reduced = (e, a, attribute, v, added);
+                    if metadata::might_update_metadata(&reduced) {
```
rnewman: This function doesn't use most of the arguments. Can you just pass `e` by value instead?

https://medium.com/@robertgrosse/how-copying-an-int-made-my-code-11-times-faster-f76c66312e0f#.m3xlziaht

db/src/metadata.rs
ncalexan: I'd be happy to optimize this eventually.  I believe that Rust implements this via a jump table in some form, so it's ~free already.

db/src/metadata.rs
ncalexan: I wasn't precise: by "some form" I was thinking perhaps `:db.unique/nil` or `:db/notUnique`, for precisely the reasons you note about `nil`.  I was not aware you could retract to unset; my current patch doesn't allow that.  I'll consider allowing it once I work out some of the details.

db/src/schema.rs
ncalexan: This is not being consumed in a fluent interface (like `Builder::new().thing1().thing2().build()`), but in a loop that grabs the builder reference, does a thing, and then finally builds everything at the end.  It's equivalent to consume `self` and update the `&mut Builder` in the map entry, I guess.

ncalexan:
@rnewman OK, this is ready for review.  This shares very little code with what you saw before; I hope I captured the spirit (or explicitly addressed) your major concerns, but I apologize if I'm missing stuff in the churn.  I've been working on this too long without landing pieces, for which I apologise.  Turned out to be a complex thing to get right.

db/src/debug.rs
```diff
@@ -147,13 +155,17 @@ pub fn transactions_after<S: Borrow<Schema>>(conn: &rusqlite::Connection, schema
         let v: rusqlite::types::Value = row.get_checked(2)?;
         let value_type_tag: i32 = row.get_checked(3)?;
 
-        let typed_value = TypedValue::from_sql_value_pair(v, value_type_tag)?;
+        let typed_value = match TypedValue::from_sql_value_pair(v, value_type_tag)? {
+            TypedValue::Ref(e) => {
+                borrowed_schema.get_ident(e).cloned().map(TypedValue::Keyword).unwrap_or(TypedValue::Ref(e))
```
rnewman: You do this twice, so why not add a trait to `TypedValue` that ident-maps with a provided schema?

```rust
trait ToIdent {
  fn map_ident(self, schema: &Schema) -> Self;
}
```

db/src/diff_set.rs
```diff
@@ -0,0 +1,50 @@
...
+    pub fn witness(&mut self, key: K, value: V, added: bool) {
+        if added {
+            if let Some(retracted_value) = self.retracted.remove(&key) {
+                self.altered.insert(key, (retracted_value, value));
+            } else {
+                self.asserted.insert(key, value);
+            }
+        } else {
+            if let Some(asserted_value) = self.asserted.remove(&key) {
+                self.altered.insert(key, (value, asserted_value));
```
rnewman: I don't understand this. If there was an asserted value, and we get a removal, store an alteration for the original value instead?

rnewman: Oh, I see. If you get the addition before the retraction, retract the old value and assert the new one.

db/src/diff_set.rs
```diff
@@ -0,0 +1,50 @@
...
+// specific language governing permissions and limitations under the License.
+
+#![allow(dead_code)]
+
+use std::collections::BTreeMap;
+
+/// Witness assertions and retractions, folding (assertion, retraction) pairs into alterations.
+/// Assumes that no assertion or retraction will be witnessed more than once.
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialOrd, PartialEq)]
+pub struct DiffSet<K, V> {
```
rnewman: This looks like a much more general data structure than it actually is! I was having trouble reasoning about what happens in some orderings, what happens if there is or isn't assumed to be data outside… but all it really does is filter out and categorize pairs, so given `[assert(2), retract(1)]` you don't get `[]`, you get `[alter(1, 2)]`.

rnewman: Could you add a little more commentary and/or a test to explain what happens for various inputs?

db/src/db.rs
```diff
@@ -870,6 +889,90 @@ pub fn update_partition_map(conn: &rusqlite::Connection, partition_map: &Partiti
...
+        conn.execute(format!("DELETE FROM idents").as_str(),
+                     &[])?;
+        // Take the last, as determined by tx, added [e :db/ident v _ true].  This will pick up
+        // :db/idents that have been retracted and/or re-purposed.
+        conn.execute(format!("INSERT INTO idents SELECT e, a, v, value_type_tag FROM (SELECT e, a, v, value_type_tag, MAX(tx) FROM transactions WHERE a IN {} AND added = 1 GROUP BY e, a, v, value_type_tag, tx)", entids::IDENTS_SET.as_str()).as_str(),
+                     &[])?;
+    }
+
+    for &entid in &metadata_report.attributes_installed {
+        conn.execute(format!("INSERT INTO schema SELECT e, a, v, value_type_tag FROM datoms WHERE e = ? AND a IN {}", entids::SCHEMA_SET.as_str()).as_str(),
```
rnewman: Prepared statements inside loops, please!

rnewman:
Sorry, in that comment I reversed the directions; I think you get my gist, though.

With historical idents stored in the schema, how do we recognize the current mapping?

ncalexan:
> I'm not entirely sure I understand your division of entids vs idents — I would expect both idents and entids to be scoped to current data, and the for lookup in either direction to be a partial injective function. As I understand that part, you've made keyword to entid partial and injective, but entid to keyword partial and multivalued Am I misreading that commit message?

I think you are.  I think the following example will be instructive, and I'll fold it into a comment or commit message once we have shared understanding.  Suppose I transact a single ident:
```edn
[[:db/add 100 :db/ident :test/ident]]
```
Then both the "entids" and "idents" materialized views will look (symbolically) like:
```edn
[[100 :db/ident :test/ident]]
```
If I now transact the following datoms:
```edn
[[:db/retract 100 :db/ident :test/ident]]
 [:db/add 100 :db/ident :test/updated]]
```
the "entids" materialized view will look like:
```edn
[[100 :db/ident :test/updated]]
```
but the "idents" materialized view will maintain the historical values, like:
```edn
[[100 :db/ident :test/ident]
 [100 :db/ident :test/updated]]
```

In terms of the resulting maps, `entid_map` will be `{100 :test/ident}` and `ident_map` will be `{:test/ident 100, :test/updated 100}`.  Since there's a correspondence between relations and functions, it's possible to talk about injective/bijective/multivalued, but I don't think it's helpful in this case.  It's better to think about our use cases:
- if we ever want to symbolicate, `entid_map` gives the _current association_ from entid to ident;
- if we ever want to concretize, `ident_map` gives the _current and historical associations_ from ident to entid.

This is what Datomic does, and I think it's the right call.  It means that when the `:db/ident` associated to an entid changes, existing consumers can still refer to the entid with an outdated historical ident.  And you can always re-purpose an ident if you _really_ want to, breaking existing consumers that haven't updated.

Is that more clear?

db/src/db.rs
ncalexan: In general, I'm for cached statements (especially since Rusqlite makes them easy to use) but I intentionally didn't optimize this process.  I expect it to be infrequent ... oh, prepared vs. cached.  Prepared I will do!  This is not efficient at this time because walking `transactions` in this way is very expensive but it's easy to improve and I'll file follow-up to track that.

db/src/diff_set.rs
ncalexan: Correct.

db/src/diff_set.rs
ncalexan: Correct.  I will certainly elaborate and add a few tests.  It was just a thing I needed twice -- for attributes and for idents.  Feel free to suggest another name, if you prefer.

rnewman:
> Is that more clear?

I think so. I think I have two points of friction:

1. I don't really get the names: ident and entid are each end of the same bidirectional association, and I'd expect to be able to work with current or historical versions of each — that is, if I've retracted `[123 :db/ident :foo/bar]`, I want to be able to have `schema.ident(123) == None`.
2. I'm worried about the round-trip behavior. Depending on which maps are used where, you could end up with an input keyword being rewritten to be more modern, or worse — imagine a query like `[:find ?x :where [?x _ :foo/bar]]`, which could incorrectly match against an entid that _no longer_ is identified by that keyword. You'll get the wrong results!

db/src/diff_set.rs
rnewman: `AddRetractAlterSet`?

ncalexan:
@rnewman sorry about the force-push.  This gets us to the following place:
- no historical idents (so `entid_map` and `ident_map` are inverse to each other)
- "idents" and "schema" materialized views are as they were in Datomish
- no `:db.{install,alter}/partition` at all (see commit message for motivation)
- `[:db/retract e :db/ident ...]` supported
- `[:db/retract :attribute :db/* ...]` of schema characteristics not yet supported
- `[:db/retractEntity :attribute]` not yet supported

I'll file follow-up for the last two if I don't do them as part of the _next_ review cycle.  Sorry for all this churn!

core/src/lib.rs
```diff
@@ -139,6 +138,12 @@ pub enum AttributeBitFlags {
     UniqueValue   = 1 << 3,
 }
 
+#[derive(Clone,Debug,Eq,Hash,Ord,PartialOrd,PartialEq)]
+pub enum Unique {
```
rnewman: Can we call this `AttributeUniqueness` or something more specific? This is in `core`.

rnewman: Or a submodule for scoping? `attribute::Unique::Value`?

db/src/db.rs
```diff
@@ -209,18 +214,21 @@ pub fn create_current_version(conn: &mut rusqlite::Connection) -> Result<DB> {
 
     // TODO: return to transact_internal to self-manage the encompassing SQLite transaction.
     let bootstrap_schema = bootstrap::bootstrap_schema();
-    let (_report, next_partition_map, next_schema) = transact(&tx, bootstrap_partition_map, &bootstrap_schema, bootstrap::bootstrap_entities())?;
-    if next_schema.is_some() {
-        // TODO Use custom ErrorKind https://github.com/brson/error-chain/issues/117
-        bail!(ErrorKind::NotYetImplemented(format!("Initial bootstrap transaction did not produce expected bootstrap schema")));
+    let bootstrap_schema_for_mutation = Schema::default(); // The bootstrap transaction will populate this schema.
```
rnewman: :beer:

db/src/db.rs
```diff
@@ -401,56 +409,60 @@ impl TypedSQLValue for TypedValue {
...
+        }
+    }).collect()
+}
 
-    r.and_then(|triples| Schema::from_ident_map_and_triples(ident_map.clone(), triples))
+/// Read the schema materialized view from the given SQL store.
+fn read_schema_map(conn: &rusqlite::Connection) -> Result<SchemaMap> {
+    let entid_triples = read_materialized_view(conn, "schema")?;
+    let mut schema_map = SchemaMap::default();
+    metadata::update_schema_map_from_entid_triples(&mut schema_map, entid_triples)?;
```
rnewman: Can you push this into a method on `SchemaMap`? `SchemaMap::with_entid_triples`?

db/src/db.rs
```diff
@@ -827,6 +855,87 @@ pub fn update_partition_map(conn: &rusqlite::Connection, partition_map: &Partiti
...
+    // TODO: consider doing this in fewer SQLite execute() invocations.
+    // TODO: use concat! to avoid creating String instances.
+    if !metadata_report.idents_altered.is_empty() {
+        // Idents is the materialized view of the [entid :db/ident ident _ added] slice of
+        // transactions.  This keeps outdated idents around, just like Datomic does.
+        conn.execute(format!("DELETE FROM idents").as_str(),
+                     &[])?;
+        // Take the last, as determined by tx, added [e :db/ident v _ true].  This will pick up
+        // :db/idents that have been retracted and/or re-purposed.
+        conn.execute(format!("INSERT INTO idents SELECT e, a, v, value_type_tag FROM datoms WHERE a IN {}", entids::IDENTS_SET.as_str()).as_str(),
```
rnewman: Careful: I suspect that this is implemented as a table walk with a filter, at least for the case of non-unit sets, which is startlingly expensive. You probably want to do this differently in the short or medium term.

db/src/db.rs
```diff
@@ -827,6 +855,87 @@ pub fn update_partition_map(conn: &rusqlite::Connection, partition_map: &Partiti
...
+{
+    use metadata::AttributeAlteration::*;
+
+    // Populate the materialized view directly from datoms (and, potentially in the future,
+    // transactions).  This might generalize nicely as we expand the set of materialized views.
+    // TODO: consider doing this in fewer SQLite execute() invocations.
+    // TODO: use concat! to avoid creating String instances.
+    if !metadata_report.idents_altered.is_empty() {
+        // Idents is the materialized view of the [entid :db/ident ident _ added] slice of
+        // transactions.  This keeps outdated idents around, just like Datomic does.
```
rnewman: I don't think this comment is still accurate.

db/src/db.rs
rnewman: Oh, and filtering on `value_type_tag` to `keyword` is something you might consider.

db/src/db.rs
```diff
@@ -827,6 +855,87 @@ pub fn update_partition_map(conn: &rusqlite::Connection, partition_map: &Partiti
...
+    let mut stmt = conn.prepare(format!("INSERT INTO schema SELECT e, a, v, value_type_tag FROM datoms WHERE e = ? AND a IN {}", entids::SCHEMA_SET.as_str()).as_str())?;
+    for &entid in &metadata_report.attributes_installed {
+        stmt.execute(&[&entid as &ToSql])?;
+    }
+
+    let mut delete_stmt = conn.prepare(format!("DELETE FROM schema WHERE e = ? AND a IN {}", entids::SCHEMA_SET.as_str()).as_str())?;
+    let mut insert_stmt = conn.prepare(format!("INSERT INTO schema SELECT e, a, v, value_type_tag FROM datoms WHERE e = ? AND a IN {}", entids::SCHEMA_SET.as_str()).as_str())?;
+    let mut index_stmt = conn.prepare("UPDATE datoms SET index_avet = ? WHERE a = ?")?;
+    let mut unique_value_stmt = conn.prepare("UPDATE datoms SET unique_value = ? WHERE a = ?")?;
+    let mut cardinality_stmt = conn.prepare("SELECT e FROM datoms WHERE a = ? GROUP BY e HAVING COUNT(*) > 1 LIMIT 1")?;
```
rnewman: I might rephrase this to something potentially cheaper:

```
SELECT EXISTS (
  SELECT 1 FROM datoms AS one, datoms AS two
  WHERE
        one.a = ?1 AND two.a = ?1 AND
        one.e = two.e AND
        one.v <> two.v
  )
  AS dupes;
```

db/src/entids.rs
```diff
@@ -56,3 +56,58 @@ pub const DB_DOC: Entid = 35;
...
+
+lazy_static! {
+    /// Attributes that are "ident related".  These might change the "idents" materialized view.
+    pub static ref IDENTS_SET: String = {
+        format!("({})",
+                DB_IDENT)
+    };
+
+    /// Attributes that are "schema related".  These might change the "schema" materialized view.
+    pub static ref SCHEMA_SET: String = {
```
rnewman: Yikes. Call out that these are SQL lists, not Rust sets!

ncalexan: s/_SET/_SQL_LIST/.

db/src/db.rs
ncalexan: Sure, thanks for the suggestion.

db/src/db.rs
ncalexan: I could, but I want the mutation form so it's `&mut self` in general.  There's a symmetry in having the two updating functions close together (one operating on `SchemaMap`, the other on `Schema`) so I'm going to keep them in `metadata.rs` until we have more changes to make.

db/src/db.rs
ncalexan: I agree!

db/src/db.rs
ncalexan: I expected this to be efficient because there's an index on `e a v value_type_tag`.  Do you think it would be faster to have a bunch of `WHERE a = ? OR a = ?` clauses?  A little playing with SQLite suggests the query plans are the same:
```
sqlite> EXPLAIN QUERY PLAN INSERT INTO idents SELECT e, a FROM datoms WHERE a = 0 OR a = 1;
0|0|0|SEARCH TABLE datoms USING COVERING INDEX idx_datoms_aevt (a=?)
0|0|0|EXECUTE LIST SUBQUERY 1
sqlite> EXPLAIN QUERY PLAN INSERT INTO idents SELECT e, a FROM datoms WHERE a IN (0, 1);
0|0|0|SEARCH TABLE datoms USING COVERING INDEX idx_datoms_aevt (a=?)
0|0|0|EXECUTE LIST SUBQUERY 1
```
Feel free to file follow-up if you're not convinced.

core/src/lib.rs
ncalexan: Thanks for correcting me on this; I should have known better.  I've pushed it into an `attribute` submodule.

db/src/db.rs
rnewman: Query plan trumps vague suspicions!

PR #372, opened 2017-03-13
Ensure that variable bindings are used when selecting a table.

rnewman:
This PR ensures that a query like

```
[:find ?x :where [?x ?a ?v]]
```

with a binding for `?a`, is algebrized with knowledge of `?a`.

### DIFF
--- a/query-algebrizer/src/cc.rs
+++ b/query-algebrizer/src/cc.rs
@@ -176,6 +176,19 @@ impl Debug for ColumnConstraint {
     }
 }
 
+trait OptionEffect<T> {
+    fn when_not<F: FnOnce()>(self, f: F) -> Option<T>;
+}
+
+impl<T> OptionEffect<T> for Option<T> {
+    fn when_not<F: FnOnce()>(self, f: F) -> Option<T> {
+        if self.is_none() {
+            f();
+        }
+        self
+    }
+}
+
 /// A `ConjoiningClauses` (CC) is a collection of clauses that are combined with `JOIN`.
 /// The topmost form in a query is a `ConjoiningClauses`.
 ///
@@ -249,6 +262,7 @@ pub enum EmptyBecause {
     UnresolvedIdent(NamespacedKeyword),
     InvalidAttributeIdent(NamespacedKeyword),
     InvalidAttributeEntid(Entid),
+    InvalidBinding(DatomsColumn, TypedValue),
     ValueTypeMismatch(ValueType, TypedValue),
     AttributeLookupFailed,         // Catch-all, because the table lookup code is lazy. TODO
 }
@@ -269,6 +283,9 @@ impl Debug for EmptyBecause {
             },
             &InvalidAttributeEntid(entid) => {
                 write!(f, "{} is not an attribute", entid)
+            },
+            &InvalidBinding(ref column, ref tv) => {
+                write!(f, "{:?} cannot name column {:?}", tv, column)
             },
             &ValueTypeMismatch(value_type, ref typed_value) => {
                 write!(f, "Type mismatch: {:?} doesn't match attribute type {:?}",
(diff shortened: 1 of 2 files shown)


query-algebrizer/src/cc.rs
```diff
@@ -334,10 +334,25 @@ impl ConjoiningClauses {
         self.value_bindings.get(var).cloned()
     }
 
-    pub fn bind_column_to_var(&mut self, table: TableAlias, column: DatomsColumn, var: Variable) {
+    pub fn bind_column_to_var(&mut self, schema: &Schema, table: TableAlias, column: DatomsColumn, var: Variable) {
```
ncalexan: It's awkward to thread the schema-for-interpretation around piece-meal in this way.  Consider either pushing the `&Schema` reference into `CC`, or adding a `CCBuilder` to keep it around.  This could be a future clean-up/unification if you so choose.

query-algebrizer/src/cc.rs
```diff
@@ -334,10 +334,25 @@ impl ConjoiningClauses {
...
+            if column != DatomsColumn::Value {
+                if let TypedValue::Keyword(ref kw) = bound_val {
+                    if let Some(entid) = self.entid_for_ident(schema, kw) {
+                        self.constrain_column_to_entity(table, column, entid);
+                    } else {
+                        // Impossible.
+                        self.mark_known_empty(EmptyBecause::UnresolvedIdent(kw.clone()));
+                    }
+                    return;
+                }
```
ncalexan: Shouldn't there be some additional `mark_known_empty` based on non-`Keyword` bindings?  That is, if I give `[$a ...]` and `$a` is bound to `"string value"`, we can reject immediately.  Or does this happen elsewhere as well?

query-algebrizer/src/cc.rs
```diff
@@ -943,6 +979,124 @@ mod testing {
...
+        let mut cc = ConjoiningClauses::default();
+        let mut schema = Schema::default();
+        associate_ident(&mut schema, NamespacedKeyword::new("foo", "bar"), 99);
+        add_attribute(&mut schema, 99, Attribute {
+            value_type: ValueType::Boolean,
+            ..Default::default()
+        });
+
+        let x = Variable(PlainSymbol::new("?x"));
+        let a = Variable(PlainSymbol::new("?a"));
```
ncalexan: Don't Datomic's Datalog queries generally use `$a` for input values?

query-algebrizer/src/cc.rs
```diff
@@ -943,6 +979,124 @@ mod testing {
...
+        assert_eq!(cc.from, vec![SourceAlias(DatomsTable::Datoms, "datoms00".to_string())]);
+
+        // ?x must be a ref.
+        assert_eq!(cc.known_types.get(&x).unwrap(), &ValueType::Ref);
+
+        // ?x is bound to datoms0.e.
+        assert_eq!(cc.column_bindings.get(&x).unwrap(), &vec![d0_e.clone()]);
+        assert_eq!(cc.wheres, vec![
+                   ColumnConstraint::EqualsEntity(d0_a, 99),
+        ]);
```
ncalexan: Shouldn't you be asserting that `v` is boolean as well?

query-algebrizer/src/cc.rs
rnewman: `$` indicates a source variable — a point-in-time DB, the transaction log, etc.

You can supply both as inputs (though in our strongly-typed world, we'll do so through two different collections).

query-algebrizer/src/cc.rs
rnewman: No. Remember, this knife cuts both ways: now that we've bound `?a` to `:foo/bar`, which we know has a range of `boolean`, any value in `v` with that `a` must be a boolean. `1` can never mean anything but `true` if `a=99`.

This is the main reason we don't have a forest of type tag checks and tag projections in every query: whenever we know the attribute, we can safely and statically assume the type of each value.

query-algebrizer/src/cc.rs
rnewman: I reworked some of this.

rnewman:
70e5759

Issue #373, opened 2017-03-15
[query] Constraint-based algebrizing


rnewman:
The current algebrizer — mainly `ConjoiningClauses` — is deliberately _procedural_ and _predictable_. Paired with the simple query translator, we quite directly translate Datalog into SQL.

CC performs two roles: it tracks the metadata it needs to not be completely clumsy (the types deduced for variables and known bindings), and it accumulates some information that should really be part of a _planner_: table allocations and mappings between their columns and variables.

Again, that works fine for simple and predictable translation.

Extending this approach, however, to attain more power gets… hacky.

- Query simplification/optimization (_e.g._, removal of redundant patterns and constraints).
- Algebrizing queries down to static results (_e.g._, `[:find ?x :where [?x :db/ident :foo/bar]]`).
- Value space intersection and optimization (_e.g._, in `[:find ?x :where [?x :foo/age ?z][?y :foo/age ?o] [(!= ?x ?y)] [(> ?z 20)] [(< ?o 30)] [(> ?o ?z)]]` we can derive a lower bound for `?o`).
- Deducing attributes from their ranges (_e.g._, `[:find ?x :where [?x ?y :foo/bar]]` can deduce the value of `?y` in the current database from the schema alone, and `[:find ?x :where [?x ?y true] [?z ?y ?z]]` can never bind a single `?y`).

At some point it might be worthwhile to more fully separate algebrize+translate into algebrize+plan+translate, with the algebrizing stage being primarily a constraint system used to derive something like CC.

The risk of doing so is that the planner input would be less ordered, and thus the translated output would likely be less predictable and less related to the input. Small changes to the input query could cause large changes to what we can deduce.

rnewman:
@fluffyemily this is what we were chatting about yesterday.

ncalexan:
I'm not _against_ doing this, especially in the far future after we learn some lessons about what really happens in the wild, but it's worth pointing out what the fundamental design trade-off encoded into our conversion from Datalog to SQL is.  In my mind, we have always been betting that SQLite's planning (re-order clauses, remove redundant constraints, choose indices efficiently, etc) would be "better" than our own: possibly more efficient, definitely quicker to market.

rnewman:
Yes, this is far future (unless predicates become troublesome for me!). But there are three caveats.

[...]

PR #374, opened 2017-03-16
Parse, algebrize, and translate basic numeric predicate patterns

rnewman:
This PR starts to split up `cc.rs` a little, as well as doing some cleanup. (I left in some reformatting that VSCode did on its own, mostly in their own commits.)

I then added a test for some things that CC isn't smart enough to do yet (see #373), and finally moved on to the meat of the PR: extending the parser to handle predicates.

Right now `edn::Value::List` isn't 'spreadable' into a slice, which our parsing approach relies on. That means we don't yet support the syntax you usually see for predicates:

```
[(< ?x 10)]
```

Instead we partly follow the Datomic query grammar (!!):

```
[] = list or vector
…
pred-expr = [ [pred fn-arg+] ]
```

and support only vectors: `[[< ?x 10]]` is valid, and that's what we handle. I'll expand that next.

### DIFF
--- a/query-algebrizer/src/cc.rs
+++ b/query-algebrizer/src/cc.rs
@@ -14,8 +14,8 @@ extern crate mentat_query;
 use std::fmt::{
     Debug,
     Formatter,
-    Result,
 };
+
 use std::collections::{
     BTreeMap,
     BTreeSet,
@@ -37,6 +37,12 @@ use self::mentat_query::{
     PatternValuePlace,
     SrcVar,
     Variable,
+};
+
+use errors::{
+    Error,
+    ErrorKind,
+    Result,
 };
 
 use types::{
@@ -192,7 +198,7 @@ pub enum EmptyBecause {
 }
 
 impl Debug for EmptyBecause {
-    fn fmt(&self, f: &mut Formatter) -> Result {
+    fn fmt(&self, f: &mut Formatter) -> ::std::fmt::Result {
         use self::EmptyBecause::*;
         match self {
             &TypeMismatch(ref var, ref existing, ref desired) => {
@@ -223,7 +229,7 @@ impl Debug for EmptyBecause {
 }
 
 impl Debug for ConjoiningClauses {
-    fn fmt(&self, fmt: &mut Formatter) -> Result {
+    fn fmt(&self, fmt: &mut Formatter) -> ::std::fmt::Result {
         fmt.debug_struct("ConjoiningClauses")
             .field("is_known_empty", &self.is_known_empty)
             .field("from", &self.from)
--- a/query-algebrizer/src/cc.rs
+++ b/query-algebrizer/src/cc.rs
@@ -19,7 +19,10 @@ use std::fmt::{
 use std::collections::{
     BTreeMap,
     BTreeSet,
+    HashSet,
 };
+
+use std::collections::btree_map::Entry;
 
 use self::mentat_core::{
     Attribute,
@@ -36,6 +39,7 @@ use self::mentat_query::{
     Pattern,
     PatternNonValuePlace,
     PatternValuePlace,
+    PlainSymbol,
     Predicate,
     SrcVar,
     Variable,
@@ -81,6 +85,12 @@ impl<T> OptionEffect<T> for Option<T> {
         }
         self
     }
+}
+
+fn unit_type_set(t: ValueType) -> HashSet<ValueType> {
+    let mut s = HashSet::with_capacity(1);
+    s.insert(t);
+    s
 }
 
 /// A `ConjoiningClauses` (CC) is a collection of clauses that are combined with `JOIN`.
@@ -143,7 +153,7 @@ pub struct ConjoiningClauses {
     /// A map from var to type. Whenever a var maps unambiguously to two different types, it cannot
     /// yield results, so we don't represent that case here. If a var isn't present in the map, it
     /// means that its type is not known in advance.
-    pub known_types: BTreeMap<Variable, ValueType>,
+    pub known_types: BTreeMap<Variable, HashSet<ValueType>>,
 
     /// A mapping, similar to `column_bindings`, but used to pull type tags out of the store at runtime.
     /// If a var isn't present in `known_types`, it should be present here.
@@ -153,7 +163,8 @@ pub struct ConjoiningClauses {
 #[derive(PartialEq)]
 pub enum EmptyBecause {
     // Var, existing, desired.
-    TypeMismatch(Variable, ValueType, ValueType),
+    TypeMismatch(Variable, HashSet<ValueType>, ValueType),
+    NonNumericArgument,
     UnresolvedIdent(NamespacedKeyword),
     InvalidAttributeIdent(NamespacedKeyword),
     InvalidAttributeEntid(Entid),
@@ -169,6 +180,9 @@ impl Debug for EmptyBecause {
             &TypeMismatch(ref var, ref existing, ref desired) => {
                 write!(f, "Type mismatch: {:?} can't be {:?}, because it's already {:?}",
                        var, desired, existing)
+            },
+            &NonNumericArgument => {
+                write!(f, "Non-numeric argument in numeric place")
             },
             &UnresolvedIdent(ref kw) => {
                 write!(f, "Couldn't resolve keyword {}", kw)
@@ -237,7 +251,7 @@ impl ConjoiningClauses {
         // Pre-fill our type mappings with the types of the input bindings.
         cc.known_types
           .extend(cc.value_bindings.iter()
-                    .map(|(k, v)| (k.clone(), v.value_type())));
+                    .map(|(k, v)| (k.clone(), unit_type_set(v.value_type()))));
         cc
     }
 }
@@ -245,6 +259,16 @@ impl ConjoiningClauses {
 impl ConjoiningClauses {
     fn bound_value(&self, var: &Variable) -> Option<TypedValue> {
         self.value_bindings.get(var).cloned()
+    }
+
+    /// Return a single `ValueType` if the given variable is known to have a precise type.
+    /// Returns `None` if the type of the variable is known but not precise -- "double
+    /// or integer" isn't good enough.
+    pub fn known_type(&self, var: &Variable) -> Option<ValueType> {
+        match self.known_types.get(var) {
+            Some(types) if types.len() == 1 => types.iter().next().cloned(),
+            _ => None,
+        }
     }
 
     pub fn bind_column_to_var(&mut self, schema: &Schema, table: TableAlias, column: DatomsColumn, var: Variable) {
@@ -291,7 +315,7 @@ impl ConjoiningClauses {
         let needs_type_extraction =
             !late_binding &&                           // Never need to extract for bound vars.
             column == DatomsColumn::Value &&           // Never need to extract types for refs.
-            !self.known_types.contains_key(&var) &&    // We know the type!
+            self.known_type(&var).is_none() &&         // Don't need to extract if we know a single type.
             !self.extracted_types.contains_key(&var);  // We're already extracting the type.
 
         let alias = QualifiedAlias(table, column);
@@ -322,6 +346,36 @@ impl ConjoiningClauses {
             QueryValue::PrimitiveLong(value)))
     }
 
+    /// Mark the given value as one of the set of numeric types.
+    fn constrain_var_to_numeric(&mut self, variable: Variable) {
+        let mut numeric_types = HashSet::with_capacity(2);
+        numeric_types.insert(ValueType::Double);
+        numeric_types.insert(ValueType::Long);
+
+        let entry = self.known_types.entry(variable);
+        match entry {
+            Entry::Vacant(vacant) => {
+                vacant.insert(numeric_types);
+            },
+            Entry::Occupied(mut occupied) => {
+                let narrowed: HashSet<ValueType> = numeric_types.intersection(occupied.get()).cloned().collect();
+                match narrowed.len() {
+                    0 => {
+                        // TODO: can't borrow as mutable more than once!
+                        //self.mark_known_empty(EmptyBecause::TypeMismatch(occupied.key().clone(), occupied.get().clone(), ValueType::Double));   // I know…
+                    },
+                    1 => {
+                        // Hooray!
+                        self.extracted_types.remove(occupied.key());
+                    },
+                    _ => {
+                    },
+                };
+                occupied.insert(narrowed);
+            },
+        }
+    }
+
     /// Constrains the var if there's no existing type.
     /// Marks as known-empty if it's impossible for this type to apply because there's a conflicting
     /// type already known.
@@ -343,13 +397,17 @@ impl ConjoiningClauses {
         // spot it here.
         if let Some(existing) = self.known_types.get(&variable).cloned() {
             // If so, the types must match.
-            if existing != this_type {
+            if !existing.contains(&this_type) {
                 self.mark_known_empty(EmptyBecause::TypeMismatch(variable, existing, this_type));
+            } else {
+                if existing.len() > 1 {
+                    // Narrow.
+                    self.known_types.insert(variable, unit_type_set(this_type));
+                }
             }
         } else {
             // If not, record the one we just determined.
-            self.known_types.insert(variable, this_type);
-
+            self.known_types.insert(variable, unit_type_set(this_type));
         }
     }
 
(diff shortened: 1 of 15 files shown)


ncalexan:
> Right now edn::Value::List isn't 'spreadable' into a slice, which our parsing approach relies on. That means we don't yet support the syntax you usually see for predicates:

This might be an argument for backing `edn::Value::List` with a contiguous data structure.  It would not be hard to `impl Parser<T> for LinkedList<T>`, but I doubt it's worth it.

query-algebrizer/src/cc.rs
```diff
@@ -1235,4 +1235,114 @@ mod testing {
         // The type of the provided binding doesn't match the type of the attribute.
         assert!(cc.is_known_empty);
     }
+
+    #[test]
+    fn test_fulltext_type_disagreement() {
```
ncalexan: Let's start `/// Noting what the test verifies`.  I'm guilty of not doing this as well.

query-algebrizer/src/cc.rs
```diff
@@ -1235,4 +1235,114 @@ mod testing {
...
+            value: PatternValuePlace::Variable(y.clone()),
+            tx: PatternNonValuePlace::Placeholder,
+        });
+
+        // Finally, expand column bindings to get the overlaps for ?x.
+        cc.expand_column_bindings();
+
+        assert!(cc.is_known_empty);
+        assert_eq!(cc.empty_because.unwrap(),
+                   EmptyBecause::TypeMismatch(y.clone(), ValueType::String, ValueType::Boolean));
```
ncalexan: Nice, you made this more structured.

query-algebrizer/src/cc.rs
```diff
@@ -1235,4 +1235,114 @@ mod testing {
...
+    }
+
+    #[test]
+    #[should_panic(expected = "assertion failed: cc.is_known_empty")]
```
ncalexan: Is this better than explicitly asserting the current state and explaining what the future state should be?  It definitely doesn't allow you to add to the test easily.

query-parser/src/parse.rs
```diff
@@ -136,21 +178,25 @@ def_value_parser_fn!(Where, pattern, Pattern, input, {
...
+});
+
 def_value_parser_fn!(Where, clauses, Vec<WhereClause>, input, {
-    // Right now we only support patterns. See #239 for more.
-    (many1::<Vec<Pattern>, _>(Where::pattern()), eof())
-        .map(|(patterns, _)| {
-            patterns.into_iter().map(WhereClause::Pattern).collect()
-        })
+    // Right now we only support patterns and predicates. See #239 for more.
+    (many1::<Vec<WhereClause>, _>(choice([Where::pattern(), Where::pred()])), eof())
```
ncalexan: There's a weirdness using `combine` where `choice` and parsers that consume some input  require care.  Have you taken that care here?  (I don't know the details offhand so I can't determine if you have without doing some research myself, so I'd like to save that research if you have done it.)

query/src/lib.rs
```diff
@@ -129,6 +150,19 @@ pub enum FnArg {
     Constant(NonIntegerConstant),
 }
 
+impl FromValue<FnArg> for FnArg {
+    fn from_value(v: &edn::Value) -> Option<FnArg> {
+        // TODO: support SrcVars.
+        Variable::from_value(v)
```
ncalexan: Is this better than matching the `edn::Value` first?  You don't want to encode knowledge of variables into `FnArg`?

query/src/lib.rs
```diff
@@ -433,6 +467,11 @@ impl Pattern {
     }
 }
 
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Predicate {
+    pub operator: PlainSymbol,
```
ncalexan: You're purposely accepting arbitrary predicate operators, right?  You'll constrain that set after parsing?  (Perhaps during interpretation/simplification?)

query-parser/src/parse.rs
```diff
@@ -93,34 +96,73 @@ impl<I> Query<I>
...
+                             pattern_non_value_place,
+                             PatternNonValuePlace,
+                             PatternNonValuePlace::from_value);
+
+/// Take a vector Value containing one vector Value, and return the `Vec` inside the inner vector.
+fn unwrap_nested(x: edn::Value) -> Option<Vec<edn::Value>> {
+    match x {
+        edn::Value::Vector(mut v) => {
+            match v.pop() {
+                Some(edn::Value::Vector(items)) => Some(items),
```
ncalexan: You're returning consuming your inputs and returning owned values; can't you match `Some(edn::Value::List(linked_list))` and produce a `Vec` out of it?

query-parser/src/find.rs
```diff
@@ -201,25 +201,21 @@ mod test_parse {
         let truncated_input = edn::Value::Vector(vec![Value::from_keyword(None, "find")]);
         assert!(parse_find(truncated_input).is_err());
 
-        let input = edn::Value::Vector(vec![
```
ncalexan: I feel like our editors are going to do some battle royale at some point, but that's in the future.  Right?  Right!

query-algebrizer/src/errors.rs
```diff
@@ -0,0 +1,40 @@
...
+            description("no such function")
+            display("no function named {}", name)
+        }
+    
+        InvalidNumberOfArguments(name: PlainSymbol, number: usize, expected: usize) {
+            description("invalid number of arguments")
+            display("invalid number of arguments to {}: expected {}, got {}.", name, expected, number)
+        }
+
+        UnboundVariable(var: Variable) {
```
ncalexan: Can you label this with the function `name`?

query-algebrizer/src/cc.rs
```diff
@@ -822,6 +789,64 @@ impl ConjoiningClauses {
...
+        }
+    }
+
+    /// At present we support five built-in comparison binary operators: < > <= >= !=.
+    /// This function:
+    /// - Resolves variables and converts types to those more amenable to SQL.
+    /// - Ensures that the predicate functions name a known operator.
+    /// - Accumulates a `BinaryPredicate` constraint into the `wheres` list.
+    pub fn apply_predicate<'s, 'p>(&mut self, schema: &'s Schema, predicate: Predicate) -> Result<()> {
+        if predicate.args.len() != 2 {
```
ncalexan: For these finite-length destructurings, I quite like `match (it.next(), it.next(), it.next()) { (Some, Some, None) => happy; _ => sad }`.

query-algebrizer/src/types.rs
```diff
@@ -85,4 +93,99 @@ impl QualifiedAlias {
...
+impl Debug for ColumnConstraint {
+    fn fmt(&self, f: &mut Formatter) -> ::std::fmt::Result {
+        use self::ColumnConstraint::*;
+        match self {
+            &Equals(ref qa1, ref thing) => {
+                write!(f, "{:?} = {:?}", qa1, thing)
+            },
+
+            &BinaryPredicate { operator, ref left, ref right } => {
+                write!(f, "{:?} {:?} {:?}", left, operator, right)
```
ncalexan: I would be slightly to keep `{:?}` "standard", and overload `{}`.

rnewman:
Reworked with the things I could fix, rebased on top of the `unique` changes, and extended to do soup-to-nuts. Take another look, @ncalexan and @fluffyemily!

rnewman:
Oh, and this handles lists now.

query-parser/src/parse.rs
```diff
@@ -93,34 +96,75 @@ impl<I> Query<I>
...
+        }
+        _ => None,
+    }
+}
+
+/// A vector containing just a parenthesized filter expression.
+def_value_parser_fn!(Where, pred, WhereClause, input, {
+    satisfy_map(|x: edn::Value| {
+        // Accept either a list or a vector here:
+        // `[(foo ?x ?y)]` or `[[foo ?x ?y]]`
```
ncalexan: What's the motivation here?  Is the intention to agree with Datomic, which probably takes anything sequential?

query-parser/src/parse.rs
```diff
@@ -93,34 +96,75 @@ impl<I> Query<I>
...
+        unwrap_nested(x).and_then(|items| {
+            let mut p = (Query::predicate_fn(), Query::arguments(), eof()).map(|(f, args, _)| {
+                WhereClause::Pred(
+                    Predicate {
+                        operator: f.0,
+                        args: args,
+                    })
+            });
+            let r: ParseResult<WhereClause, _> = p.parse_lazy(&items[..]).into();
```
ncalexan: I wonder if we should add `result()` to `ParseResult` or something to shorthand this `.ok().map()` pattern.

query-algebrizer/src/cc.rs
```diff
@@ -822,6 +789,80 @@ impl ConjoiningClauses {
...
+            Ident(ref i) => unimplemented!(),     // TODO
+            Constant(NonIntegerConstant::Boolean(val)) => Ok(QueryValue::TypedValue(TypedValue::Boolean(val))),
+            Constant(NonIntegerConstant::Float(f)) => Ok(QueryValue::TypedValue(TypedValue::Double(f))),
+            Constant(NonIntegerConstant::Text(s)) => Ok(QueryValue::TypedValue(TypedValue::String(s.clone()))),
+            Constant(NonIntegerConstant::BigInteger(_)) => unimplemented!(),
+            SrcVar(_) => unimplemented!(),
+        }
+    }
+
+    /// There are several kinds of predicates/functions in our Datalog:
```
ncalexan: nit: Datalog dialect?  Datalog vocabulary?

query-algebrizer/src/cc.rs
```diff
@@ -822,6 +789,80 @@ impl ConjoiningClauses {
...
+        let right = self.resolve_argument(args.next().unwrap())?;
+
+        // These arguments must be variables or numeric constants.
+        // TODO: generalize argument resolution and validation for different kinds of predicates:
+        // as we process `(< ?x 5)` we are able to check or deduce that `?x` is numeric, and either
+        // simplify the pattern or optimize the rest of the query.
+        // To do so needs a slightly more sophisticated representation of type constraints — a set,
+        // not a single `Option`.
+
+        // TODO: static evaluation.
```
ncalexan: I'm glad you're punting on this.  Maybe link to a ticket?

query-algebrizer/src/types.rs
```diff
@@ -85,4 +90,117 @@ impl QualifiedAlias {
...
+
+impl Debug for NumericComparison {
+    fn fmt(&self, f: &mut Formatter) -> ::std::fmt::Result {
+        use self::NumericComparison::*;
+        f.write_str(match self {
+            &LessThan => "<",
+            &LessThanOrEquals => "<=",
+            &GreaterThan => ">",
+            &GreaterThanOrEquals => ">=",
+            &NotEquals => "!=",                // Datalog uses !=. SQL uses <>.
```
ncalexan: FWIW, I believe SQLite accepts the non-SQL-standard `!=`.  I prefer to use the SQL standard `<>`, as you have done here.

query-translator/tests/translate.rs
```diff
@@ -178,3 +178,50 @@ fn test_unknown_ident() {
...
+
+#[test]
+fn test_numeric_gte_known_attribute() {
+    let mut schema = Schema::default();
+    associate_ident(&mut schema, NamespacedKeyword::new("foo", "bar"), 99);
+    add_attribute(&mut schema, 99, Attribute {
+        value_type: ValueType::Double,
+        ..Default::default()
+    });
+
```
ncalexan: nit: you have a few extra newlines, here and below.

query-algebrizer/src/lib.rs
```diff
@@ -62,37 +68,42 @@ impl AlgebraicQuery {
...
     let mut cc = cc::ConjoiningClauses::default();
     let where_clauses = parsed.where_clauses;
     for where_clause in where_clauses {
-        if let WhereClause::Pattern(p) = where_clause {
-            cc.apply_pattern(schema, p);
-        } else {
-            unimplemented!();
+        match where_clause {
+            WhereClause::Pattern(p) => {
+                cc.apply_pattern(schema, p);
```
ncalexan: Consider making `apply_pattern -> Result<>` as well, just for uniformity.

query-algebrizer/src/cc.rs
```diff
@@ -83,6 +87,12 @@ impl<T> OptionEffect<T> for Option<T> {
     }
 }
 
+fn unit_type_set(t: ValueType) -> HashSet<ValueType> {
```
ncalexan: Don't you wish Rust made this easier?  Like `set.with(...)` or `Set::of{1,2,3}(...)`.

ncalexan: I'd take a trait in `util` for making some of these easier.  Could probably do this generically across collection types with a `FromIterator` bound.

query-algebrizer/src/cc.rs
```diff
@@ -247,6 +261,16 @@ impl ConjoiningClauses {
         self.value_bindings.get(var).cloned()
     }
 
+    /// Return a single `ValueType` if the given variable is known to have a precise type.
+    /// Returns `None` if the type of the variable is known but not precise -- "double
```
ncalexan: nit: or if the type of the variable is not known.

query-algebrizer/src/cc.rs
```diff
@@ -792,6 +856,38 @@ impl ConjoiningClauses {
...
+            },
+            // Can't be an entid.
+            EntidOrInteger(i) => Ok(QueryValue::TypedValue(TypedValue::Long(i))),
+            Ident(_) |
+            SrcVar(_) |
+            Constant(NonIntegerConstant::Boolean(_)) |
+            Constant(NonIntegerConstant::Text(_)) |
+            Constant(NonIntegerConstant::BigInteger(_)) => {
+                self.mark_known_empty(EmptyBecause::NonNumericArgument);
+                // We use Double because… well, we only have one slot!
```
ncalexan: I don't follow this comment.

query-translator/tests/translate.rs
```diff
@@ -186,9 +186,8 @@ fn test_numeric_less_than_unknown_attribute() {
     let input = r#"[:find ?x :where [?x _ ?y] [(< ?y 10)]]"#;
     let SQLQuery { sql, args } = translate(&schema, input, None);
 
-    // TODO: we don't infer numeric types from numeric predicates, because the _SQL_ type code
-    // is a single value (5), but the Datalog types are a set (Double and Long).
-    // When we do, this will correctly use `datoms` instead of `all_datoms`.
+    // Although we infer numericness from numeric predicates, we've already assigned a table to the
```
ncalexan: This seems like a TODO at least.

query-parser/src/parse.rs
rnewman: Indeed, the grammar declares that anywhere you see `[]` it means a list or vector (probably because all Clojure cares about are `seq`s). The common usages we see of things like `[(< ?x 5)]` are just convention.

I don't think we should jump to being so flexible; I'm happy to pave the cowpaths and make `[]` required in most cases.

query-algebrizer/src/cc.rs
rnewman: You can do `vec![1, 2].into_iter().collect()` with the right type bounds, but it's messy.

query-parser/src/parse.rs
rnewman: `ParseResult` is just a type alias to `Result`, so we can't easily add traits to it in `combine`. We could perhaps macroize this in parser utils… but in the mean time I'll use the `to_parsed_value` function I defined to at least make it more readable.

query-algebrizer/src/cc.rs
rnewman: That's because I simplified the code! I originally generated an error that said, in short, "this thing isn't a double", when of course it really meant "this isn't a {double, long}".

Comment removed!

query-algebrizer/src/types.rs
rnewman: It does; one of SQLite's design goals is to do the best it can with SQL written for other database systems. But they all handle `<>`, so that's what we'll use.

rnewman:
> `+    // Although we infer numericness from numeric predicates, we've already assigned a table to the`
>
> This seems like a TODO at least.

I drafted an issue for this while I was on the plane, but didn't file it yet.

I'm on the fence with this.

On the one hand, we can solve this case, and some others, simply by deferring table computation until the CC has been built. (Assuming that columns don't depend on datom table selection, anyway.)

Wait until we know types and attributes, and then go. That's equivalent to processing the numeric constraint first.

On the other hand, that's a bit of a hack, and perhaps we should be doing either a multi-pass CC reduction or multi-pass constraint-based algebrizing (basically #373). Walk all the patterns to deduce types and fixed values, then walk them again to produce the ordered constraint list.

I'm not super concerned about this, because real-world queries look like this:

```
[:find ?page ?visit :where
 [?page :visit/timestamp ?timestamp]
 [(> ?timestamp 123456)]]
```

and the attribute is enough to pick a table.

rnewman:
I also simplified the type narrowing code:

```
-        if let Some(existing) = self.known_types.get(&variable).cloned() {
-            // If so, the types must match.
+        if let Some(existing) = self.known_types.insert(variable.clone(), unit_type_set(this_type)) {
+            // There was an existing mapping. Does this type match?
             if !existing.contains(&this_type) {
                 self.mark_known_empty(EmptyBecause::TypeMismatch(variable, existing, this_type));
-            } else {
-                if existing.len() > 1 {
-                    // Narrow.
-                    self.known_types.insert(variable, unit_type_set(this_type));
-                }
             }
-        } else {
-            // If not, record the one we just determined.
-            self.known_types.insert(variable, unit_type_set(this_type));
         }
```

rnewman:
5e971f3b229f9cc9710706157f7bc9f66799f3f4

PR #375, opened 2017-03-17
Support transacting :db/fulltext true attributes. Fixes #189.

### LINKED ISSUE #189: [tx] Handle fulltext in transactor
There's enough work here to track separately.  It's identical to the Clojure implementation, but requires special care and testing.

It might be possible to implement #12 / #69 concurrently to this, since fulltext values are the first natural candidate for such a hashing scheme.

ncalexan:
This builds on #370, so be sure to look at that first.

### DIFF
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -508,6 +508,7 @@ pub trait MentatStoring {
 
     // TODO: this is not a reasonable abstraction, but I don't want to really consider non-SQL storage just yet.
     fn insert_non_fts_searches<'a>(&self, entities: &'a [ReducedEntity], search_type: SearchType) -> Result<()>;
+    fn insert_fts_searches<'a>(&self, entities: &'a [ReducedEntity], search_type: SearchType) -> Result<()>;
 
     /// Finalize the underlying storage layer after a Mentat transaction.
     ///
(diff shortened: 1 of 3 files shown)


db/src/db.rs
```diff
@@ -806,6 +807,104 @@ impl MentatStoring for rusqlite::Connection {
         results.map(|_| ())
     }
 
+    /// Insert search rows into temporary search tables.
+    ///
+    /// Eventually, the details of this approach will be captured in
```
rnewman: File issue?

db/src/db.rs
```diff
@@ -806,6 +807,104 @@ impl MentatStoring for rusqlite::Connection {
...
+    /// https://github.com/mozilla/mentat/wiki/Transacting:-entity-to-SQL-translation.
+    fn insert_fts_searches<'a>(&self, entities: &'a [ReducedEntity<'a>], search_type: SearchType) -> Result<()> {
+        let max_vars = self.limit(Limit::SQLITE_LIMIT_VARIABLE_NUMBER) as usize;
+        let bindings_per_statement = 6;
+
+        let mut outer_searchid = 2000;
+
+        let chunks: itertools::IntoChunks<_> = entities.into_iter().chunks(max_vars / bindings_per_statement);
+
+        // We'd like to flat_map here, but it's not obvious how to flat_map across Result.
```
rnewman: I think `collect` has automatic conversions for `Vec<Result<T>>` to `Result<Vec<T>>`, if that helps.

db/src/db.rs
```diff
@@ -806,6 +807,104 @@ impl MentatStoring for rusqlite::Connection {
...
+        let results: Result<Vec<()>> = chunks.into_iter().map(|chunk| -> Result<()> {
+            let mut count = 0;
+
+            // We must keep these computed values somewhere to reference them later, so we can't
+            // combine this map and the subsequent flat_map.
+            // (e0, a0, v0, value_type_tag0, added0, flags0)
+            let block: Result<Vec<(i64 /* e */,
```
rnewman: `Entid`? Or are you sticking to raw types to emphasize SQLness?

db/src/db.rs
```diff
@@ -806,6 +807,104 @@ impl MentatStoring for rusqlite::Connection {
...
+                                   i64 /* a */,
+                                   ToSqlOutput<'a> /* value */,
+                                   i32 /* value_type_tag */,
+                                   bool /* added0 */,
+                                   u8 /* flags0 */,
+                                   i64 /* searchid */)>> = chunk.map(|&(e, a, ref attribute, ref typed_value, added)| {
+                if typed_value.value_type() != ValueType::String {
+                    bail!("TODO");
```
rnewman: `unimplemented!()`?

db/src/debug.rs
```diff
@@ -21,13 +21,20 @@ use rusqlite::types::{ToSql};
 use tabwriter::TabWriter;
 
 use bootstrap;
+use db::TypedSQLValue;
```
rnewman: I think you've interleaved modules and crates here…

db/src/db.rs
ncalexan: I don't think it does.  We want to `flat_map` a bunch of `Result<Iter>` into `Result<Iter>`.  I guess it's a `fold` with `.chain()`, but that's no more clear than this pattern.

db/src/db.rs
ncalexan: I guess I am.

db/src/db.rs
ncalexan: Oh!  This was just overlooked; I'll add an error message.

db/src/debug.rs
ncalexan: No, these are all modules.  Or do you want all `use module;` before any `use module::symbol`?  I've been interleaving them.

rnewman: I mean: you have `use mentat_core::…` below here, followed by `use types::Schema`. Rather than:

```
extern crate …;

use third_party_library::…;

use mentat_foo::…;
use mentat_bar::…;

use samecratemodule_a::…;
use samecratemodule_b::…;
```

you've interleaved them.

ncalexan: I haven't been separating out `mentat_*` from `samecratemodule_*`.  I can start doing that.

PR #376, opened 2017-03-20
Check sqlite version

### LINKED ISSUE #366: Check SQLite version on first use
We use some features with minimum SQLite version support, such as partial indices (3.8.0).

rusqlite recently grew `version` and `version_number` functions. We should use those in `db/src/db.rs` to verify that we can use the linked SQLite library before going any further.

fluffyemily:
Fixes #366 

Checks whether current SQLite version is at least the minimum required version and force exits with an error message if not.

This might have been done much simpler by just having 2 constants, one for version string and one for version number but I wanted to remove the possibility of developer error by bad translation.

### DIFF
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -217,8 +217,6 @@ fn version_number_from_string(version_string: &'static str) -> i32 {
 
 // TODO: rename "SQL" functions to align with "datoms" functions.
 pub fn create_current_version(conn: &mut rusqlite::Connection) -> Result<DB> {
-    writeln!(&mut io::stderr(), "current rusqlite version is {}", rusqlite::version()).unwrap();
-
     if rusqlite::version_number() < version_number_from_string(MIN_SQLITE_VERSION) {
         writeln!(&mut io::stderr(), "Mentat requires at least sqlite {}", MIN_SQLITE_VERSION).unwrap();
         exit(1)
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -1062,4 +1062,14 @@ mod tests {
         conn.set_limit(Limit::SQLITE_LIMIT_VARIABLE_NUMBER, 222);
         assert_eq!(222, conn.limit(Limit::SQLITE_LIMIT_VARIABLE_NUMBER));
     }
+
+    #[test]
+    fn test_version_number_from_string() {
+        assert_eq!(3000000, version_number_from_string("3.0.0"));
+        assert_eq!(3008000, version_number_from_string("3.8.0"));
+        assert_eq!(3017000, version_number_from_string("3.17.0"));
+        assert_eq!(3017008, version_number_from_string("3.17.8"));
+        assert_eq!(3017012, version_number_from_string("3.17.12"));
+        assert_eq!(13017012, version_number_from_string("13.17.12"));
+    }
 }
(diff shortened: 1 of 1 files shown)


db/src/db.rs
```diff
@@ -189,8 +197,30 @@ fn get_user_version(conn: &rusqlite::Connection) -> Result<i32> {
...
+            version_number.push_str("0")
+        }
+        version_number.push_str(s);
+    }
+    return version_number.parse::<i32>().unwrap();
+}
+
 // TODO: rename "SQL" functions to align with "datoms" functions.
 pub fn create_current_version(conn: &mut rusqlite::Connection) -> Result<DB> {
+    if rusqlite::version_number() < version_number_from_string(MIN_SQLITE_VERSION) {
```
rnewman: Why not just define `MIN_SQLITE_VERSION` as `3008000` and do that here, rather than adding runtime parsing code?

I think it's quite common to express SQLite versions in this way, so I'm very happy to simply hardcode the simple thing:

```
const MIN_SQLITE_VERSION = 3008000;    // 3.8.0
```

db/src/db.rs
```diff
@@ -189,8 +197,30 @@ fn get_user_version(conn: &rusqlite::Connection) -> Result<i32> {
...
+        writeln!(&mut io::stderr(), "Mentat requires at least sqlite {}", MIN_SQLITE_VERSION).unwrap();
+        exit(1)
```
rnewman: Just do

```
panic!("Mentat requires SQLite {}", MIN_SQLITE_VERSION); 
```

which will do what you want: kill the invoking thread with that message. That way you don't need `std::process:exit` or that `writeln`.

db/src/db.rs
fluffyemily: I was working on the most readable output scenario, which is having the error print out the correct minimum SQLite version as a readable string, but yeah, if this is usual then any users would understand it I guess.

db/src/db.rs
```diff
@@ -71,9 +71,15 @@ pub fn new_connection<T>(uri: T) -> rusqlite::Result<rusqlite::Connection> where
 ///    the part range here; tie bootstrapping to the SQLite user_version.
 pub const CURRENT_VERSION: i32 = 2;
 
+/// MIN_SQLITE_VERSION should be changed when there's a new minimum version of sqlite required
+/// to build the project.
```
rnewman: Strictly speaking this is to 'work', not to 'build the project'!

db/src/db.rs
```diff
@@ -71,9 +71,15 @@ pub fn new_connection<T>(uri: T) -> rusqlite::Result<rusqlite::Connection> where
...
+const MIN_SQLITE_VERSION: i32 = 3008000;
+
 const TRUE: &'static bool = &true;
 const FALSE: &'static bool = &false;
 
+
```
rnewman: Nit: surplus newlines?

db/src/db.rs
```diff
@@ -191,6 +197,9 @@ fn get_user_version(conn: &rusqlite::Connection) -> Result<i32> {
 
 // TODO: rename "SQL" functions to align with "datoms" functions.
 pub fn create_current_version(conn: &mut rusqlite::Connection) -> Result<DB> {
+    if rusqlite::version_number() < MIN_SQLITE_VERSION {
```
rnewman: This should be in `ensure_current_version`, down below — `create_current_version` doesn't run when simply opening an existing database.

Issue #237, opened 2017-02-02
[tx] Allow setting the default partition for string temp IDs


ncalexan:
For #184, I'm going to try implementing `edn::Value::Text` string temp IDs.  By default, these will allocate in the `:db.part/user` partition.

This follows Datomic, but way down at the bottom of http://docs.datomic.com/transactions.html it shows a way to configure this "default partition".  This ticket tracks letting this be configured.  My initial implementation suggestion is to make this a datom in the store with a cached materialized view; might as well be consistent.  Something like:
```edn
[:db/default :db.default/partition :db.part/user]
```
which reads "the DB defaults entity has default partition set to the user partition".

ncalexan:
This is not a good first bug, and it's blocked on quite a few other things landing first.

PR #382, opened 2017-03-21
Lookup refs, nested vector values, map notation.

ncalexan:
The transactor starts to feel like a real child!

### DIFF
--- a/db/src/tx.rs
+++ b/db/src/tx.rs
@@ -211,57 +211,57 @@ impl<'conn, 'a> Tx<'conn, 'a> {
         let mut terms: Vec<TermWithTempIdsAndLookupRefs> = Vec::with_capacity(stack.len());
 
         while let Some(entity) = stack.pop() {
-                match entity {
-                    Entity::AddOrRetract { op, e, a, v } => {
-                        let a: i64 = match a {
-                            entmod::Entid::Entid(ref a) => *a,
-                            entmod::Entid::Ident(ref a) => self.schema.require_entid(&a)?,
-                        };
+            match entity {
+                Entity::AddOrRetract { op, e, a, v } => {
+                    let a: i64 = match a {
+                        entmod::Entid::Entid(ref a) => *a,
+                        entmod::Entid::Ident(ref a) => self.schema.require_entid(&a)?,
+                    };
 
-                        let attribute: &Attribute = self.schema.require_attribute_for_entid(a)?;
+                    let attribute: &Attribute = self.schema.require_attribute_for_entid(a)?;
 
-                        let e = match e {
-                            entmod::EntidOrLookupRefOrTempId::Entid(e) => {
-                                let e: i64 = match e {
-                                    entmod::Entid::Entid(ref e) => *e,
-                                    entmod::Entid::Ident(ref e) => self.schema.require_entid(&e)?,
-                                };
-                                Either::Left(e)
-                            },
+                    let e = match e {
+                        entmod::EntidOrLookupRefOrTempId::Entid(e) => {
+                            let e: i64 = match e {
+                                entmod::Entid::Entid(ref e) => *e,
+                                entmod::Entid::Ident(ref e) => self.schema.require_entid(&e)?,
+                            };
+                            Either::Left(e)
+                        },
 
-                            entmod::EntidOrLookupRefOrTempId::TempId(e) => {
-                                Either::Right(LookupRefOrTempId::TempId(temp_ids.intern(e)))
-                            },
+                        entmod::EntidOrLookupRefOrTempId::TempId(e) => {
+                            Either::Right(LookupRefOrTempId::TempId(temp_ids.intern(e)))
+                        },
 
-                            entmod::EntidOrLookupRefOrTempId::LookupRef(lookup_ref) => {
-                                Either::Right(LookupRefOrTempId::LookupRef(intern_lookup_ref(&mut lookup_refs, lookup_ref)?))
-                            },
-                        };
+                        entmod::EntidOrLookupRefOrTempId::LookupRef(lookup_ref) => {
+                            Either::Right(LookupRefOrTempId::LookupRef(intern_lookup_ref(&mut lookup_refs, lookup_ref)?))
+                        },
+                    };
 
-                        let v = match v {
-                            entmod::AtomOrLookupRef::Atom(v) => {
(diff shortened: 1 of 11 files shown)


tx-parser/src/lib.rs
```diff
@@ -83,86 +95,91 @@ def_parser_fn!(Tx, temp_id, Value, String, input, {
...
+        let add = token(Value::NamespacedKeyword(NamespacedKeyword::new("db", "add")))
+            .map(|_| OpType::Add);
+        let retract = token(Value::NamespacedKeyword(NamespacedKeyword::new("db", "retract")))
+            .map(|_| OpType::Retract);
+        let mut p = (add.or(retract),
+                     Tx::<&[Value]>::entid_or_lookup_ref_or_temp_id(),
+                     Tx::<&[Value]>::entid(),
+                     Tx::<&[Value]>::atom_or_lookup_ref_or_vector(),
+                     eof())
+            .map(|(op, e, a, v, _)| {
```
rnewman: This comes up a lot. Can we fork [between](https://docs.rs/combine/2.3.1/combine/combinator/fn.between.html) or figure out a way to make it into `terminated`, so we could just do

```
let mut p = terminated((add.or(retract), Tx::<…>…)).map(Entity::AddOrRetract);
```

tx-parser/src/lib.rs
```diff
@@ -83,86 +95,91 @@ def_parser_fn!(Tx, temp_id, Value, String, input, {
...
+fn value_to_map_notation(val: &Value) -> Option<MapNotation> {
+    val.as_map().cloned().and_then(|map| {
+        // Parsing pairs is tricky; parsing sequences is easy.
+        let avseq: Vec<Value> = map.into_iter().flat_map(|(a, v)| once(a).chain(once(v))).collect();
+
+        let av = (Tx::<&[Value]>::entid(),
+                  Tx::<&[Value]>::atom_or_lookup_ref_or_vector())
+            .map(|(a, v)| -> (Entid, AtomOrLookupRefOrVectorOrMapNotation) { (a, v) });
+        let mut p = (many(av),
+                     eof())
```
rnewman: See!

tx-parser/src/lib.rs
```diff
@@ -175,13 +192,61 @@ impl<'a> Tx<&'a [edn::Value]> {
...
+
+/// Remove any :db/id value from the given map notation, converting the returned value into
+/// something suitable for the entity position rather than something suitable for a value position.
+///
+/// This is here simply to not expose some of the internal APIs of the tx-parser.
+pub fn remove_db_id(map: &mut MapNotation) -> std::result::Result<Option<EntidOrLookupRefOrTempId>, errors::Error> {
+    // TODO: extract lazy defined constant.
+    let db_id_key = Entid::Ident(edn::NamespacedKeyword::new("db", "id"));
+
+    let db_id: Option<EntidOrLookupRefOrTempId> = if let Some(id) = map.remove(&db_id_key) {
```
rnewman: Prefer `Option::map` to `if let` here, I think — that `else None` at the end caused a "huh" and a skim back.

db/src/tx.rs
```diff
@@ -183,61 +188,204 @@ impl<'conn, 'a> Tx<'conn, 'a> {
...
+                            if op != OpType::Add {
+                                bail!(ErrorKind::NotYetImplemented(format!("Cannot explode nested map value in :db/retract for attribute {}", a)));
+                            }
+
+                            if attribute.value_type != ValueType::Ref {
+                                bail!(ErrorKind::NotYetImplemented(format!("Cannot explode nested map value for attribute {} that is not :db/valueType :db.type/ref", a)))
+                            }
+
+                            // :db/id is optional; if it's not given, we generate a special internal tempid
+                            // to use for upserting.  This tempid will not be reported in the TxReport.
```
ncalexan: This was laziness.  The thing is that it's not splitting the `TempId` at parse time, it's splitting the `TempId` into external and internal at resolution time, which is a more complicated change.  (Possibly?)  However, it leads into associating partitions with tempids (ala #237) so I'll see how easy it is.

db/src/tx.rs
```diff
@@ -323,6 +471,10 @@ impl<'conn, 'a> Tx<'conn, 'a> {
             assert!(tempids.contains_key(&**tempid));
         }
 
+        // Any tempid starting with MENTAT_TEMPID_PREFIX has been allocated by the system and is a private
+        // implementation detail; it shouldn't be exposed in the final transaction report.
+        let tempids = tempids.into_iter().filter(|&(ref tempid, _)| !tempid.starts_with(MENTAT_TEMPID_PREFIX)).collect();
```
ncalexan: I concur -- see previous comment.

edn/src/types.rs
```diff
@@ -307,6 +307,29 @@ macro_rules! def_common_value_methods {
...
+                $t::NamespacedKeyword(_) => false,
+                $t::Vector(_) => true,
+                $t::List(_) => true,
+                $t::Set(_) => true,
+                $t::Map(_) => true,
+            }
+        }
+
+        pub fn is_atom(&self) -> bool {
+            return !self.is_collection()
```
ncalexan: Dropped.

tx-parser/src/lib.rs
```diff
@@ -83,86 +95,91 @@ def_parser_fn!(Tx, temp_id, Value, String, input, {
...
-                match p.parse_lazy(&y[..]).into() {
-                    Ok((r, _)) => Some(r),
-                    _ => None,
-                }
-            } else {
-                None
-            }
-        })
+def_parser_fn!(Tx, atom, Value, Value, input, {
+    satisfy_map(|x: Value| if x.is_atom() { Some(x) } else { None })
```
ncalexan: Done.

tx-parser/src/lib.rs
ncalexan: Per Marwes, the pattern is (...).skip(eof()), which I'll start using in this file.

rnewman: Sounds good!

tx-parser/src/lib.rs
ncalexan: This is a "huh" moment: the inner `map` is actually a `Result`.  I don't know a slick way of doing the right thing from `Option<Result<..>>` to `Result<Option, ...>` via `collect` or some other way, so I'm going to leave this as is and we can follow-up if and when I learn this pattern.

Issue #180, opened 2017-01-21
[tx] Handle map assertion shorthand


ncalexan:
Right now, we don't handle assertions in the map shorthand, like
```edn
[{:db/id IDENT :db/attr1 VALUE1 :db/attr2 VALUE2 ...}]
```
That is equivalent to
```edn
[[:db/add IDENT :db/attr1 VALUE1]]
[[:db/add IDENT :db/attr2 VALUE2]]
```
This ticket tracks parsing such assertions into `Entity` instances, and transacting such `Entity` instances.

Datomic allows to drop the `:db/id IDENT` in nested entities (marked with `:db/isComponent`?) like
```edn
[{:db/id IDENT :db/ref-attr {:db/attr1 VALUE1 ...} ...}]
```
A tempid in the same partition as `IDENT` will be generated for the inner entity.  And it's possible that even the outer `:db/id` can be dropped, automatically selecting `:db.part/user`, although I haven't tested this.

In the Clojure implementation, I expanded from a map into a vector before transacting.  I'd very much like to avoid doing that, both to save allocations, and to potentially save tempid lookups; but it's not critical that we're optimal initially.

ncalexan:
This relies on #169, but the tempid parsing should land in #283 so we should just do the map parsing as part of this ticket.  This is similar to #284; do that one first, since it's simpler.

> In the Clojure implementation, I expanded from a map into a vector before transacting. I'd very much like to avoid doing that, both to save allocations, and to potentially save tempid lookups; but it's not critical that we're optimal initially.

This is no longer relevant -- the approach taken by #283 is to share `Rc` handles to expensive-to-clone things (like lookup refs and tempid strings), so it's reasonable to just expand map entities into many assertions and be done with it.  However, this is still a multi-part process.  You'll need to:

1. Add a `Map` case to https://github.com/mozilla/mentat/blob/6831925af0eafe9e647a462364b6188748f1e840/tx/src/entities.rs#L57.  Map notation is recursive, so the transaction processor will have to handle arbitrary nested `edn::Value::Map` instances, so there's no advantage to specializing the representation here -- just make it `Map { map: edn::Value }`.
1. Parse maps around https://github.com/mozilla/mentat/blob/6831925af0eafe9e647a462364b6188748f1e840/tx-parser/src/lib.rs#L140.
[...]

PR #384, opened 2017-03-22
#260 Convert Schema into edn::Value

fluffyemily:
Here is what I have done so far. Can I please check that this is right?

Also, I am not sure the best way to test the output. Advice welcome.

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -298,6 +298,7 @@ impl Schema {
         self.get_entid(x).map(|e| self.is_attribute(e)).unwrap_or(false)
     }
 
+    /// Returns an edn::Value::Vector representation of the schema map.
     pub fn to_edn_value(&self) -> edn::Value {
         let mut all_values = Vec::new();
         for (entid, attribute) in &self.schema_map {
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -8,13 +8,13 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
-extern crate edn;
-extern crate ordered_float;
 #[macro_use]
 extern crate lazy_static;
+extern crate ordered_float;
+
+extern crate edn;
 
 pub mod values;
-
 
 use std::collections::BTreeMap;
 use self::ordered_float::OrderedFloat;
@@ -41,6 +41,20 @@ pub enum ValueType {
     Double,
     String,
     Keyword,
+}
+
+impl ValueType {
+    pub fn to_edn_value(&self) -> edn::Value {
+        match self {
+            &ValueType::Ref => values::DB_TYPE_REF.clone(),
+            &ValueType::Boolean => values::DB_TYPE_BOOLEAN.clone(),
+            &ValueType::Instant => values::DB_TYPE_INSTANT.clone(),
+            &ValueType::Long => values::DB_TYPE_LONG.clone(),
+            &ValueType::Double => values::DB_TYPE_DOUBLE.clone(),
+            &ValueType::String => values::DB_TYPE_STRING.clone(),
+            &ValueType::Keyword => values::DB_TYPE_KEYWORD.clone(),
+        }
+    }
 }
 
 /// Represents a Mentat value in a particular value set.
@@ -219,6 +233,35 @@ impl Attribute {
         }
         flags
     }
+
+    pub fn to_edn_value(&self, ident: Option<&NamespacedKeyword>) -> edn::Value {
+        let mut attribute_map: BTreeMap<edn::Value, edn::Value> = BTreeMap::default();
+        if let Some(ident) = ident {
+            attribute_map.insert(values::DB_IDENT.clone(), edn::Value::NamespacedKeyword(ident.clone()));
+        }
+
+        attribute_map.insert(values::DB_VALUE_TYPE.clone(), self.value_type.to_edn_value());
+
+        attribute_map.insert(values::DB_CARDINALITY.clone(), if self.multival { values::DB_CARDINALITY_MANY.clone() } else { values::DB_CARDINALITY_ONE.clone() });
+
+        if self.unique == Some(attribute::Unique::Value) {
+            attribute_map.insert(values::DB_UNIQUE.clone(), values::DB_UNIQUE_IDENTITY.clone());
+        }
+        
+        if self.index {
+            attribute_map.insert(values::DB_INDEX.clone(), edn::Value::Boolean(true));
+        }
+
+        if self.fulltext {
+            attribute_map.insert(values::DB_FULLTEXT.clone(), edn::Value::Boolean(true));
+        }
+
+        if self.component {
+            attribute_map.insert(values::DB_IS_COMPONENT.clone(), edn::Value::Boolean(true));
+        }
+
+        edn::Value::Map(attribute_map)
+    }
 }
 
 impl Default for Attribute {
@@ -300,37 +343,10 @@ impl Schema {
 
     /// Returns an symbolic representation of the schema suitable for applying across Mentat stores.
     pub fn to_edn_value(&self) -> edn::Value {
-        let mut all_values = Vec::new();
-        for (entid, attribute) in &self.schema_map {
-            let mut attribute_map: BTreeMap<edn::Value, edn::Value> = BTreeMap::default();
-            if let Some(ident) = self.get_ident(*entid) {
-                attribute_map.insert(values::DB_IDENT.clone(), edn::Value::NamespacedKeyword(ident.clone()));
-            }
-
-            let value_type = format!("{:?}", attribute.value_type);
-            attribute_map.insert(values::DB_VALUE_TYPE.clone(), edn::Value::NamespacedKeyword(NamespacedKeyword::new("db.type", &value_type.to_lowercase())));
-
-            attribute_map.insert(values::DB_CARDINALITY.clone(), if attribute.multival { values::DB_CARDINALITY_MANY.clone() } else { values::DB_CARDINALITY_ONE.clone() });
-
-            if attribute.unique == Some(attribute::Unique::Value) {
-                attribute_map.insert(values::DB_UNIQUE.clone(), values::DB_UNIQUE_IDENTITY.clone());
-            }
-            
-            if attribute.index {
-                attribute_map.insert(values::DB_INDEX.clone(), edn::Value::Boolean(true));
-            }
-
-            if attribute.fulltext {
-                attribute_map.insert(values::DB_FULLTEXT.clone(), edn::Value::Boolean(true));
-            }
-
-            if attribute.component {
-                attribute_map.insert(values::DB_IS_COMPONENT.clone(), edn::Value::Boolean(true));
-            }
-
-            all_values.push(edn::Value::Map(attribute_map));
-        }
-        edn::Value::Vector(all_values)
+        edn::Value::Vector((&self.schema_map).into_iter()
+            .map(|(entid, attribute)| 
+                attribute.to_edn_value(self.get_ident(*entid)))
+            .collect::<Vec<_>>())
     }
 }
 
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -14,9 +14,6 @@ extern crate ordered_float;
 use std::collections::BTreeMap;
 use self::ordered_float::OrderedFloat;
 use self::edn::NamespacedKeyword;
-use self::edn::types::Value;
-
-use std::io::{self, Write};
 
 /// Core types defining a Mentat knowledge base.
 
@@ -299,15 +296,15 @@ impl Schema {
         let mut s = "[ ".to_string(); 
         for (entid, attribute) in &self.schema_map {
             s.push_str("{");
-            s.push_str(&format!("\t:db/id     :{:?}", entid));
+            s.push_str(&format!(" :db/id :{:?}", entid));
             let some_ident = self.get_ident(entid.clone());
             if some_ident.is_some() {
                 let ident = some_ident.unwrap();
-                s.push_str(&format!("\n\t:db/ident     :{}/{}", ident.namespace, ident.name));
+                s.push_str(&format!(" :db/ident :{}/{}", ident.namespace, ident.name));
             }
             let value_type = format!("{:?}", attribute.value_type).to_lowercase();
-            s.push_str(&format!("\n\t:db/valueType :db.type/{}", value_type));
-            s.push_str("\n\t:db/cardinality :db.cardinality/");
+            s.push_str(&format!(" :db/valueType :db.type/{}", value_type));
+            s.push_str(" :db/cardinality :db.cardinality/");
 
             if attribute.multival {
                 s.push_str("many");
@@ -316,22 +313,21 @@ impl Schema {
             }
 
             if attribute.unique == Some(attribute::Unique::Value) {
-                s.push_str("\n\t:db/unique :db.unique/identity");
+                s.push_str(" :db/unique :db.unique/identity");
             }
 
             if attribute.index {
-                s.push_str("\n\t:db/index true");
+                s.push_str(" :db/index true");
             }
             if attribute.fulltext {
-                s.push_str("\n\t:db/fulltext true");
+                s.push_str(" :db/fulltext true");
             }
             if attribute.component {
-                s.push_str("\n\t:db/isComponent true");
+                s.push_str(" :db/isComponent true");
             }
             s.push_str(" },");  
         }
         s.push_str(" ]");
-        writeln!(&mut io::stderr(), "{}", s).unwrap();    
         edn::parse::value(&s).unwrap().without_spans()
     }
 }
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -10,6 +10,11 @@
 
 extern crate edn;
 extern crate ordered_float;
+#[macro_use]
+extern crate lazy_static;
+
+pub mod values;
+
 
 use std::collections::BTreeMap;
 use self::ordered_float::OrderedFloat;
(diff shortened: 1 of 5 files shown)


rnewman:
Here's the simple way to test: build the EDN you expect the schema fragment to produce, and check they're equal.

https://github.com/mozilla/mentat/blob/rust/edn/tests/tests.rs#L853

@ncalexan might have other ideas.

core/src/lib.rs
```diff
@@ -291,6 +291,45 @@ impl Schema {
     pub fn identifies_attribute(&self, x: &NamespacedKeyword) -> bool {
         self.get_entid(x).map(|e| self.is_attribute(e)).unwrap_or(false)
     }
+
+    pub fn as_edn_value(&self) -> edn::Value {
+        let mut s = "[ ".to_string(); 
+        for (entid, attribute) in &self.schema_map {
+            s.push_str("{");
+            s.push_str(&format!(" :db/id :{:?}", entid));
```
rnewman: You're converting to an EDN string then parsing. That's probably not the right approach; the caller of this function is probably just going to turn it into a string again, perhaps first embedding it in another data structure!

Instead, you should build up an `edn::Value::Map`, which is simply a `BTreeMap` from keys to `edn::Value`s.

fluffyemily: ahh. I think I got wrong-footed by this link in the original bug [https://github.com/mozilla/mentat/blob/c585715224c180ff84630c05aa763d0d271f350a/db/src/bootstrap.rs#L86](https://github.com/mozilla/mentat/blob/c585715224c180ff84630c05aa763d0d271f350a/db/src/bootstrap.rs#L86)

core/src/lib.rs
```diff
@@ -291,6 +292,45 @@ impl Schema {
     pub fn identifies_attribute(&self, x: &NamespacedKeyword) -> bool {
         self.get_entid(x).map(|e| self.is_attribute(e)).unwrap_or(false)
     }
+
+    pub fn as_edn_value(&self) -> edn::Value {
```
ncalexan: Rust generally uses `as` for cheap casts (views) and `to` for expensive conversions.  See https://aturon.github.io/style/naming/conversions.html.

core/src/lib.rs
```diff
@@ -291,6 +292,45 @@ impl Schema {
...
+        let mut all_values = Vec::new();
+        for (entid, attribute) in &self.schema_map {
+            let mut attribute_map: BTreeMap<edn::Value, edn::Value> = BTreeMap::default();
+            attribute_map.insert(edn::Value::NamespacedKeyword(NamespacedKeyword::new("db", "id")), edn::Value::Keyword(Keyword::new(entid.to_string())));
```
ncalexan: These values are all (lazily) declared in `src/db/values.rs`.  Let's move that to `core` and use them here.

You should find that you only need those values; anything else reveals a mismatch between what you're doing and what the system is doing elsewhere, and we should figure out what it is.

core/src/lib.rs
```diff
@@ -341,6 +381,76 @@ mod test {
...
+{   :db/id     :99
+    :db/ident     :foo/bat
+    :db/valueType :db.type/boolean
+    :db/cardinality :db.cardinality/one
+    :db/component true }, ]"#;
+        let expected_value = edn::parse::value(&expected_output).unwrap().without_spans();
+        assert_eq!(expected_value, value);
+    }
+
+    fn associate_ident(schema: &mut Schema, i: NamespacedKeyword, e: Entid) {
```
ncalexan: nit: put these before (all) the tests, just 'cuz the test set will grow downwards.

core/src/lib.rs
```diff
@@ -341,6 +381,76 @@ mod test {
...
+    :db/valueType :db.type/string
+    :db/cardinality :db.cardinality/many
+    :db/unique :db.unique/identity
+    :db/fulltext true },
+{   :db/id     :99
+    :db/ident     :foo/bat
+    :db/valueType :db.type/boolean
+    :db/cardinality :db.cardinality/one
+    :db/component true }, ]"#;
+        let expected_value = edn::parse::value(&expected_output).unwrap().without_spans();
```
ncalexan: It's best to use `.expect("to be able to parse")` rather than `.unwrap()`, 'cuz `.unwrap()` doesn't give a helpful line number when it fails.

core/src/lib.rs
fluffyemily: So, there are no values defined for `index`, `component` and `fulltext`, but they are present as entids in `entids.rs`.

```

pub const DB_IS_COMPONENT: Entid = 10;
pub const DB_INDEX: Entid = 11;
pub const DB_FULLTEXT: Entid = 12;
```

I am therefore going to create entries for them in `values` as clearly they exist as concepts here. If you feel that is not correct, let me know.

There is no entry anywhere for `db/id` so I shall remove that entry from the output.

core/src/lib.rs
```diff
@@ -10,10 +10,16 @@
 
 extern crate edn;
 extern crate ordered_float;
+#[macro_use]
+extern crate lazy_static;
```
rnewman: Nit: newline above `#[macro_use]`, move this above `ordered_float`, and shuffle `edn` down to here. We do this order:

- extern crate: libs
- extern crate: other mentat modules
- `use` libs
- `use` other mentat modules
- `use` modules in the same crate

db/src/bootstrap.rs
```diff
@@ -22,10 +22,10 @@ use mentat_core::{
     IdentMap,
     Schema,
     TypedValue,
+    values,
```
ncalexan: Namespaced identifiers for the win!

core/src/lib.rs
```diff
@@ -10,10 +10,16 @@
...
+
+pub mod values;
+
 
 use std::collections::BTreeMap;
 use self::ordered_float::OrderedFloat;
 use self::edn::NamespacedKeyword;
+use self::edn::Keyword;
```
rnewman:
```
use self::edn::{
    Keyword,
    NamespacedKeyword,
};
```

core/src/lib.rs
```diff
@@ -293,38 +298,37 @@ impl Schema {
...
     }
 
-    pub fn as_edn_value(&self) -> edn::Value {
+    pub fn to_edn_value(&self) -> edn::Value {
         let mut all_values = Vec::new();
         for (entid, attribute) in &self.schema_map {
             let mut attribute_map: BTreeMap<edn::Value, edn::Value> = BTreeMap::default();
-            attribute_map.insert(edn::Value::NamespacedKeyword(NamespacedKeyword::new("db", "id")), edn::Value::Keyword(Keyword::new(entid.to_string())));
-
             let some_ident = self.get_ident(entid.clone());
```
ncalexan: You probably don't need to clone the `entid` here; `get_ident` should take it by reference.

It is more idiomatic to pattern match here, like:
```rust
if let Some(ref ident) = self.get_ident(entid) {
    ...
} else {
    ...
}
```
That way you don't need to fetch-test-unwrap.  Try it and see!

core/src/lib.rs
```diff
@@ -291,12 +297,60 @@ impl Schema {
...
+
+            if attribute.fulltext {
+                attribute_map.insert(values::DB_FULLTEXT.clone(), edn::Value::Boolean(true));
+            }
+
+            if attribute.component {
+                attribute_map.insert(values::DB_IS_COMPONENT.clone(), edn::Value::Boolean(true));
+            }
+
+            all_values.push(edn::Value::Map(attribute_map));
```
rnewman: Rather than doing this with mutable vecs, use comprehensions:

```rust
pub fn to_edn_value(&self) -> edn::Value {
    edn::Value::Vector(
        self.schema_map
            .map(|(entid, attribute)| self.schema_to_edn(entid, attribute))
            .collect())
}
```

core/src/lib.rs
```diff
@@ -291,12 +297,60 @@ impl Schema {
...
+        }
+        edn::Value::Vector(all_values)
+    }
 }
 
 #[cfg(test)]
 mod test {
     use super::*;
 
+    fn associate_ident(schema: &mut Schema, i: NamespacedKeyword, e: Entid) {
```
rnewman: These are coming in owned — you only need to clone once.

core/src/lib.rs
```diff
@@ -291,12 +297,60 @@ impl Schema {
...
+    pub fn to_edn_value(&self) -> edn::Value {
+        let mut all_values = Vec::new();
+        for (entid, attribute) in &self.schema_map {
+            let mut attribute_map: BTreeMap<edn::Value, edn::Value> = BTreeMap::default();
+            let some_ident = self.get_ident(entid.clone());
+            if some_ident.is_some() {
+                let ident = some_ident.unwrap();
+                attribute_map.insert(values::DB_IDENT.clone(), edn::Value::NamespacedKeyword(ident.clone()));
+            } else {
+                attribute_map.insert(values::DB_IDENT.clone(), edn::Value::Keyword(Keyword::new(entid.to_string())));
```
rnewman: This isn't right: you'll be making something like

```edn
  :db/ident :1234
```

Just don't do it at all. This whole block can become:

```rust
if let Some(ident) = self.get_ident(entid) {
    attribute_map.insert(values::DB_IDENT.clone(), edn::Value::NamespacedKeyword(ident.clone()));
}
```

core/src/lib.rs
```diff
@@ -293,38 +298,37 @@ impl Schema {
...
         let mut all_values = Vec::new();
         for (entid, attribute) in &self.schema_map {
             let mut attribute_map: BTreeMap<edn::Value, edn::Value> = BTreeMap::default();
-            attribute_map.insert(edn::Value::NamespacedKeyword(NamespacedKeyword::new("db", "id")), edn::Value::Keyword(Keyword::new(entid.to_string())));
-
             let some_ident = self.get_ident(entid.clone());
             if some_ident.is_some() {
                 let ident = some_ident.unwrap();
-                attribute_map.insert(edn::Value::NamespacedKeyword(NamespacedKeyword::new("db", "ident")), edn::Value::NamespacedKeyword(ident.clone()));
+                attribute_map.insert(values::DB_IDENT.clone(), edn::Value::NamespacedKeyword(ident.clone()));
```
ncalexan: Sorry, I wasn't clear, and now I realize I wasn't even correct.  `:db/id` is special, in that it's not really "part of the store", it's syntax for the map notation that Datomic supports.  (See http://docs.datomic.com/transactions.html and my about-to-land PR for #180 for more on the map notation.)  It doesn't assign an ident (like `:db/ident`), it (optionally) specifies an entid for the resulting map.  That is, it's assigning an internal identifier (entid) rather than an external identifier (ident).

Now, I think it would be correct to produce `{:db/id entid}` all the time, but it ties the abstract symbolic representation that is all built on external identifiers (all the `values::*` are keywords -- idents -- not entids) that you produce in this function to the exact store (with the exact same internal identifiers -- entids).  It will be more useful to be able to transplant the representation across stores by not having the internal identifiers.

So: no `:db/id` at all, and a `:db/ident` only if there is a registered keyword.  Which is what I think you had!  Sorry to double back on you.

core/src/values.rs
```diff
@@ -42,8 +42,11 @@ lazy_static_namespaced_keyword_value!(DB_ALTER_ATTRIBUTE, "db.alter", "attribute
 lazy_static_namespaced_keyword_value!(DB_CARDINALITY, "db", "cardinality");
 lazy_static_namespaced_keyword_value!(DB_CARDINALITY_MANY, "db.cardinality", "many");
 lazy_static_namespaced_keyword_value!(DB_CARDINALITY_ONE, "db.cardinality", "one");
+lazy_static_namespaced_keyword_value!(DB_FULLTEXT, "db", "fulltext");
```
ncalexan: Very good.

core/src/lib.rs
```diff
@@ -298,6 +298,7 @@ impl Schema {
         self.get_entid(x).map(|e| self.is_attribute(e)).unwrap_or(false)
     }
 
+    /// Returns an edn::Value::Vector representation of the schema map.
```
ncalexan: nit: `edn::value::Vector` in backticks.  (This will get formatted by Rustdoc eventually; it's some flavour of Markdown.)

nit: this uses the more than `schema_map`, so let's say "of the schema".  I think it's work saying that it returns a "symbolic representation", too; and maybe even "symbolic representation suitable for applying across Mentat stores".

core/src/lib.rs
```diff
@@ -291,12 +297,60 @@ impl Schema {
...
+            let mut attribute_map: BTreeMap<edn::Value, edn::Value> = BTreeMap::default();
+            let some_ident = self.get_ident(entid.clone());
+            if some_ident.is_some() {
+                let ident = some_ident.unwrap();
+                attribute_map.insert(values::DB_IDENT.clone(), edn::Value::NamespacedKeyword(ident.clone()));
+            } else {
+                attribute_map.insert(values::DB_IDENT.clone(), edn::Value::Keyword(Keyword::new(entid.to_string())));
+            }
+
+            let value_type = format!("{:?}", attribute.value_type);
```
rnewman: You're relying on the debug format (`{:?}`) of an enum here.

Given a `ValueType`, you're trying to get a `NamespacedKeyword`. There are two ways to do that:

- Go from the `ValueType` to an entid, and look up the ident.
- Go directly from a `ValueType` to a `NamespacedKeyword`.

Either way, break that out as a helper…

core/src/lib.rs
```diff
@@ -8,12 +8,18 @@
...
 extern crate ordered_float;
 
+extern crate edn;
+
+pub mod values;
+
 use std::collections::BTreeMap;
 use self::ordered_float::OrderedFloat;
 use self::edn::NamespacedKeyword;
+use self::edn::Keyword;
```
rnewman: See previous comment.

core/src/lib.rs
```diff
@@ -213,6 +233,35 @@ impl Attribute {
         }
         flags
     }
+
+    pub fn to_edn_value(&self, ident: Option<&NamespacedKeyword>) -> edn::Value {
```
rnewman: Make that `Option<NamespacedKeyword>`, so you don't need to use `ident.clone()` on line 240; see below for the neat trick when calling it.

core/src/lib.rs
```diff
@@ -291,12 +340,29 @@ impl Schema {
     pub fn identifies_attribute(&self, x: &NamespacedKeyword) -> bool {
         self.get_entid(x).map(|e| self.is_attribute(e)).unwrap_or(false)
     }
+
+    /// Returns an symbolic representation of the schema suitable for applying across Mentat stores.
+    pub fn to_edn_value(&self) -> edn::Value {
+        edn::Value::Vector((&self.schema_map).into_iter()
+            .map(|(entid, attribute)| 
+                attribute.to_edn_value(self.get_ident(*entid)))
```
rnewman: Neat trick: `self.get_ident(*entid).cloned()` turns an `Option<&NamespacedKeyword>` into `Option<NamespacedKeyword>`.

core/src/lib.rs
```diff
@@ -291,12 +340,29 @@ impl Schema {
...
+            .collect::<Vec<_>>())
```
rnewman: I'm surprised you need a turbofish when you're collecting into an `edn::Value::Vector`. Did you need this because you had this in a `let`, then reorganized?

core/src/lib.rs
```diff
@@ -291,12 +340,29 @@ impl Schema {
...
+    }
 }
 
 #[cfg(test)]
 mod test {
     use super::*;
 
+    fn associate_ident(schema: &mut Schema, i: NamespacedKeyword, e: Entid) {
+        schema.entid_map.insert(e, i.clone());
+        schema.ident_map.insert(i.clone(), e);
```
rnewman: No need for the second `clone`.

core/src/lib.rs
```diff
@@ -18,8 +18,9 @@ pub mod values;
 
 use std::collections::BTreeMap;
 use self::ordered_float::OrderedFloat;
-use self::edn::NamespacedKeyword;
-use self::edn::Keyword;
+use self::edn::{
+    NamespacedKeyword,
```
rnewman: Hah, didn't realize the other one was no longer needed!

core/src/lib.rs
```diff
@@ -291,12 +339,29 @@ impl Schema {
     pub fn identifies_attribute(&self, x: &NamespacedKeyword) -> bool {
         self.get_entid(x).map(|e| self.is_attribute(e)).unwrap_or(false)
     }
+
+    /// Returns an symbolic representation of the schema suitable for applying across Mentat stores.
+    pub fn to_edn_value(&self) -> edn::Value {
+        edn::Value::Vector((&self.schema_map).into_iter()
```
rnewman: I'm surprised that this works — you're moving out of a borrowed thing. `into_iter` consumes it`self`.

rnewman: Can you make sure you can call `to_edn_value` twice in a row on the same `Schema`? E.g., a test that doing so twice yields the same result.

core/src/lib.rs
```diff
@@ -300,37 +343,10 @@ impl Schema {
...
-            }
-
-            if attribute.component {
-                attribute_map.insert(values::DB_IS_COMPONENT.clone(), edn::Value::Boolean(true));
-            }
-
-            all_values.push(edn::Value::Map(attribute_map));
-        }
-        edn::Value::Vector(all_values)
+        edn::Value::Vector((&self.schema_map).into_iter()
```
rnewman: Yeah, it's tricky.

The reason for doing it this way is that it's making it very clear what you're doing: turning a thing into an iterator, walking it in order (the only thing you can do with an iterator!), and collecting the results into a vector. It might initially seem less readable than a loop and a mutable collection — that's closures for ya — but it imposes obvious compiler-enforced constraints on the program that _later_ mean you can read it with much more confidence.

Having a `let mut foo = Vec::new()` _implies_ to a Rusty reader several things:

- You don't know how big the output will be, otherwise you'd use `with_capacity`. (This is a sidenote, 'cos of course you could use `w_c` here.)
- You're not doing something that could be expressed with `collect`, otherwise you would be using `collect` — e.g., you're adding two things in some steps of the iteration.
- You might be passing that mutable reference somewhere — manually implementing your own flat map, say, or collecting into the same vector from iterations over two things.

The `collect` form is also more efficient: `collect` can ask the iterator for a hint (see `size_hint`), and an iterator over a fixed collection knows its size, and can pass it through a `Map` iterator. You'd need to manually remember to use `with_capacity` to get the same effect. Rust closures are very cheap, usually eliminated entirely by the compiler, so there's no perf worry here.

In a broader context, you'll find that the iterator/map form is amenable to switching to parallel computation with a single-word change; a `for` loop not so much, precisely because of the constraints the iterator approach opts into. Fun stuff!

core/src/lib.rs
fluffyemily: so, without the `&` in front of the `self.schema_map` call, I get the following error:
```
file: 'file:///Users/emilytoop/Development/mentat/core/src/lib.rs'
severity: 'Error'
message: 'E0507: cannot move out of borrowed content
  label: cannot move out of borrowed content'
at: '345,29'
source: '
```

Turning the `schema_map` into a reference (like I did when I used it in the loop previously) was the only way I could then call `into_iter` without that error. I am not sure how else I should have done it. Looking at all other references to `into_iter` elsewhere in the code, it is always either called on a reference to an object, or directly on a locally created object. If I have misunderstood something fundamental around ownership and references here, please let me know what it is!

core/src/lib.rs
```diff
@@ -341,6 +406,64 @@ mod test {
...
+{   :db/ident     :foo/bas
+    :db/valueType :db.type/string
+    :db/cardinality :db.cardinality/many
+    :db/unique :db.unique/identity
+    :db/fulltext true },
+{   :db/ident     :foo/bat
+    :db/valueType :db.type/boolean
+    :db/cardinality :db.cardinality/one
+    :db/component true }, ]"#;
+        let expected_value = edn::parse::value(&expected_output).expect("to be able to parse").without_spans();
```
fluffyemily: Just relooking at my test and I think I have discovered a bug.

In `attr2`, I create a `Unique::Value` unique key, and in `attr3` I create a `Unique::Identity` unique key. In my expected output, however, I expect `db.unique/identity` for `attr2` and no uniqueness for `attr3`.  And this test passes.

I need to correct my `to_edn_value` code inside `Attribute` to 
```
        if self.unique == Some(attribute::Unique::Value) {
            attribute_map.insert(values::DB_UNIQUE.clone(), values::DB_UNIQUE_VALUE.clone());
        } else if self.unique == Some(attribute::Unique::Identity) {
            attribute_map.insert(values::DB_UNIQUE.clone(), values::DB_UNIQUE_IDENTITY.clone());
        }
```

and my expected output to
```
r#"[ {   :db/ident     :foo/bar
    :db/valueType :db.type/ref
    :db/cardinality :db.cardinality/one
    :db/index true },
{   :db/ident     :foo/bas
    :db/valueType :db.type/string
    :db/cardinality :db.cardinality/many
    :db/unique :db.unique/value
    :db/fulltext true },
{   :db/ident     :foo/bat
    :db/valueType :db.type/boolean
    :db/cardinality :db.cardinality/one
    :db/unique :db.unique/identity
    :db/component true }, ]"#;
```

core/src/lib.rs
rnewman: There are two different ways to get an iterator: `into_iter` consumes the container, giving you an iterator over `T`, and `iter`, which _doesn't_ consume the container, but gives you an iterator over `&T`.

In general, use `into_iter` to consume temporary values or transform sequences. Use `iter` when you just want to peek at the contents.

core/src/lib.rs
```diff
@@ -244,6 +244,8 @@ impl Attribute {
         attribute_map.insert(values::DB_CARDINALITY.clone(), if self.multival { values::DB_CARDINALITY_MANY.clone() } else { values::DB_CARDINALITY_ONE.clone() });
 
         if self.unique == Some(attribute::Unique::Value) {
+            attribute_map.insert(values::DB_UNIQUE.clone(), values::DB_UNIQUE_VALUE.clone());
+        } else if self.unique == Some(attribute::Unique::Identity) {
```
rnewman:
```rust
match self.unique {
    Some(attribute::Unique::Value) => {
        attribute_map.insert…
    },
    Some(attribute::Unique::Identity) => …,
    None => {},
}
```

If we later add an additional type of uniqueness, the `match` form will tell you about an unhandled case — which it would also have here if you missed out `Identity`.

Issue #386, opened 2017-03-22
Use SELECT DISTINCT for multi-row projections


rnewman:
See comments in `query/src/lib.rs`.

Datalog returns distinct results, which means most of the time — unless we're returning only one result, or we can deduce that duplicates are impossible — we need to use `SELECT DISTINCT`.

PR #387, opened 2017-03-22
Correctly generate DISTINCT and LIMIT. (#386)

rnewman:
- The projector figures out whether distinct is needed from the find spec. Scalar and tuple don't need distinct. Rel and coll do.
- If an external limit was provided, and it's `1`, we don't need distinct.
- The translator pulls the value out of the projector and into the SQL intermediate format.
- The SQL outputter writes out the right string.

If, in the future, we add a form of projection for which there is no 1:1 correspondence between SQL rows and Datalog results (_e.g._, in-memory programmatic aggregation), then we will need to revise some of this — user-provided limits can no longer be used in the SQL query. That work can be driven quite easily by tests, and we don't know if we'll ever reach that point.

### DIFF
--- a/query-translator/tests/translate.rs
+++ b/query-translator/tests/translate.rs
@@ -49,48 +49,63 @@ fn translate<T: Into<Option<u64>>>(schema: &Schema, input: &'static str, limit: 
     select.query.to_sql_query().unwrap()
 }
 
-#[test]
-fn test_coll() {
+fn prepopulated_schema() -> Schema {
     let mut schema = Schema::default();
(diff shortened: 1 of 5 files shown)


query-translator/tests/translate.rs
```diff
@@ -49,48 +49,63 @@ fn translate<T: Into<Option<u64>>>(schema: &Schema, input: &'static str, limit:
     select.query.to_sql_query().unwrap()
 }
 
-#[test]
-fn test_coll() {
+fn prepopulated_schema() -> Schema {
```
ncalexan: Thanks, I was going to request this at various times and figured you'd get there at some point.  You did!

PR #390, opened 2017-03-23
Crudely parse `or` and `or-join`. (#388)

rnewman:
I'm really unhappy about the amount of boilerplate in this parser. But it works.

Note that unlike DataScript's parser, this one doesn't validate the contents of the variable lists.

### DIFF
--- a/query-algebrizer/src/errors.rs
+++ b/query-algebrizer/src/errors.rs
@@ -40,6 +40,12 @@ error_chain! {
             description("invalid argument")
             display("invalid argument to {}: expected numeric in position {}.", function, position)
         }
+
+        NonMatchingVariablesInOrClause {
+            // TODO: flesh out.
+            description("non-matching variables in 'or' clause")
+            display("non-matching variables in 'or' clause")
+        }
     }
 }
 
--- a/query-algebrizer/src/lib.rs
+++ b/query-algebrizer/src/lib.rs
@@ -75,7 +75,7 @@ pub fn algebrize(schema: &Schema, parsed: FindQuery) -> Result<AlgebraicQuery> {
     let mut cc = cc::ConjoiningClauses::default();
     let where_clauses = parsed.where_clauses;
     for where_clause in where_clauses {
-        cc.apply_clause(schema, where_clause);
+        cc.apply_clause(schema, where_clause)?;
     }
 
     let limit = if parsed.find_spec.is_unit_limited() { Some(1) } else { None };
--- a/query-algebrizer/src/lib.rs
+++ b/query-algebrizer/src/lib.rs
@@ -75,15 +75,7 @@ pub fn algebrize(schema: &Schema, parsed: FindQuery) -> Result<AlgebraicQuery> {
     let mut cc = cc::ConjoiningClauses::default();
     let where_clauses = parsed.where_clauses;
     for where_clause in where_clauses {
-        match where_clause {
-            WhereClause::Pattern(p) => {
-                cc.apply_pattern(schema, p);
-            },
-            WhereClause::Pred(p) => {
-                cc.apply_predicate(schema, p)?;
-            },
-            _ => unimplemented!(),
-        }
+        cc.apply_clause(schema, where_clause);
     }
 
     let limit = if parsed.find_spec.is_unit_limited() { Some(1) } else { None };
--- a/query-algebrizer/src/lib.rs
+++ b/query-algebrizer/src/lib.rs
@@ -16,6 +16,7 @@ extern crate mentat_query;
 
 mod errors;
 mod types;
+mod validate;
 mod cc;
 
 use mentat_core::{
--- a/query-algebrizer/src/validate.rs
+++ b/query-algebrizer/src/validate.rs
@@ -13,7 +13,6 @@ use std::collections::BTreeSet;
 use mentat_query::{
     ContainsVariables,
     OrJoin,
-    Pattern,
     Variable,
     WhereClause,
     UnifyVars,
@@ -54,9 +53,9 @@ pub fn validate_or_join(or_join: &OrJoin) -> Result<()> {
                 Ok(())
             } else {
                 let mut clauses = or_join.clauses.iter();
-                let template = clauses.next().unwrap().mentioned_variables();
+                let template = clauses.next().unwrap().collect_mentioned_variables();
                 for clause in clauses {
-                    if template != clause.mentioned_variables() {
+                    if template != clause.collect_mentioned_variables() {
                         bail!(ErrorKind::NonMatchingVariablesInOrClause);
                     }
                 }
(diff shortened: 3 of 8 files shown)


query-algebrizer/src/errors.rs
```diff
@@ -40,6 +40,12 @@ error_chain! {
             description("invalid argument")
             display("invalid argument to {}: expected numeric in position {}.", function, position)
         }
+
+        NonMatchingVariablesInOrClause {
+            // TODO: flesh out.
+            description("non-matching variables in 'or' clause")
```
ncalexan: How does `non-matching` differ from `unmatched`?  Or `mismatched`?

query/src/lib.rs
```diff
@@ -515,3 +516,79 @@ pub struct FindQuery {
     pub where_clauses: Vec<WhereClause>,
     // TODO: in_rules;
 }
+
+pub trait ContainsVariables {
+    fn acc_mentioned_variables(&self, acc: &mut BTreeSet<Variable>);
```
ncalexan: Consider `accumulate_mentioned_variables`.

ncalexan: Or `mentioned_variables`, and then make the other function `collect_mentioned_variables`.

query-algebrizer/src/validate.rs
```diff
@@ -0,0 +1,234 @@
...
+                        :where (or [?artist :artist/type :artist.type/group]
+                                   (and [?artist :artist/type :artist.type/person]
+                                        [?artist :artist/gender :artist.gender/female]))]"#;
+        let parsed = parse_find_string(query).unwrap();
+        let clauses = valid_or_join(parsed, UnifyVars::Implicit);
+
+        // Let's do some detailed parse checks.
+        let mut arms = clauses.into_iter();
+        match (arms.next(), arms.next(), arms.next()) {
+            (Some(left),
```
ncalexan: nit: one line or different indentation.  (Maybe this is an artifact of GH's layout, but it looks dedented one space.)

query-algebrizer/src/validate.rs
```diff
@@ -0,0 +1,234 @@
...
+
+    /// Test that an `or` with differing variable sets in each arm will fail to validate.
+    #[test]
+    fn test_invalid_implicit_or() {
+        let query = r#"[:find [?artist ...]
+                        :where (or [?artist :artist/type :artist.type/group]
+                                   [?artist :artist/type ?type])]"#;
+        let parsed = parse_find_string(query).unwrap();
+        match parsed.where_clauses.into_iter().next().unwrap() {
+            WhereClause::OrJoin(or_join) => assert!(validate_or_join(&or_join).is_err()),
```
ncalexan: This style can be quite difficult to debug when you break the tests, since you'll get unhelpful panics from the `unwrap` invocations.

query-algebrizer/src/validate.rs
```diff
@@ -0,0 +1,234 @@
...
+            _ => panic!(),
+        }
+    }
+
+    /// Test that two arms of an `or-join` can contain different variables if they both
```
ncalexan: These descriptions really helped.  Thanks!

query-algebrizer/src/lib.rs
```diff
@@ -75,15 +75,7 @@ pub fn algebrize(schema: &Schema, parsed: FindQuery) -> Result<AlgebraicQuery> {
...
-        match where_clause {
-            WhereClause::Pattern(p) => {
-                cc.apply_pattern(schema, p);
-            },
-            WhereClause::Pred(p) => {
-                cc.apply_predicate(schema, p)?;
-            },
-            _ => unimplemented!(),
-        }
+        cc.apply_clause(schema, where_clause);
```
ncalexan: This ignores the result, doesn't it?  You want the trailing `?` or some other chaining operator.

ncalexan: I'm surprised the compiler doesn't yell about this.

query-parser/src/parse.rs
```diff
@@ -110,21 +113,104 @@ def_value_satisfy_parser_fn!(Where,
...
+def_value_parser_fn!(Where, or, (), input, {
+    matches_plain_symbol!("or", input)
+});
+
+def_value_parser_fn!(Where, or_join, (), input, {
+    matches_plain_symbol!("or-join", input)
+});
+
+def_value_parser_fn!(Where, rule_vars, Vec<Variable>, input, {
+    satisfy_map(|x: edn::Value| {
```
ncalexan: It took a long time, but I think I've figured out improvements to this pattern, which should allow us to write something like:
```rust
or(vector(), list()).of(many1::<Vec<_>, _>(Query::variable())
```
which I think is a significant improvement.

query/src/lib.rs
```diff
@@ -474,13 +474,30 @@ pub struct Predicate {
     pub args: Vec<FnArg>,
 }
 
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum UnifyVars {
+    Implicit,
```
ncalexan: Worth a comment on the difference here.  Am I correct that `Implicit` is equivalent to `Explicit` with the complete set of variables occurring in all (equivalently, each) clause?

query-algebrizer/src/errors.rs
rnewman: I think you're asking about wording.

IMO there's not a lot of clarity in the distinction, if any, but:

- "Unmatched variables" kinda means we were looking for one or two and didn't get them.
- "Mismatched variables" means we were combining two things and they didn't match. A mismatch is bidirectional.
- "Non-matching variables" implies quite weakly that some expected property didn't hold.

That's the weak reasoning behind this wording: this error means that the expected property — that every arm in the `or` uses all of the variables in the rule vars, which might themselves be derived from the arms — doesn't hold. There can be an arbitrary number of arms.

query-algebrizer/src/validate.rs
rnewman: Line 135 has a relative 4-space indent, but `match` is five characters long.

query-algebrizer/src/lib.rs
rnewman: Me too!

query/src/lib.rs
rnewman: I did both: `accumulate_mentioned_variables`, called by `collect_mentioned_variables`.

query-algebrizer/src/validate.rs
rnewman: Oh, I see — you were referring to 136. Fixed.

query-algebrizer/src/validate.rs
rnewman: Switched for `expect`s.

rnewman:
8adb6d9

Issue #268, opened 2017-02-08
Test schema validations


jsantell:
Follow up from https://github.com/mozilla/mentat/pull/267#issuecomment-278481471

ncalexan:
This is a pretty _good first bug_, and gets easier after #260 is implemented.

What you'll need to do is define some `Schema` values, either manually (creating `IdentMap` and `Attribute` instances in code), or by using #260 to parse `edn::Value` into a "symbolic schema" and then turning that into a `Schema`.  Test the good case (validations pass!) and the bad cases (validations fail!) using `Schema::validate_schema` (see the code around https://github.com/mozilla/mentat/blob/42580539b8a58f9d10e5891a22112f8c1b697bff/db/src/schema.rs#L18).

PR #391, opened 2017-03-28
Add tests for `validate_schema_map`

fluffyemily:
#268

### DIFF
--- a/db/src/schema.rs
+++ b/db/src/schema.rs
@@ -250,3 +250,88 @@ impl SchemaTypeChecking for Schema {
         }
     }
 }
+
+
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use self::edn::NamespacedKeyword;
+
+    fn add_attribute(schema: &mut Schema, 
+            ident: NamespacedKeyword, 
+            entid: Entid, 
+            index: bool,
+            value_type: ValueType,
+            fulltext: bool,
+            unique: Option<attribute::Unique>,
+            multival: bool,
+            component: bool) {
+
+        schema.entid_map.insert(entid, ident.clone());
+        schema.ident_map.insert(ident.clone(), entid);
+
+        schema.schema_map.insert(entid, Attribute {
+            index: index,
+            value_type: value_type,
+            fulltext: fulltext,
+            unique: unique,
+            multival: multival,
+            component: component,
+        });
+    }
+
+    #[test]
+    fn validate_schema_map_success() {
+        let mut schema = Schema::default();
+        // attribute that is not an index has no uniqueness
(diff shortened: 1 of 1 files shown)


db/src/schema.rs
```diff
@@ -250,3 +250,88 @@ impl SchemaTypeChecking for Schema {
...
+        assert!(validate_schema_map(&schema.entid_map, &schema.schema_map).is_err());
+    }
+
+    #[test]
+    fn invalid_schema_component_not_ref() {
+        let mut schema = Schema::default();
+        // attribute that is a component is not a `Ref`
+        add_attribute(&mut schema, NamespacedKeyword::new("foo", "bar"), 97, false, ValueType::Boolean, false, None, false, true);
+        
+        assert!(validate_schema_map(&schema.entid_map, &schema.schema_map).is_err());
```
rnewman: @ncalexan at this point would say: could you check which error is returned?

You can use `.err()` to get the error out:

```
assert_eq!(Some(MyErrorType::Foo(foo_bar.clone()), validate_schema_map(…).err());
```

db/src/schema.rs
```diff
@@ -250,3 +250,88 @@ impl SchemaTypeChecking for Schema {
...
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use self::edn::NamespacedKeyword;
+
+    fn add_attribute(schema: &mut Schema, 
+            ident: NamespacedKeyword, 
+            entid: Entid, 
+            index: bool,
```
rnewman: Rather than all of these opaque positional arguments, could you just take the ident/entid and an `Attribute` instance? That allows callers to name the fields as they construct the attribute.

db/src/schema.rs
```diff
@@ -250,3 +250,185 @@ impl SchemaTypeChecking for Schema {
...
+            unique: Some(attribute::Unique::Value),
+            multival: false,
+            component: false,
+        });
+        
+        let err = validate_schema_map(&schema.entid_map, &schema.schema_map).err();
+        assert!(err.is_some());
+        
+        match err.unwrap() {
+            Error(ErrorKind::BadSchemaAssertion(message), _) => { assert_eq!(message, ":db/unique :db/unique_value without :db/index true for entid: :foo/bar"); },
```
rnewman: You can do this:

```
let err = validate_schema_map(…).err();
assert_eq!(err,
    Some(Error(ErrorKind::BadSchemaAssertion(":db/unique :db/unique_value without :db/index true for entid: :foo/bar"), _)));
```

fluffyemily: Ah, I tried almost that (without the `Some`) and it kept telling me I couldn't use `==` on `Error` as it didn't implement `PartialEq`, so i came up with this. It didn't occur to me to use `Some`. Doh!

Issue #337, opened 2017-02-20
[edn] Make EDN allocation more flexible


ncalexan:
This is a general ticket tracking making our EDN representation more efficient.  We have talked about interning in a variety of contexts (see, for example, #318) but expect to need to intern at the EDN level.  So this ticket tracks making our EDN parser take a factory of some sort.

There's a nice builder/allocator pattern for this in `pretty.rs` which I think is worth understanding: see https://github.com/freebroccolo/pretty.rs/blob/bb5e5aa3e93485c8b1c8314ff24b189deca0301b/src/pretty/lib.rs#L41

ncalexan:
We've dropped links before but the following might be useful:
- https://github.com/servo/string-cache
- https://github.com/SimonSapin/rust-typed-arena

ncalexan:
See also https://github.com/insaneinside/symtern.

PR #393, opened 2017-03-29
Improve parsing of nested `edn::ValueAndSpan` streams

ncalexan:
This is work in progress, but you can see how it simplifies a lot of `tx-parser/`.


ncalexan:
This is not a perfect PR by any means, but the expression is so much better than the old code (in places) that it's worth moving forward and fixing up the remaining issues.

I'm quite worried that we're now cloning our `edn::SpanAndValue` instances _all over the place_, but hopefully we can address that with some strategic `Rc` usage.  (And/or #337.)

parser-utils/src/value_and_span.rs
```diff
@@ -0,0 +1,282 @@
...
+
+use edn;
+
+/// A wrapper to let us order `edn::Span` in whatever way is appropriate for parsing with `combine`.
+#[derive(Clone, Debug)]
+pub struct SpanPosition(edn::Span);
+
+impl Display for SpanPosition {
+    fn fmt(&self, f: &mut Formatter) -> ::std::fmt::Result {
+        write!(f, "{:?}", self.0)
```
rnewman: Is this equivalent to `self.0.fmt(f)`?

parser-utils/src/value_and_span.rs
```diff
@@ -0,0 +1,282 @@
...
+    fn update_position(&self, position: &mut Self::Position) {
+        *position = SpanPosition(self.span.clone())
+    }
+
+    fn into_child_stream_iter(self) -> IntoIter {
+        match self.inner {
+            edn::SpannedValue::Vector(values) => IntoIter::Vector(values.into_iter()),
+            edn::SpannedValue::List(values) => IntoIter::List(values.into_iter()),
+            // Parsing pairs with `combine` is tricky; parsing sequences is easy.
+            edn::SpannedValue::Map(map) => IntoIter::Map(map.into_iter().flat_map(|(a, v)| std::iter::once(a).chain(std::iter::once(v))).collect::<Vec<_>>().into_iter()),
```
rnewman: File an issue to improve

parser-utils/src/value_and_span.rs
```diff
@@ -0,0 +1,282 @@
...
+    fn into_child_stream(self) -> Stream;
+    fn into_atom_stream_iter(self) -> IntoIter;
+    fn into_atom_stream(self) -> Stream;
+}
+
+impl Item for edn::ValueAndSpan {
+    type Position = SpanPosition;
+
+    fn start(&self) -> Self::Position {
+        SpanPosition(self.span.clone())
```
rnewman: I think `Span` should derive `Copy`. After all, it's a pair of `usize`…

That makes all of these `clone` calls go away.

I quote:

> A type can implement `Copy` if all of its components implement `Copy`…
> … A struct can be `Copy`, and `i32` is `Copy`, therefore `Point` is eligible to be `Copy` …
> Generally speaking, if your type can implement `Copy`, it should.

If you change `Span` to be a pair of `u32`, then Rust should make that a single 64-bit value on 64-bit platforms, and the copy should be as cheap as can be. Even as a pair of `usize`, the copy is very cheap.

parser-utils/src/value_and_span.rs
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    fn update_position(&self, position: &mut Self::Position) {
+        *position = SpanPosition(self.span.clone())
```
rnewman: What a fascinatingly backwards way of expressing this! 😁

parser-utils/src/value_and_span.rs
```diff
@@ -0,0 +1,282 @@
...
+        }
+    }
+
+    fn add_error(&mut self, errors: &mut ParseError<Self::Input>) {
+        self.0.add_error(errors);
+    }
+}
+
+#[inline(always)]
+pub fn of<P, N, O>(p: P, n: N) -> Of<P, Skip<N, Eof<Stream>>>
```
rnewman: Further to our discussion the other day, I think this might best be named `of_only` in order to emphasize the `skip(eof())`. Concur?

```
vector().of_only(Query::variable)
```
?

parser-utils/src/value_and_span.rs
```diff
@@ -0,0 +1,282 @@
...
+use combine::combinator::{
+    Eof,
+    Skip,
+};
+
+use edn;
+
+/// A wrapper to let us order `edn::Span` in whatever way is appropriate for parsing with `combine`.
+#[derive(Clone, Debug)]
+pub struct SpanPosition(edn::Span);
```
rnewman: When you make `Span` `Copy`, do so here, too.

parser-utils/src/value_and_span.rs
```diff
@@ -0,0 +1,282 @@
...
+/// to `combine::IteratorStream` as produced by `combine::from_iter`, but specialized to
+/// `edn::ValueAndSpan`.
+#[derive(Clone)]
+pub struct Stream(IntoIter, SpanPosition);
+
+/// Things specific to parsing with `combine` and our `Stream` that need a trait to live outside of
+/// the `edn` crate.
+pub trait Item: Clone + PartialEq + Sized {
+    /// Position could be specialized to `SpanPosition`.
+    type Position: Clone + Ord + std::fmt::Display;
```
rnewman: If you require `Copy` too…

parser-utils/src/value_and_span.rs
```diff
@@ -0,0 +1,282 @@
...
+                x.update_position(&mut self.1);
+                Ok(x)
+            },
+            None => Err(primitives::Error::end_of_input()),
+        }
+    }
+
+    #[inline(always)]
+    fn position(&self) -> Self::Position {
+        self.1.clone()
```
rnewman: … `Copy` makes this nicer.

parser-utils/src/value_and_span.rs
```diff
@@ -280,3 +291,107 @@ pub fn value(value: edn::Value) -> Box<Parser<Input=Stream, Output=edn::ValueAnd
     // trees together, we could avoid creating garbage.
     satisfy(move |v: edn::ValueAndSpan| value == v.inner.into()).boxed()
 }
+
+pub fn keyword_map_(input: Stream) -> ParseResult<edn::ValueAndSpan, Stream>
+{
+    // One run is a keyword followed by at one or more non-keywords.
```
rnewman: "at one or more"

parser-utils/src/value_and_span.rs
```diff
@@ -280,3 +291,107 @@ pub fn value(value: edn::Value) -> Box<Parser<Input=Stream, Output=edn::ValueAnd
...
+    let run = (satisfy(|v: edn::ValueAndSpan| v.inner.is_keyword()),
```
rnewman: Do these really need their type annotations?

parser-utils/src/value_and_span.rs
rnewman: Example doc comment please!

db/src/tx.rs
```diff
@@ -254,13 +254,13 @@ impl<'conn, 'a> Tx<'conn, 'a> {
 
                     let v = match v {
                         entmod::AtomOrLookupRefOrVectorOrMapNotation::Atom(v) => {
-                            if attribute.value_type == ValueType::Ref && v.is_text() {
-                                Either::Right(LookupRefOrTempId::TempId(temp_ids.intern(v.as_text().cloned().map(TempId::External).unwrap())))
+                            if attribute.value_type == ValueType::Ref && v.inner.is_text() {
```
rnewman: Can we simplify some of this by making `ValueAndSpan` implement `Deref<Value>`? That way you can treat it like a `Value` for method call purposes…

tx-parser/src/lib.rs
```diff
@@ -39,165 +42,114 @@ use mentat_tx::entities::{
...
-
-def_parser_fn!(Tx, atom_or_lookup_ref_or_vector, Value, AtomOrLookupRefOrVectorOrMapNotation, input, {
-    Tx::<I>::lookup_ref().map(AtomOrLookupRefOrVectorOrMapNotation::LookupRef)
-        .or(Tx::<I>::nested_vector().map(AtomOrLookupRefOrVectorOrMapNotation::Vector))
-        .or(Tx::<I>::map_notation().map(AtomOrLookupRefOrVectorOrMapNotation::MapNotation))
-        .or(Tx::<I>::atom().map(AtomOrLookupRefOrVectorOrMapNotation::Atom))
-        .parse_lazy(input)
-        .into()
+def_parser!(Tx, atom, edn::ValueAndSpan, {
+    satisfy_map(|x: edn::ValueAndSpan| x.into_atom().map(|v| v))
```
rnewman: What's that identity `map` for?

tx-parser/src/lib.rs
```diff
@@ -39,165 +42,114 @@ use mentat_tx::entities::{
...
-    })
-}
-def_value_satisfy_parser_fn!(Tx, map_notation, MapNotation, value_to_map_notation);
+def_parser!(Tx, add_or_retract, Entity, {
+    let add = value(edn::Value::NamespacedKeyword(NamespacedKeyword::new("db", "add")))
+        .map(|_| OpType::Add);
+    let retract = value(edn::Value::NamespacedKeyword(NamespacedKeyword::new("db", "retract")))
+        .map(|_| OpType::Retract);
+    let p = (add.or(retract),
+             Tx::entid_or_lookup_ref_or_temp_id(),
```
rnewman: it s beautiful

tx-parser/src/lib.rs
```diff
@@ -39,165 +42,114 @@ use mentat_tx::entities::{
...
+
+    vector().of(p)
+});
 
-def_parser_fn!(Tx, entity, Value, Entity, input, {
-    let mut p = Tx::<I>::add_or_retract()
-        .or(Tx::<I>::map_notation().map(Entity::MapNotation));
-    p.parse_stream(input)
+def_parser!(Tx, map_notation, MapNotation, {
+    map()
```
rnewman: Mm, confusing name.

tx-parser/src/lib.rs
```diff
@@ -253,108 +210,107 @@ mod tests {
...
-        map.insert(kw("db", "id"), Value::Text("t".to_string()));
-        map.insert(kw("db", "ident"), kw("test", "attribute"));
-        let input = [Value::Map(map.clone())];
-        let mut parser = Tx::entity();
-        let result = parser.parse(&input[..]);
-        assert_eq!(result,
-                   Ok((Entity::MapNotation(expected),
-                       &[][..])));
-    }
+    // #[test]
```
rnewman: File an issue to fix these.

query-parser/src/parse.rs
```diff
@@ -83,320 +96,348 @@ error_chain! {
...
+
+def_parser!(Where, pattern_non_value_place, PatternNonValuePlace, {
+    satisfy_map(PatternNonValuePlace::from_value)
+});
+
+
+def_parser!(Where, and, edn::ValueAndSpan, {
+    satisfy(|v: edn::ValueAndSpan| {
+        if let edn::SpannedValue::PlainSymbol(ref s) = v.inner {
+            s.0.as_str() == "and"
```
rnewman: Can we keep `matches_plain_symbol` or similar?

query-parser/src/parse.rs
```diff
@@ -83,320 +96,348 @@ error_chain! {
...
+            false
+        }
+    })
 });
 
-def_value_parser_fn!(Where, or_join, (), input, {
-    matches_plain_symbol!("or-join", input)
+def_parser!(Where, rule_vars, Vec<Variable>, {
+    seq()
+        .of(many1(Query::variable()))
```
rnewman: Join lines

query-parser/src/parse.rs
```diff
@@ -83,320 +96,348 @@ error_chain! {
...
-    .skip(eof())
-    .parse_stream(input)
 });
 
-pub struct Find<I>(::std::marker::PhantomData<fn(I) -> I>);
+pub struct Find;
 
-def_value_parser_fn!(Find, period, (), input, {
-    matches_plain_symbol!(".", input)
+/// TODO: extract macro for matching these `PlainSymbol` instances.
```
rnewman: Heh, there we go.

rnewman:
```
macro_rules! def_matches_plain_symbol {
  ( $struct: ty, $name: ident, $match: expr ) => {
    {
      def_parser!($struct, $name, edn::ValueAndSpan, {
          satisfy(|v: edn::ValueAndSpan| {
              match v.inner {
                  edn::SpannedValue::PlainSymbol(ref s) => s.0.as_str() == $match,
                  _ => false,
              }
          }
      }
}}
```

should do the trick…

query-parser/src/parse.rs
```diff
@@ -83,320 +96,348 @@ error_chain! {
...
+            false
+        }
+    })
 });
 
-def_value_parser_fn!(Find, find_scalar, FindSpec, input, {
+def_parser!(Find, find_scalar, FindSpec, {
     Query::variable()
+        .map(|var| FindSpec::FindScalar(Element::Variable(var)))
         .skip(Find::period())
```
rnewman: I don't think this is in the right place. You should skip the `period` before the `.map` — you're matching 

```edn
[:find ?x . :in …]
```

In this order, we'll construct the `FindScalar` instance, then fail to match the period in the failure case, which isn't free.

query-parser/src/parse.rs
```diff
@@ -83,320 +96,348 @@ error_chain! {
...
-                        .map(|var| FindSpec::FindColl(Element::Variable(var)));
-            let r: ParseResult<FindSpec, _> = p.parse_lazy(&y[..]).into();
-            Query::to_parsed_value(r)
-        })
-        .parse_stream(input)
+def_parser!(Find, find_coll, FindSpec, {
+    vector()
+        .of(Query::variable()
+            .map(|var| FindSpec::FindColl(Element::Variable(var)))
+            .skip(Find::ellipsis()))
```
rnewman: Same here.

query-parser/src/parse.rs
```diff
@@ -83,320 +96,348 @@ error_chain! {
...
+    vector()
+        .of(Find::elements().map(FindSpec::FindTuple))
 });
 
-def_value_parser_fn!(Find, find, FindSpec, input, {
-    // Any one of the four specs might apply, so we combine them with `choice`.
-    // Our parsers consume input, so we need to wrap them in `try` so that they
-    // operate independently.
-    choice::<[&mut Parser<Input = I, Output = FindSpec>; 4], _>
+/// Parse a stream values into one of four find specs.
```
rnewman: stream of values

query-parser/src/parse.rs
```diff
@@ -83,320 +96,348 @@ error_chain! {
...
+        .with(vector().of(Find::spec().map(FindQueryPart::FindSpec)));
+
+    let p_with_vars = Find::literal_with()
+        .with(vector().of(many(Query::variable()).map(FindQueryPart::With)));
+
+    let p_where_clauses = Find::literal_where()
+        .with(vector().of(Where::clauses().map(FindQueryPart::WhereClauses))).expected(":where clauses");
+
+    (or(map(), keyword_map()))
+        .of(many(choice::<[&mut Parser<Input = ValueStream, Output = FindQueryPart>; 3], _>([
```
rnewman: Hm.

I have to say, the components of this parser do not build up to be used like this; this looks like it's inefficient and produces worse error messages.

`or(map(), keyword_map())` is acceptable, I think, but beyond that I wouldn't use `choice`.

I think I'd write a `satisfy_map` parser that consumed the output of the `or`, programmatically looked at the expected keys, and processed the values accordingly using three individual parsers. Here you've got a much more expensive parse by using `choice` — it'll typically try `p_find_spec` four times, once for `:find`, once for `:in`, once for `:with`, and once for `:where`, then `p_with_vars` three times, and so on.

You're also losing the opportunity for the precisely typed, useful error messages that come out of the code you deleted, replacing them with whatever `combine` provides as an error when given this tree of parsers.

Given input like `{:in [] :where [[?x _ ?y]]}` the old code returned `ErrorKind::MissingFieldError(:find)`.

I believe this code now says "unexpected token: expected :find at position -2123981209243", or something like that. No span is meaningful here, and it's not an unexpected token…

Can we do better by accepting that this is not well phrased as a `combine` parser?

query-parser/src/parse.rs
```diff
@@ -83,320 +96,348 @@ error_chain! {
...
+                }
+            }
+
+            Ok(FindQuery {
+                find_spec: find_spec.clone().ok_or(combine::primitives::Error::Unexpected("expected :find".into()))?,
+                default_source: SrcVar::DefaultSrc,
+                with: with_vars.unwrap_or(vec![]),
+                in_vars: vec![],       // TODO
+                in_sources: vec![],    // TODO
+                where_clauses: where_clauses.ok_or(combine::primitives::Error::Unexpected("expected :where".into()))?,
```
rnewman: Wait a minute, doesn't this discard any parse error that comes out of parsing `p_where_clauses`, defeating the whole point of stacking this whole thing into parser-land?

parser-utils/src/value_and_span.rs
ncalexan: Done.

parser-utils/src/value_and_span.rs
ncalexan: Yes, I think it is.  Replaced.

parser-utils/src/value_and_span.rs
ncalexan: I think the tests are sufficient and the comment attached to the public function are sufficient.

parser-utils/src/value_and_span.rs
ncalexan: In general, no; in practice, often enough that I started doing it uniformly.

parser-utils/src/value_and_span.rs
ncalexan: I've done this, using `as u32` to unsafely cast `usize` to `u32`.  In practice, we won't be parsing 4Gb inputs.

parser-utils/src/value_and_span.rs
ncalexan: Yeah, this is a weirdness arising from trying to agree with `combine`.  It's probably not worth agreeing with `combine`, but it's also not a big issue to do it.

parser-utils/src/value_and_span.rs
ncalexan: OK, although I think this might be a `combine`-level issue.  https://github.com/mozilla/mentat/issues/400

parser-utils/src/value_and_span.rs
ncalexan: I'm fine with this, caveat that I prefer `of_exactly`.  I'll change this throughout.

db/src/tx.rs
ncalexan: I don't think we can, since we'd need to provide an `&Value`.  To do so you need to recurse into the children of `SpannedValue` in order to convert them to `&Value`, and I don't think we can produce an owned container with a sufficiently long lifetime to collect the deref-ed children.  I'd be happy to be proven wrong -- and thanks for bringing `Deref` to my attention!

tx-parser/src/lib.rs
ncalexan: History.  Gone now!

tx-parser/src/lib.rs
ncalexan: I just did it in this PR.

query-parser/src/parse.rs
ncalexan: I like this style.  I'll let you reformat as you like while you're working.

query-parser/src/parse.rs
ncalexan: Done.  Just FYI, `$struct` and `$match` are both forbidden, since they parse as keywords.

query-parser/src/parse.rs
ncalexan: I prefer the `map` to be as close to what's mapped as possible, but you're correct.  Since this isn't lazy, I'll revert, here and below.

query-parser/src/parse.rs
ncalexan: Done.

query-parser/src/parse.rs
ncalexan: Fixed.

query-parser/src/parse.rs
ncalexan: I'm going to punt this to follow-up.  I've lost the will to live with this patch :(

https://github.com/mozilla/mentat/issues/401

ncalexan:
@rnewman OK, back over to you.

PR #394, opened 2017-03-29
Split cc.rs, analyze or/or-join

rnewman:
Sorry for the big PR.

This includes:

- Some "pre" commits. The "am I a pattern?" one isn't used, so I'm on the fence about deleting it.
- Intersection and alternation for constraints. This will allow us to describe `OR` and `AND` in a CC's constraint list.
- `table_for_places` no longer automatically fails the pattern if a lookup fails, because we use this to consider the arms of an `or`.
- The logic for analyzing an `or`, complete with a ton of comments. This narrows down the entire space of `or` into:
  - Failed (no pattern can match)
  - Succeeded (empty to start with)
  - Unit and fully unified — treat as a single pattern
  - Simple
  - Complex.

The last commit is the coal face of applying simple `or`, which I hope to finish today. Complex `or` will be a subsequent PR.


rnewman:
Exhausted, but most of the way to bashing `Rc` into the codebase, @ncalexan.

query-algebrizer/src/types.rs
```diff
@@ -194,6 +194,88 @@ pub enum ColumnConstraint {
...
+    Alternation(ColumnAlternation),
+}
+
+impl From<ColumnConstraint> for ColumnConstraintOrAlternation {
+    fn from(thing: ColumnConstraint) -> Self {
+        ColumnConstraintOrAlternation::Constraint(thing)
+    }
+}
+
+/// An ordered collection of constraints that must all match.
```
ncalexan: Consider: "The `ColumnIntersection` constraint is satisfied if all of the given constraints are satisfied."  What happens if the underlying collection is empty?  Usually `and([])` is `true`, is that what you want?

query-algebrizer/src/types.rs
```diff
@@ -194,6 +194,88 @@ pub enum ColumnConstraint {
...
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    pub fn also(&mut self, constraint: ColumnConstraint) {
+        self.0.push(ColumnConstraintOrAlternation::Constraint(constraint));
+    }
+}
+
+/// An ordered collection of a collection of constraints, one of which must entirely match.
```
ncalexan: Consider: "The `ColumnAlteration` constraint is satisfied if at least one of the given constraints are satisfied."  What happens if the underlying vec is empty?  Usually `or([])` is `false`, is that what you want?

query-algebrizer/src/types.rs
```diff
@@ -194,6 +194,88 @@ pub enum ColumnConstraint {
...
+    type Item = ColumnIntersection;
+    type IntoIter = ::std::vec::IntoIter<ColumnIntersection>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.0.into_iter()
+    }
+}
+
+impl ColumnAlternation {
+    pub fn instead(&mut self, intersection: ColumnIntersection) {
```
ncalexan: `instead` implies replacement to me.  Why not `or`, `or_else`, or `alternately`?

query-sql/src/lib.rs
```diff
@@ -268,6 +271,15 @@ impl QueryFragment for Constraint {
                 Ok(())
             },
 
+            &Or { ref constraints } => {
+                out.push_sql("(");
```
ncalexan: Does SQLite do good things with `OR ()`?

query-algebrizer/src/types.rs
```diff
@@ -194,6 +194,88 @@ pub enum ColumnConstraint {
...
+
+impl From<ColumnConstraint> for ColumnConstraintOrAlternation {
+    fn from(thing: ColumnConstraint) -> Self {
+        ColumnConstraintOrAlternation::Constraint(thing)
+    }
+}
+
+/// An ordered collection of constraints that must all match.
+#[derive(PartialEq, Eq)]
+pub struct ColumnIntersection(pub Vec<ColumnConstraintOrAlternation>);
```
ncalexan: Your data types aren't symmetric: `Intersection` is many `Constraint|Alteration`, but `Alteration` is many `Intersection`.  This can express many `Constraint`, but it's another layer of indirection.

Is this just an accident of how the code grew?  Is this intentional?  I.e., are you intending to normalize to a conjunction-of-disjunctions?  (Hope I got that right!)

query-algebrizer/src/cc.rs
```diff
@@ -534,8 +531,11 @@ impl ConjoiningClauses {
     /// `is_known_empty`.
     fn alias_table<'s, 'a>(&mut self, schema: &'s Schema, pattern: &'a Pattern) -> Option<SourceAlias> {
```
ncalexan: I only see one consumer of `alias_table`, and it immediately checks for `None` and marks the CC as known empty (potentially for a different reason).  So: should this just be `Result` the whole way through, rather than introducing `Option` and then reverting to `Result`?  And do you expect to want to _chain_ `EmptyBecause` reasons to provide more feedback?

query/src/lib.rs
```diff
@@ -559,6 +559,22 @@ pub struct FindQuery {
...
 
+impl OrJoin {
+    /// Return true if either the `OrJoin` is `UnifyVars::Implicit`, or if
+    /// every variable mentioned inside the join is also mentioned in the `UnifyVars` list.
+    pub fn is_fully_unified(&self) -> bool {
+        match &self.unify_vars {
+            &UnifyVars::Implicit => true,
+            &UnifyVars::Explicit(ref vars) => {
+                // We know that the join list must be a subset of the vars in the pattern, or
+                // it would have failed validation. That allows us to simply compare counts here.
```
ncalexan: Sure, but consider `assert!`-ing the full set equality anyway.  Never know what you'll catch!

query-algebrizer/src/clauses/or.rs
```diff
@@ -78,3 +78,289 @@ fn _simply_matches_value_place(left: &PatternValuePlace, right: &PatternValuePla
...
+    Unit(OrWhereClause),
+    UnitPattern(Pattern),
+    Simple(Vec<Pattern>),
+    Complex(OrJoin),
+}
+
+/// Application of `or`. Note that this is recursive!
+impl ConjoiningClauses {
+    fn apply_or_join(&mut self, schema: &Schema, mut or_join: OrJoin) -> Result<()> {
+        // Simple optimization. Empty `or` clauses disappear. Unit `or` clauses
```
ncalexan: Ah, this partially answers my earlier questions about empty `Intersection` and `Alternation.`  You're going to enforce normalizations that make some of your choices and representations safe.

query-algebrizer/src/clauses/or.rs
```diff
@@ -78,3 +78,289 @@ fn _simply_matches_value_place(left: &PatternValuePlace, right: &PatternValuePla
...
+    /// - Only one pattern can match: the `or` can be simplified away.
+    fn deconstruct_or_join(&self, schema: &Schema, mut or_join: OrJoin) -> DeconstructedOrJoin {
+        // If we have explicit non-maximal unify-vars, we *can't* simply run this as a
+        // single pattern --
+        // ```
+        // [:find ?x :where [?x :foo/bar ?y] (or-join [?x] [?x :foo/baz ?y])]
+        // ```
+        // is *not* equivalent to
+        // ```
+        // [:find ?x :where [?x :foo/bar ?y] [?x :foo/baz ?y]]
```
ncalexan: It would be equivalent to `[:find ?x :where [?x :foo/bar ?y] [?x :foo/baz (gen-sym)]]`, right?  Is there value in deconstructing `or-join` clauses in this way generally?  (Perhaps you do this -- I will see.)

query-algebrizer/src/clauses/or.rs
```diff
@@ -78,3 +78,289 @@ fn _simply_matches_value_place(left: &PatternValuePlace, right: &PatternValuePla
...
+            // same CC. TODO: pay attention to the unify list when generating
+            // constraints. Temporarily shadow variables within each `or` branch.
+            return DeconstructedOrJoin::Complex(or_join);
+        }
+
+        match or_join.clauses.len() {
+            0 => DeconstructedOrJoin::KnownSuccess,
+
+            // It's safe to simply 'leak' the entire clause, because we know every var in it is
+            // supposed to unify with the enclosing form.
```
ncalexan: That's 'cuz of the `is_fully_unified` above, right?

query-algebrizer/src/clauses/or.rs
```diff
@@ -78,3 +78,289 @@ fn _simply_matches_value_place(left: &PatternValuePlace, right: &PatternValuePla
...
+            1 => DeconstructedOrJoin::Unit(or_join.clauses.pop().unwrap()),
+            _ => self._deconstruct_or_join(schema, or_join),
+        }
+    }
+
+    fn _deconstruct_or_join(&self, schema: &Schema, mut or_join: OrJoin) -> DeconstructedOrJoin {
```
ncalexan: Can we find a more informative name?  `deconstruct_multi_clause_or_join`?

query-algebrizer/src/clauses/or.rs
```diff
@@ -78,3 +78,289 @@ fn _simply_matches_value_place(left: &PatternValuePlace, right: &PatternValuePla
...
+        // reconstruct and return a complex `OrJoin`.
+        let mut patterns: Vec<Pattern> = Vec::with_capacity(or_join.clauses.len());
+
+        // Keep track of the table we need every pattern to use.
+        let mut expected_table: Option<DatomsTable> = None;
+
+        // Technically we might have several reasons, but we take the last -- that is, that's the
+        // reason we don't have at least one pattern!
+        // We'll return this as our reason if no pattern can return results.
+        let mut empty_because: Option<EmptyBecause> = None;
```
ncalexan: OK, this is fine, and addresses an earlier comment.

query-algebrizer/src/clauses/or.rs
ncalexan: A little note explaining the goal of this might help, or just changing `deconstruct` to something that's more like `simplify_...` or `to_simple_...`, since that's what you're really doing: trying to get a simple join at the end.

query-algebrizer/src/clauses/or.rs
```diff
@@ -78,3 +78,289 @@ fn _simply_matches_value_place(left: &PatternValuePlace, right: &PatternValuePla
...
+                                _simply_matches_place(&template.entity, &p.entity) &&
+                                _simply_matches_place(&template.attribute, &p.attribute) &&
+                                _simply_matches_value_place(&template.value, &p.value) &&
+                                _simply_matches_place(&template.tx, &p.tx);
+                        }
+
+                        // All of our clauses that _do_ yield a table -- that are possible --
+                        // must use the same table in order for this to be a simple `or`!
+                        if same_shape {
+                            if expected_table == Some(table) {
```
ncalexan: `match expected_table`?

query-algebrizer/src/clauses/or.rs
```diff
@@ -78,3 +78,289 @@ fn _simply_matches_value_place(left: &PatternValuePlace, right: &PatternValuePla
...
+
+            // If we get here, it means one of our checks above failed. Reconstruct and bail.
+            let reconstructed: Vec<OrWhereClause> =
+                // Non-empty patterns already collected…
+                patterns.into_iter()
+                        .map(|p| OrWhereClause::Clause(WhereClause::Pattern(p)))
+                // … then the clause we just considered…
+                        .chain(::std::iter::once(last))
+                // … then the rest of the iterator.
+                        .chain(clauses).collect();
```
ncalexan: nit: prefer to trail the `.collect()` on its own line.

query-algebrizer/src/clauses/or.rs
```diff
@@ -271,25 +281,23 @@ impl ConjoiningClauses {
...
 
         match self.deconstruct_or_join(schema, or_join) {
+            DeconstructedOrJoin::KnownSuccess => {
+                // The pattern came to us empty -- `(or)`. Do nothing.
+                Ok(())
+            },
             DeconstructedOrJoin::KnownEmpty(reason) => {
                 // There were no arms of the join that could be mapped to a table.
-                // The entire `or` can yield no results.
+                // The entire `or`, and thus the CC, cannot yield results.
```
ncalexan: Is this true with nested `or`?  With nested `and`, which is written as concatentation in Datomic IIRC?

query-algebrizer/src/clauses/or.rs
```diff
@@ -0,0 +1,80 @@
+// Copyright 2016 Mozilla
```
ncalexan: I'm not really reviewing this `Split cc.rs` commit -- do what you think makes sense.

query/src/lib.rs
```diff
@@ -44,6 +44,20 @@ pub type SrcVarName = String;          // Do not include the required syntactic
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct Variable(pub PlainSymbol);
 
+impl Variable {
+    pub fn as_str(&self) -> &str {
```
ncalexan: `Into` throughout?  Generally, I find named functions easier to follow, but Rust style is definitely to use `Into`.

query/src/lib.rs
```diff
@@ -75,13 +76,20 @@ impl FromValue<Variable> for Variable {
 }
 
 impl Variable {
-    pub fn from_symbol(sym: &PlainSymbol) -> Option<Variable> {
```
ncalexan: But here, you could take anything that's `Into<Rc<PlainSymbol>>`, since `Rc<T>: From<T>`.  I guess you want `try_into`?

query-algebrizer/src/clauses/pattern.rs
```diff
@@ -292,14 +294,18 @@ mod testing {
         SourceAlias,
     };
 
+    fn variable_named(s: &'static str) -> Variable {
```
ncalexan: This would definitely be more pleasant if you wrote the transitive closure of `Into`/`From` declarations.  You could end up with `Variable("name".into())` or similar.

db/tests/value_tests.rs
```diff
@@ -14,11 +14,24 @@ extern crate mentat_db;
...
+use std::rc::Rc;
+
 use ordered_float::OrderedFloat;
+
 use edn::symbols;
 
+use mentat_core::{TypedValue, ValueType};
+use mentat_db::db::TypedSQLValue;
+
+fn typed_keyword(ns: &'static str, name: &'static str) -> TypedValue {
```
ncalexan: Shouldn't we be exposing and using these helpers on `TypedValue` more generally?

rnewman: Yes, probably. I was moving quickly.

query/src/lib.rs
rnewman: Yes, when Rust finishes stabilizing those things…

rnewman:
> why you don't expect dangling references to persist

Remember that `Rc<Foo>` with one strong reference is equivalent to `Box<Foo>` — and this commit stack is mostly doing exactly that, because we're not interning anywhere.

When an `Rc` is dropped, its shared internal strong reference count drops. When the last `Rc` is dropped, and will take the strong reference count down to 0, it also drops the value. It's impossible to leak values, and it's impossible for a reference to dangle. So long as you have a clone of the `Rc`, you have the value, and when you don't the value is gone.

The worst case to which you seem to be alluding is no worse than the standard `clone`-everything case: one copy sticks around somewhere.

After this we'll have more cheaply copyable CCs, because the things the CC cares about — variables and values — will now be cheap `Rc` wrappers. A clone of the `CC` won't require cloning strings.

We can then move on to intern values and variables in the query parser, so we save duplication earlier in the process. We might end up using `Arc` inside `edn::Value`, but that would be opaque from the perspective of this PR.

My reasoning for where in the pile to insert the `Rc` is roughly:

- Solely at the `edn::Value` layer is inappropriate, because we don't work with `Value` after parsing. It simply wouldn't help.
- At the `PlainSymbol` and `NamespacedKeyword` layer is slightly more expensive from a comparison standpoint, makes these simple types more complex, and gives us no easy `Send` 'out' for error handling. `TypedValue` will already need an `Rc` 'cos it contains `String`s, so we might as well take a coherent approach.
- Putting `Rc` inside `Variable` and `TypedValue` means that calling code (which in the algebrizer's case works pretty much exclusively with those) can effectively treat them as `Copy` — in all enum cases they contain small, cheaply cloneable values. Code that works with these values doesn't need to change at all: you can see this PR mostly affects tests that construct them.

All that said, I wish there was a little more magic here to avoid spreading the knowledge of inner types around, and of course we still need to pick points in the flow to do interning. To achieve end-to-end memory usage improvements requires touching the EDN parser, which I didn't want to do until your parser changes landed.

query-algebrizer/src/types.rs
rnewman: Yes, this is deliberate — another way of phrasing your point is "these types aren't recursive", which at present is exactly what I want. This is enough to very simply implement simple `or`, and it won't be painful to extend this to `not`. I doubt we'll need more than that.

query-algebrizer/src/types.rs
rnewman: With `instead` and `also` I was trying to avoid using common control flow words. I think I might opt for `add_alternate` and `add_intersection`…

query-sql/src/lib.rs
rnewman: It's actually `AND ()` we need to worry about — this will be joined into a sequence of `AND`.

I'll add some error checking, probably producing `1` or something.

query-algebrizer/src/clauses/or.rs
rnewman: Yes, exactly that. For `not` we do this through a subquery and `NOT EXISTS` — projection is our boundary. Complex `or` is similar: it produces a `UNION`.

Using a scoped variable might shuffle a small set of complex `or`s into the simple camp, and is actually relatively straightforward: control the set of variables transferred between CCs when algebrizing. But I'll do that as a follow-up, some weeks from now!

query-algebrizer/src/clauses/or.rs
rnewman: Yup.

rnewman:
I merged the first part of this in 92cdd72.

You now have `when_*` helpers in `mentat_core::util`, @ncalexan.

I'm making a separate PR for the `Rc` stuff, with comments addressed. I do want to chew on that a little more and see it in isolation before merging it.

PR #402, opened 2017-04-03
Implement querying fulltext values.

ncalexan:
@rnewman you can take a look at the first commit, if you care to.

@fluffyemily if you're using the new parsing apparatus (yet to land!), this is a self-contained example of how to use it.

### DIFF
--- a/query-algebrizer/src/clauses/where_fn.rs
+++ b/query-algebrizer/src/clauses/where_fn.rs
@@ -1,0 +1,384 @@
+// Copyright 2016 Mozilla
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+use std::rc::Rc;
+
+use mentat_core::{
+    Schema,
+    TypedValue,
+    ValueType,
+};
+
+use mentat_query::{
+    Binding,
+    FnArg,
+    NonIntegerConstant,
+    SrcVar,
+    Variable,
+    WhereFn,
+};
+
+use clauses::{
+    ConjoiningClauses,
+    PushComputed,
+};
+
+use errors::{
+    BindingError,
+    ErrorKind,
+    Result,
+};
+
+use types::{
+    ComputedTable,
+    SourceAlias,
+    VariableColumn,
+};
+
+trait FnArgInto<F> {
+    fn into_typed_value(self, lookup: F) -> Result<TypedValue>;
+    fn into_column_vector(self, lookup: F) -> Result<Vec<TypedValue>>;
+
+    fn into_unit_matrix(self, lookup: F) -> Result<Vec<Vec<TypedValue>>>;
+    fn into_row_matrix(self, lookup: F) -> Result<Vec<Vec<TypedValue>>>;
+    fn into_column_matrix(self, lookup: F) -> Result<Vec<Vec<TypedValue>>>;
+    fn into_block_matrix(self, lookup: F) -> Result<Vec<Vec<TypedValue>>>;
+}
+
+impl<F> FnArgInto<F> for FnArg where F: Fn(Variable) -> Result<TypedValue> {
+    fn into_typed_value(self, lookup: F) -> Result<TypedValue> {
+        match self {
+            FnArg::EntidOrInteger(x) => Ok(TypedValue::Long(x)),
+            FnArg::Ident(x) => Ok(TypedValue::Keyword(Rc::new(x))),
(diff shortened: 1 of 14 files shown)


query-parser/src/parse.rs
```diff
@@ -391,6 +414,51 @@ def_parser!(Find, query, FindQuery, {
         })
 });
 
+pub struct Bind;
+
+def_matches_plain_symbol!(Bind, placeholder, "_");
+
+def_matches_plain_symbol!(Bind, ellipsis, "...");
```
rnewman: Just use `Find.ellipsis()`. `Find` is the catch-all top-level parser struct here.

rnewman:
Also this is a fantastic example of extending the query parser. 👍

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -243,6 +252,172 @@ impl ConjoiningClauses {
...
+        if let VariableOrPlaceholder::Variable(ref var) = bindings[3] {
+            if self.bound_value(&var).is_some() || self.input_variables.contains(&var) {
+                bail!(ErrorKind::InvalidBinding(where_fn.operator.clone(), BindingError::BoundInputVariable));
+            }
+            self.constrain_var_to_type(var.clone(), ValueType::Double);
+
+            // Right now we don't support binding a column to a constant value.  Therefore, we
+            // introduce a computed table with exactly the constant value we require.  See the
+            // translate tests for some edge cases in this space.
+            // TODO: optimize this by binding the value, like:
```
rnewman: Yup. But as with `ground`, doing so requires awareness of whether there are any existing patterns. In that case you can spit out a simple scalar `FROM` like `(SELECT 0.0 AS ?match) AS c00` so that it'll join with existing columns.

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -243,6 +252,172 @@ impl ConjoiningClauses {
...
+
+        if !where_fn.binding.is_valid() {
+            // The binding must not duplicate bound variables.
+            bail!(ErrorKind::InvalidBinding(where_fn.operator.clone(), BindingError::RepeatedBoundVariable));
+        }
+
+        for var in &where_fn.binding.variables() {
+            if let &Some(ref var) = var {
+                if self.input_variables.contains(var) {
+                    bail!(ErrorKind::InvalidBinding(where_fn.operator.clone(), BindingError::BoundInputVariable));
```
rnewman: Can you give a citation?

Why is this not OK?

```
[:find ?val
 :in $ ?entity
 :where [(fulltext $ :foo/txt "hello") [?entity ?val]]]
```

— that is, find values for the given entity and attribute pair…

I would expect variable unification to occur here.

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -243,6 +252,172 @@ impl ConjoiningClauses {
...
+                }
+            }
+        }
+
+        let bindings = match where_fn.binding {
+            Binding::BindRel(bindings) => {
+                if bindings.len() != 4 {
```
rnewman: In practice, Datomic uses implicit blanks for bindings, not just data patterns. Example:

http://stackoverflow.com/a/24753310/22003

I don't think we have enough specificity in the parser to flesh these out there, so we should allow it here.

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -243,6 +252,172 @@ impl ConjoiningClauses {
...
+        self.constrain_attribute(datoms_table_alias.clone(), a);
+
+        self.wheres.add_intersection(ColumnConstraint::Equals(
+            QualifiedAlias(datoms_table_alias.clone(), Column::Fixed(DatomsColumn::Value)),
+            QueryValue::Column(QualifiedAlias(fulltext_values_alias.clone(), Column::Fulltext(FulltextColumn::Rowid)))));
+
+        // `search` is either text or a variable.
+        let search: TypedValue = match args.next().unwrap() {
+            FnArg::Variable(in_var) => {
+                if !self.input_variables.contains(&in_var) {
```
rnewman: It could also be bound by a text-typed variable elsewhere in the query, allowing lookups:

```
[:find ?e
 :where
 [?x :foo/name "John"]
 [?x :foo/searchkey ?key]
 [(fulltext $ :foo/text ?key) [[?e ?val]]]]
```

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -243,6 +252,172 @@ impl ConjoiningClauses {
...
+                    bail!(ErrorKind::UnboundVariable((*in_var.0).clone()));
+                }
+                match self.bound_value(&in_var) {
+                    Some(ref in_value) => in_value.clone(),
+                    None => bail!(ErrorKind::UnboundVariable((*in_var.0).clone())),
+                }
+            },
+            FnArg::Constant(NonIntegerConstant::Text(s)) => {
+                TypedValue::typed_string(s.as_str())
```
rnewman: Do this without cloning. `typed_string` basically exists for tests.

At this point we own `s`, and it's an `Rc`, so just do `TypedValue::String(s)`.

rnewman:
Obsoleted by #477.

Issue #304, opened 2017-02-15
[query] Algebrize simple `or` patterns


rnewman:
A simple `or` clause is one in which each branch can be evaluated against a single pattern match. That means that all the variables are in the same places.

We can produce a `ConjoiningClauses` in that case — the `wheres` will suffice for alternation.

For example:

```edn
[:find ?joe
 :where
 (or [?joe :person/name "Joe"]
     [?joe :person/name "Joseph"])]
```

is simply

```sql
WHERE datoms123.a = 456       -- :person/name
 AND (datoms123.v = "Joe" OR datoms123.v = "Joseph")
```

This is notably different from a complex `or-join` in which each branch must be implemented as a branch of a `UNION`.

rnewman:
Fixed by #304.

PR #403, opened 2017-04-05
WIP: Algebrize simple `or` joins. (#304)

rnewman:
This is most of the way there. I need to flesh out the rest of the tests, skim for TODOs, and then double-check the translator by adding tests for the CCs this produces.

I also need to add tests for pernicious patterns like `[?x ?x ?x]`.

### DIFF
--- a/query-algebrizer/src/clauses/mod.rs
+++ b/query-algebrizer/src/clauses/mod.rs
@@ -599,12 +599,26 @@ impl ConjoiningClauses {
             .ok()
     }
 
+    fn get_attribute_for_value<'s>(&self, schema: &'s Schema, value: &TypedValue) -> Option<&'s Attribute> {
+        match value {
+            &TypedValue::Ref(id) => schema.attribute_for_entid(id),
+            &TypedValue::Keyword(ref kw) => schema.attribute_for_ident(kw),
+            _ => None,
+        }
+    }
+
     fn get_attribute<'s, 'a>(&self, schema: &'s Schema, pattern: &'a Pattern) -> Option<&'s Attribute> {
         match pattern.attribute {
             PatternNonValuePlace::Entid(id) =>
                 schema.attribute_for_entid(id),
             PatternNonValuePlace::Ident(ref kw) =>
                 schema.attribute_for_ident(kw),
+            PatternNonValuePlace::Variable(ref var) =>
+                // If the pattern has a variable, we've already determined that the binding -- if
+                // any -- is acceptable and yields a table. Here, simply look to see if it names
+                // an attribute so we can find out the type.
+                self.value_bindings.get(var)
+                                   .and_then(|val| self.get_attribute_for_value(schema, val)),
             _ =>
                 None,
         }
--- a/query-algebrizer/src/clauses/mod.rs
+++ b/query-algebrizer/src/clauses/mod.rs
@@ -648,6 +648,19 @@ impl ConjoiningClauses {
         }
     }
 
+    /// Eliminate any type extractions for variables whose types are definitely known.
+    pub fn prune_extracted_types(&mut self) {
+        if self.extracted_types.is_empty() || self.known_types.is_empty() {
+            return;
+        }
+        for (var, types) in self.known_types.iter() {
+            if types.len() == 1 {
+                self.extracted_types.remove(var);
+            }
+        }
+    }
+
+
     /// When a CC has accumulated all patterns, generate value_type_tag entries in `wheres`
     /// to refine value types for which two things are true:
     ///
--- a/query-algebrizer/src/clauses/mod.rs
+++ b/query-algebrizer/src/clauses/mod.rs
@@ -8,11 +8,6 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
-use std::fmt::{
-    Debug,
-    Formatter,
-};
-
 use std::collections::{
     BTreeMap,
     BTreeSet,
@@ -20,6 +15,11 @@ use std::collections::{
 };
 
 use std::collections::btree_map::Entry;
+
+use std::fmt::{
+    Debug,
+    Formatter,
+};
 
 use mentat_core::{
     Attribute,
@@ -89,6 +89,33 @@ fn unit_type_set(t: ValueType) -> HashSet<ValueType> {
     let mut s = HashSet::with_capacity(1);
     s.insert(t);
     s
+}
+
+trait Contains<K, T> {
+    fn contains_then<F: FnOnce() -> T>(&self, k: &K, f: F) -> Option<T>;
+}
+
+trait Intersection<K> {
+    fn with_intersected_keys(&self, ks: &BTreeSet<K>) -> Self;
+}
+
+impl<K: Ord, T> Contains<K, T> for BTreeSet<K> {
+    fn contains_then<F: FnOnce() -> T>(&self, k: &K, f: F) -> Option<T> {
+        if self.contains(k) {
+            Some(f())
+        } else {
+            None
+        }
+    }
+}
+
+impl<K: Clone + Ord, V: Clone> Intersection<K> for BTreeMap<K, V> {
+    /// Return a clone of the map with only keys that are present in `ks`.
+    fn with_intersected_keys(&self, ks: &BTreeSet<K>) -> Self {
+        self.iter()
+            .filter_map(|(k, v)| ks.contains_then(k, || (k.clone(), v.clone())))
+            .collect()
+    }
 }
 
 /// A `ConjoiningClauses` (CC) is a collection of clauses that are combined with `JOIN`.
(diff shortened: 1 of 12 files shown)


query-algebrizer/src/clauses/mod.rs
```diff
@@ -355,6 +334,73 @@ impl ConjoiningClauses {
...
+                                                                            .next()
+                                                                            .cloned()
+                                                                             .unwrap()));
+                } else {
+                    e.insert(intersected);
+                }
+            },
+        }
+
+        if let Some(e) = empty_because {
```
ncalexan: This formulation is a little awkward; did you expect something different to come here?  Why not just set this inline (and return)?

query-algebrizer/src/clauses/mod.rs
```diff
@@ -355,6 +334,73 @@ impl ConjoiningClauses {
...
+        }
+    }
+
+    fn narrow_types(&mut self, additional_types: BTreeMap<Variable, HashSet<ValueType>>) {
+        if additional_types.is_empty() {
+            return;
+        }
+        for (var, new_types) in additional_types {
+            self.narrow_types_for_var(var, new_types);
+            if self.is_known_empty {
```
ncalexan: Consider dropping this stateful flag and computing `is_known_empty()` from the underlying reason.

query-algebrizer/src/clauses/mod.rs
```diff
@@ -355,6 +334,73 @@ impl ConjoiningClauses {
...
+                    e.insert(new_types);
+                },
+                Entry::Occupied(mut e) => {
+                    e.get_mut().extend(new_types.into_iter());
+                },
+            }
+        }
+    }
+
+    fn narrow_types_for_var(&mut self, var: Variable, types: HashSet<ValueType>) {
```
ncalexan: The asymmetry between `broad_types`, which takes _additional_ types, and `narrow_types`, which _sets_ the types, surprises me.  Overall I think you need to grow a `TypeSet` and do some of this a little more robustly.  I find the interaction with `extracted_types` particularly difficult to follow; I'm not sure what your case is.  If you think this work isn't necessary, you'll need to write docstrings to elaborate on what these functions actually achieve.

query-algebrizer/src/clauses/mod.rs
```diff
@@ -191,6 +191,48 @@ impl Default for ConjoiningClauses {
...
+        template.extracted_types = intersect_keys(&self.extracted_types, &vars);
+
+        template
+    }
+}
+
+/// Return a clone of the provided map with only keys that are present in `ks`.
+fn intersect_keys<K: Clone + Ord, V: Clone>(m: &BTreeMap<K, V>, ks: &BTreeSet<K>) -> BTreeMap<K, V> {
+    m.iter()
+     .filter_map(|(k, v)| if ks.contains(k) {
```
ncalexan: `ks.get(k).map(|_| (k.clone(), v.clone())|)` might be a little shorter.

query-algebrizer/src/clauses/or.rs
```diff
@@ -343,41 +354,381 @@ impl ConjoiningClauses {
...
+            multival: false,
+            ..Default::default()
+        });
+        schema
+    }
+
+    /// Test that if all the attributes in an `or` fail to resolve, the entire thing fails.
+    #[test]
+    fn test_schema_based_failure() {
+        let schema = Schema::default();
```
ncalexan: This really threw me.  Can you use attributes that aren't in the prepopulated schema?

query-parser/tests/find_tests.rs
```diff
@@ -70,9 +70,9 @@ fn can_parse_simple_or() {
                FindSpec::FindScalar(Element::Variable(Variable::from_valid_name("?x"))));
     assert_eq!(p.where_clauses,
                vec![
-                   WhereClause::OrJoin(OrJoin {
-                       unify_vars: UnifyVars::Implicit,
-                       clauses: vec![
+                   WhereClause::OrJoin(OrJoin::new(
```
ncalexan: Consider doing this `OrJoin` constructor change as a Pre: patch.

query-algebrizer/src/clauses/mod.rs
rnewman: `narrow_types` doesn't set the types — it intersects them with the types that are already there. `broaden` broadens, `narrow` narrows.

```
                let intersected: HashSet<_> = types.intersection(e.get()).cloned().collect();
```

query-algebrizer/src/clauses/mod.rs
rnewman: This pattern — `mut` `Option` outside, set inside block, check outside block — seems to be the only sensible way to use `Entry` and potentially mutate the containing object (`mark_as_empty`).

It's not possible to simply inline the `mark_as_empty`; you'll get an error like `cannot mutably borrow foo; immutably borrowed here…`.

query-algebrizer/src/clauses/mod.rs
rnewman: Yeah, seems reasonable.

query-algebrizer/src/clauses/or.rs
rnewman: Done.

query-algebrizer/src/clauses/mod.rs
rnewman: I pushed a commit concurrent with your review that alters `extracted_types` a little.

We used to maintain `extracted_types` alongside `known_types`: if a var's type (only one!) was known, we didn't need a tag extraction column. If we later found out the type of an untyped var, then we'd remove the extraction column.

Now, of course, types can narrow and broaden. I killed most of the removals, and replaced them with a phase at the end.

The map still needs to exist, because it tracks which columns supply the tag.

query-algebrizer/src/clauses/mod.rs
rnewman: There would need to be some kind of `filter` in there — it's an intersection, not an extraction. But yeah, there are a bunch of ways to golf this.

rnewman:
Ready for a final pass, @ncalexan.

query-algebrizer/src/clauses/mod.rs
```diff
@@ -91,6 +91,33 @@ fn unit_type_set(t: ValueType) -> HashSet<ValueType> {
...
+trait Contains<K, T> {
+    fn contains_then<F: FnOnce() -> T>(&self, k: &K, f: F) -> Option<T>;
+}
+
+trait Intersection<K> {
+    fn with_intersected_keys(&self, ks: &BTreeSet<K>) -> Self;
+}
+
+impl<K: Ord, T> Contains<K, T> for BTreeSet<K> {
+    fn contains_then<F: FnOnce() -> T>(&self, k: &K, f: F) -> Option<T> {
```
ncalexan: `when_contains`?

query-algebrizer/src/clauses/mod.rs
```diff
@@ -91,6 +91,33 @@ fn unit_type_set(t: ValueType) -> HashSet<ValueType> {
...
+            None
+        }
+    }
+}
+
+impl<K: Clone + Ord, V: Clone> Intersection<K> for BTreeMap<K, V> {
+    /// Return a clone of the map with only keys that are present in `ks`.
+    fn with_intersected_keys(&self, ks: &BTreeSet<K>) -> Self {
+        self.iter()
+            .filter_map(|(k, v)| ks.contains_then(k, || (k.clone(), v.clone())))
```
ncalexan: Honestly, I'm not sure this expression is better than the earlier one :(

query-algebrizer/src/clauses/mod.rs
```diff
@@ -191,6 +218,38 @@ impl Default for ConjoiningClauses {
...
+        concrete.value_bindings = self.value_bindings.clone();
+        concrete.known_types = self.known_types.clone();
+        concrete.extracted_types = self.extracted_types.clone();
+
+        concrete
+    }
+
+    /// Make a new CC populated with the relevant variable associations in this CC.
+    /// Note that the CC's table aliaser is not yet usable. That's not a problem for templating for
+    /// simple `or`.
```
ncalexan: Is this still valid after your later commits?  Make sure the commit agrees with the final reality.

query-algebrizer/src/clauses/or.rs
```diff
@@ -407,20 +407,24 @@ impl ConjoiningClauses {
...
-                                let reason = receptacle.empty_because;
-                                None
-                            } else {
-                                Some(receptacle)
-                            }
+                            receptacle
                         })
                         .peekable();
 
+            // Let's see if we can grab a reason if every pattern failed.
```
ncalexan: I don't think this expression is wrong, but it's really pretty awkward.  Consider just building a `Vec` to avoid `peekable`.

query-algebrizer/src/clauses/or.rs
```diff
@@ -8,29 +8,19 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
-// WIP
```
ncalexan: Whee!

query-algebrizer/src/clauses/mod.rs
rnewman: This comment is overwritten by the alias commit.

rnewman:
Merged: 1636134

Issue #389, opened 2017-03-23
Parse :not and :not-join


rnewman:
Part of #239.

fluffyemily:
Examples of Not and Not Join from Datomic. Do we want to match this here?

Not
```
[:find ?name
 :where [?eid :artist/name ?name]
        (not [?eid :artist/country :country/CA])]
```

Not-join
```
[:find (count ?artist) .
 :where [?artist :artist/name]
        (not-join [?artist]
          [?release :release/artists ?artist]
          [?release :release/year 1970])]
```

fluffyemily:
The Datomic docs on not have pretty much exactly the same wording as for or and so we probably need to validate in the same way.

* unify the variables. 
    * In `not` that means ensuring that any variable in the not clause is also represented in the other clauses (`?eid` in the example above). 
    * In `not-join`, ensuring that the variables stated in the join list (`?artist` in the example above) are represented in the other clauses.
    * In the `not-join` that the variables in each leg of the not match (`?release` in the example above`)

rnewman:
> Do we want to match this here?

Yes.


Validation and variable extraction of `or`, which you can reuse here, are in #394; I hope to land those today.

Note that `not` doesn't have legs/arms in the way that `or` does. `not` and `not-join` simply have a _body_, so there's no analogue of making sure that each arm is similar. That's nice.

Algebrizing simple `not` is pretty much equivalent to computing the non-negated changes, wrapping them in a negation, and adding them to the destination CC.

Algebrizing simple `not-join` with non-exhaustive unification is much the same, but of course bearing in mind the vars that aren't allowed to unify!

Algebrizing complex `not` — clauses that themselves contain a complex join or generate bindings — involves adding `NOT EXISTS` to the final query.

https://github.com/mozilla/mentat/blob/master/src/common/datomish/query/clauses.cljc#L178

and

https://github.com/mozilla/mentat/blob/master/src/common/datomish/query/clauses.cljc#L300

and

https://github.com/mozilla/mentat/blob/master/src/common/datomish/query/clauses.cljc#L323

should be a pretty good inspiration.

The algebrizing part will be much easier with my `or` work to follow, so I recommend you start with parsing once you're satisfied that you understand the concepts!

fluffyemily:
Do we have to take into account Not And and Not Or?

rnewman:
Yes! The grammar is here:

http://docs.datomic.com/query.html

But the CLJS implementation already gets this right, I think.

[...]

PR #405, opened 2017-04-06
parse not &  not-join (#389)

fluffyemily:
This deals with parsing and the pre-algebrizing validation.

I'm also looking at how Nicks changes to the parser affect this, and also at applying and algebrizing the clauses.

### DIFF
--- a/query-algebrizer/src/clauses/mod.rs
+++ b/query-algebrizer/src/clauses/mod.rs
@@ -62,8 +62,8 @@ mod predicate;
 mod resolve;
 
 use validate::{
-    validate_or_join,
     validate_not_join,
+    validate_or_join,
 };
 
 // We do this a lot for errors.
--- a/query-algebrizer/src/clauses/mod.rs
+++ b/query-algebrizer/src/clauses/mod.rs
@@ -56,11 +56,15 @@ use types::{
 };
 
 mod or;
+mod not;
 mod pattern;
 mod predicate;
 mod resolve;
 
-use validate::validate_or_join;
+use validate::{
+    validate_or_join,
+    validate_not_join,
+};
 
 // We do this a lot for errors.
 trait RcCloned<T> {
@@ -585,6 +589,9 @@ impl ConjoiningClauses {
                 validate_or_join(&o)
                 //?;
                 //self.apply_or_join(schema, o)
+            },
+            WhereClause::NotJoin(n) => {
+                validate_not_join(&n)
             },
             _ => unimplemented!(),
         }
--- a/query-algebrizer/src/validate.rs
+++ b/query-algebrizer/src/validate.rs
@@ -13,6 +13,7 @@ use std::collections::BTreeSet;
 use mentat_query::{
     ContainsVariables,
     OrJoin,
+    NotJoin,
     Variable,
     UnifyVars,
 };
@@ -74,6 +75,46 @@ pub fn validate_or_join(or_join: &OrJoin) -> Result<()> {
     }
 }
 
+pub fn validate_not_join(not_join: &NotJoin) -> Result<()> {
+    // Grab our mentioned variables and ensure that the rules are followed.
+    match not_join.unify_vars {
+        UnifyVars::Implicit => {
+            if not_join.clauses.len() < 2 {
+                Ok(())
+            } else {
+                let mut clauses = not_join.clauses.iter();
+                let template = clauses.next().unwrap().collect_mentioned_variables();
+                for clause in clauses {
+                    if template != clause.collect_mentioned_variables() {
+                        bail!(ErrorKind::NonMatchingVariablesInNotClause);
+                    }
+                }
+                Ok(())
+            }
+        },
+        UnifyVars::Explicit(ref vars) => {
+            // The joined vars must each appear somewhere in the clauses mentioned variables.
+            let var_set: BTreeSet<Variable> = vars.iter().cloned().collect();
+            for var in &var_set {
+                if !&not_join.clauses.iter().any(|clause| clause.collect_mentioned_variables().contains(&var))  {
+                    bail!(ErrorKind::NonMatchingVariablesInNotClause);
+                }
+            }
+
+            // any other mentioned variables must also unify
+            let mut clauses = not_join.clauses.iter();
+            let template: BTreeSet<Variable> = clauses.next().unwrap().collect_mentioned_variables().difference(&var_set).cloned().collect();
+            for clause in clauses {
+                let mentioned_vars = clause.collect_mentioned_variables().difference(&var_set).cloned().collect();
+                if template != mentioned_vars {
+                    bail!(ErrorKind::NonMatchingVariablesInOrClause);
+                }
+            }
+            Ok(())
+        },
+    }
+}
+
 #[cfg(test)]
 mod tests {
     extern crate mentat_core;
@@ -84,6 +125,7 @@ mod tests {
         FindQuery,
         NamespacedKeyword,
         OrWhereClause,
+        WhereNotClause,
         Pattern,
         PatternNonValuePlace,
         PatternValuePlace,
@@ -96,7 +138,10 @@ mod tests {
 
     use clauses::ident;
 
-    use super::validate_or_join;
+    use super::{
+        validate_or_join,
+        validate_not_join,
+    };
 
     fn value_ident(ns: &str, name: &str) -> PatternValuePlace {
         PatternValuePlace::IdentOrKeyword(::std::rc::Rc::new(NamespacedKeyword::new(ns, name)))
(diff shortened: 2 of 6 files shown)


query-algebrizer/src/clauses/mod.rs
```diff
@@ -56,11 +56,15 @@ use types::{
...
 mod or;
+mod not;
 mod pattern;
 mod predicate;
 mod resolve;
 
-use validate::validate_or_join;
+use validate::{
+    validate_or_join,
+    validate_not_join,
```
rnewman: Nit: `n` before `o` (alphabetical ordering).

query-algebrizer/src/validate.rs
```diff
@@ -74,6 +75,46 @@ pub fn validate_or_join(or_join: &OrJoin) -> Result<()> {
...
+    match not_join.unify_vars {
+        UnifyVars::Implicit => {
+            if not_join.clauses.len() < 2 {
+                Ok(())
+            } else {
+                let mut clauses = not_join.clauses.iter();
+                let template = clauses.next().unwrap().collect_mentioned_variables();
+                for clause in clauses {
+                    if template != clause.collect_mentioned_variables() {
+                        bail!(ErrorKind::NonMatchingVariablesInNotClause);
```
rnewman: This whole clause can go.

A `not` has only a single body, made up of multiple clauses; it's the negated version of an implied `and`. Simply being `Implicit` satisfies intrinsic validation.

(The clause might be _extrinsically_ invalid if its unify list isn't already bound outside the pattern, but you can't answer that question here.)

query-algebrizer/src/validate.rs
```diff
@@ -74,6 +75,46 @@ pub fn validate_or_join(or_join: &OrJoin) -> Result<()> {
...
+                    }
+                }
+                Ok(())
+            }
+        },
+        UnifyVars::Explicit(ref vars) => {
+            // The joined vars must each appear somewhere in the clauses mentioned variables.
+            let var_set: BTreeSet<Variable> = vars.iter().cloned().collect();
+            for var in &var_set {
```
rnewman: Careful of cloning: you're cloning here then immediately only using a reference, which isn't necessary.

query-algebrizer/src/validate.rs
```diff
@@ -74,6 +75,46 @@ pub fn validate_or_join(or_join: &OrJoin) -> Result<()> {
...
+                if !&not_join.clauses.iter().any(|clause| clause.collect_mentioned_variables().contains(&var))  {
```
rnewman: As in the `Implicit` case: the parts of a `not` must all match together; they're not considered separately.

Here you want to make sure that `vars` is a subset of the union of the mentioned variables in each clause. You can use `accumulate_mentioned_variables` to accumulate from each, or you can try to figure out a cheaper way to do it.

In any case, you should probably implement `ContainsVariables` for `NotJoin`, then you can just say

```
if !var_set.is_subset(self.collect_mentioned_variables()) {
    …
```

query/src/lib.rs
```diff
@@ -569,11 +569,30 @@ pub struct OrJoin {
...
+            &WhereNotClause::Clause(WhereClause::Pattern(_)) => true,
+            _ => false,
+        }
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct NotJoin {
+    pub unify_vars: UnifyVars,
+    pub clauses: Vec<WhereNotClause>,
```
rnewman: This can just be a vector of `WhereClause`, and `WhereNotClause` can go away.

query-algebrizer/src/validate.rs
```diff
@@ -96,7 +138,10 @@ mod tests {
 
     use clauses::ident;
 
-    use super::validate_or_join;
+    use super::{
+        validate_or_join,
+        validate_not_join,
```
rnewman: Same.

query-algebrizer/src/validate.rs
```diff
@@ -229,4 +274,177 @@ mod tests {
             _ => panic!(),
         };
     }
+
+
+    /// Tests that the top-level form is a valid `not`, returning the clauses.
+    fn valid_not_join(parsed: FindQuery, expected_unify: UnifyVars) -> Vec<WhereNotClause> {
+        // filter all the not clauses
+        let mut nots = parsed.where_clauses.iter().filter(|&x| match *x {
```
rnewman: If you use `into_iter()` here, you won't need to clone below, and you'll have fewer reference operations inside the closure.

query-algebrizer/src/validate.rs
```diff
@@ -229,4 +274,177 @@ mod tests {
...
+            WhereClause::NotJoin(_) => true,
+            _ => false,
+        });
+
+        // there should be only one not clause
+        let clause = nots.next().unwrap().clone();
+        assert_eq!(None, nots.next());
```
rnewman: Style opinion: I might prefer `assert!(nots.next().is_none())` here.

query-algebrizer/src/validate.rs
```diff
@@ -229,4 +274,177 @@ mod tests {
...
+                // It's valid: the variables are the same in each branch.
+                assert_eq!((), validate_not_join(&not_join).unwrap());
+                assert_eq!(expected_unify, not_join.unify_vars);
+                not_join.clauses
+            },
+            _ => panic!(),
+        }
+    }
+
+    /// Test that a `not` is valid if all of its arms refer to the same variables.
```
rnewman: Sadly this is testing a restriction that doesn't apply to `not`!

The docs say:

> When more than one clause is supplied to `not`, you should read the clauses as if they are connected by 'and', just as they are in `:where`.

query-algebrizer/src/validate.rs
```diff
@@ -229,4 +274,177 @@ mod tests {
...
+                            tx: PatternNonValuePlace::Placeholder,
+                        }),
+                    ));
+            },
+            _ => panic!(),
+        };
+    }
+
+
+    /// Test that a `not` with differing variable sets in each body part will fail to validate.
```
rnewman: This is invalid, but not for this reason. It's invalid because there's no binding for `?release` by the time the `not` is executed.

See this DataScript test for an example:

https://github.com/tonsky/datascript/blob/master/test/datascript/test/query_not.cljc#L183

(Strictly speaking there is an interpretation for this query: it's "find artist names for non-Canadian artists, if no albums were released by anybody in 1970". We prefer you to be explicit about that interpretation by using `not-join`!)

query-algebrizer/src/validate.rs
```diff
@@ -229,4 +274,177 @@ mod tests {
...
+        let clause = nots.next().unwrap().clone();
+        assert_eq!(None, nots.next());
+
+        match clause {
+            WhereClause::NotJoin(not_join) => assert!(validate_not_join(&not_join).is_err()),
+            _ => panic!(),
+        }
+    }    
+    
+    /// Test that a `not-join` that does not match any vars declared inside it fails to validate.
```
rnewman: This query is valid. Again, it's almost certain not to be a meaningful query — the variable named `?year` will be bound to releases, and won't join against the rest of the clauses — but it's valid. It means "Find all artists, but only artists with no releases unless anybody released a record in 1970".

query-parser/src/parse.rs
```diff
@@ -213,6 +223,53 @@ def_value_parser_fn!(Where, or_join_clause, WhereClause, input, {
...
+    Where::clause().map(|clause| WhereNotClause::Clause(clause)).parse_stream(input)
+});
+
+def_value_parser_fn!(Where, where_not_clause, WhereNotClause, input, {
+    choice([Where::not_pattern_clause()]).parse_stream(input)
+});
+
+def_value_parser_fn!(Where, not_clause, WhereClause, input, {
+    satisfy_map(|x: edn::Value| {
+        seq(x).and_then(|items| {
```
rnewman: @ncalexan, is your parser stuff ready to land yet?

query-algebrizer/src/validate.rs
ncalexan: One advantage of `assert_eq!` is that it will show you the `Some(...)` in the failure case.  For this reason, I prefer asserting equality.

ncalexan: nit: prefer full sentence comments (throughout).  @rnewman and I are pretty consistent about this.

query-parser/src/parse.rs
ncalexan: I thought it landed and see now that it hasn't!  It's waiting on your final review: https://github.com/mozilla/mentat/pull/393#issuecomment-291260771.

rnewman:
Closing in favor of #422.

PR #410, opened 2017-04-08
WIP: applying complex 'or'

rnewman:
This is proceeding in stages.

1. Because complex `or`, and other kinds of binding-introducing patterns, are expressed as subqueries — a `UNION` of subqueries in this case — we need to extend several components:
  - CC, which now needs to understand that some of its data isn't just coming from a named table, but a computed table. In order to avoid churning too much, I opted to do this by adding a simple indexed `Computed` table to `DatomsTable`.
  - query-sql, which now needs to represent nested queries.
  - The translator, which needs to turn those nested queries into a string.

2. We need to build an internal Datalog/SQL projection representation for these subqueries. They contribute column bindings, just like a simple table — `"c00"` — but the columns are named for variables. Furthermore, we need to unify the static types of these variables with the surrounding query, and (for each arm individually) project type columns, preserving a single SQL projection. The end result might have a query like

```edn
[:find ?v
 :where
 (or-join [?x]
   [?x 99 ?v]
   (and
     [?x ?a ?v]
     [(> ?v 10)]
     [?x 96 16]))
  [?x 101 17]]
```

turning into the SQL

```sql
SELECT c0.`?v` AS `?v`,
       c0.`?v_value_type_tag` AS `?v_value_type_tag`
FROM
  (
  SELECT datoms01.e AS `?x`,
         datoms01.v AS `?v`,
         4 AS `?v_value_type_tag`     -- This clause's attribute determines this.
  FROM datoms AS datoms01
  WHERE datoms01.a = 99

  UNION

  SELECT datoms02.e AS `?x`,
         datoms02.v AS `?v`,
         datoms02.value_type_tag AS `?v_value_type_tag`  -- This one's doesn't.
  FROM datoms AS datoms02,
       datoms AS datoms03
  WHERE datoms02.v > 10
    AND datoms03.e = datoms02.e
    AND datoms03.a = 96
    AND datoms03.v = 16
  ) AS c0,

  datoms AS datoms00
WHERE datoms00.a = 101
  AND datoms00.v = 17
  AND datoms00.e = c0.`?x`
```

3. Finally, we need to propagate shared type deductions out of the arms of the `or`, just as we did for simple `or`s.

Some of this work will generalize nicely to @fluffyemily's `NOT EXISTS` work in handling `not-join`, and perhaps to binding-generation work that @ncalexan will get to.

### DIFF
--- a/query-algebrizer/src/clauses/mod.rs
+++ b/query-algebrizer/src/clauses/mod.rs
@@ -49,6 +49,7 @@ use types::{
     ColumnConstraint,
     ColumnIntersection,
     ComputedTable,
+    Column,
     DatomsColumn,
     DatomsTable,
     EmptyBecause,
@@ -286,35 +287,55 @@ impl ConjoiningClauses {
         }
     }
 
-    pub fn bind_column_to_var(&mut self, schema: &Schema, table: TableAlias, column: DatomsColumn, var: Variable) {
+    pub fn bind_column_to_var<C: Into<Column>>(&mut self, schema: &Schema, table: TableAlias, column: C, var: Variable) {
+        let column = column.into();
         // Do we have an external binding for this?
         if let Some(bound_val) = self.bound_value(&var) {
             // Great! Use that instead.
             // We expect callers to do things like bind keywords here; we need to translate these
             // before they hit our constraints.
-            // TODO: recognize when the valueType might be a ref and also translate entids there.
-            if column == DatomsColumn::Value {
-                self.constrain_column_to_constant(table, column, bound_val);
-            } else {
-                match bound_val {
-                    TypedValue::Keyword(ref kw) => {
-                        if let Some(entid) = self.entid_for_ident(schema, kw) {
+            match column {
+                Column::Variable(_) => {
+                    // We don't need to handle expansion of attributes here. The subquery that
+                    // produces the variable projection will do so.
+                    self.constrain_column_to_constant(table, column, bound_val);
+                },
+
+                Column::Fixed(DatomsColumn::ValueTypeTag) => {
+                    // I'm pretty sure this is meaningless.
+                    unimplemented!();
+                },
+
+                // TODO: recognize when the valueType might be a ref and also translate entids there.
+                Column::Fixed(DatomsColumn::Value) => {
+                    self.constrain_column_to_constant(table, column, bound_val);
+                },
+
+                // These columns can only be entities, so attempt to translate keywords. If we can't
+                // get an entity out of the bound value, the pattern cannot produce results.
+                Column::Fixed(DatomsColumn::Attribute) |
+                Column::Fixed(DatomsColumn::Entity) |
+                Column::Fixed(DatomsColumn::Tx) => {
+                    match bound_val {
+                        TypedValue::Keyword(ref kw) => {
+                            if let Some(entid) = self.entid_for_ident(schema, kw) {
+                                self.constrain_column_to_entity(table, column, entid);
+                            } else {
+                                // Impossible.
+                                // For attributes this shouldn't occur, because we check the binding in
+                                // `table_for_places`/`alias_table`, and if it didn't resolve to a valid
+                                // attribute then we should have already marked the pattern as empty.
+                                self.mark_known_empty(EmptyBecause::UnresolvedIdent(kw.cloned()));
+                            }
+                        },
+                        TypedValue::Ref(entid) => {
                             self.constrain_column_to_entity(table, column, entid);
-                        } else {
-                            // Impossible.
-                            // For attributes this shouldn't occur, because we check the binding in
-                            // `table_for_places`/`alias_table`, and if it didn't resolve to a valid
-                            // attribute then we should have already marked the pattern as empty.
-                            self.mark_known_empty(EmptyBecause::UnresolvedIdent(kw.cloned()));
-                        }
-                    },
-                    TypedValue::Ref(entid) => {
-                        self.constrain_column_to_entity(table, column, entid);
-                    },
-                    _ => {
-                        // One can't bind an e, a, or tx to something other than an entity.
-                        self.mark_known_empty(EmptyBecause::InvalidBinding(column, bound_val));
-                    },
+                        },
+                        _ => {
+                            // One can't bind an e, a, or tx to something other than an entity.
+                            self.mark_known_empty(EmptyBecause::InvalidBinding(column, bound_val));
+                        },
+                    }
                 }
             }
 
@@ -328,10 +349,13 @@ impl ConjoiningClauses {
         // If this is a value, and we don't already know its type or where
         // to get its type, record that we can get it from this table.
         let needs_type_extraction =
-            !late_binding &&                           // Never need to extract for bound vars.
-            column == DatomsColumn::Value &&           // Never need to extract types for refs.
-            self.known_type(&var).is_none() &&         // Don't need to extract if we know a single type.
-            !self.extracted_types.contains_key(&var);  // We're already extracting the type.
+            !late_binding &&                                // Never need to extract for bound vars.
+
+            // Never need to extract types for refs, and var columns are handled elsewhere:
+            // a subquery will be projecting a type tag.
+            column == Column::Fixed(DatomsColumn::Value) &&
+            self.known_type(&var).is_none() &&              // Don't need to extract if we know a single type.
+            !self.extracted_types.contains_key(&var);       // We're already extracting the type.
 
         let alias = QualifiedAlias(table, column);
 
@@ -343,11 +367,13 @@ impl ConjoiningClauses {
         self.column_bindings.entry(var).or_insert(vec![]).push(alias);
     }
 
-    pub fn constrain_column_to_constant(&mut self, table: TableAlias, column: DatomsColumn, constant: TypedValue) {
+    pub fn constrain_column_to_constant<C: Into<Column>>(&mut self, table: TableAlias, column: C, constant: TypedValue) {
+        let column = column.into();
         self.wheres.add_intersection(ColumnConstraint::Equals(QualifiedAlias(table, column), QueryValue::TypedValue(constant)))
     }
 
-    pub fn constrain_column_to_entity(&mut self, table: TableAlias, column: DatomsColumn, entity: Entid) {
+    pub fn constrain_column_to_entity<C: Into<Column>>(&mut self, table: TableAlias, column: C, entity: Entid) {
+        let column = column.into();
         self.wheres.add_intersection(ColumnConstraint::Equals(QualifiedAlias(table, column), QueryValue::Entid(entity)))
     }
 
@@ -357,7 +383,7 @@ impl ConjoiningClauses {
 
     pub fn constrain_value_to_numeric(&mut self, table: TableAlias, value: i64) {
         self.wheres.add_intersection(ColumnConstraint::Equals(
-            QualifiedAlias(table, DatomsColumn::Value),
+            QualifiedAlias(table, Column::Fixed(DatomsColumn::Value)),
             QueryValue::PrimitiveLong(value)))
     }
 
@@ -586,7 +612,7 @@ impl ConjoiningClauses {
                     Some(v) => {
                         // This pattern cannot match: the caller has bound a non-entity value to an
                         // attribute place.
-                        Err(EmptyBecause::InvalidBinding(DatomsColumn::Attribute, v.clone()))
+                        Err(EmptyBecause::InvalidBinding(Column::Fixed(DatomsColumn::Attribute), v.clone()))
                     },
                 }
             },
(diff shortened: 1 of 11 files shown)


query-algebrizer/src/types.rs
```diff
@@ -35,6 +36,14 @@ pub enum DatomsTable {
     FulltextValues,     // The virtual table mapping IDs to strings.
     FulltextDatoms,     // The fulltext-datoms view.
     AllDatoms,          // Fulltext and non-fulltext datoms.
+    Computed(usize),    // A computed table, tracked elsewhere in the query.
```
ncalexan: Is this the generated ID from the `RcCounter`?  Let's introduce a type-alias for this.

query-algebrizer/src/types.rs
```diff
@@ -35,6 +36,14 @@ pub enum DatomsTable {
...
+}
+
+/// A source of rows.
```
ncalexan: Elaborate: `/// A source of rows produced dynamically by the underlying SQL store.` or similar.

query-algebrizer/src/types.rs
```diff
@@ -35,6 +36,14 @@ pub enum DatomsTable {
...
+pub enum ComputedTable{
```
ncalexan: nit: space before `{`.

query-algebrizer/src/types.rs
```diff
@@ -35,6 +36,14 @@ pub enum DatomsTable {
...
+    // TODO: how do we represent projecting type columns?
+    // Subquery(BTreeSet<Variable>, ::clauses::ConjoiningClauses),
+    Union(BTreeSet<Variable>, Vec<::clauses::ConjoiningClauses>),
```
ncalexan: nit: Do you want a bit for `UnionDistinct` here?  You probably want to make some statements about preconditions.  I don't see exactly what you're representing, but that'll probably be clear from later context.

query-algebrizer/src/clauses/or.rs
```diff
@@ -193,7 +197,7 @@ impl ConjoiningClauses {
         let mut empty_because: Option<EmptyBecause> = None;
 
         // Walk each clause in turn, bailing as soon as we know this can't be simple.
-        let (join_clauses, mentioned_vars) = or_join.dismember();
+        let (join_clauses, _, mentioned_vars) = or_join.dismember();
```
ncalexan: nit: prefer to `_name` these things.

query-algebrizer/src/clauses/or.rs
```diff
@@ -481,6 +489,107 @@ impl ConjoiningClauses {
...
+    ///    (or [?page :page/title ?string]
+    ///        [?page :page/excerpt ?string]
+    ///        (and [?save :save/string ?string]
+    ///             [?page :page/save ?save]))]
+    /// ```edn
+    ///
+    /// would expand to something like
+    ///
+    /// ```sql
+    /// SELECT or123.page AS page FROM
```
ncalexan: Are you sure you want to accept arbitrary `edn::Value::PlainSymbol` here?  I expect you'll need to escape some things for this to work.

query-algebrizer/src/clauses/or.rs
```diff
@@ -481,6 +489,107 @@ impl ConjoiningClauses {
...
+        };
+
+        let template = self.use_as_template(&projected);
+
+        let mut acc = Vec::with_capacity(join_clauses.len());
+        let mut empty_because: Option<EmptyBecause> = None;
+
+        for clause in join_clauses.into_iter() {
+            let mut receptacle = template.make_receptacle();
+            match clause {
```
ncalexan: Keep in mind that you could make `OrWhereClause` iterable.  Not sure it's worth it -- this is lovely and explicit.

query-algebrizer/src/clauses/or.rs
```diff
@@ -481,6 +489,107 @@ impl ConjoiningClauses {
...
+
+        if acc.is_empty() {
+            self.mark_known_empty(empty_because.expect("empty for a reason"));
+            return Ok(());
+        }
+
+        let projection: BTreeSet<Variable> = projected.into_iter().collect();
+        let union = ComputedTable::Union(projection, acc);
+        let i = self.computed_tables.len();
+        let table = DatomsTable::Computed(i);
```
ncalexan: Eh?  This isn't what I expected.

query-sql/src/lib.rs
```diff
@@ -343,7 +345,15 @@ impl QueryFragment for TableOrSubquery {
     fn push_sql(&self, out: &mut QueryBuilder) -> BuildQueryResult {
         use self::TableOrSubquery::*;
         match self {
-            &Table(ref sa) => source_alias_push_sql(out, sa)
+            &Table(ref sa) => source_alias_push_sql(out, sa),
+            &Union(ref subqueries, ref table_alias) => {
+                out.push_sql("(");
+                interpose!(subquery, subqueries,
+                           { subquery.push_sql(out)? },
+                           { out.push_sql(" UNION ") });
```
ncalexan: Again -- `UNION DISTINCT`?  Or are you confident that you'll be able to handle such considerations at the root, and there's never a performance advantage to pushing the `DISTINCT` down to the leaves?

query-algebrizer/src/types.rs
rnewman: Strictly speaking it's not an ID, it's an index into the computed tables vector, hence the other chunk of code you asked about.

query-algebrizer/src/clauses/or.rs
rnewman: All of our SQL output is escaped. In that respect it's different from the ClojureScript version — in the Rust code our columns are named `?x` not `x`.

query-algebrizer/src/clauses/or.rs
rnewman: We produce a union. We find the index it _would be_ in the computed table vector (that is, the current length of that vector). We use that as part of the alias we create.

Then we put the union and the alias in `from`, and put the union in the right place in the computed table vector.

What I'd like is for `Vec::push` to return the index pushed into, but it doesn't do that. On Android I'd use a `SparseArray` to give effectively keyed-by-integer access to an array, but we don't have that either.

rnewman: The end result is that these computed queries appear in the expected place in `from`, and they have an index that can be used to retrieve them from a separate computed table collection. In the future we might use CTEs instead of inlining the query into `FROM` clauses.

rnewman: I've tidied this up a lot with a little trait.

query-algebrizer/src/clauses/mod.rs
```diff
@@ -283,35 +284,41 @@ impl ConjoiningClauses {
...
-                self.constrain_column_to_constant(table, column, bound_val);
-            } else {
-                match bound_val {
-                    TypedValue::Keyword(ref kw) => {
-                        if let Some(entid) = self.entid_for_ident(schema, kw) {
+            match column {
+                Column::Fixed(DatomsColumn::Value) | Column::Variable(_) => {
+                    self.constrain_column_to_constant(table, column, bound_val);
+                }
+                _ => {
```
ncalexan: In general, I've found wildcard matches to be worse than listing all the possibilities.  It's really nice to have the compiler tell you where you need to handle a newly added enum tag.

rnewman:
Yes, this is a bit of a complicated pile of code. Some of it is complicated but good, and some of it is complicated and I'm not super happy with.

Could you throw out some words or concepts that you found the most muddy?

> `ConsumableVec` is a big red flag that something's not right with this approach.

You would think so, but that kind of swapping is often discussed as the way around the borrow checker for exactly this kind of purpose.

I didn't really want to hang the computed table itself off the table alias: those get cloned around so that they can appear in clauses, and `CC` isn't `Clone`, of course. So I store the sub-CCs — subqueries, computed tables — in a collection on `CC`, and I refer to them by index.

The trouble is how to store them in such a way that we can consume them later. One could trust that we'll always see aliases in order — cc0, cc1, cc2 — in the `from` list, and just consume a vector of computed tables directly, as one often does with a vector. Rather than making that assumption, I instead consume them by index in the order the indices appear, which requires that trickery. We know for sure that we'll see each index, see each once, and they start at 0 — after all, we're allocating them as we push them into the vector — but the compiler doesn't know that.

This lets us refer to computed tables by integer, keeping table references cheap and small, but allows `from` clauses to safely appear in any order. I don't expect to use that capability, but it's the one that seems most likely to break…

I could also use an integer-keyed map of some kind, which has index-ish mutability, but is less clear. Or I could hang `Rc<ConjoiningClauses>` inside the table alias itself, dispensing with the computed tables vec. That slightly impedes our ability to translate these easily into `WITH` clauses if we later choose to do so, and it means some unnecessary refcounting, but I'd be OK with that approach.

query-algebrizer/src/clauses/or.rs
```diff
@@ -193,7 +211,7 @@ impl ConjoiningClauses {
         let mut empty_because: Option<EmptyBecause> = None;
 
         // Walk each clause in turn, bailing as soon as we know this can't be simple.
-        let (join_clauses, mentioned_vars) = or_join.dismember();
+        let (join_clauses, _unify_vars, mentioned_vars) = or_join.dismember();
```
rnewman: @ncalexan explicitly asked for it to be named. The leading underscore shows that it's unused.

query-translator/tests/translate.rs
```diff
@@ -259,3 +259,97 @@ fn test_simple_or_join() {
...
+                    (or-join [?page]
+                      [?page :page/url "http://foo.com/"]
+                      [?page :page/title "Foo"]
+                      (and
+                        [?page :page/save ?save]
+                        [?save :save/title "Foo"]))
+                    [?page :page/url ?url]
+                    [?page :page/description ?description]]"#;
+    let SQLQuery { sql, args } = translate(&schema, input, None);
+    assert_eq!(sql, "SELECT `datoms04`.v AS `?url`, \
```
ncalexan: You can use `r#"..."` to avoid the line break markers.

query-algebrizer/src/clauses/mod.rs
```diff
@@ -286,35 +287,55 @@ impl ConjoiningClauses {
...
-                        if let Some(entid) = self.entid_for_ident(schema, kw) {
+            match column {
+                Column::Variable(_) => {
+                    // We don't need to handle expansion of attributes here. The subquery that
+                    // produces the variable projection will do so.
+                    self.constrain_column_to_constant(table, column, bound_val);
+                },
+
+                Column::Fixed(DatomsColumn::ValueTypeTag) => {
+                    // I'm pretty sure this is meaningless.
```
rnewman: I think this is the _only_ bit of the space that's not valid — and it might eventually become so, if we allow something like `(= (typeof ?x) :db.valueType/instant)`. Right now it's meaningless because we don't allow you to say that `datoms00.value_type_tag = ?x`.

query-algebrizer/src/clauses/or.rs
```diff
@@ -481,6 +503,151 @@ impl ConjoiningClauses {
...
+        // each ready to project the same variables.
+        // At this point we can lift out any common type information (and even constraints) to the
+        // destination CC.
+        // We must also contribute type extraction information for any variables that aren't
+        // concretely typed for all union arms.
+        //
+        // For each variable in the list of variables to unify -- which will become our projection
+        // list -- find out its type info. This might be:
+        // 1. We knew the type from the outset.
+        // 2. We didn't, but each arm bound it to a single known type.
```
rnewman: 3 and 4 are both special cases of the negation of 2, or 4 is a special case of a generalization of 3.

3 is where no arm knows a single type: the type tags will be projected from every arm. 4 is where one or more arms know a single type, and those arms project a constant.

And so 2 means a single type shared by all arms:

```
[:find ?s :where
  (or [15 :foo/name ?s]
      [15 :foo/description ?s])]
```

— `?s` is a string for sure. 

I'll add some clarifying examples.

query-translator/tests/translate.rs
rnewman: This is actually deliberate: `r` preserves whitespace, and linebreaks do not, so by using linebreaks I can indent this query to make it readable, but still have it be equal to the raw query output!

query-algebrizer/src/clauses/mod.rs
```diff
@@ -49,6 +49,7 @@ use types::{
     ColumnConstraint,
     ColumnIntersection,
     ComputedTable,
+    Column,
```
rnewman: Yes, cheap. Will fix.

query-translator/src/translate.rs
```diff
@@ -148,14 +154,69 @@ pub struct ProjectedSelect{
...
+            // each arm will have different columns.
+            // TODO: type extraction.
+            let queries = arms.into_iter()
+                              .map(|cc| {
+                                    let var_columns = projection.iter().map(|var| {
+                                        let col = cc.column_bindings.get(&var).unwrap()[0].clone();
+                                        ProjectedColumn(ColumnOrExpression::Column(col), var.to_string())
+                                    }).collect();
+                                    let projection = Projection::Columns(var_columns);
+                                    cc_to_select_query(projection, cc, false, None)
```
rnewman: They're used in the next commit; see the commit message that explains we haven't got to that yet.

rnewman:
d8f7619

Issue #406, opened 2017-04-06
Set up a first benchmark


bgrins:
AIUI a good first candidate for benchmarking might be the tx-parser: https://github.com/mozilla/mentat/blob/rust/tx-parser/tests/parser.rs.

There's built in support for benchmark tests: https://doc.rust-lang.org/book/benchmark-tests.html, so let's try to use them if possible.

bgrins:
After adding a benchmark test for tx-parser, it can be run with `cargo bench --package mentat_tx_parser`

bgrins:
@ncalexan wanted to get some feedback on the approach here before opening a PR.

1) Any opinion on whether to incorporate these in the code or an external folder?  Here's some code that does [benchmarks in-file](https://github.com/mozilla/mentat/compare/rust...bgrins:initial-benchmarks?expand=1) and here's some that does [benchmarks in a separate file](https://github.com/mozilla/mentat/compare/rust...bgrins:initial-benchmarks-2?expand=1).  I'm leaning towards the latter myself so that we could see all benchmarks in one place, but I'm not sure if there's a criteria you've been using for when a test should be inline vs external that might be relevant here.
2) Any opinion on what to actually benchmark? Both examples above use the same code which I sniped / modified from some of the tests in `tx-parser/src/lib.rs` and are basically running `Tx::parse` through a situation with a small and large sample of edn.

ncalexan:
> @ncalexan wanted to get some feedback on the approach here before opening a PR.
> 
>     Any opinion on whether to incorporate these in the code or an external folder? Here's some code that does benchmarks in-file and here's some that does benchmarks in a separate file. I'm leaning towards the latter myself so that we could see all benchmarks in one place, but I'm not sure if there's a criteria you've been using for when a test should be inline vs external that might be relevant here.

Slight preference for external folder (`benches`?), like we have `tests` (for some things).

>     Any opinion on what to actually benchmark? Both examples above use the same code which I sniped / modified from some of the tests in tx-parser/src/lib.rs and are basically running Tx::parse through a situation with a small and large sample of edn.

This useful, for sure.  I've been thinking of the combinatorial product of:

* {N assertion,N retraction} for N=1,2,3,4,5,10,50,500,5000
* `:db/cardinality :db.cardinality/{one,many}`
* {`:db/unique :db.unique/identity`} where it makes sense

Into an empty store, and into a store with N=1,10,100,1000,10000,...1e6 existing datoms/transactions.

[...]

PR #413, opened 2017-04-13
Add an initial benchmark for the tx-parser crate. (#406) r=nalexander

bgrins:
This is an example first benchmark for the tx-parser.  It can be run with:

```cargo bench --package mentat_tx_parser```

To take measurements and print it in a nicer format I used this command:

```for i in `seq 1 5`; do echo '---'; cargo bench --package mentat_tx_parser 2> /dev/null | sed -n -e 's/test \(.*\)\( ... bench: \)/\1/p'; done;```

Here are my results for that locally:

    bench_parse1      7,462 ns/iter (+/- 7,514)
    bench_parse2    384,068 ns/iter (+/- 48,037)
    ---
    bench_parse1      7,171 ns/iter (+/- 1,431)
    bench_parse2    381,971 ns/iter (+/- 59,735)
    ---
    bench_parse1      6,860 ns/iter (+/- 1,742)
    bench_parse2    352,454 ns/iter (+/- 177,286)
    ---
    bench_parse1      6,787 ns/iter (+/- 3,107)
    bench_parse2    348,041 ns/iter (+/- 146,625)
    ---
    bench_parse1      7,113 ns/iter (+/- 961)
    bench_parse2    376,373 ns/iter (+/- 52,412)

### DIFF
--- a/tx-parser/benches/bench.rs
+++ b/tx-parser/benches/bench.rs
@@ -14,9 +14,7 @@ use mentat_tx_parser::Tx;
 fn bench_parse1(b: &mut Bencher) {
     let input = r#"[[:db/add 1 :test/val "a"]]"#;
     let parsed_edn = edn::parse::value(input).expect("to parse test input");
-    b.iter(|| {
-        return Tx::parse(parsed_edn.clone());
-    });
+    b.iter(|| Tx::parse(parsed_edn.clone()));
 }
 
 #[bench]
@@ -49,7 +47,5 @@ fn bench_parse2(b: &mut Bencher) {
          [:db/add 25 :test/val "y"]
          [:db/add 26 :test/val "z"]]"#;
     let parsed_edn = edn::parse::value(input).expect("to parse test input");
-    b.iter(|| {
-        return Tx::parse(parsed_edn.clone());
-    });
+    b.iter(|| Tx::parse(parsed_edn.clone()));
 }
--- a/tx-parser/benches/bench.rs
+++ b/tx-parser/benches/bench.rs
@@ -1,0 +1,55 @@
+#![feature(test)]
+
+// These benchmarks can be run from the project root with:
+// > cargo bench --package mentat_tx_parser
+
+extern crate test;
+extern crate edn;
+extern crate mentat_tx_parser;
+
+use test::Bencher;
+use mentat_tx_parser::Tx;
+
+#[bench]
+fn bench_parse1(b: &mut Bencher) {
+    let input = r#"[[:db/add 1 :test/val "a"]]"#;
+    let parsed_edn = edn::parse::value(input).expect("to parse test input");
+    b.iter(|| {
+        return Tx::parse(parsed_edn.clone());
+    });
+}
+
+#[bench]
+fn bench_parse2(b: &mut Bencher) {
+    let input = r#"
+        [[:db/add 1  :test/val "a"]
+         [:db/add 2  :test/val "b"]
+         [:db/add 3  :test/val "c"]
+         [:db/add 4  :test/val "d"]
+         [:db/add 5  :test/val "e"]
+         [:db/add 6  :test/val "f"]
+         [:db/add 7  :test/val "g"]
+         [:db/add 8  :test/val "h"]
+         [:db/add 9  :test/val "i"]
+         [:db/add 10 :test/val "j"]
+         [:db/add 11 :test/val "k"]
+         [:db/add 12 :test/val "l"]
+         [:db/add 13 :test/val "m"]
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by rnewman** [comment]
lol, I think that script doesn't like zsh:

```
$ for i in `seq 1 5`; do echo '---'; cargo bench --package mentat_tx_parser 2> /dev/null | sed -n -e 's/test \(.*\)\( ... bench: \)/\1/p'; done;
---
---
---
---
---
```

**Comment by rnewman** [comment]
Oh, it fails 'cos I'm not using Nightly.

**Comment by rnewman** [comment]
Older dual-core 13" rMBP:

```
test bench_parse1 ... bench:       8,959 ns/iter (+/- 1,450)
test bench_parse2 ... bench:     594,676 ns/iter (+/- 66,751)
---
test bench_parse1 ... bench:       8,634 ns/iter (+/- 4,373)
test bench_parse2 ... bench:     575,320 ns/iter (+/- 71,528)
```

File: tx-parser/benches/bench.rs [hunk 1]
```diff
@@ -0,0 +1,55 @@
...
+
+use test::Bencher;
+use mentat_tx_parser::Tx;
+
+#[bench]
+fn bench_parse1(b: &mut Bencher) {
+    let input = r#"[[:db/add 1 :test/val "a"]]"#;
+    let parsed_edn = edn::parse::value(input).expect("to parse test input");
+    b.iter(|| {
+        return Tx::parse(parsed_edn.clone());
```
rnewman: Nit: drop `return` and `;` (and below).

Concern: we're cloning the string here, inside the benchmark. Is it fairer to clone it outside? Or is it more realistic to call `edn::parse` inside the loop, and not explicitly clone?

bgrins: > Concern: we're cloning the string here, inside the benchmark. Is it fairer to clone it outside?

That seems ideal, but without the clone we get this error:
```
53 |         return Tx::parse(parsed_edn);
   |                          ^^^^^^^^^^ cannot move out of captured outer variable in an `FnMut` closure
```

> Or is it more realistic to call edn::parse inside the loop, and not explicitly clone?

We could - I thought cloning the parsed edn would be more narrow than doing the actual edn parsing in the loop.  If we wanted to fold edn parsing into this benchmark that'd be OK but it turns into a broader test - I figured we might want a separate benchmark for just parsing out the EDN.

With my test script and edn parsing in the loop I get:

```
bench_parse1      6,814 ns/iter (+/- 2,250)
bench_parse2    604,382 ns/iter (+/- 70,221)
---
bench_parse1      7,526 ns/iter (+/- 2,517)
bench_parse2    631,710 ns/iter (+/- 86,643)
---
bench_parse1      7,073 ns/iter (+/- 2,645)
bench_parse2    624,657 ns/iter (+/- 74,293)
---
bench_parse1      7,124 ns/iter (+/- 2,509)
bench_parse2    636,105 ns/iter (+/- 83,272)
---
bench_parse1      7,118 ns/iter (+/- 3,830)
bench_parse2    620,871 ns/iter (+/- 116,928)
```

And with the clone in the loop and parsing outside I get:

```
---
bench_parse1      6,925 ns/iter (+/- 1,087)
bench_parse2    382,473 ns/iter (+/- 63,815)
---
bench_parse1      7,073 ns/iter (+/- 1,339)
bench_parse2    428,671 ns/iter (+/- 185,621)
---
bench_parse1      7,392 ns/iter (+/- 2,354)
bench_parse2    398,395 ns/iter (+/- 134,769)
---
bench_parse1      7,044 ns/iter (+/- 792)
bench_parse2    394,561 ns/iter (+/- 44,045)
---
bench_parse1      6,685 ns/iter (+/- 1,071)
bench_parse2    387,890 ns/iter (+/- 70,388)
```

bgrins: Note: on those numbers I had only made the change to bench_parse2 so ignore bench_parse1

tx-parser/benches/bench.rs
```diff
@@ -0,0 +1,55 @@
...
+         [:db/add 20 :test/val "t"]
+         [:db/add 21 :test/val "u"]
+         [:db/add 22 :test/val "v"]
+         [:db/add 23 :test/val "w"]
+         [:db/add 24 :test/val "x"]
+         [:db/add 25 :test/val "y"]
+         [:db/add 26 :test/val "z"]]"#;
+    let parsed_edn = edn::parse::value(input).expect("to parse test input");
+    b.iter(|| {
+        return Tx::parse(parsed_edn.clone());
```
ncalexan: As @rnewman suggested, just `|| Tx::parse(...)` is idiomatic.

PR #416, opened 2017-04-15
Implement ordering

rnewman:
Built on top of #414.

Note that this _should_ order numbers correctly, because both `Long` and `Double` have the same type tag (5). But there's no test for that yet.

### DIFF
--- a/query-algebrizer/src/errors.rs
+++ b/query-algebrizer/src/errors.rs
@@ -31,7 +31,7 @@ error_chain! {
         }
 
         UnboundVariable(name: PlainSymbol) {
-            description("unbound variable in function call")
+            description("unbound variable in order clause or function call")
             display("unbound variable: {}", name)
         }
 
--- a/query-algebrizer/src/lib.rs
+++ b/query-algebrizer/src/lib.rs
@@ -30,6 +30,7 @@ use mentat_core::counter::RcCounter;
 use mentat_query::{
     FindQuery,
     FindSpec,
+    Order,
     SrcVar,
     Variable,
 };
@@ -44,8 +45,9 @@ pub use errors::{
 pub struct AlgebraicQuery {
     default_source: SrcVar,
     pub find_spec: FindSpec,
-    pub with: BTreeSet<Variable>,
     has_aggregates: bool,
+    pub with: BTreeSet<Variable>,
+    pub order: Option<Vec<OrderBy>>,
     pub limit: Option<u64>,
     pub cc: clauses::ConjoiningClauses,
 }
@@ -84,6 +86,42 @@ pub fn algebrize(schema: &Schema, parsed: FindQuery) -> Result<AlgebraicQuery> {
     algebrize_with_cc(schema, parsed, clauses::ConjoiningClauses::default())
 }
 
+/// Take an ordering list. Any variables that aren't fixed by the query are used to produce
+/// a vector of `OrderBy` instances, including type comparisons if necessary. This function also
+/// returns a set of variables that should be added to the `with` clause to make the ordering
+/// clauses possible.
+fn validate_and_simplify_order(cc: &ConjoiningClauses, order: Option<Vec<Order>>)
+    -> Result<(Option<Vec<OrderBy>>, BTreeSet<Variable>)> {
+    match order {
+        None => Ok((None, BTreeSet::default())),
+        Some(order) => {
+            let mut order_bys: Vec<OrderBy> = Vec::with_capacity(order.len() * 2);   // Space for tags.
+            let mut vars: BTreeSet<Variable> = BTreeSet::default();
+
+            for Order(direction, var) in order.into_iter() {
+                // Eliminate any ordering clauses that are bound to fixed values.
+                if cc.bound_value(&var).is_some() {
+                    continue;
+                }
+
+                // Fail if the var isn't bound by the query.
+                if !cc.column_bindings.contains_key(&var) {
+                    bail!(ErrorKind::UnboundVariable(var.name()));
+                }
+
+                // Otherwise, determine if we also need to order by type…
+                if cc.known_type(&var).is_none() {
+                    order_bys.push(OrderBy(direction.clone(), VariableColumn::VariableTypeTag(var.clone())));
+                }
+                order_bys.push(OrderBy(direction, VariableColumn::Variable(var.clone())));
+                vars.insert(var.clone());
+            }
+
+            Ok((if order_bys.is_empty() { None } else { Some(order_bys) }, vars))
+        }
+    }
+}
+
 #[allow(dead_code)]
 pub fn algebrize_with_cc(schema: &Schema, parsed: FindQuery, mut cc: ConjoiningClauses) -> Result<AlgebraicQuery> {
     // TODO: integrate default source into pattern processing.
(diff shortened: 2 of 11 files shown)


query-parser/src/parse.rs
```diff
@@ -119,6 +121,28 @@ def_parser!(Query, arguments, Vec<FnArg>, {
     (many::<Vec<FnArg>, _>(Query::fn_arg()))
 });
 
+def_parser!(Query, direction, Direction, {
+    satisfy_map(|v: edn::ValueAndSpan| {
```
ncalexan: Huh, I don't care about this, but I would have expected this to have been `variable().optional(or("asc", desc"))`.  Where you worried about backtracing?

query/src/lib.rs
```diff
@@ -129,6 +129,15 @@ impl PredicateFn {
 }
 
 #[derive(Clone, Debug, Eq, PartialEq)]
+pub enum Direction {
+    Ascending,
+    Descending,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Order(pub Direction, pub Variable);   // Future: Element instead of Variable?
```
ncalexan: Had to look up `Element`, but that's interesting -- ordering by an aggregate.  Is there a way to bind an aggregate to a named variable, so that you don't have to repeat the aggregate in the ordering clause?

query-algebrizer/src/errors.rs
```diff
@@ -31,7 +31,7 @@ error_chain! {
         }
 
         UnboundVariable(name: PlainSymbol) {
-            description("unbound variable in function call")
```
ncalexan: I expect this to happen a lot -- it will happen in function applications, etc.  Why don't we include some `context` argument to ease this?

query-parser/src/parse.rs
ncalexan: Oh, you want `?x` or `[desc ?x]`.  I see.

query-parser/tests/find_tests.rs
```diff
@@ -208,3 +210,27 @@ fn can_parse_simple_or_and_join() {
...
+fn can_parse_order_by() {
+    let invalid = "[:find ?x :where [?x :foo/baz ?y] :order]";
+    assert!(parse_find_string(invalid).is_err());
+
+    // Defaults to ascending.
+    let default = "[:find ?x :where [?x :foo/baz ?y] :order ?y]";
+    assert_eq!(parse_find_string(default).unwrap().order,
+               Some(vec![Order(Direction::Ascending, Variable::from_valid_name("?y"))]));
+
+    let ascending = "[:find ?x :where [?x :foo/baz ?y] :order (asc ?y)]";
```
ncalexan: I find the direction first awkward, especially since SQL is `ORDER BY variable ASC`.  Why not follow suit?

query-parser/tests/find_tests.rs
```diff
@@ -208,3 +210,27 @@ fn can_parse_simple_or_and_join() {
                    )),
                ]);
 }
+
+#[test]
+fn can_parse_order_by() {
+    let invalid = "[:find ?x :where [?x :foo/baz ?y] :order]";
```
ncalexan: Can you assert that repeating an ordered variable fails?  (Or, if you want to file follow-up, test that it _doesn't_ fail and that the later direction takes precedent?)

query-algebrizer/src/lib.rs
```diff
@@ -84,6 +86,42 @@ pub fn algebrize(schema: &Schema, parsed: FindQuery) -> Result<AlgebraicQuery> {
     algebrize_with_cc(schema, parsed, clauses::ConjoiningClauses::default())
 }
 
+/// Take an ordering list. Any variables that aren't fixed by the query are used to produce
```
ncalexan: This comment helped.  Thanks!

query-algebrizer/src/lib.rs
```diff
@@ -84,6 +86,42 @@ pub fn algebrize(schema: &Schema, parsed: FindQuery) -> Result<AlgebraicQuery> {
...
+/// a vector of `OrderBy` instances, including type comparisons if necessary. This function also
+/// returns a set of variables that should be added to the `with` clause to make the ordering
+/// clauses possible.
+fn validate_and_simplify_order(cc: &ConjoiningClauses, order: Option<Vec<Order>>)
+    -> Result<(Option<Vec<OrderBy>>, BTreeSet<Variable>)> {
```
ncalexan: Is there an advantage to differentiating `None` from `vec![]`?  I can't see it.

query-translator/tests/translate.rs
```diff
@@ -369,4 +369,25 @@ fn test_with_without_aggregate() {
...
+                     WHERE `datoms00`.a = 99 \
+                     ORDER BY `?y` DESC");
+
+    // Unknown type.
+    let input = r#"[:find ?x :with ?y :where [?x _ ?y] :order ?y ?x]"#;
+    let SQLQuery { sql, args } = translate(&schema, input, None);
+    assert_eq!(sql, "SELECT DISTINCT `all_datoms00`.e AS `?x`, `all_datoms00`.v AS `?y`, \
+                                     `all_datoms00`.value_type_tag AS `?y_value_type_tag` \
+                     FROM `all_datoms` AS `all_datoms00` \
+                     ORDER BY `?y_value_type_tag` ASC, `?y` ASC, `?x` ASC");
```
ncalexan: Can you assert that the `value_type_tag` ordering is constant independent of the `value` ordering?  (I think it should be.)

query-algebrizer/src/lib.rs
```diff
@@ -84,6 +86,42 @@ pub fn algebrize(schema: &Schema, parsed: FindQuery) -> Result<AlgebraicQuery> {
...
+                }
+
+                // Fail if the var isn't bound by the query.
+                if !cc.column_bindings.contains_key(&var) {
+                    bail!(ErrorKind::UnboundVariable(var.name()));
+                }
+
+                // Otherwise, determine if we also need to order by type…
+                if cc.known_type(&var).is_none() {
+                    order_bys.push(OrderBy(direction.clone(), VariableColumn::VariableTypeTag(var.clone())));
```
ncalexan: I think the type tags should be ordered independently of the direction.  That is, I always want booleans-before-longs, regardless of how I order the results.  I think this does the right thing for comparable types (long, double) too -- since they share a type tag.

query-parser/tests/find_tests.rs
rnewman: Because everything else in this formulation of Datalog is prefixed — aggregate calls are `(max ?x)`, inequalities are `(< ?x ?y)`, etc.

If we support ordering by expressions, we'll have things like `(asc (+ ?x ?y))` here.

I think you're suggesting `(?x :asc)`, which would extend to `((+ ?x ?y) asc)`, which is really unusual for a parenthesized language. Is that what you meant, or am I misunderstanding?

ncalexan: OK, I think you are right: this is lisp-y.  Withdrawn :)

query-algebrizer/src/lib.rs
rnewman: Not really; just tried to avoid allocating an empty vector in some places. This is just shitty code.

query-algebrizer/src/lib.rs
rnewman: Interesting idea. That's not the way it's normally done. SQLite, for example:

https://sqlite.org/datatype3.html#sort_order

> An INTEGER or REAL value is less than any TEXT or BLOB value. When an INTEGER or REAL is compared to another INTEGER or REAL, a numerical comparison is performed.
>
> A TEXT value is less than a BLOB value. When two TEXT values are compared an appropriate collating sequence is used to determine the result.

That is, if you order by `?x ASC` in SQL, you get numbers before text; if you order `?x DESC`, you get text before numbers.

Can you go into more detail about your motivation?

query/src/lib.rs
rnewman: We could do it by allowing you to name aggregate expressions, as SQL allows:

```
[:find [(max ?age) ?oldest]
 :where [_ :foo/age ?age] …
 :order ?oldest]
```

but at present there's no way to do it.

query-algebrizer/src/lib.rs
ncalexan: I feel like that's an SQLite-ism, not a SQL-ism, although perhaps I'm incorrect.  But now that I think more about it, I'm not sure why I felt so clearly that we should order by type tags independently of values.  I guess if you flip `ASC` to `DESC`, you expect the result list to be provided in the opposite order.

It's worth calling out that we're moving away from Datomic's set model with this change, although I don't quite know where to capture this.

I think I'm wrong here.  Keep it as it is.

rnewman:
35d73d5

PR #418, opened 2017-04-17
Finish passing input parameters through to queries

rnewman:
In the course of starting another feature, I realized this wasn't finished.

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -17,6 +17,7 @@ extern crate edn;
 pub mod values;
 
 use std::collections::BTreeMap;
+use std::fmt;
 use std::rc::Rc;
 use self::ordered_float::OrderedFloat;
 use self::edn::NamespacedKeyword;
@@ -44,16 +45,30 @@ pub enum ValueType {
 }
 
 impl ValueType {
-    pub fn to_edn_value(&self) -> edn::Value {
+    pub fn to_edn_value(self) -> edn::Value {
         match self {
-            &ValueType::Ref => values::DB_TYPE_REF.clone(),
-            &ValueType::Boolean => values::DB_TYPE_BOOLEAN.clone(),
-            &ValueType::Instant => values::DB_TYPE_INSTANT.clone(),
-            &ValueType::Long => values::DB_TYPE_LONG.clone(),
-            &ValueType::Double => values::DB_TYPE_DOUBLE.clone(),
-            &ValueType::String => values::DB_TYPE_STRING.clone(),
-            &ValueType::Keyword => values::DB_TYPE_KEYWORD.clone(),
+            ValueType::Ref => values::DB_TYPE_REF.clone(),
+            ValueType::Boolean => values::DB_TYPE_BOOLEAN.clone(),
+            ValueType::Instant => values::DB_TYPE_INSTANT.clone(),
+            ValueType::Long => values::DB_TYPE_LONG.clone(),
+            ValueType::Double => values::DB_TYPE_DOUBLE.clone(),
+            ValueType::String => values::DB_TYPE_STRING.clone(),
+            ValueType::Keyword => values::DB_TYPE_KEYWORD.clone(),
         }
+    }
+}
+
+impl fmt::Display for ValueType {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", match *self {
+            ValueType::Ref =>     "db.type/ref",
+            ValueType::Boolean => "db.type/boolean",
+            ValueType::Instant => "db.type/instant",
+            ValueType::Long =>    "db.type/long",
+            ValueType::Double =>  "db.type/double",
+            ValueType::String =>  "db.type/string",
+            ValueType::Keyword => "db.type/keyword",
+        })
     }
 }
 
--- a/query-algebrizer/src/clauses/inputs.rs
+++ b/query-algebrizer/src/clauses/inputs.rs
@@ -1,0 +1,72 @@
+// Copyright 2016 Mozilla
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+use std::collections::BTreeMap;
+
+use mentat_core::{
+    TypedValue,
+    ValueType,
+};
+
+use mentat_query::{
+    Variable,
+};
+
+pub use errors::{
+    Error,
+    ErrorKind,
+    Result,
+};
+
+/// Define the inputs to a query. This is in two parts: a set of values known now, and a set of
+/// types known now.
+/// The separate map of types is to allow queries to be algebrized without full knowledge of
+/// the bindings that will be used at execution time.
+/// When built correctly, `types` is guaranteed to contain the types of `values` -- use
+/// `QueryInputs::new` or `QueryInputs::with_values` to construct an instance.
+pub struct QueryInputs {
+    pub types: BTreeMap<Variable, ValueType>,
+    pub values: BTreeMap<Variable, TypedValue>,
+}
+
+impl Default for QueryInputs {
+    fn default() -> Self {
+        QueryInputs { types: BTreeMap::default(), values: BTreeMap::default() }
+    }
+}
+
+impl QueryInputs {
+    pub fn with_value_sequence(vals: Vec<(Variable, TypedValue)>) -> QueryInputs {
+        let values: BTreeMap<Variable, TypedValue> = vals.into_iter().collect();
+        QueryInputs::with_values(values)
+    }
+
+    pub fn with_values(values: BTreeMap<Variable, TypedValue>) -> QueryInputs {
+        QueryInputs {
+            types: values.iter().map(|(var, val)| (var.clone(), val.value_type())).collect(),
+            values: values,
+        }
+    }
+
+    pub fn new(mut types: BTreeMap<Variable, ValueType>,
+               values: BTreeMap<Variable, TypedValue>) -> Result<QueryInputs> {
+        // Make sure that the types of the values agree with those in types, and collect.
+        for (var, v) in values.iter() {
+            let t = v.value_type();
+            let old = types.insert(var.clone(), t);
+            if let Some(old) = old {
+                if old != t {
+                    bail!(ErrorKind::InputTypeDisagreement(var.name(), old, t));
+                }
+            }
+        }
+        Ok(QueryInputs { types: types, values: values })
+    }
+}
--- a/query-algebrizer/src/clauses/mod.rs
+++ b/query-algebrizer/src/clauses/mod.rs
@@ -89,6 +89,7 @@ trait Contains<K, T> {
 
 trait Intersection<K> {
     fn with_intersected_keys(&self, ks: &BTreeSet<K>) -> Self;
+    fn keep_intersected_keys(&mut self, ks: &BTreeSet<K>);
 }
 
 impl<K: Ord, T> Contains<K, T> for BTreeSet<K> {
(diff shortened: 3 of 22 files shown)


rnewman:
I just realized this allows for inputs not named in `:in`. Fixing.

rnewman:
Oh, and we don't parse `:in` yet. Heh.

rnewman:
@ncalexan: part 1 needs some advice from you on how to get things to compile when returning an error; I can't get it to work.

core/src/lib.rs
```diff
@@ -45,18 +46,32 @@ pub enum ValueType {
...
     pub fn to_edn_value(&self) -> edn::Value {
-        match self {
-            &ValueType::Ref => values::DB_TYPE_REF.clone(),
-            &ValueType::Boolean => values::DB_TYPE_BOOLEAN.clone(),
-            &ValueType::Instant => values::DB_TYPE_INSTANT.clone(),
-            &ValueType::Long => values::DB_TYPE_LONG.clone(),
-            &ValueType::Double => values::DB_TYPE_DOUBLE.clone(),
-            &ValueType::String => values::DB_TYPE_STRING.clone(),
-            &ValueType::Keyword => values::DB_TYPE_KEYWORD.clone(),
+        match *self {
```
rnewman: I did this on the way, but I suspect it should actually be `to_edn_value(self)` — `ValueType` is `Copy`, so avoiding references wherever possible is best.

query-algebrizer/src/clauses/mod.rs
```diff
@@ -223,6 +243,41 @@ impl ConjoiningClauses {
...
+
+    pub fn with_inputs_and_alias_counter<T>(in_variables: BTreeSet<Variable>,
+                                            inputs: T,
+                                            alias_counter: RcCounter) -> ConjoiningClauses
+    where T: Into<Option<QueryInputs>> {
+        match inputs.into() {
+            None => ConjoiningClauses::with_alias_counter(alias_counter),
+            Some(QueryInputs { mut types, mut values }) => {
+                println!("Got inputs {:?}", values);
+                println!("Got input vars {:?}", in_variables);
```
rnewman: I'll remove these locally.

query-parser/src/parse.rs
```diff
@@ -65,6 +67,11 @@ error_chain! {
     }
 
     errors {
+        DuplicateVariableError {
```
ncalexan: Worth decorating with the variable, I think.  (I see your TODO below.)

query/src/lib.rs
```diff
@@ -138,7 +147,7 @@ pub enum Direction {
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct Order(pub Direction, pub Variable);   // Future: Element instead of Variable?
 
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
```
ncalexan: nit: prefer alphabetical order.

query-algebrizer/src/clauses/inputs.rs
```diff
@@ -0,0 +1,72 @@
...
+use mentat_core::{
+    TypedValue,
+    ValueType,
+};
+
+use mentat_query::{
+    Variable,
+};
+
+pub use errors::{
```
ncalexan: `pub`?

query-algebrizer/src/clauses/inputs.rs
```diff
@@ -0,0 +1,72 @@
...
+};
+
+/// Define the inputs to a query. This is in two parts: a set of values known now, and a set of
+/// types known now.
+/// The separate map of types is to allow queries to be algebrized without full knowledge of
+/// the bindings that will be used at execution time.
+/// When built correctly, `types` is guaranteed to contain the types of `values` -- use
+/// `QueryInputs::new` or `QueryInputs::with_values` to construct an instance.
+pub struct QueryInputs {
+    pub types: BTreeMap<Variable, ValueType>,
```
ncalexan: Do these need to be `pub`?  You could make a stronger statement if they were hidden, and only exposed via accessors.

query-algebrizer/src/clauses/inputs.rs
```diff
@@ -0,0 +1,72 @@
...
+    pub values: BTreeMap<Variable, TypedValue>,
+}
+
+impl Default for QueryInputs {
+    fn default() -> Self {
+        QueryInputs { types: BTreeMap::default(), values: BTreeMap::default() }
```
ncalexan: Didn't we settle on
```rust
Name {
    x: X,
    y: Y,
    ...
}
```
(Here and below.)

query-algebrizer/src/clauses/inputs.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+}
+
+impl QueryInputs {
+    pub fn with_value_sequence(vals: Vec<(Variable, TypedValue)>) -> QueryInputs {
```
ncalexan: Both of these could be `From`... although I'm happy with explicit constructors.  Easier to code complete, that's for sure!

core/src/lib.rs
```diff
@@ -44,19 +45,33 @@ pub enum ValueType {
...
+            ValueType::String => values::DB_TYPE_STRING.clone(),
+            ValueType::Keyword => values::DB_TYPE_KEYWORD.clone(),
         }
     }
 }
 
+impl fmt::Display for ValueType {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", match *self {
+            ValueType::Ref =>     "db.type/ref",
```
ncalexan: Is there a reason you're not including the leading colon?  I don't feel strongly but it might be nice if `x.to_string() == x.to_edn_value().to_string()`.

query-algebrizer/src/clauses/mod.rs
```diff
@@ -108,6 +109,22 @@ impl<K: Clone + Ord, V: Clone> Intersection<K> for BTreeMap<K, V> {
             .filter_map(|(k, v)| ks.when_contains(k, || (k.clone(), v.clone())))
             .collect()
     }
+
+    /// Remove all keys from the map that are not present in `ks`.
+    /// This implementation is terrible because there's no mutable iterator for BTreeMap.
+    fn keep_intersected_keys(&mut self, ks: &BTreeSet<K>) {
+        let mut to_remove = Vec::with_capacity(self.len() - ks.len());
+        {
```
ncalexan: Annotate scope braces with what they're holding/dropping, and is there a reason to not indent?

query-algebrizer/src/clauses/mod.rs
```diff
@@ -108,6 +109,22 @@ impl<K: Clone + Ord, V: Clone> Intersection<K> for BTreeMap<K, V> {
             .filter_map(|(k, v)| ks.when_contains(k, || (k.clone(), v.clone())))
             .collect()
     }
+
+    /// Remove all keys from the map that are not present in `ks`.
+    /// This implementation is terrible because there's no mutable iterator for BTreeMap.
+    fn keep_intersected_keys(&mut self, ks: &BTreeSet<K>) {
+        let mut to_remove = Vec::with_capacity(self.len() - ks.len());
```
ncalexan: I expect this to panic when `self.len()` is less than `ks.len()`.  That might be okay, but can you be safe about it?

query-algebrizer/src/clauses/mod.rs
```diff
@@ -89,6 +89,7 @@ trait Contains<K, T> {
 
 trait Intersection<K> {
     fn with_intersected_keys(&self, ks: &BTreeSet<K>) -> Self;
+    fn keep_intersected_keys(&mut self, ks: &BTreeSet<K>);
```
ncalexan: You have one mutating and one non-mutating method.  Is there a compelling reason to have different API regimes?

query-algebrizer/src/clauses/mod.rs
```diff
@@ -240,6 +243,38 @@ impl ConjoiningClauses {
...
+    }
+
+    pub fn with_inputs_and_alias_counter<T>(in_variables: BTreeSet<Variable>,
+                                            inputs: T,
+                                            alias_counter: RcCounter) -> ConjoiningClauses
+    where T: Into<Option<QueryInputs>> {
+        match inputs.into() {
+            None => ConjoiningClauses::with_alias_counter(alias_counter),
+            Some(QueryInputs { mut types, mut values }) => {
+                // Discard any bindings not mentioned in our :in clause.
```
ncalexan: Do we test somewhere that all `:in` variables appear?  I feel like Datomic probably doesn't care, but we should probably fail in this case, since it's likely to be an error.

src/errors.rs
```diff
@@ -40,6 +42,11 @@ error_chain! {
     }
 
     errors {
+        UnboundVariables(names: BTreeSet<String>) {
+            description("unbound variables at execution time")
```
ncalexan: nit: "query execution time"?  There are many things being executed at many times.

src/query.rs
```diff
@@ -75,6 +78,13 @@ pub fn q_once<'sqlite, 'schema, 'query, T, U>
 
     algebrized.apply_limit(limit.into());
 
+    // Because this is q_once, we can check that all of our `:in` variables are bound at this point.
+    // If they aren't, the user has made an error -- perhaps writing the wrong variable in `:in`, or
+    // not binding in the `QueryInput`.
+    let unbound = algebrized.unbound_variables();
```
ncalexan: Ah!  I think this closes an open loop from earlier.

rnewman:
Thanks for the help, Nick!

aa14a71

query-parser/src/parse.rs
rnewman: Unfortunately it's a little tricky with the way I've implemented it; I hope this can be part of an ergonomics pass in a few months. It should be pretty obvious to users when there are duplicated variables.

query/src/lib.rs
rnewman: This was copypasta'd from elsewhere, but sure.

query-algebrizer/src/clauses/inputs.rs
rnewman: It means "reexporting".

query-algebrizer/src/clauses/inputs.rs
rnewman: There'll be another one for specifying types, I expect, and perhaps a builder pattern, so `From` might get confusing!

core/src/lib.rs
rnewman: Just omission. I'll fix.

query-algebrizer/src/clauses/inputs.rs
rnewman: Oh, I get your question. Yeah, probably unnecessary.

query-algebrizer/src/clauses/inputs.rs
rnewman: I'd like them not to be, but there's no concept of crate-private in Rust. I could flatten `inputs.rs` into `mod.rs`…

query-algebrizer/src/clauses/mod.rs
rnewman: Well, it means we can write:

```
            value_bindings: self.value_bindings.with_intersected_keys(&vars),
```

rather than

```
let new_value_bindings = self.value_bindings.clone();
new_value_bindings.keep_intersected_keys(&vars);
…
    value_bindings: new_value_bindings,
```

The reason for `k_i_k` is because we're already taking ownership of the inputs, so it's wasteful to clone into a new map.

query-algebrizer/src/clauses/mod.rs
rnewman: Good catch!

query-algebrizer/src/clauses/mod.rs
rnewman: They were no longer necessary, apparently!

query-algebrizer/src/clauses/mod.rs
rnewman: We do check that they're all bound. I don't think we check that they're a subset of the variables in the query. I'll file.

src/query.rs
rnewman: I did this because I wrote a test that failed because I typoed the `:in` variable :)

rnewman:
Review comments addressed in bffefe7.

PR #420, opened 2017-04-19
Add :limit to queries

rnewman:
At present we take limits as an argument alongside the query.

That's flawed: the limit is used when algebrizing (no need to use `DISTINCT` when only fetching one result); in some cases it can be included in the generated SQL; and it breaks our ability to associate cached prepared statements back to input Datalog. Quite apart from all that, it's nice to be able to stuff everything — including ordering and limiting — into the same blob, and supporting limit arguments makes our code more complex.

I plan to:

- Parse `:limit` alongside `:order`.
- Allow the use of variable bindings for `:limit`.
- Remove the existing limit argument mechanism.

### DIFF
--- a/query-algebrizer/src/errors.rs
+++ b/query-algebrizer/src/errors.rs
@@ -47,6 +47,11 @@ error_chain! {
             display("invalid argument to {}: expected numeric in position {}.", function, position)
         }
 
+        InvalidLimit(val: String, kind: ValueType) {
+            description("invalid limit")
+            display("invalid limit {} of type {}: expected natural number.", val, kind)
+        }
+
         NonMatchingVariablesInOrClause {
             // TODO: flesh out.
             description("non-matching variables in 'or' clause")
--- a/query-parser/src/parse.rs
+++ b/query-parser/src/parse.rs
@@ -18,7 +18,7 @@ use std; // To refer to std::result::Result.
 use std::collections::BTreeSet;
 
 use self::combine::{eof, many, many1, optional, parser, satisfy, satisfy_map, Parser, ParseResult, Stream};
-use self::combine::combinator::{choice, or, try};
+use self::combine::combinator::{any, choice, or, try};
 
 use self::mentat_parser_utils::{
     ResultParser,
@@ -43,6 +43,7 @@ use self::mentat_query::{
     FindSpec,
     FnArg,
     FromValue,
+    Limit,
     Order,
     OrJoin,
     OrWhereClause,
@@ -102,6 +103,16 @@ error_chain! {
             description("missing field")
             display("missing field: '{}'", value)
         }
+
+        UnknownLimitVar(var: edn::PlainSymbol) {
+            description("limit var not present in :in")
+            display("limit var {} not present in :in", var)
+        }
+
+        InvalidLimit(val: edn::Value) {
+            description("limit value not valid")
+            display("expected natural number, got {}", val)
+        }
     }
 }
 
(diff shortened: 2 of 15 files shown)


rnewman:
This PR goes part-way to allowing algebrizing to occur without all variables being bound.

You'll see a TODO in the test where we do the wrong thing. But variable limits are working just fine!

query-translator/tests/translate.rs
```diff
@@ -255,7 +306,7 @@ fn test_simple_or_join() {
                       [?page :page/title "Foo"])
                     [?page :page/url ?url]
                     [?page :page/description ?description]]"#;
-    let SQLQuery { sql, args } = translate(&schema, query, None);
```
ncalexan: I was going to moan about all of these `None` instances... but they're deletions \o/

query-algebrizer/src/errors.rs
```diff
@@ -47,6 +47,11 @@ error_chain! {
             display("invalid argument to {}: expected numeric in position {}.", function, position)
         }
 
+        InvalidLimit(val: String, kind: ValueType) {
```
ncalexan: Fold this into the earlier patch; I think it'll be easier to parse.

query-translator/tests/translate.rs
```diff
@@ -171,6 +171,28 @@ fn test_bound_variable_limit_affects_distinct() {
...
+    let algebrized = algebrize(&schema, parsed).expect("algebrize failed");
+
+    // The type is known.
+    assert_eq!(Some(ValueType::Long),
+               algebrized.cc.known_type(&Variable::from_valid_name("?limit")));
+
+    let select = query_to_select(algebrized);
+    let SQLQuery { sql, args } = select.query.to_sql_query().unwrap();
+
+    // TODO: this query isn't actually correct -- we don't yet algebrize for variables that are
```
ncalexan: Ah, I was looking for this case in your tests.

query-parser/src/parse.rs
```diff
@@ -402,17 +431,28 @@ def_parser!(Find, query, FindQuery, {
...
                     FindQueryPart::FindSpec(x) => find_spec = Some(x),
                     FindQueryPart::In(x) => in_vars = Some(x),
+                    FindQueryPart::Limit(x) => limit = x,
                     FindQueryPart::Order(x) => order_clauses = Some(x),
                     FindQueryPart::WhereClauses(x) => where_clauses = Some(x),
                     FindQueryPart::With(x) => with_vars = Some(x),
                 }
             }
 
+            // Make sure that if we have `:limit ?x`, `?x` appears in `:in`.
```
rnewman: I opted for a kind of internal consistency: everything between those `[]` is in the same namespace of variables, is bound in the same way, and so on.

Only a step away from that is ensuring that a limit isn't used elsewhere within the query, as you suggest — yes, it might well be human error.

But that's an additional bit of metadata that would need to be passed around, an additional bit of validation, and it raises good questions: can I project the limit from the query to allow for pipelined processing of results, even if I'm not allowed to use a limit var inside `:where`?

There are also legitimate reasons why a user might want to use a limit as part of `:where`:

```
[:find ?x ?person
 :in ?limit
 :where
 [?x :foo/knows ?person]
 (or-join [?person]
   [?person :foo/employment "faculty"]
   (> ?limit 10))
 :limit ?limit]
```

— that is, find _everyone_ `?x` knows if we're fetching more than 10, but only include faculty if we're fetching 10 or fewer. You can make a similar case for including limits inside `:order` clauses to control ordering based on how many results will be returned.

query/src/lib.rs
```diff
@@ -377,6 +377,13 @@ pub enum Element {
     // Pull(Pull),             // TODO
 }
 
+#[derive(Clone,Debug,Eq,PartialEq)]
+pub enum Limit {
+    None,
+    Fixed(u64),
+    Variable(Variable),
```
rnewman: No — only named variables.

query-parser/tests/find_tests.rs
```diff
@@ -234,3 +235,35 @@ fn can_parse_order_by() {
...
+
+    let onethousand = "[:find ?x :where [?x :foo/baz ?y] :limit 1000]";
+    assert_eq!(parse_find_string(onethousand).unwrap().limit,
+               Limit::Fixed(1000));
+
+    let variable_with_in = "[:find ?x :in ?limit :where [?x :foo/baz ?y] :limit ?limit]";
+    assert_eq!(parse_find_string(variable_with_in).unwrap().limit,
+               Limit::Variable(Variable::from_valid_name("?limit")));
+
+    let variable_with_in_used = "[:find ?x :in ?limit :where [?x :foo/baz ?limit] :limit ?limit]";
```
rnewman: See previous (long!) comment. Note also that SQLite allows this for both positional and named parameters.

Issue #303, opened 2017-02-15
[query] Translate `not` patterns


rnewman:
On top of #301 and #302, translating a `not-join` to SQL involves one of two branches:

* If the enclosed CC doesn't generate any bindings, then it can simply be an expression negating the CC's `wheres`:

    ```
    NOT (datoms123.e = 456 AND …)
    ```

* If the enclosed CC does generate bindings, then we must generate a subquery:

    ```
    NOT EXISTS (SELECT …)
    ```

This ticket will reuse a lot of code from #301, and will make it clear whether the abstractions in that ticket were well chosen!

fluffyemily:
https://github.com/mozilla/mentat/commit/bd389d2f0de395ef1bc46ea4363def098e2bd687

Issue #302, opened 2017-02-15
[query] Algebrize `not` patterns


rnewman:
After parsing these in #239, we need to build a simple wrapper around a CC to denote negation.

`not-join` is the simplest extension: it's simply a set of patterns (which we term a CC) and a list of variables to unify with the enclosing query.

fluffyemily:
https://github.com/mozilla/mentat/commit/bd389d2f0de395ef1bc46ea4363def098e2bd687

Issue #293, opened 2017-02-14
[query] Type code expansion


rnewman:
As we process a collection of patterns into a CC, we collect constraints. For example,

```edn
[?x ?a true]
```

yields a constraint like

```
ColumnConstraint::EqualsValue(d0_v, TypedValue::Boolean(true))
```

Some types (including boolean) share a SQLite value representation with other types — `1` could be a ref, an integer, an FTS string, or a boolean.

This ticket tracks expanding appropriate `value_type_tag` constraints out of a set of column constraints once all value constraints are known. We naturally must wait until _all_ value constraints are known, because `?a` might later be bound, which will eliminate the need to constrain `value_type_tag`.

It's possible at this point for the space of acceptable type tags to not intersect: _e.g._, the query

```edn
[:find ?x :where
 [?x ?y true]
 [?z ?y ?x]]
```

where `?y` must simultaneously be a ref-typed attribute and a boolean-typed attribute, which is impossible. This function should deduce that and call `ConjoiningClauses.mark_known_empty`.

See also #292.

fluffyemily:
Can I clarify I need to do 3 things here.
1. Identify all constratins with ambiguous values and add `value_type_tag` constraints to clarify.
2. Identify all constraints with values assigned that do not match their `known_type` and `mark_known_empty`.
3. Identify all `extracted_types` and add `value_type_tag` constraints.

rnewman:
There are three kinds of queries that you need to make behave.

The first is a join between type-ambiguous values.

```
[:find ?x ?y ?z
 :where
 [?x _ ?y]
 [?z _ ?y]]
```

— `?y` could be anything, so we not only need `datoms00.v = datoms01.v`, but we also need `datoms00.value_type_tag = datoms01.value_type_tag`. We know all of those columns in `column_bindings`, and we expand those in `expand_column_bindings`; you need to do a similar thing when you fill out `expand_type_tags`.

The second is a join between ambiguous and unambiguous.

```
[:find ?x ?y ?z
 :where
 [?x _ ?y]
 [?y _ ?z]]
```
— here `?y` must be `ValueType::Ref`, so we need to generate `datoms00.value_type_tag = 1 AND datoms00.v = datoms01.e`. I'm not sure if a place for this already exists: after all, we think that `?y` is unambiguous, but we won't be filtering the first table correctly.

The third is a type-imposing constraint.

```
[:find ?x ?v
 :where
 [?x _ ?v]
 [(< ?v 10)]]
```

[...]

PR #422, opened 2017-04-19
Parse and Algebrize `not` & `not-join`.

fluffyemily:
Parse, Algebrize and Translate `not` and `not-join`

https://github.com/mozilla/mentat/issues/389
https://github.com/mozilla/mentat/issues/302
https://github.com/mozilla/mentat/issues/303

### DIFF
--- a/query-algebrizer/src/clauses/mod.rs
+++ b/query-algebrizer/src/clauses/mod.rs
@@ -833,7 +833,8 @@ impl ConjoiningClauses {
                 self.apply_or_join(schema, o)
             },
             WhereClause::NotJoin(n) => {
-                validate_not_join(&n)
+                validate_not_join(&n)?;
+                self.apply_not_join(schema, n)
             },
             _ => unimplemented!(),
         }
--- a/query-algebrizer/src/clauses/mod.rs
+++ b/query-algebrizer/src/clauses/mod.rs
@@ -63,11 +63,15 @@ use types::{
 
 mod inputs;
 mod or;
+mod not;
 mod pattern;
 mod predicate;
 mod resolve;
 
-use validate::validate_or_join;
+use validate::{
+    validate_or_join,
+    validate_not_join,
+};
 
 pub use self::inputs::QueryInputs;
 
@@ -811,6 +815,9 @@ impl ConjoiningClauses {
             WhereClause::OrJoin(o) => {
                 validate_or_join(&o)?;
                 self.apply_or_join(schema, o)
+            },
+            WhereClause::NotJoin(n) => {
+                validate_not_join(&n)
             },
             _ => unimplemented!(),
         }
--- a/query-algebrizer/src/clauses/mod.rs
+++ b/query-algebrizer/src/clauses/mod.rs
@@ -69,8 +69,8 @@ mod predicate;
 mod resolve;
 
 use validate::{
-    validate_or_join,
     validate_not_join,
+    validate_or_join,
 };
 
 pub use self::inputs::QueryInputs;
--- a/query-algebrizer/src/clauses/mod.rs
+++ b/query-algebrizer/src/clauses/mod.rs
@@ -207,6 +207,22 @@ pub struct ConjoiningClauses {
     pub extracted_types: BTreeMap<Variable, QualifiedAlias>,
 }
 
+impl PartialEq for ConjoiningClauses {
+    fn eq(&self, other: &ConjoiningClauses) -> bool {
+        self.empty_because.eq(&other.empty_because) &&
+        self.from.eq(&other.from) &&
+        self.computed_tables.eq(&other.computed_tables) &&
+        self.wheres.eq(&other.wheres) &&
+        self.column_bindings.eq(&other.column_bindings) &&
+        self.input_variables.eq(&other.input_variables) &&
+        self.value_bindings.eq(&other.value_bindings) &&
+        self.known_types.eq(&other.known_types) &&
+        self.extracted_types.eq(&other.extracted_types)
+    }
+}
+
+impl Eq for ConjoiningClauses {}
+
 impl Debug for ConjoiningClauses {
     fn fmt(&self, fmt: &mut Formatter) -> ::std::fmt::Result {
         fmt.debug_struct("ConjoiningClauses")
--- a/query-algebrizer/src/clauses/not.rs
+++ b/query-algebrizer/src/clauses/not.rs
@@ -31,18 +31,12 @@ impl ConjoiningClauses {
         };
 
         let mut template = self.use_as_template(&unified);
-        let mut empty_because: Option<EmptyBecause> = None;
 
         for clause in not_join.clauses.into_iter() {
             template.apply_clause(&schema, clause)?;
-            // This is disgusting, but I can't think of anything else right now that works.
-            if template.is_known_empty() {
-                empty_because = template.empty_because;
-                template.empty_because = None;
-            }
         }
 
-        if template.wheres.is_empty() && empty_because.is_some() {
+        if template.is_known_empty() {
             return Ok(());
         } else {            
             for v in unified.iter() {
@@ -472,8 +466,9 @@ mod testing {
             ]));
     }
 
+    // Test that if any single clause in the `not` fails to resolve the whole clause is considered empty
     #[test]
-    fn test_only_one_clause_succeeds() {
+    fn test_fails_if_any_clause_invalid() {
         let schema = prepopulated_schema();
         let query = r#"
             [:find ?x
@@ -485,7 +480,7 @@ mod testing {
         assert!(!cc.is_known_empty());
         compare_ccs(cc,
                     alg(&schema,
-                        r#"[:find ?x :where [?x :foo/knows "Bill"] (not [?x :foo/parent "Ámbar"])]"#));
+                        r#"[:find ?x :where [?x :foo/knows "Bill"]]"#));
     }
 
     /// Test that if all the attributes in an `not` fail to resolve, the `cc` isn't considered empty.
--- a/query-algebrizer/src/clauses/not.rs
+++ b/query-algebrizer/src/clauses/not.rs
@@ -85,6 +85,7 @@ mod testing {
 
     use mentat_query::{
         NamespacedKeyword, 
+        PlainSymbol,
         Variable
     };
 
@@ -94,6 +95,11 @@ mod testing {
         QueryInputs, 
         add_attribute, 
         associate_ident,
+    };
+
+    use errors::{
+        Error,
+        ErrorKind,
     };
 
     use types::{
@@ -522,13 +528,18 @@ mod testing {
     }
 
     #[test]
-    #[should_panic]
     fn test_unbound_var_fails() {
         let schema = prepopulated_schema();
         let query = r#"
         [:find ?x
          :in ?y
          :where (not [?x :foo/knows ?y])]"#;
-        alg(&schema, query);
+        let parsed = parse_find_string(query).expect("parse failed");
+        let err = algebrize(&schema, parsed).err();
+        assert!(err.is_some());
+         match err.unwrap() {
+            Error(ErrorKind::UnboundVariable(var), _) => { assert_eq!(var, PlainSymbol("?x".to_string())); },
+            x => panic!("expected Unbound Variable error, got {:?}", x),
+        }
     }
 }
--- a/query-algebrizer/src/clauses/not.rs
+++ b/query-algebrizer/src/clauses/not.rs
@@ -12,7 +12,11 @@ use std::collections::btree_map::Entry;
 
 use mentat_core::Schema;
 
-use mentat_query::{NotJoin, UnifyVars, ContainsVariables};
+use mentat_query::{
+    NotJoin, 
+    UnifyVars, 
+    ContainsVariables,
+};
 
 use clauses::ConjoiningClauses;
 
@@ -21,7 +25,10 @@ use errors::{
     ErrorKind,
 };
 
-use types::{ColumnConstraint, ComputedTable, EmptyBecause};
+use types::{
+    ColumnConstraint, 
+    ComputedTable,
+};
 
 impl ConjoiningClauses {
     pub fn apply_not_join(&mut self, schema: &Schema, not_join: NotJoin) -> Result<()> {
@@ -31,6 +38,15 @@ impl ConjoiningClauses {
         };
 
         let mut template = self.use_as_template(&unified);
+        
+        for v in unified.iter() {
+            if self.value_bindings.contains_key(&v) {
+                let val = self.value_bindings.get(&v).unwrap().clone();
+                template.value_bindings.insert(v.clone(), val);
+            } else if !self.column_bindings.contains_key(&v) {
+                bail!(ErrorKind::UnboundVariable(v.name()));
+            }
+        }
 
         for clause in not_join.clauses.into_iter() {
             template.apply_clause(&schema, clause)?;
@@ -38,26 +54,14 @@ impl ConjoiningClauses {
 
         if template.is_known_empty() {
             return Ok(());
-        } else {            
-            for v in unified.iter() {
-                if self.column_bindings.contains_key(&v) {
-                    let col = self.column_bindings.get(&v).unwrap()[0].clone();
-                    match template.column_bindings.entry(v.clone()) {
-                        Entry::Vacant(e) => {
-                            e.insert(vec![col]);
-                        }
-                        Entry::Occupied(mut e) => {
-                            e.get_mut().append(&mut vec![col]);
-                        }
-                    }
-                } else if self.value_bindings.contains_key(&v) {
-                    let val = self.value_bindings.get(&v).unwrap().clone();
-                    template.value_bindings.insert(v.clone(), val);
-                } else {
-                    bail!(ErrorKind::UnboundVariable(v.name()));
-                }
-            }
         }
+
+        for v in unified.iter() {
+           if self.column_bindings.contains_key(&v) {
+                let col = self.column_bindings.get(&v).unwrap()[0].clone();
+                template.column_bindings.entry(v.clone()).or_insert(vec![]).push(col);
+            }
+        } 
 
         // We are only expanding column bindings here and not pruning extracted types as we are not projecting values.
         template.expand_column_bindings();
@@ -79,19 +83,43 @@ mod testing {
 
     use super::*;
 
-    use mentat_core::{Attribute, TypedValue, ValueType};
+    use mentat_core::{
+        Attribute, 
+        TypedValue, 
+        ValueType,
+    };
 
-    use mentat_query::{NamespacedKeyword, Variable};
+    use mentat_query::{
+        NamespacedKeyword, 
+        Variable
+    };
 
     use self::mentat_query_parser::parse_find_string;
 
-    use clauses::{QueryInputs, add_attribute, associate_ident};
+    use clauses::{
+        QueryInputs, 
+        add_attribute, 
+        associate_ident,
+    };
 
-    use types::{ColumnAlternation, ColumnConstraint, ColumnConstraintOrAlternation,
-                ColumnIntersection, DatomsColumn, DatomsTable, NumericComparison, QualifiedAlias,
-                QueryValue, SourceAlias, ValueTypeSet};
+    use types::{
+        ColumnAlternation, 
+        ColumnConstraint, 
+        ColumnConstraintOrAlternation,
+        ColumnIntersection, 
+        DatomsColumn, 
+        DatomsTable, 
+        NumericComparison, 
+        QualifiedAlias,
+        QueryValue, 
+        SourceAlias, 
+        ValueTypeSet,
+    };
 
-    use {algebrize, algebrize_with_inputs};
+    use {
+        algebrize, 
+        algebrize_with_inputs,
+    };
 
     fn alg(schema: &Schema, input: &str) -> ConjoiningClauses {
         let parsed = parse_find_string(input).expect("parse failed");
@@ -496,5 +524,6 @@ mod testing {
         assert!(!cc.is_known_empty());
         compare_ccs(cc,
                     alg(&schema, r#"[:find ?x :where [?x :foo/knows "John"]]"#));
+                    
     }
 }
--- a/query-algebrizer/src/clauses/not.rs
+++ b/query-algebrizer/src/clauses/not.rs
@@ -11,16 +11,16 @@
 use mentat_core::Schema;
 
 use mentat_query::{
+    ContainsVariables,
     NotJoin, 
     UnifyVars, 
-    ContainsVariables,
 };
 
 use clauses::ConjoiningClauses;
 
 use errors::{
-    Result,
     ErrorKind,
+    Result,
 };
 
 use types::{
--- a/query-algebrizer/src/clauses/not.rs
+++ b/query-algebrizer/src/clauses/not.rs
@@ -520,4 +520,15 @@ mod testing {
                     alg(&schema, r#"[:find ?x :where [?x :foo/knows "John"]]"#));
                     
     }
+
+    #[test]
+    #[should_panic]
+    fn test_unbound_var_fails() {
+        let schema = prepopulated_schema();
+        let query = r#"
+        [:find ?x
+         :in ?y
+         :where (not [?x :foo/knows ?y])]"#;
+        alg(&schema, query);
+    }
 }
--- a/query-algebrizer/src/clauses/not.rs
+++ b/query-algebrizer/src/clauses/not.rs
@@ -1,0 +1,450 @@
+// Copyright 2016 Mozilla
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+use std::collections::btree_map::Entry;
+
+use mentat_core::{
+    Schema,
+};
+
+use mentat_query::{
+    NotJoin,
+    UnifyVars,
+    ContainsVariables,
+};
+
+use clauses::ConjoiningClauses;
+
+use errors::{
+    Result,
+};
+
+use types::{
+    ColumnConstraint,
+    ComputedTable,
+    EmptyBecause,
+};
+
+impl ConjoiningClauses {
+    pub fn apply_not_join(&mut self, schema: &Schema, not_join: NotJoin) -> Result<()> {
+        let projected = match not_join.unify_vars {
+            UnifyVars::Implicit => not_join.collect_mentioned_variables(),
+            UnifyVars::Explicit(vs) => vs.into_iter().collect(),
+        };
+        let mut template = self.use_as_template(&projected);
+        let mut empty_because: Option<EmptyBecause> = None;
+        
+        for clause in not_join.clauses.into_iter() {
+            template.apply_clause(&schema, clause)?;
+            // This is disgusting, but I can't think of anything else right now that works.
+            if template.is_known_empty() {
+                empty_because = template.empty_because;
+                template.empty_because = None;
+            }
+        }
+
+        if template.wheres.is_empty() && empty_because.is_some() {
+            return Ok(());
+        } else {
+            for (v, cols) in (&self.column_bindings).iter().filter(|&(key, _)| projected.contains(&key) ) {
+                match template.column_bindings.entry(v.clone()) {
+                    Entry::Vacant(e) => {
+                        e.insert(cols.clone());
+                    },
+                    Entry::Occupied(mut e) => {
+                        e.get_mut().append(&mut cols.clone());
+                    },
+                }
+            }
+        }
+
+        // We are only expanding column bindings here and not pruning extracted types as we are not projecting values.
+        template.expand_column_bindings();
+
+        let subquery = ComputedTable::Subquery(template);
+
+        self.wheres.add_intersection(ColumnConstraint::NotExists(subquery));
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod testing {
+    extern crate mentat_query_parser;
+
+    use std::rc::Rc;
+    use std::collections::BTreeSet;
+
+    use super::*;
+
+    use mentat_core::{
+        Attribute,
+        TypedValue,
+        ValueType,
+    };
+
+    use mentat_query::{
+        NamespacedKeyword,
+        Variable,
+    };
+
+    use self::mentat_query_parser::{
+        parse_find_string,
+    };
+
+    use clauses::{
+        QueryInputs,
+        add_attribute,
+        associate_ident,
+    };
+
+    use types::{
+        ColumnAlternation,
+        ColumnConstraint,
+        ColumnConstraintOrAlternation,
+        ColumnIntersection,
+        DatomsColumn,
+        DatomsTable,
+        NumericComparison,
+        QualifiedAlias,
+        QueryValue,
+        SourceAlias,
+        ValueTypeSet,
+    };
+
+    use {
+        algebrize,
+        algebrize_with_inputs,
+    };
+
+    fn alg(schema: &Schema, input: &str) -> ConjoiningClauses {
+        let parsed = parse_find_string(input).expect("parse failed");
+        algebrize(schema.into(), parsed).expect("algebrize failed").cc
+    }
+
+    fn alg_with_inputs(schema: &Schema, input: &str, inputs: QueryInputs) -> ConjoiningClauses {
+        let parsed = parse_find_string(input).expect("parse failed");
+        algebrize_with_inputs(schema.into(), parsed, 0, inputs).expect("algebrize failed").cc
+    }
+
+    fn prepopulated_schema() -> Schema {
+        let mut schema = Schema::default();
+        associate_ident(&mut schema, NamespacedKeyword::new("foo", "name"), 65);
+        associate_ident(&mut schema, NamespacedKeyword::new("foo", "knows"), 66);
+        associate_ident(&mut schema, NamespacedKeyword::new("foo", "parent"), 67);
+        associate_ident(&mut schema, NamespacedKeyword::new("foo", "age"), 68);
+        associate_ident(&mut schema, NamespacedKeyword::new("foo", "height"), 69);
+        add_attribute(&mut schema, 65, Attribute {
+            value_type: ValueType::String,
+            multival: false,
+            ..Default::default()
+        });
+        add_attribute(&mut schema, 66, Attribute {
+            value_type: ValueType::String,
+            multival: true,
+            ..Default::default()
+        });
+        add_attribute(&mut schema, 67, Attribute {
+            value_type: ValueType::String,
+            multival: true,
+            ..Default::default()
+        });
+        add_attribute(&mut schema, 68, Attribute {
+            value_type: ValueType::Long,
+            multival: false,
+            ..Default::default()
+        });
+        add_attribute(&mut schema, 69, Attribute {
+            value_type: ValueType::Long,
+            multival: false,
+            ..Default::default()
+        });
+        schema
+    }
+
+    fn compare_ccs(left: ConjoiningClauses, right: ConjoiningClauses) {
+        assert_eq!(left.wheres, right.wheres);
+        assert_eq!(left.from, right.from);
+    }
+
+    // not.
+    #[test]
+    fn test_successful_not() {
+        let schema = prepopulated_schema();
+        let query = r#"
+            [:find ?x
+             :where [?x :foo/knows "John"]
+                    (not [?x :foo/parent "Ámbar"]
+                         [?x :foo/knows "Daphne"])]"#;
+        let cc = alg(&schema, query);
+
+        let vx = Variable::from_valid_name("?x");
+        let d0 = "datoms00".to_string();
+        let d1 = "datoms01".to_string();
+        let d2 = "datoms02".to_string();
+        let d0e = QualifiedAlias::new(d0.clone(), DatomsColumn::Entity);
+        let d0a = QualifiedAlias::new(d0.clone(), DatomsColumn::Attribute);
+        let d0v = QualifiedAlias::new(d0.clone(), DatomsColumn::Value);
+        let d1e = QualifiedAlias::new(d1.clone(), DatomsColumn::Entity);
+        let d1a = QualifiedAlias::new(d1.clone(), DatomsColumn::Attribute);
+        let d1v = QualifiedAlias::new(d1.clone(), DatomsColumn::Value);
+        let d2e = QualifiedAlias::new(d2.clone(), DatomsColumn::Entity);
+        let d2a = QualifiedAlias::new(d2.clone(), DatomsColumn::Attribute);
+        let d2v = QualifiedAlias::new(d2.clone(), DatomsColumn::Value);
+
+        let knows = QueryValue::Entid(66);
+        let parent = QueryValue::Entid(67);
+        let john = QueryValue::TypedValue(TypedValue::typed_string("John"));
+        let ambar = QueryValue::TypedValue(TypedValue::typed_string("Ámbar"));
+        let daphne = QueryValue::TypedValue(TypedValue::typed_string("Daphne"));
+
+        let mut subquery = ConjoiningClauses::default();
+        subquery.from.append(&mut vec![SourceAlias(DatomsTable::Datoms, d1),
+                                  SourceAlias(DatomsTable::Datoms, d2)]);
+        subquery.column_bindings.insert(vx.clone(), vec![d1e.clone(), d2e.clone(), d0e.clone()]);
+        subquery.wheres = ColumnIntersection(vec![ColumnConstraintOrAlternation::Constraint(ColumnConstraint::Equals(d1a.clone(), parent)),
+                               ColumnConstraintOrAlternation::Constraint(ColumnConstraint::Equals(d1v.clone(), ambar)),
+                               ColumnConstraintOrAlternation::Constraint(ColumnConstraint::Equals(d2a.clone(), knows.clone())),
+                               ColumnConstraintOrAlternation::Constraint(ColumnConstraint::Equals(d2v.clone(), daphne)),
+                               ColumnConstraintOrAlternation::Constraint(ColumnConstraint::Equals(d1e.clone(), QueryValue::Column(d2e.clone()))),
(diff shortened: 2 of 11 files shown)


query-algebrizer/src/clauses/not.rs
```diff
@@ -0,0 +1,63 @@
+// // Copyright 2016 Mozilla
```
rnewman: Nit: duplicated comment slashes.

query-algebrizer/src/clauses/not.rs
```diff
@@ -0,0 +1,63 @@
...
+    ColumnConstraint,
+    ComputedTable,
+    PushComputed,
+    SourceAlias,
+};
+
+impl ConjoiningClauses {
+    pub fn apply_not_join(&mut self, schema: &Schema, not_join: NotJoin) -> Result<()> {
+        let projected = match not_join.unify_vars {
+            UnifyVars::Implicit => not_join.collect_mentioned_variables ().into_iter().collect(),
```
rnewman: Nit: extra space.

rnewman: `collect_mentioned_variables` returns a `BTreeSet<Variable>`. `use_as_template` takes a `BTreeSet<Variable>`. You can remove the `into_iter().collect()` altogether — it's converting from `T` to `T`!

query-algebrizer/src/clauses/not.rs
```diff
@@ -0,0 +1,63 @@
...
+            UnifyVars::Explicit(vs) => vs.into_iter().collect(),
+        };
+        let template = self.use_as_template(&projected);
+
+        let mut receptacle = template.make_receptacle();
```
rnewman: If you're only doing one, you can use the `template` directly. You'll note that `use_as_template` is essentially an intersected `clone`, and `make_receptacle` is a full `clone` of the same parts.

query-algebrizer/src/clauses/not.rs
```diff
@@ -0,0 +1,63 @@
...
+        for clause in not_join.clauses.into_iter() {
+            receptacle.apply_clause(&schema, clause)?;
+        }
+        if receptacle.is_known_empty() {
+            self.mark_known_empty(receptacle.empty_because.expect("empty for a reason"));
```
rnewman: Actually the opposite! If you have a `(not …)` expression that's known-empty — that is, it can never match — it's equivalent to discarding the `not` altogether.

For example:

```
[:find ?x
 :where
 [?x :foo/int ?y]
 (not [?x :foo/string ?y])]
```

is equivalent to

```
[:find ?x
 :where
 [?x :foo/int ?y]]
```

because `?y` cannot simultaneously be both a number and a string.

If you get here, you can just `return`.

query-algebrizer/src/clauses/not.rs
```diff
@@ -0,0 +1,63 @@
...
+            return Ok(());
+        } else {
+            receptacle.expand_column_bindings();
+            receptacle.prune_extracted_types();
```
rnewman: I don't think you need to bother with pruning: we're not projecting values from the `not`. (Comment this, of course.)

query-sql/src/lib.rs
```diff
@@ -143,7 +148,7 @@ pub struct Join {
 pub enum TableOrSubquery {
     Table(SourceAlias),
     Union(Vec<SelectQuery>, TableAlias),
-    // TODO: Subquery.
+    Subquery(Rc<SelectQuery>),
```
rnewman: You can probably get rid of `Rc` here.

query-sql/src/lib.rs
```diff
@@ -323,6 +328,11 @@ impl QueryFragment for Constraint {
                 out.push_sql(")");
                 Ok(())
             },
+            &NotExists { ref subquery } => {
+                out.push_sql(" NOT EXISTS ");
```
rnewman: You'll need `()` around the subquery in the generated SQL.

query-sql/src/lib.rs
fluffyemily: Without the `Rc` I get a recursion problem as `SelectQuery` contains a `FromClause` contains a `Join` which contains a `TableOrSubquery` which contains a `SelectQuery` which contains a `FromClause` and so on. `Rc` seemed to be the most appropriate way to fix this, but if you have another suggestion I am open to it.

query-sql/src/lib.rs
fluffyemily: Doh! A remnant of a previous iteration that had the subquery outputting itself with it's braces.

query-algebrizer/src/clauses/not.rs
```diff
@@ -56,3 +55,146 @@ impl ConjoiningClauses {
...
+        let d0v = QualifiedAlias::new(d0.clone(), DatomsColumn::Value);
+        let d1e = QualifiedAlias::new(d1.clone(), DatomsColumn::Entity);
+        let d1a = QualifiedAlias::new(d1.clone(), DatomsColumn::Attribute);
+        let d1v = QualifiedAlias::new(d1.clone(), DatomsColumn::Value);
+        let d2e = QualifiedAlias::new(d2.clone(), DatomsColumn::Entity);
+        let d2a = QualifiedAlias::new(d2.clone(), DatomsColumn::Attribute);
+        let d2v = QualifiedAlias::new(d2.clone(), DatomsColumn::Value);
+
+        let knows = QueryValue::Entid(66);
+        let subquery_knows = QueryValue::Entid(66);
```
rnewman: You can just use `knows.clone()`…

query-algebrizer/src/clauses/not.rs
```diff
@@ -56,3 +55,146 @@ impl ConjoiningClauses {
...
+        let parent = QueryValue::Entid(67);
+        let john = QueryValue::TypedValue(TypedValue::typed_string("John"));
+        let ambar = QueryValue::TypedValue(TypedValue::typed_string("Ámbar"));
+        let daphne = QueryValue::TypedValue(TypedValue::typed_string("Daphne"));
+
+        let mut subquery = ConjoiningClauses::default();
+        subquery.from.append(&mut vec![SourceAlias(DatomsTable::Datoms, d1),
```
rnewman: If you're constructing the CC directly, you can just do:

```
subquery.from = vec![…]
```

query-algebrizer/src/clauses/not.rs
```diff
@@ -56,3 +55,146 @@ impl ConjoiningClauses {
...
+                                  SourceAlias(DatomsTable::Datoms, d2)]);
+        subquery.column_bindings.insert(vx.clone(), vec![d1e.clone(), d2e.clone()]);
+        subquery.wheres.add(ColumnConstraintOrAlternation::Constraint(ColumnConstraint::Equals(d1a.clone(), parent)));
+        subquery.wheres.add(ColumnConstraintOrAlternation::Constraint(ColumnConstraint::Equals(d1v.clone(), ambar)));
+        subquery.wheres.add(ColumnConstraintOrAlternation::Constraint(ColumnConstraint::Equals(d2a.clone(), subquery_knows)));
+        subquery.wheres.add(ColumnConstraintOrAlternation::Constraint(ColumnConstraint::Equals(d2v.clone(), daphne)));
+        subquery.wheres.add(ColumnConstraintOrAlternation::Constraint(ColumnConstraint::Equals(d1e.clone(), QueryValue::Column(d2e.clone()))));
+
+        let mut s = HashSet::with_capacity(1);
+        s.insert(ValueType::Ref);
```
rnewman: You might want to expose `unit_type_set`.

query-algebrizer/src/clauses/not.rs
```diff
@@ -56,3 +55,146 @@ impl ConjoiningClauses {
...
+
+        let mut subquery = ConjoiningClauses::default();
+        subquery.from.append(&mut vec![SourceAlias(DatomsTable::Datoms, d1),
+                                  SourceAlias(DatomsTable::Datoms, d2)]);
+        subquery.column_bindings.insert(vx.clone(), vec![d1e.clone(), d2e.clone()]);
+        subquery.wheres.add(ColumnConstraintOrAlternation::Constraint(ColumnConstraint::Equals(d1a.clone(), parent)));
+        subquery.wheres.add(ColumnConstraintOrAlternation::Constraint(ColumnConstraint::Equals(d1v.clone(), ambar)));
+        subquery.wheres.add(ColumnConstraintOrAlternation::Constraint(ColumnConstraint::Equals(d2a.clone(), subquery_knows)));
+        subquery.wheres.add(ColumnConstraintOrAlternation::Constraint(ColumnConstraint::Equals(d2v.clone(), daphne)));
+        subquery.wheres.add(ColumnConstraintOrAlternation::Constraint(ColumnConstraint::Equals(d1e.clone(), QueryValue::Column(d2e.clone()))));
```
rnewman: Missing here is the link between the `NOT EXISTS` and the body of the enclosing query. The CC you've built here actually encodes this query:

```
[:find ?x :where [?x :foo/name "John"] (not [?y :foo/parent "Ámbar"] [?y :foo/knows "Daphne"])]
```

— note that it's `?y` inside, not `?x`! — which is "find people named John unless there's a person who's Ámbar's child and knows Daphne".

I'd be expecting to see `ColumnConstraint::Equals(d0e.clone(), QueryValue::Column(d1e.clone())` here, and for the `column_bindings` to be pre-populated from the parent with mention of `d0`.

query-algebrizer/src/clauses/not.rs
```diff
@@ -56,3 +55,146 @@ impl ConjoiningClauses {
...
+
+        let mut s = HashSet::with_capacity(1);
+        s.insert(ValueType::Ref);
+        subquery.known_types.insert(vx.clone(), s);
+
+        assert!(!cc.is_known_empty());
+        assert_eq!(cc.wheres, ColumnIntersection(vec![
+                ColumnConstraintOrAlternation::Constraint(ColumnConstraint::Equals(d0a.clone(), knows.clone())),
+                ColumnConstraintOrAlternation::Constraint(ColumnConstraint::Equals(d0v.clone(), john)),
+                ColumnConstraintOrAlternation::Constraint(ColumnConstraint::NotExists(ComputedTable::Subquery{ clauses: subquery })),
```
rnewman: I suspect `ComputedTable::Subquery` should change from being a "struct type" to a "tuple type", 'cos the name is confusing ("clauses") and it only has one field. That is, rather than

```
    Subquery: { clauses: ConjoiningClauses },
```

you'd have

```
    Subquery(ConjoiningClauses),
```

and use it that way here:

```
ColumnConstraint::NotExists(ComputedTable::Subquery(subquery))
```

query-algebrizer/src/clauses/not.rs
fluffyemily: Hmm, I am pretty sure I tried that and it didn't work, but now it does so obviously I didn't.

query-algebrizer/src/clauses/not.rs
fluffyemily: OK, so I seem to be missing the injection of that binding into the subquery. I'd missed that. Doh!

rnewman: If you're templating correctly — that is, passing in all the variables you might need, which is all of them — then things should Just Work®.

query-algebrizer/src/clauses/mod.rs
```diff
@@ -329,6 +329,7 @@ impl ConjoiningClauses {
             value_bindings: self.value_bindings.with_intersected_keys(&vars),
             known_types: self.known_types.with_intersected_keys(&vars),
             extracted_types: self.extracted_types.with_intersected_keys(&vars),
+            column_bindings: self.column_bindings.with_intersected_keys(&vars),
```
rnewman: I don't think this will necessarily do what we want in all cases. In particular it'll generate the expanded column comparisons (`datoms00.e = datoms01.e AND datoms00.e = datoms02.v AND…`) within every child CC.

Give this more thought?

query-algebrizer/src/clauses/not.rs
```diff
@@ -177,22 +170,64 @@ mod testing {
...
+
+    // not-join.
+    #[test]
+    fn test_successful_not_join() {
+        let schema = prepopulated_schema();
+        let query = r#"
+            [:find ?x
+             :where [?x :foo/knows ?y]
+                    (not-join [?x] 
+                              [?x :foo/parent ?y])]"#;
```
rnewman: If you extend this query to

```
[:find ?x
 :where
 [?x :foo/knows ?y]
 [?x :foo/age 11]
 [?x :foo/name "Bill"]
 (not-join [?x]
   [?x :foo/parent ?y])]
```

you should see what I mean about expanding column bindings inside the child.

query-sql/src/lib.rs
rnewman: Oh, you have `Sized` problems. `Box` is your friend; `Rc` is specifically for reference counting. `Box` just means "put this on the heap and let the `Box` own it".

You'll see that `Vec` above accidentally serves the same purpose.

query-algebrizer/src/clauses/not.rs
```diff
@@ -0,0 +1,450 @@
...
+    ColumnConstraint,
+    ComputedTable,
+    EmptyBecause,
+};
+
+impl ConjoiningClauses {
+    pub fn apply_not_join(&mut self, schema: &Schema, not_join: NotJoin) -> Result<()> {
+        let projected = match not_join.unify_vars {
+            UnifyVars::Implicit => not_join.collect_mentioned_variables(),
+            UnifyVars::Explicit(vs) => vs.into_iter().collect(),
```
rnewman: I think there's a reasonable case to be made for making `Explicit` hold a `BTreeSet<Variable>`, and checking for uniqueness in the parser just as we do for `:in` and other lists of variables. That allows you to avoid the work here and elsewhere in the code. Please file a follow-up issue.

query-algebrizer/src/clauses/not.rs
```diff
@@ -0,0 +1,450 @@
...
+        };
+        let mut template = self.use_as_template(&projected);
+        let mut empty_because: Option<EmptyBecause> = None;
+        
+        for clause in not_join.clauses.into_iter() {
+            template.apply_clause(&schema, clause)?;
+            // This is disgusting, but I can't think of anything else right now that works.
+            if template.is_known_empty() {
```
rnewman: Just check for `is_known_empty` after applying all of the patterns inside the `not`. So make the template, apply all of the clauses, and if known empty, return OK. I don't think there's any reason you need to extract the empty option at each step in the iterator.

query-algebrizer/src/clauses/not.rs
```diff
@@ -0,0 +1,450 @@
...
+                empty_because = template.empty_because;
+                template.empty_because = None;
+            }
+        }
+
+        if template.wheres.is_empty() && empty_because.is_some() {
+            return Ok(());
```
rnewman: This just becomes

```
if template.is_known_empty() {
    // If the `not` can never match, then we can optimize it away entirely.
    return Ok(())
}
```

Why are you checking the template `wheres`?

query-algebrizer/src/clauses/not.rs
```diff
@@ -0,0 +1,450 @@
...
+        } else {
```
rnewman: No need for `else` after `return`.

query-algebrizer/src/clauses/not.rs
```diff
@@ -0,0 +1,450 @@
...
+
+use types::{
+    ColumnConstraint,
+    ComputedTable,
+    EmptyBecause,
+};
+
+impl ConjoiningClauses {
+    pub fn apply_not_join(&mut self, schema: &Schema, not_join: NotJoin) -> Result<()> {
+        let projected = match not_join.unify_vars {
```
rnewman: Let's clarify this. `not` is a filter, so it doesn't actually project. We're just unifying some variables with the enclosing environment. Rename `projected` to `unified`.

query-algebrizer/src/clauses/not.rs
```diff
@@ -0,0 +1,450 @@
...
+            if template.is_known_empty() {
+                empty_because = template.empty_because;
+                template.empty_because = None;
+            }
+        }
+
+        if template.wheres.is_empty() && empty_because.is_some() {
+            return Ok(());
+        } else {
+            for (v, cols) in (&self.column_bindings).iter().filter(|&(key, _)| projected.contains(&key) ) {
```
rnewman: `projected` is almost guaranteed to be shorter than `column_bindings`, so walk that first.

query-algebrizer/src/clauses/not.rs
```diff
@@ -0,0 +1,450 @@
...
+                match template.column_bindings.entry(v.clone()) {
+                    Entry::Vacant(e) => {
+                        e.insert(cols.clone());
```
rnewman: Here you're collecting every column binding into the `not`. A query like this:

```
[:find ?x ?age ?height
 :where
 [?x :foo/name "John"]
 [?x :foo/age ?age]
 [?x :foo/height ?height]]
```

has `column_bindings` like this:

```
{?x [datoms00.e, datoms01.e, datoms02.e]}
```

and produces a query like

```
SELECT datoms00.e AS `?x`,
       datoms01.v AS `?age`,
       datoms02.v AS `?height`
FROM datoms datoms00, datoms datoms01, datoms datoms02
WHERE datoms00.e = datoms01.e
  AND datoms00.e = datoms02.e
  AND …         -- Attribute checks go here.
```

If you add a `not` in there, and do what you've done here:

```
[:find ?x ?age ?height
 :where
 [?x :foo/name "John"]
 [?x :foo/age ?age]
 [?x :foo/height ?height]
 (not-join [?x] [?x :foo/degree :bsc])]
```

you'll get SQL like:

```
SELECT datoms00.e AS `?x`,
       datoms01.v AS `?age`,
       datoms02.v AS `?height`
FROM datoms datoms00, datoms datoms01, datoms datoms02
WHERE datoms00.e = datoms01.e
  AND datoms00.e = datoms02.e
  AND …         -- Attribute checks go here.
  AND NOT EXISTS (
    SELECT 1 FROM datoms datoms03
    WHERE datoms03.a = 99         -- :foo/degree
      AND datoms03.v = $bsc
      AND datoms03.e = datoms00.e
      AND datoms03.e = datoms01.e
      AND datoms03.e = datoms02.e
  )
```

Notice that because you've copied the full set of column bindings into the inner query, you'll get the full keyed expansion of external columns. That's redundant.

I believe a sufficient fix is to only use here the first column binding, yielding:

```
SELECT datoms00.e AS `?x`,
       datoms01.v AS `?age`,
       datoms02.v AS `?height`
FROM datoms datoms00, datoms datoms01, datoms datoms02
WHERE datoms00.e = datoms01.e
  AND datoms00.e = datoms02.e
  AND …         -- Attribute checks go here.
  AND NOT EXISTS (
    SELECT 1 FROM datoms datoms03
    WHERE datoms03.a = 99         -- :foo/degree
      AND datoms03.v = $bsc
      AND datoms03.e = datoms00.e
  )
```

What do you think?

query-algebrizer/src/clauses/not.rs
```diff
@@ -0,0 +1,450 @@
...
+                    },
+                    Entry::Occupied(mut e) => {
+                        e.get_mut().append(&mut cols.clone());
+                    },
+                }
+            }
+        }
+
+        // We are only expanding column bindings here and not pruning extracted types as we are not projecting values.
```
rnewman: However, we do need to do some type work. Consider this query:

```
[:find ?x
 :where
 [?x :foo/name]
 [?x _ ?y]
 (not-join [?y] [?a _ ?y])]
```

We require this SQL:

```
SELECT datoms00.e AS `?x`
  FROM datoms datoms00, datoms datoms01
 WHERE datoms00.a = 99
   AND datoms00.e = datoms01.e
  AND NOT EXISTS (
    SELECT 1 FROM datoms datoms02
     WHERE datoms02.v = datoms01.v
       AND datoms02.value_type_tag = datoms01.value_type_tag
  )
```

Could you write a test for that, and then fix it?

query-algebrizer/src/clauses/not.rs
rnewman: Comment still applies.

query-algebrizer/src/clauses/not.rs
rnewman: Comment still applies.

query-algebrizer/src/clauses/not.rs
```diff
@@ -0,0 +1,450 @@
...
+                ColumnConstraintOrAlternation::Constraint(ColumnConstraint::NotExists(ComputedTable::Subquery(subquery))),
+            ]));
+    }
+
+    // not-join.
+    #[test]
+    fn test_not_with_in() {
+        let schema = prepopulated_schema();
+        let query = r#"
+            [:find ?x
```
rnewman: This query should fail! The projection list is not bound inside the query. `not` doesn't bind, so `?x` is unbound.

It should work if you change this to `:find ?y`. Please duplicate this test to do that, and make sure it works, and then make this one fail!

query-algebrizer/src/types.rs
```diff
@@ -407,6 +409,9 @@ impl Debug for ColumnConstraint {
             &HasType(ref qa, value_type) => {
                 write!(f, "{:?}.value_type_tag = {:?}", qa, value_type)
             },
+            &NotExists(ref ct) => {
+                write!(f, "{:?}", ct)
```
rnewman: Make this `"NOT EXISTS {:?}"`.

query-algebrizer/src/clauses/not.rs
```diff
@@ -0,0 +1,450 @@
...
+        let mut subquery = ConjoiningClauses::default();
+        subquery.from.append(&mut vec![SourceAlias(DatomsTable::Datoms, d0)]);
+        subquery.column_bindings.insert(vx.clone(), vec![d0e.clone()]);
+        subquery.wheres = ColumnIntersection(vec![ColumnConstraintOrAlternation::Constraint(ColumnConstraint::Equals(d0a.clone(), knows.clone())),
+                                                  ColumnConstraintOrAlternation::Constraint(ColumnConstraint::Equals(d0v.clone(), john))]);
+
+        subquery.known_types.insert(vx.clone(), ValueTypeSet::of_one(ValueType::Ref));
+        subquery.known_types.insert(vy.clone(), ValueTypeSet::of_one(ValueType::String));
+
+        let mut input_vars:BTreeSet<Variable> = BTreeSet::default();
```
rnewman: Nit: space after `:` in type declaration, throughout.

query-algebrizer/src/clauses/not.rs
```diff
@@ -0,0 +1,450 @@
...
+    EmptyBecause,
+};
+
+impl ConjoiningClauses {
+    pub fn apply_not_join(&mut self, schema: &Schema, not_join: NotJoin) -> Result<()> {
+        let projected = match not_join.unify_vars {
+            UnifyVars::Implicit => not_join.collect_mentioned_variables(),
+            UnifyVars::Explicit(vs) => vs.into_iter().collect(),
+        };
+        let mut template = self.use_as_template(&projected);
```
rnewman: Just before this point you should validate:

```
for var in projected.iter() {
    if !self.column_bindings.contains_key(var) &&
       !self.input_variables.contains(var) {
        bail!(ErrorKind::UnboundVariable(var.name()));
    }
}
```

You should probably "weaken" the error message for `UnboundVariable` to accommodate its use here.

rnewman: This will cause the test I discuss below to fail, as it should.

query-algebrizer/src/clauses/not.rs
fluffyemily: In which case, this might be a bug somewhere.

I found that if a clause fills the `empty_because` value when applied to the template, no subsequent clause, even if valid, will populate the `wheres` for the `cc`. Only by removing that `empty_because` value were subsequent valid clauses successfully applied.

I found this when trying to get the `test_only_one_clause_succeeds` test to work.

query-algebrizer/src/clauses/not.rs
fluffyemily: so above. It should be possible, if at least one out of the clauses in the not are valid, for the template to execute successfully. If even one of those clauses creates an `empty_because`, the whole template is marked `is_known_empty` and the not returns, despite there being other valid clauses present in the `not`.

query-algebrizer/src/clauses/not.rs
fluffyemily: So. this is because of these lines in `apply_pattern_clause_for_alias`
```
        if self.is_known_empty() {
            return;
        }
```

I am looking at a different way to apply the clauses so that we don't encounter this.

rnewman: That behavior is correct!

If I have a pattern like:

```
[:find …
 :where [?x :foo/bar ?y]
 (not [?x :never/succeeds 10]
      [?y :noo/noz 11])]
```

then the entire `not` 'fails', and is optimized away. Continue applying `[?y :noo/noz 11]`, and let the inner CC drop it on the floor. The end result will be a CC that is known-empty, with no `wheres`, with minimal work done.

query-algebrizer/src/clauses/not.rs
rnewman: > It should be possible, if at least one out of the clauses in the not are valid, for the template to execute successfully

OK, this is the misunderstanding.

That's not true: the contents of the `not` are joined with an implicit `and`, just like the contents of the top-level query. If any of the patterns in the `not` are impossible, then the conjunction of the patterns is impossible, and the `not` will never ‘fire’.

What you're describing is true for `or`: we can prune options that can never succeed.

The equivalent for `not` — and for top-level patterns! — is to prune arms that _always_ succeed and never contribute bindings:

```
[:find …
 :where …
 (not
   [:db/valueType :db/ident :db/valueType]  ; Always true, doesn’t bind: can prune.
   [:db/valueType :db/ident :not/that]      ; Always false; fails the `not`.
   …
   )]
```

query-algebrizer/src/clauses/not.rs
```diff
@@ -85,45 +85,19 @@ mod testing {
...
-        QueryValue,
-        SourceAlias,
-        ValueTypeSet,
-    };
-
-    use {
-        algebrize,
-        algebrize_with_inputs,
-    };
+    use mentat_core::{Attribute, TypedValue, ValueType};
```
rnewman: Oh no! You accidentally reformatted the whole file! Best to `git reset` and re-add, skipping these changes.

query-algebrizer/src/clauses/not.rs
```diff
@@ -52,15 +44,23 @@ impl ConjoiningClauses {
...
-                    },
-                    Entry::Occupied(mut e) => {
-                        e.get_mut().append(&mut cols.clone());
-                    },
+        } else {            
+            for v in unified.iter() {
+                if self.column_bindings.contains_key(&v) {
+                    let col = self.column_bindings.get(&v).unwrap()[0].clone();
+                    match template.column_bindings.entry(v.clone()) {
+                        Entry::Vacant(e) => {
```
rnewman: You should look up `Entry`'s `or_insert` method.

query-algebrizer/src/clauses/not.rs
fluffyemily: So, this is what I am getting out for this query from the code that I currently have:
```
SELECT 
    DISTINCT `datoms00`.e AS `?x` 
FROM 
    `datoms` AS `datoms00`, 
    `all_datoms` AS `all_datoms01` 
WHERE 
    `datoms00`.a = 99 
    AND NOT EXISTS (
        SELECT 1 
        FROM `all_datoms` AS `all_datoms02` 
        WHERE `all_datoms02`.v = `all_datoms01`.v LIMIT 1) 
    AND `datoms00`.e = `all_datoms01`.e
```

It seems that the differences here are referring to `alldatoms` instead of `datoms` for `datoms01` & `datoms02`, and linking in the subquery through `datomsX.v` instead of `datomsX.value_type_tag`. Can I confirm that they are the right places to be concentrating?

rnewman: `all_datoms` is right — that just means we don't know if the value could be a string or not.

The missing part there is the type tag. `all_datoms02.v` might be `db.type/ref` = `25`, the entid, and `all_datoms01.v` might be a `db.type/long` = `25`, the number. Those are the same in SQLite, so we need to use the `value_type_tag` column to compare them — a ref will be `1` and a long will be `5`, so they won't unify.

It's possible that this will be entirely solved by #293; see the reference to that issue in `mod.rs`.

rnewman: To be clear: we need to compare _both_ `v` and `value_type_tag` — they're a pair. All the type information we propagate around means that we're often able to ignore the tag!

query-translator/src/translate.rs
```diff
@@ -150,6 +151,13 @@ impl ToConstraint for ColumnConstraint {
                 let column = QualifiedAlias::new(table, DatomsColumn::ValueTypeTag).to_column();
                 Constraint::equal(column, ColumnOrExpression::Integer(value_type.value_type_tag()))
             },
+
+            NotExists(computed_table) => {
+                let subquery = table_for_computed(computed_table, TableAlias::new());
+                Constraint::NotExists {
+                    subquery: subquery
```
rnewman: Nit: trailing commas are encouraged in Rust.

query-translator/tests/translate.rs
```diff
@@ -428,6 +427,61 @@ fn test_complex_or_join_type_projection() {
...
+        value_type: ValueType::Boolean,
+        ..Default::default()
+    });
+
+    let query = r#"[:find ?title
+                    :where [?page :page/title ?title]
+                           (not [?page :page/url "http://foo.com/"]
+                                [?page :page/bookmarked true])]"#;
+    let SQLQuery { sql, args } = translate(&schema, query);
+    assert_eq!(sql, "SELECT DISTINCT `datoms00`.v AS `?title` FROM `datoms` AS `datoms00` WHERE `datoms00`.a = 98 AND NOT EXISTS (SELECT 1 FROM `datoms` AS `datoms01`, `datoms` AS `datoms02` WHERE `datoms01`.a = 97 AND `datoms01`.v = $v0 AND `datoms02`.a = 99 AND `datoms02`.v = 1 AND `datoms01`.e = `datoms02`.e AND `datoms01`.e = `datoms00`.e LIMIT 1)");
```
rnewman: Nit: if you use `\` to separate lines, you can indent the start of the next line as much as you like, and this'll still work.

rnewman: Substantive: you don't need `LIMIT 1` in a subquery used inside `EXISTS` — SQLite will only look at the first result anyway.

query-translator/tests/translate.rs
```diff
@@ -428,6 +427,61 @@ fn test_complex_or_join_type_projection() {
...
+        ..Default::default()
+    });
+
+    let query = r#"[:find ?url
+                        :where [?url :page/url]
+                               (not-join [?url]
+                                   [?page :bookmarks/page ?url]
+                                   [?page :bookmarks/date_created "4/4/2017"])]"#;
+    let SQLQuery { sql, args } = translate(&schema, query);
+    assert_eq!(sql, "SELECT DISTINCT `datoms00`.e AS `?url` FROM `datoms` AS `datoms00` WHERE `datoms00`.a = 97 AND NOT EXISTS (SELECT 1 FROM `datoms` AS `datoms01`, `datoms` AS `datoms02` WHERE `datoms01`.a = 98 AND `datoms02`.a = 99 AND `datoms02`.v = $v0 AND `datoms01`.e = `datoms02`.e AND `datoms01`.v = `datoms00`.e LIMIT 1)");
```
rnewman: Same.

query-translator/tests/translate.rs
```diff
@@ -428,6 +427,61 @@ fn test_complex_or_join_type_projection() {
...
+    add_attribute(&mut schema, 99, Attribute {
+        value_type: ValueType::String,
+        ..Default::default()
+    });
+
+    let query = r#"[:find ?url
+                        :where [?url :page/url]
+                               (not-join [?url]
+                                   [?page :bookmarks/page ?url]
+                                   [?page :bookmarks/date_created "4/4/2017"])]"#;
```
rnewman: Nice US/international conflict avoidance!

query-algebrizer/src/validate.rs
```diff
@@ -229,4 +250,125 @@ mod tests {
...
+                        entity: PatternNonValuePlace::Variable(Variable::from_valid_name("?release")),
+                        attribute: ident("release", "artists"),
+                        value: PatternValuePlace::Variable(Variable::from_valid_name("?artist")),
+                        tx: PatternNonValuePlace::Placeholder,
+                    }));
+                assert_eq!(
+                    clause2,
+                    WhereClause::Pattern(Pattern {
+                        source: None,
+                        entity: PatternNonValuePlace::Variable(Variable::from_valid_name("?release")),
```
rnewman: Pull out the `?release` bit for reuse (and `?artist`, too) — you use each at least twice.

query-algebrizer/src/validate.rs
```diff
@@ -74,6 +75,23 @@ pub fn validate_or_join(or_join: &OrJoin) -> Result<()> {
...
 }
 
+pub fn validate_not_join(not_join: &NotJoin) -> Result<()> {
+    // Grab our mentioned variables and ensure that the rules are followed.
+    match not_join.unify_vars {
+        UnifyVars::Implicit => {
+            Ok(())
+        },
+        UnifyVars::Explicit(ref vars) => {
+            // The joined vars must each appear somewhere in the clauses mentioned variables.
```
rnewman: s/clauses/clause's

query-algebrizer/src/clauses/not.rs
```diff
@@ -0,0 +1,500 @@
...
+        let mut template = self.use_as_template(&unified);
+
+        for clause in not_join.clauses.into_iter() {
+            template.apply_clause(&schema, clause)?;
+        }
+
+        if template.is_known_empty() {
+            return Ok(());
+        } else {            
+            for v in unified.iter() {
```
rnewman: I just realized: you should do this _before_ you apply clauses.

There are two reasons for that.

Firstly, here you're providing static values! We want to use them when applying clauses. If you copy in `value_bindings` after applying patterns, it's too late! CC won't be able to use them to determine failure or inline values.

Secondly, you'll pull the value that we know — the external column association — towards the front of the mess of conditionals that we produce. That should produce a slightly more efficient query.

Doing that also turns the `is_known_empty` check into a simple gate, rather than having this `else`.

query-algebrizer/src/clauses/not.rs
fluffyemily: nice!

query-algebrizer/src/clauses/not.rs
fluffyemily: I am going to work on this under #293 as I don't want to hold this up.

query-algebrizer/src/clauses/not.rs
```diff
@@ -0,0 +1,529 @@
...
+        for clause in not_join.clauses.into_iter() {
+            template.apply_clause(&schema, clause)?;
+        }
+
+        if template.is_known_empty() {
+            return Ok(());
+        }
+
+        for v in unified.iter() {
+           if self.column_bindings.contains_key(&v) {
```
rnewman: Put this in line 46.

rnewman: And note that if you do this _before_ you apply clauses — which you should — then you can simply unilaterally insert into `column_bindings`, because you know it's a blank template.

query-algebrizer/src/clauses/not.rs
rnewman: Remember to add the test that

```
[:find ?x
 :in ?y
 :where (not [?x :foo/knows ?y])]
```

is rejected.

query-algebrizer/src/types.rs
```diff
@@ -294,6 +295,7 @@ pub enum ColumnConstraint {
         right: QueryValue,
     },
     HasType(TableAlias, ValueType),
+    NotExists(ComputedTable),
```
rnewman: I don't see where this is used. Is this some forgotten cleanup?

query-algebrizer/src/clauses/not.rs
fluffyemily: When I put that in 46, things started going a bit wrong.

For
```
            [:find ?x ?age
             :where
             [?x :foo/age ?age]
             [[< ?age 30]]
             (not [?x :foo/knows "John"]
                  [?x :foo/knows "Daphne"])]
```
The output stopped joining `datoms01.e` with `datoms02.e` inside the subquery, but joined `datoms02.e` with `datoms00.e` instead, meaning both `datoms01.e` and `datoms02.e` are joined with `datoms00.e`, which strikes me as wrong. 

New Output:
```
Constraint(datoms00.a = entity(68)), Constraint(datoms00.v < value(Long(30))), Constraint(NOT EXISTS Subquery(ConjoiningClauses { empty_because: None, from: [SourceAlias(Datoms, datoms01), SourceAlias(Datoms, datoms02)], wheres: [Constraint(datoms01.a = entity(66)), Constraint(datoms01.v = value(String("John"))), Constraint(datoms02.a = entity(66)), Constraint(datoms02.v = value(String("Daphne"))), Constraint(datoms00.e = datoms01.e), Constraint(datoms00.e = datoms02.e)], column_bindings: {var(?x): [datoms00.e, datoms01.e, datoms02.e]}, input_variables: {}, value_bindings: {}, known_types: {var(?x): ValueTypeSet({Ref})}, extracted_types: {} }))
```

Expected Output:
```
Constraint(datoms00.a = entity(68)), Constraint(datoms00.v < value(Long(30))), Constraint(NOT EXISTS Subquery(ConjoiningClauses { empty_because: None, from: [SourceAlias(Datoms, datoms01), SourceAlias(Datoms, datoms02)], wheres: [Constraint(datoms01.a = entity(66)), Constraint(datoms01.v = value(String("John"))), Constraint(datoms02.a = entity(66)), Constraint(datoms02.v = value(String("Daphne"))), Constraint(datoms00.e = datoms01.e)], Constraint(datoms01.e = datoms02.e), column_bindings: {var(?x): [datoms00.e, datoms01.e, datoms02.e]}, input_variables: {}, value_bindings: {}, known_types: {var(?x): ValueTypeSet({Ref})}, extracted_types: {} }))
```

query-algebrizer/src/types.rs
fluffyemily: see line 70 in `not.rs`

rnewman:
> The output stopped joining datoms01.e with datoms02.e inside the subquery, but joined datoms02.e with datoms00.e instead, meaning both datoms01.e and datoms02.e are joined with datoms00.e, which strikes me as wrong.

That depends how you look at it 🐗 

Because of the unification list, `?x` unifies across the query boundary. That means `datoms01.e` and `datoms00.e` and `datoms02.e` all have to be the same. In the formulation you wrote SQLite will generate rows for which `datoms01.e` = `datoms02.e`, and then check that they match `datoms00.e`.

Given that `datoms00.e` is already known by the time SQLite executes the subquery, using that saves us potentially generating results from `datoms01` that will later be excluded for not matching `datoms00`.

There is another way to execute `not`, which is to join two queries on the unification list:

```
[:find ?x
 :where
 [?x :foo/name "Bill"]
 (not [?x :foo/age ?age]
      [(> ?age 10)])]
```

can be translated as

```
SELECT datoms00.e FROM datoms datoms00
WHERE datoms00.a = 99
  AND datoms00.v = "Bill"
  AND NOT EXISTS (
    SELECT 1 FROM datoms datoms01
    WHERE datoms01.e = datoms00.e
      AND datoms01.a = 98
      AND datoms01.v > 10
  )
```

— that's what we do now. Sometimes this is the best query plan, and sometimes not; if the subquery turns into a simple index lookup, this'll be pretty good, but if there are lots of candidate results that need to be merged then it might not be.

Another way to do this is via a `RIGHT EXCLUDING JOIN`:

```
SELECT left.x
FROM
(SELECT datoms00.e AS x
 FROM datoms datoms00
 WHERE datoms00.a = 99 AND datoms00.v = "Bill")
AS left
LEFT JOIN
(SELECT datoms01.e AS x
 FROM datoms datoms01
 WHERE datoms01.a = 98 AND datoms01.v > 10)
AS right
ON left.x = right.x
WHERE right.x IS NULL
```

Essentially this finds all people named Bill, and all people older than 10, and subtracts the latter set from the former. This is better if there aren't many old people in the database.

You can see in this formulation that it would be unnatural to refer to `datoms00` within `right`: the two sets are expressed individually and joined.

In the `NOT EXISTS` formulation it's natural to refer to columns bound in the rest of the query.

query-algebrizer/src/types.rs
rnewman: Nah, that's `ColumnConstraint::NotExists`, right?

fluffyemily: fuck yeah. How did that creep in there?

fluffyemily: especially as it doesn't seem to be in my local copy....

fluffyemily: yes, this line is `ColumnConstraint::NotExists`

rnewman: This also confused me — I checked out your branch and this part of the diff isn't there!

rnewman: Crazy GitHub bug: this line was showing as a third arm in `ComputedTable`, so there was `ComputedTable::Subquery` and `ComputedTable::NotExists`.

query-algebrizer/src/clauses/not.rs
```diff
@@ -0,0 +1,534 @@
...
+                         [?x :foo/nope "Daphne"])]"#;
+        let cc = alg(&schema, query);
+        assert!(!cc.is_known_empty());
+        compare_ccs(cc,
+                    alg(&schema, r#"[:find ?x :where [?x :foo/knows "John"]]"#));
+                    
+    }
+
+    #[test]
+    #[should_panic]
```
rnewman: Let's not use `alg`; algebrize it directly so you can use `assert!(….is_err())`. We want to make sure that a bad query like this doesn't panic, but fails, and this test hides that distinction.

query-translator/src/translate.rs
```diff
@@ -7,6 +7,7 @@
 // under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
+use std::rc::Rc;
```
rnewman: Nit: newline before.

query-algebrizer/src/clauses/not.rs
```diff
@@ -0,0 +1,534 @@
...
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+use mentat_core::Schema;
+
+use mentat_query::{
+    NotJoin, 
+    UnifyVars, 
+    ContainsVariables,
```
rnewman: Nit: alphabetize.

query-algebrizer/src/clauses/not.rs
ncalexan: Huh, I would have expected that to bind `x`.  Live and learn.

PR #423, opened 2017-04-20
Ground

ncalexan:


query-algebrizer/src/clauses/resolve.rs
```diff
@@ -54,7 +54,8 @@ impl ConjoiningClauses {
             SrcVar(_) |
```
rnewman: Commit message comment: "are all value" -> "are all valid".

query-algebrizer/src/clauses/predicate.rs
```diff
@@ -81,13 +94,181 @@ impl ConjoiningClauses {
         self.wheres.add_intersection(constraint);
         Ok(())
     }
+
+
+    /// There are several kinds of functions binding variables in our Datalog:
```
rnewman: functions that bind variables

query-algebrizer/src/clauses/predicate.rs
```diff
@@ -81,13 +94,181 @@ impl ConjoiningClauses {
...
+    /// - A set of functions like `fulltext` and `get-else` that are translated into
+    ///   SQL `MATCH`es or joins, yielding bindings.
+    /// - In the future, some functions that are implemented via function calls in SQLite.
+    ///
+    /// At present we have implemented only the `fulltext` operator.
```
rnewman: s/fulltext/ground

query-algebrizer/src/clauses/predicate.rs
```diff
@@ -81,13 +94,181 @@ impl ConjoiningClauses {
...
+        // ultimately allowing user-specified functions, we match on the function name first.
+        match where_fn.operator.0.as_str() {
+            "ground" => self.apply_ground(schema, where_fn),
+            _ => bail!(ErrorKind::UnknownFunction(where_fn.operator.clone())),
+        }
+    }
+
+    /// This function:
+    /// - Resolves variables and converts types to those more amenable to SQL.
+    /// - Ensures that the predicate functions name a known operator.
```
rnewman: It's not really a predicate at this point.

query-algebrizer/src/clauses/predicate.rs
```diff
@@ -81,13 +94,181 @@ impl ConjoiningClauses {
...
+                    FnArg::Variable(_) | FnArg::SrcVar(_) => panic!("bad things happened"),
+                    FnArg::Collection(_) => panic!("more bad things happened"),
+                };
+
+                projected.push(var.clone());
+                self.constrain_var_to_type(var.clone(), constant.value_type());
+
+                values.push(vec![constant]); // 1x1.
+
+                // // TODO: this optimization.
```
rnewman: There're two kinds of optimization to do here, no?

1. This one: if we haven't seen this variable yet, and we are binding it to a scalar, then just stuff it into `value_bindings` and pretend we never saw the `ground`. In that respect `ground` becomes no different to `:in`.
2. If the variable has already been bound earlier in the query, `ground` _becomes a constraint_:

```
[:find ?x 
 :where [?x :foo/name ?y]
 [(ground "foo") [?y .]]]
```

is

```
SELECT datoms00.e FROM datoms datoms00
WHERE datoms00.a = 99 AND datoms00.v = 'foo'
```

I'm not sure whether further substitution is valuable after that — SQLite might be able to do index optimizations with a column that we can't do by pushing in a value.


While I'm typing, we also should handle this:

```
[:find … :in ?x :where [(ground "foo") ?x]
```

— it doesn't really make sense to ground a constant to an input variable. (Not sure if you've already addressed this.)

query-algebrizer/src/clauses/predicate.rs
```diff
@@ -81,13 +94,181 @@ impl ConjoiningClauses {
...
+                // self.value_bindings.insert(var.clone(), constant.clone());
+
+                // let score_alias = QualifiedAlias(datoms_table_alias.clone(), Column::Fixed(DatomsColumn::Tx));
+            },
+            _ => bail!(ErrorKind::InvalidArgument(where_fn.operator.clone(), "bindings".into(), 999)),
+        };
+
+        let projection: BTreeSet<Variable> = projected.into_iter().collect();
```
rnewman: See comment I left on Emily's PR; we should probably make the various var lists in the parser all spit out `BTreeSet<Variable>` and err on dupes.

query-sql/src/lib.rs
```diff
@@ -143,7 +143,7 @@ pub struct Join {
 pub enum TableOrSubquery {
     Table(SourceAlias),
     Union(Vec<SelectQuery>, TableAlias),
-    // TODO: Subquery.
+    Subquery(Box<SelectQuery>, TableAlias), // Box to break cycle.
```
rnewman: @fluffyemily has this without an alias. Fight it out.

query-sql/src/lib.rs
```diff
@@ -378,6 +378,21 @@ impl QueryFragment for TableOrSubquery {
                 out.push_sql(") AS ");
                 out.push_identifier(table_alias.as_str())
             },
+            &Subquery(ref subquery, ref table_alias) => {
```
rnewman: @fluffyemily's PR has this done, I think.

ncalexan:
@rnewman OK, this is ready for your review now.

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -0,0 +1,384 @@
...
+            }
+        };
+
+        // Typecheck and drop columns that are not actually bound.  By restricting to homogeneous
+        // columns, we greatly simplify projection.  In the future, we could loosen this
+        // restriction, at the cost of projecting (some) value type tags.  If and when we want to
+        // algebrize in two phases and allow for late-binding input variables, we'll probably be
+        // able to loosen this restriction with little penalty.
+        let variables: Vec<Option<Variable>> = where_fn.binding.variables();
+        let (values, column_types) = reduced_rows_and_column_types(rows, &variables.iter().map(|x| x.is_some()).collect())?;
```
rnewman: If you switch 214 and 215, you can use `names` instead of calling `is_some` and walking twice.

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -0,0 +1,384 @@
...
+        let names: Vec<Variable> = variables.into_iter().filter_map(|x| x).collect();
+
+        let named_values = ComputedTable::NamedValues {
```
rnewman: I wonder if this should be called `NamedRel` or `NamedStaticRel` instead — what you're defining here is a relation to join into the query. That just happens to be implemented on the SQL side as `VALUES`, but it could also be implemented as `IN` (for coll), as simple substitution (for scalar), or even as a materialized table.

The `Static` bit might indicate that there's a different kind of rel, which is computed from a function…

query-translator/src/translate.rs
```diff
@@ -241,7 +242,13 @@ fn table_for_computed(computed: ComputedTable, alias: TableAlias) -> TableOrSubq
         },
         ComputedTable::Subquery(subquery) => {
             TableOrSubquery::Subquery(Box::new(cc_to_exists(subquery)))
-        }
+        },
+        ComputedTable::NamedValues {
+            names, values,
+        } => {
+            // We assume column homogeneity, so we won't have any type tag columns.
```
rnewman: As noted elsewhere, this isn't true for functions in general.

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -0,0 +1,384 @@
...
+                    bail!(ErrorKind::UnboundVariable((*in_var.0).clone()));
+                }
+                match self.bound_value(&in_var) {
+                    Some(ref in_value) => Ok(in_value.clone()),
+                    None => bail!(ErrorKind::UnboundVariable((*in_var.0).clone())),
+                }
+            };
+
+            match where_fn.binding {
+                Binding::BindScalar(_) => constant.into_unit_matrix(&lookup)?,
```
rnewman: I expect that a single value will be the most common use of `ground`. E.g.,

https://gist.github.com/vvvvalvalval/5547d8b46414b955c88f#file-datomic_datalog_count-related-entities-with-zero-sums-clj-L27

It seems that you've effectively translated the tuple form

`(ground [1 2] [?x ?y])`

to be equivalent to

`(ground 1 ?x) (ground 2 ?y)`

which is correct.

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -0,0 +1,384 @@
...
+        let names: Vec<Variable> = variables.into_iter().filter_map(|x| x).collect();
+
+        let named_values = ComputedTable::NamedValues {
+            names: names.clone(),
+            values: values,
+        };
+        let table = self.computed_tables.push_computed(named_values);
+        let alias = self.next_alias_for_table(table);
+
+        // Constrain the types of the bound variables we saw.
```
rnewman: For scalar and tuple binding forms, I think we should statically bind the value in the CC right here.

Furthermore, I think you should check to see whether any of the variables have already been used in the query. If not, there's no point in producing SQL at all; any use of the grounded variable (either in projection or in a pattern) will pick up the constant from the CC. Treat `ground` very much like `:in` at that point.

That makes a query like

```
[:find ?x ?v :where
 [(ground 10) ?v]
 [?x :foo/v ?v]]
```

very efficient… and you might think that's silly, but it's an approach used for branching, even with `:in`:

```
[:find ?x ?v
 :in ?v1 ?v2
 :where
 (or (and [(ground ?v1) ?v]
          [?x :foo/v ?v])
    …)]
```

I think it's also worth thinking about different representations for each of the binding lists.

A scalar can be expressed as a very efficient subquery, even if you don't manage to substitute it, and you can also project the type code at this point.

```
sqlite> select x from (SELECT 5 AS x);
x
----------
5
```

query-algebrizer/src/clauses/where_fn.rs
rnewman: Also: if `values` is empty, what happens? I would expect the pattern to return no results.

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -0,0 +1,384 @@
...
+
+    for row in rows {
+        if column_is_bound.len() != row.len() {
+            bail!(ErrorKind::InvalidGroundConstant)
+        }
+
+        let mut retained_row = Vec::with_capacity(column_is_bound.len());
+        for (is_bound, value) in column_is_bound.into_iter().zip(row.into_iter()) {
+            if *is_bound {
+                retained_row.push(value);
```
rnewman: Given that `rows` is passed owned, I would rather see this implemented as in-place modification:

```
let expected = column_is_bound.len();
let to_remove = column_is_bound.iter()
                               .enumerate()
                               .rev()         // So we work from the end for remove.
                               .filter_map(|i, is_bound| if is_bound { Some(i) } else { None })
                               .collect();
for row in rows {
    if expected != row.len() {
        bail!(…);
    }
    for i in to_remove {
        row.remove(*i);
    }
    matrix.push(row);
}
```

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -0,0 +1,384 @@
...
+            bail!(ErrorKind::InvalidGroundConstant)
+        },
+        Some(first_row) => {
+            first_row.iter().map(|x| x.value_type()).collect()
+        },
+    };
+
+    for row in matrix.iter() {
+        for (column_type, value) in column_types.iter().zip(row.into_iter()) {
+            if *column_type != value.value_type() {
```
rnewman: If you maintain a vector of type sets, you can broaden as you go during the initial walk on line 110.

When you get to here you can throw if you wish if any of them are non-unit, or you can implement tag projection.

Preserving type sets means you can do something useful for:

```
[:find …
 :where
 [(ground [[10 :foo] ["string" :bar]] [?x ?y]]     ;; ?x could be ref, long, or string.
 [?x :foo/keyword ?y]]]                            ;; ?x can only be ref!
```

There is a little bit of trickiness here, because `ground` can introduce multiple sets of bindings, and not all of them can be congruent with the rest of the query! Those can be pruned at this point. But the type codes must be projected for ambiguous columns.

Note that you can't quite get away from handling this ambiguity. If the input is EDN, `10` could be `TypedValue::Ref(10)` or `TypedValue::Long(10)`. A `ground` implementation that doesn't project the type tag, and doesn't join that tag against the rest of the query, will produce incorrect results.

query-translator/tests/translate.rs
```diff
@@ -550,4 +550,125 @@ fn test_complex_nested_or_join_type_projection() {
...
+
+#[test]
+fn test_ground_rel() {
+    let schema = prepopulated_schema();
+
+    // Verify that we accept inline constants.
+    let query = r#"[:find ?x ?y :where [(ground $ [[1 "xxx"] [2 "yyy"]]) [[?x ?y]]]]"#;
+    let SQLQuery { sql, args } = translate(&schema, query);
+    assert_eq!(sql, "SELECT DISTINCT `c00`.`?x` AS `?x`, `c00`.`?y` AS `?y` FROM \
+                         (SELECT 0 AS `?x`, 0 AS `?y` WHERE 0 UNION ALL VALUES (1, $v0), (2, $v1)) AS `c00`");
```
rnewman: Ooh, tricky. What is the type of `1`? Do we expect the projector to give you `Ref` or `Long`?

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -0,0 +1,384 @@
...
+    fn into_unit_matrix(self, lookup: F) -> Result<Vec<Vec<TypedValue>>>;
+    fn into_row_matrix(self, lookup: F) -> Result<Vec<Vec<TypedValue>>>;
+    fn into_column_matrix(self, lookup: F) -> Result<Vec<Vec<TypedValue>>>;
+    fn into_block_matrix(self, lookup: F) -> Result<Vec<Vec<TypedValue>>>;
+}
+
+impl<F> FnArgInto<F> for FnArg where F: Fn(Variable) -> Result<TypedValue> {
+    fn into_typed_value(self, lookup: F) -> Result<TypedValue> {
+        match self {
+            FnArg::EntidOrInteger(x) => Ok(TypedValue::Long(x)),
```
rnewman: Okay, so if you `[(ground 1) ?x]` it's always a `Long`. That's unambiguous, but it's not really correct. That makes these two queries return different results:

```
[?find ?x :where [?x _ 1]]
[?find ?x :where [(ground 1) ?y] [?x _ ?y]]
```

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -0,0 +1,384 @@
...
+            FnArg::Ident(x) => Ok(TypedValue::Keyword(Rc::new(x))),
```
rnewman: Similarly:

```
[:find ?x :where [?x :foo/bar ?y]]                    ; Works.
[:find ?x :where [(ground :foo/bar) ?a] [?x ?a ?y]]   ; Doesn't work; always a keyword.
```

query-sql/src/lib.rs
```diff
@@ -152,6 +152,14 @@ pub enum TableOrSubquery {
...
     Union(Vec<SelectQuery>, TableAlias),
     Subquery(Box<SelectQuery>),
+    Values(Values, TableAlias),
+}
+
+pub enum Values {
+    /// Like "VALUES (0, 1), (2, 3), ...".
+    Unnamed(Vec<Vec<TypedValue>>),
+    /// Like "SELECT 0 AS x, SELECT 0 AS y WHERE 0 UNION ALL VALUES (0, 1), (2, 3), ...".
+    Named(Vec<Variable>, Vec<Vec<TypedValue>>),
```
rnewman: It also occurs to me that the preferred way of doing this kind of thing on modern computers is not to have multiple vecs, but to have a single vec and a dimension argument. Better locality.

query/src/lib.rs
```diff
@@ -523,6 +532,35 @@ pub enum Binding {
...
+    }
+
+    /// Return `true` if no variable is bound twice, i.e., each binding entry is either a
+    /// placeholder or unique.
+    pub fn is_valid(&self) -> bool {
+        match self {
+            &Binding::BindScalar(_) | &Binding::BindColl(_) => true,
+            &Binding::BindRel(ref vars) | &Binding::BindTuple(ref vars) =>
+                vars.iter().filter_map(|x| x.var()).collect::<Vec<_>>().len() ==
+                vars.iter().filter_map(|x| x.var()).collect::<BTreeSet<_>>().len()
```
rnewman: We can't make _these_ varlists sets on the way out of the parser, because they include placeholders, and placeholders shouldn't dupe away.

query-algebrizer/src/clauses/where_fn.rs
rnewman: For `ground`, we must always have a fallback: we might have a query like

```
[:find ?x :where [(ground 5) ?x]]
```

This implies that there are _three_ possible behaviors here. Substitution is one obvious approach: behave as if the first query were entered.

Another is when a type is known:

```
[:find ?x :where [?x :foo/ref ?y] [(ground 10) ?y]]   -- 10 is an entid
```

The third is fallback.

rnewman: EDN is pretty expressive, so the main concern here is about entids/longs and idents/keywords. I'm still thinking about how best to address this. We already have mechanisms that can apply for the scalar case, but it's less clear for the more complex value shapes.

rnewman: Here's my proposed approach.

Datomic discusses entity resolution in the 'v' place:

http://docs.datomic.com/query.html#sec-8-1

and this problem is similar. If, for ambiguous values, we can delay concretization until all other patterns have been algebrized, then we can make a type-aware determination, falling back to a default if necessary.

If a user wants an ident to always resolve to an entid, never to a keyword, then they can do so themselves. Our algorithm, then:

1. If a value is of an unambiguous type (e.g., instant), then we're done.
2. If the value is of an ambiguous type (e.g., ident) but can't resolve as one kind, then it must be the other — `:foo/notattribute` is a keyword. We have the schema at alg time.
3. If the value is of an ambiguous type, but the query determines the type (e.g., by use of a keyword in the entity or attribute place), then we're done.
4. Finally, default to a single type.

We might consider supporting an `#entid` EDN feature.

query-algebrizer/src/clauses/where_fn.rs
rnewman: First of my comments is nonsense. Second is: it's an invalid query.

PR #424, opened 2017-04-21
[query] Complex `or` does not widen `known_types` correctly when arms agree

rnewman:
In `query-translator/tests/translate.rs`, I get a panic about known and extracted types with:
```rust
#[test]
fn test_complex_nested_or_join_type_projection() {
    let mut schema = Schema::default();
    associate_ident(&mut schema, NamespacedKeyword::new("page", "title"), 98);
    add_attribute(&mut schema, 98, Attribute {
        value_type: ValueType::String,
        ..Default::default()
    });
    associate_ident(&mut schema, NamespacedKeyword::new("page", "title2"), 97);
    add_attribute(&mut schema, 97, Attribute {
        value_type: ValueType::String,
        ..Default::default()
    });

    let input = r#"[:find [?y]
                    :where
                    (or
                      (or
                        [_ :page/title ?y])
                      (or
                        [_ :page/title ?y]))]"#;

    let SQLQuery { sql, args } = translate(&schema, input, None);
    assert_eq!(args, vec![]);
}
```

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -4,9 +4,10 @@ version = "0.0.1"
 workspace = ".."
 
 [dependencies]
+enum-set = { git = "https://github.com/rnewman/enum-set" }
+lazy_static = "0.2.2"
 num = "0.1.35"
 ordered-float = "0.4.0"
-lazy_static = "0.2.2"
 
 [dependencies.edn]
 path = "../edn"
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -8,6 +8,8 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
+extern crate enum_set;
+
 #[macro_use]
 extern crate lazy_static;
 extern crate ordered_float;
@@ -33,7 +35,8 @@ pub type Entid = i64;
 
 /// The attribute of each Mentat assertion has a :db/valueType constraining the value to a
 /// particular set.  Mentat recognizes the following :db/valueType values.
-#[derive(Clone,Copy,Debug,Eq,Hash,Ord,PartialOrd,PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialOrd, PartialEq)]
+#[repr(u32)]
 pub enum ValueType {
     Ref,
     Boolean,
@@ -42,6 +45,16 @@ pub enum ValueType {
     Double,
     String,
     Keyword,
+}
+
+impl enum_set::CLike for ValueType {
+    fn to_u32(&self) -> u32 {
+        *self as u32
+    }
+
+    unsafe fn from_u32(v: u32) -> ValueType {
+        std::mem::transmute(v)
+    }
 }
 
 impl ValueType {
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -21,6 +21,9 @@ pub mod values;
 use std::collections::BTreeMap;
 use std::fmt;
 use std::rc::Rc;
+
+use enum_set::EnumSet;
+
 use self::ordered_float::OrderedFloat;
 use self::edn::NamespacedKeyword;
 
@@ -46,6 +49,22 @@ pub enum ValueType {
     String,
     Keyword,
 }
+
+impl ValueType {
+    pub fn all_enums() -> EnumSet<ValueType> {
+        // TODO: lazy_static.
+        let mut s = EnumSet::new();
+        s.insert(ValueType::Ref);
+        s.insert(ValueType::Boolean);
+        s.insert(ValueType::Instant);
+        s.insert(ValueType::Long);
+        s.insert(ValueType::Double);
+        s.insert(ValueType::String);
+        s.insert(ValueType::Keyword);
+        s
+    }
+}
+
 
 impl enum_set::CLike for ValueType {
     fn to_u32(&self) -> u32 {
--- a/query-algebrizer/src/clauses/or.rs
+++ b/query-algebrizer/src/clauses/or.rs
@@ -9,7 +9,10 @@
 // specific language governing permissions and limitations under the License.
 
 use std::collections::btree_map::Entry;
-use std::collections::BTreeSet;
+use std::collections::{
+    BTreeMap,
+    BTreeSet,
+};
 
 use mentat_core::{
     Schema,
@@ -41,6 +44,7 @@ use types::{
     EmptyBecause,
     QualifiedAlias,
     SourceAlias,
+    ValueTypeSet,
     VariableColumn,
 };
 
--- a/query-algebrizer/src/clauses/or.rs
+++ b/query-algebrizer/src/clauses/or.rs
@@ -658,6 +658,21 @@ impl ConjoiningClauses {
             type_associations = type_needed.iter().cloned().collect();
         }
 
+        // Collect the new type information from the arms. There's some redundant work here --
+        // they already have all of the information from the parent.
+        // Note that we start with the first clause's type information.
+        {
+            let mut clauses = acc.iter();
+            let mut additional_types = clauses.next()
+                                              .expect("there's at least one clause")
+                                              .known_types
+                                              .clone();
+            for cc in clauses {
+                union_types(&mut additional_types, &cc.known_types);
+            }
+            self.broaden_types(additional_types);
+        }
+
         let union = ComputedTable::Union {
             projection: projection,
             type_extraction: type_needed,
@@ -676,6 +691,46 @@ impl ConjoiningClauses {
         self.from.push(SourceAlias(table, alias));
         Ok(())
     }
+}
+
+/// Helper to fold together a set of type maps.
+fn union_types(into: &mut BTreeMap<Variable, ValueTypeSet>,
+               additional_types: &BTreeMap<Variable, ValueTypeSet>) {
+    // We want the exclusive disjunction -- any variable not mentioned in both sets -- to default
+    // to ValueTypeSet::Any.
+    // This is necessary because we lazily populate known_types, so sometimes the type set will
+    // be missing a `ValueTypeSet::Any` for a variable, and we want to broaden rather than
+    // accidentally taking the other side's word for it!
+    // The alternative would be to exhaustively pre-fill `known_types` with all mentioned variables
+    // in the whole query, which is daunting.
+    let mut any: BTreeMap<Variable, ValueTypeSet>;
+    // Scoped borrow of `into`.
+    {
+        let i: BTreeSet<&Variable> = into.keys().collect();
+        let a: BTreeSet<&Variable> = additional_types.keys().collect();
+        any = i.symmetric_difference(&a)
+               .map(|v| ((*v).clone(), ValueTypeSet::Any))
+               .collect();
+    }
+
+    // Collect the additional types.
+    for (var, new_types) in additional_types {
+        match into.entry(var.clone()) {
+            Entry::Vacant(e) => {
+                e.insert(new_types.clone());
+            },
+            Entry::Occupied(mut e) => {
+                let new;
+                {
+                    new = e.get().union(&new_types);
+                }
+                e.insert(new);
+            },
+        }
+    }
+
+    // Blat in those that are disjoint.
+    into.append(&mut any);
 }
 
 #[cfg(test)]
--- a/query-algebrizer/src/clauses/or.rs
+++ b/query-algebrizer/src/clauses/or.rs
@@ -709,7 +709,7 @@ fn union_types(into: &mut BTreeMap<Variable, ValueTypeSet>,
         let i: BTreeSet<&Variable> = into.keys().collect();
         let a: BTreeSet<&Variable> = additional_types.keys().collect();
         any = i.symmetric_difference(&a)
-               .map(|v| ((*v).clone(), ValueTypeSet::Any))
+               .map(|v| ((*v).clone(), ValueTypeSet::any()))
                .collect();
     }
 
--- a/query-algebrizer/src/clauses/or.rs
+++ b/query-algebrizer/src/clauses/or.rs
@@ -664,7 +664,7 @@ impl ConjoiningClauses {
         {
             let mut clauses = acc.iter();
             let mut additional_types = clauses.next()
-                                              .expect("there's at least one clause")
+                                              .expect("there to be at least one clause")
                                               .known_types
                                               .clone();
             for cc in clauses {
@@ -720,10 +720,7 @@ fn union_types(into: &mut BTreeMap<Variable, ValueTypeSet>,
                 e.insert(new_types.clone());
             },
             Entry::Occupied(mut e) => {
-                let new;
-                {
-                    new = e.get().union(&new_types);
-                }
+                let new = e.get().union(&new_types);
                 e.insert(new);
             },
         }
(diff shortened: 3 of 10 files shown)


rnewman:
The `ValueTypeSet` change isn't strictly necessary, but I think it's worthwhile; Part 3 shows an improvement in clarity. I'm going to spend a bit more time shifting that over to an `EnumSet`, which will make it less expensive.

Part 4 has the real fix. With just the change you suggested, @ncalexan, a different test began failing, so this does two things: it computes the union of types aware of `::Any`, and _then_ it broadens the destination CC. The `::Any` change is necessary because we're lazy about noting types (which avoids taking mutable references to `known_types` all over the place); a union where one arm has an unbound type needs that to be taken into account when reducing.

rnewman:
And now we have an `EnumSet`. Note that I forked the upstream `enum-set` to fix some old outstanding issues. We can shift back when my PR merges.

query-algebrizer/src/types.rs
```diff
@@ -461,4 +464,140 @@ impl Debug for EmptyBecause {
...
     }
+}
+
+/// A set of `ValueType`s.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum ValueTypeSet {
+    None,
+    Any,
+    One(ValueType),
+    Many(EnumSet<ValueType>),
```
ncalexan: You could enforce your invariant statically and make this `Many(ValueType, EnumSet<ValueType>)`.  As it is, you need to comment on your invariant, and I think you should `assert` liberally that you never get `Many(empty())`.

I think this representation would make your cases cleaner almost everywhere.

ncalexan: But I'm OK with this; once the bugs get hammered out, and the assumptions documented, it will basically never be touched.  Consider making it `DistinguishedSet<T>` or something, since there's no dependence on `ValueType`.

query-algebrizer/src/clauses/or.rs
```diff
@@ -658,6 +658,21 @@ impl ConjoiningClauses {
             type_associations = type_needed.iter().cloned().collect();
         }
 
+        // Collect the new type information from the arms. There's some redundant work here --
+        // they already have all of the information from the parent.
+        // Note that we start with the first clause's type information.
+        {
+            let mut clauses = acc.iter();
+            let mut additional_types = clauses.next()
+                                              .expect("there's at least one clause")
```
ncalexan: nit: "there to be at least ...".

query-algebrizer/src/clauses/or.rs
```diff
@@ -678,6 +693,46 @@ impl ConjoiningClauses {
...
+    }
+
+    // Collect the additional types.
+    for (var, new_types) in additional_types {
+        match into.entry(var.clone()) {
+            Entry::Vacant(e) => {
+                e.insert(new_types.clone());
+            },
+            Entry::Occupied(mut e) => {
+                let new;
```
ncalexan: Do you need to scope?  And if you do, couldn't you `let new = { }`?

query-algebrizer/src/clauses/or.rs
```diff
@@ -678,6 +693,46 @@ impl ConjoiningClauses {
...
 
+/// Helper to fold together a set of type maps.
+fn union_types(into: &mut BTreeMap<Variable, ValueTypeSet>,
+               additional_types: &BTreeMap<Variable, ValueTypeSet>) {
+    // We want the exclusive disjunction -- any variable not mentioned in both sets -- to default
+    // to ValueTypeSet::Any.
+    // This is necessary because we lazily populate known_types, so sometimes the type set will
+    // be missing a `ValueTypeSet::Any` for a variable, and we want to broaden rather than
+    // accidentally taking the other side's word for it!
+    // The alternative would be to exhaustively pre-fill `known_types` with all mentioned variables
```
ncalexan: This is complicated because our "type map" isn't Python's `defaultdict`.  Why not make it so?  (Assuming we find a Rust implementation.)  Of course, we could also produce our own newtype for this.

query-algebrizer/src/clauses/or.rs
```diff
@@ -658,6 +658,21 @@ impl ConjoiningClauses {
             type_associations = type_needed.iter().cloned().collect();
         }
 
+        // Collect the new type information from the arms. There's some redundant work here --
+        // they already have all of the information from the parent.
```
ncalexan: The fact that we push type information down is strictly an optimization, right?  That is, this problem decomposes as a post-order tree walk followed by a type narrowing union or intersection.  We can cull parts of the tree by doing the intersection inline if we push type information into the children.

query-translator/tests/translate.rs
```diff
@@ -466,3 +466,39 @@ fn test_order_by() {
...
+    associate_ident(&mut schema, NamespacedKeyword::new("page", "title2"), 97);
+    add_attribute(&mut schema, 97, Attribute {
+        value_type: ValueType::String,
+        ..Default::default()
+    });
+
+    let input = r#"[:find [?y]
+                    :where
+                    (or
+                      (or
```
ncalexan: This test is a little fragile, since it's likely we'll eventually optimize this away:
1) a repeat `or` can be dropped;
2) even if it's not a repeat `or`, we can turn this into a single `WHERE a = 98 OR a = 99`.

query-translator/tests/translate.rs
```diff
@@ -466,3 +466,39 @@ fn test_order_by() {
...
+
+#[test]
+fn test_complex_nested_or_join_type_projection() {
+    let mut schema = Schema::default();
+    associate_ident(&mut schema, NamespacedKeyword::new("page", "title"), 98);
+    add_attribute(&mut schema, 98, Attribute {
+        value_type: ValueType::String,
+        ..Default::default()
+    });
+    associate_ident(&mut schema, NamespacedKeyword::new("page", "title2"), 97);
```
ncalexan: This extra attribute isn't actually used in the test.  Remove it or use it -- see the comment below.

rnewman:
> I feel there is a declarative formulation of this problem that will simplify
> a lot of this book-keeping and imperative work, and reduce the likelihood that
> we do the wrong thing for types needing type extraction. (Right now known_types
> is "more authoritative" than needs_extraction, I think.)

There are a few different ways to frame this. I think you're essentially
talking about constraint-based algebrizing.

What we've implemented here is a fairly straightforward somewhat-optimizing
compiler from Datalog to SQL.

Rather than solve a Datalog program to yield solutions that unify when the
query and the store are juxtaposed, we produce a SQL query that performs
equivalent unification.

We descend from top to bottom, accumulating the intersection of patterns. We
track enough metadata to do this correctly:

- Enough joined tables to represent the search space.
- Correlations between variables and database columns.
- Fixed values.
- Subqueries (recursively) where necessary.

We track enough type information about variables to accomplish three goals:

1. Unify correctly without having to project and compare every type tag.
2. Project only ambiguous type tag columns.
3. Prune clauses that can't return results.

The passing of types between recursive layers -- down when templating, in order
to optimize types, and up in order to accumulate type information for
contributed bindings -- simply serves to produce more efficient output. If we
made `known_types` a `BlackHoleMap`, we'd simply project and compare type tags
for every variable column in the query; the results should be the same.

(I might try that as an experiment!)

But even if we didn't pass types, we would have to pass column bindings, and do
so with care. Inter-clause boundaries need to be navigated regardless.
 
An alternative -- more thorough, but more intellectually demanding -- way to
solve this problem is to:

- Frame it as declarative constraints, which can be trivially collected.
- Simplify the constraints by combining them with the schema (as rules) and optimizations (as rules).
- Generate an intermediate query that's fully explicit about all needed checks.
- Translate that to SQL, or some other execution strategy.

This involves writing a bunch of logic like:

```
;; [?e ?attr ?x] where it's an integer-ranged attribute => ?x is an integer.
(<- (integer ?x)
  (?attr ?e ?x)
  (= (range-of ?attr integer)))

;; [?x ?attr ?y] => ?x is a ref.
(<- (ref ?x)
  (?attr ?x ?y))

;; [?x ?attr ?y] => ?attr is a ref.
(<- (ref ?attr)
  (?attr ?x ?y))
```

leading to a query that's a little easier to translate

`[:find ?x ?x_type ?y ?y_type
  :where
  [?x _ ?y]
  [(ground (type ?y) ?y_type)]
  [(ground :db.type/ref ?x_type)]]`

That still involves some careful thinking about what knowledge propagates
between scopes, which might be phrased as term rewriting, and it still requires
some coordination between parts of queries. It also makes preserving properties
of the input query, such as clause order, more challenging.

I've thought about representations for these kinds of constraints that allow
some amount of structure-preserving optimization -- mainly dependency trees and
pruning -- but only went far enough to understand that this was a complicated
problem.


> I think we can formulate this problem in a recursive descent, iterative
> narrowing framework. This is possible
> because the domain of discourse is finite and we don't need to push
> information from the root down into the leaves -- it is strictly an
> optimization. That is, we don't need the bi-directional widening of
> Hindley-Milner style type systems.

Splitting this in two:

- We don't strictly need to push column bindings down for non-filter clauses.
However, this will typically make SQL joins more expensive. For `not`,
expressed as `NOT EXISTS`, we need to share column bindings; if we didn't do so
we'd need to do a `LEFT EXCLUDING JOIN` and hope that SQLite would make the set
operation efficient.

- We don't strictly need to push type information into and out of sub-clauses.
However, if we don't we will be unable to prune the search space as we do now:
remember, thanks to the schema data we have, and the rules we're procedurally
encoding, we have more information than is available to SQLite.


In the simplest world where we directly translate to SQL, doing no type
analysis, a query like:

```
[:find ?x ?v
 :where
 [?x :foo/integer ?v]
 (or
   [?x :foo/string ?v]
   [?x :foo/keyword ?v])]
```

is extremely expensive, turning into:

```
SELECT datoms00.e AS x,
       datoms00.v AS v,
       datoms00.value_type_tag AS v_type_tag
FROM datoms datoms00
JOIN
(SELECT datoms01.e AS x,
        datoms01.v AS v,
        datoms01.value_type_tag AS v_type_tag
 FROM datoms datoms01
 WHERE datoms00.a = 67     -- :foo/string
 UNION
 SELECT datoms02.e AS x,
        datoms02.v AS v,
        datoms02.value_type_tag AS v_type_tag
 FROM datoms datoms02
 WHERE datoms00.a = 68)    -- :foo/keyword
AS or01
ON or01.x = datoms00.e
WHERE or01.v = datoms00.v
  AND or01.v_type_tag = datoms00.v_type_tag
  AND datoms00.a = 66      -- :foo/integer
```

… only to return no results.

If we only propagate type information *out of* subqueries we will know that this
returns no results. Things get 'tighter' sooner -- simplifying the `or`
immediately -- if we propagate in both directions, as we do now.

However, if we only propagate types out of subqueries, we will sometimes
project type tags unnecessarily. To avoid that we'd need to walk the tree again
re-propagating types: there's no free lunch.

With bidirectional propagation this can still happen, of course, thanks to
sequential processing -- think about sibling subqueries -- but I think it's
likely to be uncommon in real-world queries.

 
> If we make types authoritative, and make it a Python defaultdict like
> structure that defaults to TypeSet::All, can't we simplify almost all of this?
> EnumSet is a bitfield, so the space cost of keeping types for all mentioned
> variables isn't large. Then we can do the needs_extraction parts much later.

We could -- indeed, that was my first implementation strategy.

Unfortunately that requires every access to `cc.known_types`, and every call to
`cc.known_type()`, to take a mutable reference to `cc`. And that means all of
their callers must, and in the same scope they can't call a function that takes
an immutable borrow. Eventually that pushes mutable references into a lot of
places.

To fix this bug didn't require that work; it simply required being a little
more careful in one spot. And defaulting a type set to `Any` for every variable
actually doesn't solve parts of this bug: you can see where I don't use an
empty set to reduce, because doing so in a union of course broadens to `Any`
for every variable!

 
> Am I missing something? Can this get simpler, or is there some essential
> feature I'm neglecting?

No, you're not missing something. It's simply the case that I've made two choices:

- To build a predictable, procedural, verbose/nested transformation rather than
  a more demanding (but 'flatter' and more general) constraint-analysis system.
- To go the extra distance of propagating the maximal amount of information in
  order to be able to (procedurally) simplify queries as much as possible.

That results in a certain amount of verbose, sometimes error-prone bookkeeping,
but for the moment I think it's adequate.

At some point I'd like to go a small step further, which is to separate the
collecting/simplification/result stages. You've alluded to this in the past
with reference to a kind of builder pattern.

I'd also consider a simple re-algebrizing process: we are likely to need this
for partially prepared statements, where input bindings aren't necessarily
available when the query is first parsed, but it would also allow us to achieve
simpler queries by passing information to prior sibling clauses, and
potentially allow us to simplify this code.

And eventually, if the rest of the system proves to be sound, we can think
about more sophisticated query planning.

My bet is that we will make significant SQL-schema or storage level changes
that require a complete rewrite of this code before there's a query-specific
reason to do so.

core/src/lib.rs
```diff
@@ -33,7 +35,8 @@ pub type Entid = i64;
 
 /// The attribute of each Mentat assertion has a :db/valueType constraining the value to a
 /// particular set.  Mentat recognizes the following :db/valueType values.
-#[derive(Clone,Copy,Debug,Eq,Hash,Ord,PartialOrd,PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialOrd, PartialEq)]
+#[repr(u32)]
```
rnewman: It's in the "do this" section for `EnumSet`. I think it means "please represent this enum, which must be C-like — that is, with no associated values — as 32 bits, so I can be sure of my conversion to u32 later".

http://rustbyexample.com/custom_types/enum/c_like.html

A C-like enum gets a discriminant of 0, 1, 2, etc. for the cases.

`CLike::bit` then left-shifts 1 that many times.

This is equivalent to having a discriminant of 1, 2, 4, 8, 16, etc, which pack neatly into a bitfield. This is why we have a limitation of 31 cases in the enum: we only have 32 bit flags to flip.

Without forcing `repr(u32)` Rust might choose to use a single byte to represent this enum, potentially giving us junk or a crash when we do the unsafe `CLike::from_u32` inside `enum_set`.

Does that make sense?

core/Cargo.toml
```diff
@@ -4,9 +4,10 @@ version = "0.0.1"
 workspace = ".."
 
 [dependencies]
+enum-set = { git = "https://github.com/rnewman/enum-set" }
```
rnewman: Shell copy and paste!

query-algebrizer/src/types.rs
```diff
@@ -461,4 +464,140 @@ impl Debug for EmptyBecause {
...
+    pub fn exemplar(&self) -> Option<ValueType> {
+        match self {
+            &ValueTypeSet::None => None,
+            &ValueTypeSet::Any => Some(ValueType::Ref),
+            &ValueTypeSet::One(t) => Some(t),
+            &ValueTypeSet::Many(ref s) => s.iter().next(),
+        }
+    }
+
+    pub fn contains(&self, vt: ValueType) -> bool {
```
rnewman: The `contains` call for collections usually does, but `ValueType` is `Copy` and probably only 32 or 64 bits wide, so it's cheaper to pass by copy than by reference.

query-algebrizer/src/types.rs
```diff
@@ -461,4 +464,140 @@ impl Debug for EmptyBecause {
...
+        match self {
+            &ValueTypeSet::None => false,
+            &ValueTypeSet::Any => true,
+            &ValueTypeSet::One(t) => t == vt,
+            &ValueTypeSet::Many(ref s) => s.contains(&vt),
+        }
+    }
+
+    pub fn is_none(&self) -> bool {
```
rnewman: `is_empty` is better, I think.

I was probably going for symmetry with `Any`, `None`, `Many` — it depends on whether you're looking at this as a set of types, or as a possible single value with some additional "I'm not a single value" states.

Issue #426, opened 2017-04-21
[meta] Support UUIDs


rnewman:
Part of #201.

Meta:
- Define value type
- Translate to and from SQL
- Support in queries
- Check transactor support

rnewman:
Fixes by #438.

PR #438, opened 2017-04-28
UUIDs and instants. Fixes #44, #45, #426, #427.

### LINKED ISSUE #44: Store UUIDs as BLOBs
16-byte blob beats 36 bytes of string space-wise.

rnewman:
At present this means you can't write a query comparing timestamps (to timestamps or integers!). I'll do that in a follow-up.

### DIFF
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -81,7 +81,7 @@ pub fn new_connection<T>(uri: T) -> rusqlite::Result<rusqlite::Connection> where
 
 /// Version history:
 ///
-/// 1: initial Mentat schema.
+/// 1: initial Rust Mentat schema.
 pub const CURRENT_VERSION: i32 = 1;
 
 /// MIN_SQLITE_VERSION should be changed when there's a new minimum version of sqlite required
@@ -354,7 +354,10 @@ impl TypedSQLValue for TypedValue {
         match (value_type_tag, value) {
             (0, rusqlite::types::Value::Integer(x)) => Ok(TypedValue::Ref(x)),
             (1, rusqlite::types::Value::Integer(x)) => Ok(TypedValue::Boolean(0 != x)),
+
+            // Negative integers are simply times before 1970.
             (4, rusqlite::types::Value::Integer(x)) => Ok(TypedValue::Instant(DateTime::<UTC>::from_micros(x))),
+
             // SQLite distinguishes integral from decimal types, allowing long and double to
             // share a tag.
             (5, rusqlite::types::Value::Integer(x)) => Ok(TypedValue::Long(x)),
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -26,7 +26,14 @@ use rusqlite::limits::Limit;
 
 use ::{repeat_values, to_namespaced_keyword};
 use bootstrap;
-use edn::types::Value;
+
+use edn::{
+    DateTime,
+    UTC,
+    Uuid,
+    Value,
+};
+
 use entids;
 use mentat_core::{
     attribute,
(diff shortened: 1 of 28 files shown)


rnewman:
Fixes #44, #45, #426, #427.

edn/src/edn.rustpeg
```diff
@@ -14,8 +14,14 @@ use std::collections::{BTreeSet, BTreeMap, LinkedList};
 use std::iter::FromIterator;
 use std::f64::{NAN, INFINITY, NEG_INFINITY};
 
+use chrono::{
+    DateTime,
+    TimeZone,
+    UTC
```
rnewman: rustpeg won't allow a trailing comma here, btw.

edn/src/edn.rustpeg
```diff
@@ -142,6 +148,52 @@ pub text -> ValueAndSpan =
...
+              "Z" / (("+" / "-") [0-2][0-9] ":" [0-5][0-9])
+            )
+    "\"" {?
+        DateTime::parse_from_rfc3339(d)
+            .map(|t| t.with_timezone(&UTC))
+            .map_err(|_| "invalid datetime")        // Oh, rustpeg.
+    }
+
+pub inst_millis -> DateTime<UTC> =
+    d:$( digit+ ) {
```
rnewman: I could be persuaded that we accept:

```
#inst "2017…"
#instmillis 1493410721802
#instmicros 1493410721802123
```

instead of this.

sql/src/lib.rs
```diff
@@ -136,11 +140,16 @@ impl QueryBuilder for SQLiteQueryBuilder {
             &Boolean(v) => self.push_sql(if v { "1" } else { "0" }),
             &Long(v) => self.push_sql(v.to_string().as_str()),
             &Double(OrderedFloat(v)) => self.push_sql(v.to_string().as_str()),
-
-            // These are both `Rc`. We can just clone an `Rc<String>`, but we
-            // must make a new single `String`, wrapped in an `Rc`, for keywords.
-            &String(ref s) => self.push_static_arg(s.clone()),
-            &Keyword(ref s) => self.push_static_arg(Rc::new(s.as_ref().to_string())),
+            // These are both `Rc`. Unfortunately, we can't use that fact when
```
ncalexan: Yeah, this is awkward.  It does seem that you could rework this to use `rusqlite::ToSqlOutput<'a>`, since the `Rc` will have at least as long a lifetime as `&'a self`.  But it's work, and I'm not 100% sure I'm actually correct.

sql/src/lib.rs
```diff
@@ -136,11 +140,16 @@ impl QueryBuilder for SQLiteQueryBuilder {
...
+            // turning these into rusqlite Values.
+            &String(ref s) => {
+                let v = Rc::new(rusqlite::types::Value::Text(s.as_ref().clone()));
```
ncalexan: But this will destroy the value of interning strings in the query parser, right?  Not that we intern yet; once we do, we should be able to coalesce repeated values and avoid cloning the same string more than once.  I guess that'll be good enough.

db/src/db.rs
```diff
@@ -352,6 +354,7 @@ impl TypedSQLValue for TypedValue {
         match (value_type_tag, value) {
             (0, rusqlite::types::Value::Integer(x)) => Ok(TypedValue::Ref(x)),
             (1, rusqlite::types::Value::Integer(x)) => Ok(TypedValue::Boolean(0 != x)),
+            (4, rusqlite::types::Value::Integer(x)) => Ok(TypedValue::Instant(DateTime::<UTC>::from_micros(x))),
```
rnewman: Negative is fine; that's how one represents times earlier than 1970!

db/src/types.rs
```diff
@@ -96,3 +95,13 @@ pub struct TxReport {
     /// literal tempids to all unify to a single freshly allocated entid.)
     pub tempids: BTreeMap<String, Entid>,
 }
+
+impl Default for TxReport {
+    fn default() -> Self {
+        TxReport {
+            tx_id: Entid::default(),
+            tx_instant: UTC::now(),
```
rnewman: I wonder why `TxReport` needs `Default` anyway…

sql/src/lib.rs
rnewman: We'll have to do that work here, I think. We could even use an `Rc`-keyed map to grab `Rc<Value>`s.

edn/src/types.rs
```diff
@@ -337,14 +345,15 @@ macro_rules! def_common_value_methods {
...
                 $t::Text(_) => 5,
-                $t::PlainSymbol(_) => 6,
-                $t::NamespacedSymbol(_) => 7,
-                $t::Keyword(_) => 8,
-                $t::NamespacedKeyword(_) => 9,
-                $t::Vector(_) => 10,
-                $t::List(_) => 11,
-                $t::Set(_) => 12,
-                $t::Map(_) => 13,
+                $t::Uuid(_) => 6,
```
rnewman: AIUI it doesn't make a difference, so I was simply putting atoms together, and similar-ish types together. Am I missing some consequence of this?

db/src/types.rs
rnewman: Yeah, `Default` is gone!

PR #444, opened 2017-05-05
Parse without copying.  Fixes #436.

### LINKED ISSUE #436: [tx] Speed up transaction parser
When importing a map structure via transactions that contains around 1000 lines I'm getting around 3 seconds on the parse

mentat-profile

ncalexan:


ncalexan:
https://github.com/mozilla/mentat/commit/8a8fcedd1c9d57ed6673f398b84bf72aa9f0bd5f

ncalexan:
> My god that's a lot of code. I haven't had the amount of attention needed to do a good review of this, and I'm unlikely to in the next week, so if you're confident, that's fine.

I'm confident enough to land, for sure.

> My initial concern: parsing with references doesn't allow us to steal the inner values from the edn::ValueAndSpans. That means this isn't entirely without copying: it's without incidental copying within the parsing framework.
> 
> Is there a phrasing of this that can parse by reference, and then take ownership? You could do this with an Rc<T>, with Rc::try_unwrap — turn a ValueAndSpan into an Rc<ValueAndSpan>, parse against those (with cheap clones but not Copy), and then when we've left the parser and the refcount is back to 1, grab the real value — no cloning of edn values themselves, only cloning of Rc.

I thought about this, and couldn't see how to achieve it without adding support to the underlying `edn::ValueAndSpan` instances.
 
> Anything else spring to mind? Is this likely to be a win?

Yes, I absolutely think it will be a win -- I just don't see how to get to that point just now.  I expect we will push a lot of the shared references into the EDN parser so the "expensive `String` clone" is not expensive rather than try to finesse this at the `ValueStream` level.

> I would like to get to understand the keyword map stuff better, but don't currently have time.

Understood.  It's really not complicated, it's just onerous to understand -- and relies on tricks (culted from `combine`) that work-around the hygiene system Rust's macro system fails to enforce.

PR #447, opened 2017-05-08
[tx] Forbid insertion of unknown entity IDs

rnewman:
The Datomic docs state the following:

>  There are three ways to sepcify an entity id:
> - a temporary id for a new entity being added to the database
> - an existing id for an entity that's already in the database
> - an identifier for an entity that's already in the database

Note that _not_ mentioned is "by making up a new entity ID".

The problem with simply making up a new entity ID is that it'll collide with tempid allocations, either within a transact itself, or with a later tempid allocation colliding with an existing entity.

I've written a couple of failing tests for this. I believe the most straightforward and robust solution is for non-syncing code to simply be forbidden from transacting new raw entity IDs: it should use upserting or tempids to define new entities.

A simple and cheap check is to ensure that every entid in transact input is within the used part ranges. A more thorough solution would be to maintain a set of every mentioned entity ID. We might be doing this anyway for syncing.

### DIFF
--- a/db/src/tx.rs
+++ b/db/src/tx.rs
@@ -57,10 +57,14 @@ use db::{
     MentatStoring,
     PartitionMapping,
 };
+use edn::{
+    NamespacedKeyword,
+};
 use entids;
 use errors::{ErrorKind, Result};
 use internal_types::{
     Either,
+    KnownEntid,
     LookupRef,
     LookupRefOrTempId,
     TempIdHandle,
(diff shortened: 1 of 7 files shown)


rnewman:
@ncalexan, what do you think? This seems like a subtle footgun that we ought to disarm…

rnewman:
N.B., if we only ever allocate new IDs via tempid allocation, and the only way to allocate tempids is within transact (or sync ID remapping), then we can probably preserve a nice property: that every ID in the contiguous part space has been mentioned in the tx log. I was a little worried about having a sparse identifier space, but this would make it contiguous — much, much more efficient.

db/src/tx.rs
```diff
@@ -103,6 +103,14 @@ use types::{
 };
 use upsert_resolution::Generation;
 
+struct PartRanges(Vec<(i64, i64)>);
```
ncalexan: You could profitably newtype `PartitionMap`, simplifying this slightly.  Either that or add it to the `PartitionMapping` trait that already exists.

rnewman:
@ncalexan: thoughts on this?

src/conn.rs
```diff
@@ -179,6 +179,59 @@ mod tests {
...
+
+        match conn.transact(&mut sqlite, t.as_str()).unwrap_err() {
+            Error(ErrorKind::DbError(::mentat_db::errors::ErrorKind::UnrecognizedEntid(e)), _) => {
+                // All this, despite this being the ID we were about to allocate!
+                assert_eq!(e, next);
+            },
+            x => panic!("expected transact error, got {:?}", x),
+        }
+
+        // And if we subsequently transact in a way that allocates one ID, we _will_ use that one.
```
ncalexan: Maybe worth noting that 10 is allocated by the bootstrap process.

PR #448, opened 2017-05-09
Begin moving web server to a separate crate.

rnewman:
This doesn't yet introduce a working Cargo.toml for 'mentatweb', but it
does allow RLS to build correctly without errors, and it reduces the
core library's dependency space, which is more important in the short
term.


bgrins:
Looks good to me

PR #461, opened 2017-05-09
Implement code path for re-opening databases

rnewman:
This means we can't actually reopen DBs yet, which is a bit of a snag.

```
pub fn ensure_current_version(conn: &mut rusqlite::Connection) -> Result<DB> {
    if rusqlite::version_number() < MIN_SQLITE_VERSION {
        panic!("Mentat requires at least sqlite {}", MIN_SQLITE_VERSION);
    }

    let user_version = get_user_version(&conn)?;
    match user_version {
        0 => create_current_version(conn),
        // TODO: support updating or re-opening an existing store.
        v => bail!(ErrorKind::NotYetImplemented(format!("Opening databases with Mentat version: {}", v))),
    }
}
```


rnewman:
Misfire with `hub`.

PR #462, opened 2017-05-09
Allow opening a DB.

rnewman:
Still needs a test…

### DIFF
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -335,8 +335,10 @@ pub fn ensure_current_version(conn: &mut rusqlite::Connection) -> Result<DB> {
 
     let user_version = get_user_version(&conn)?;
     match user_version {
-        0 => create_current_version(conn),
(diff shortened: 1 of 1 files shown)


db/src/db.rs
```diff
@@ -335,8 +335,10 @@ pub fn ensure_current_version(conn: &mut rusqlite::Connection) -> Result<DB> {
 
     let user_version = get_user_version(&conn)?;
     match user_version {
-        0 => create_current_version(conn),
-        // TODO: support updating or re-opening an existing store.
+        0               => create_current_version(conn),
+        CURRENT_VERSION => read_db(conn),
+
+        // TODO: support updating an existing store.
         v => bail!(ErrorKind::NotYetImplemented(format!("Opening databases with Mentat version: {}", v))),
```
fluffyemily: Do we want to also indent the `=>` to match here, if we're formatting?

Issue #453, opened 2017-05-09
Accept mentat commands from inside CLI


fluffyemily:
* Determine if a mentat command is being entered. A mentat command starts with a `[` or `{` and encompases everything typed until a matching `]` or `}` is entered, at which point the command is ready executed.
* Examine the entered command before execution and determine if it is a query (`:find`) or a transaction (everything else).

rnewman:
Do bear in mind that technically `[:where [?x :foo/knows ?y] :find ?y]` is valid, so we can't just check the first symbol.

PR #465, opened 2017-05-16
Accept and parse EDN Query and Transact commands (#453)

fluffyemily:
https://github.com/mozilla/mentat/issues/453

depends on https://github.com/mozilla/mentat/pull/463

### DIFF
--- a/tools/cli/src/mentat_cli/command_parser.rs
+++ b/tools/cli/src/mentat_cli/command_parser.rs
@@ -30,6 +30,8 @@ use combine::combinator::{
 
 use errors as cli;
 
+use edn;
+
 pub static HELP_COMMAND: &'static str = &"help";
 pub static OPEN_COMMAND: &'static str = &"open";
 pub static CLOSE_COMMAND: &'static str = &"close";
@@ -52,8 +54,10 @@ impl Command {
     /// TODO: for query and transact commands, they will be considered complete if a parsable EDN has been entered as an argument
     pub fn is_complete(&self) -> bool {
         match self {
-            &Command::Query(_) |
-            &Command::Transact(_) => false,
+            &Command::Query(ref args) |
+            &Command::Transact(ref args) => {
+                edn::parse::value(&args).is_ok()
+            },
             &Command::Help(_) |
             &Command::Open(_) |
             &Command::Close => true
--- a/tools/cli/src/mentat_cli/command_parser.rs
+++ b/tools/cli/src/mentat_cli/command_parser.rs
@@ -9,7 +9,9 @@
 // specific language governing permissions and limitations under the License.
 
 use combine::{
+    any,
     eof, 
+    many,
     many1, 
     satisfy, 
     sep_end_by, 
@@ -31,11 +33,13 @@ use errors as cli;
 pub static HELP_COMMAND: &'static str = &"help";
 pub static OPEN_COMMAND: &'static str = &"open";
 pub static CLOSE_COMMAND: &'static str = &"close";
+pub static QUERY_COMMAND: &'static str = &"q";
+pub static TRANSACT_COMMAND: &'static str = &"t";
 
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub enum Command {
-    Transact(Vec<String>),
-    Query(Vec<String>),
+    Transact(String),
+    Query(String),
     Help(Vec<String>),
     Open(String),
     Close,
@@ -88,13 +92,38 @@ pub fn command(s: &str) -> Result<Command, cli::Error> {
                         }
                         Ok(Command::Close)
                     });
+    
+    let opening_brace_parser = try(string("["))
+                    .or(try(string("{")));
+    
+    let edn_arg_parser = spaces()
+                    .and(opening_brace_parser.clone()
+                    .and(many::<Vec<_>, _>(try(any()))));
+
+    let query_parser = string(QUERY_COMMAND)
+                    .and(edn_arg_parser.clone())
+                    .map( |x| {
+                        let args = (x.1).1;
+                        let content: String = args.1.iter().collect();
+                        Ok(Command::Query(format!("{}{}", args.0, content)))
+                    });
+
+    let transact_parser = string(TRANSACT_COMMAND)
+                    .and(edn_arg_parser.clone())
+                    .map( |x| {
+                        let args = (x.1).1;
+                        let content: String = args.1.iter().collect();
+                        Ok(Command::Transact(format!("{}{}", args.0, content)))
+                    });
 
     spaces()
     .skip(token('.'))
-    .with(choice::<[&mut Parser<Input = _, Output = Result<Command, cli::Error>>; 3], _>
+    .with(choice::<[&mut Parser<Input = _, Output = Result<Command, cli::Error>>; 5], _>
           ([&mut try(help_parser),
             &mut try(open_parser),
-            &mut try(close_parser),]))
+            &mut try(close_parser),
+            &mut try(query_parser),
+            &mut try(transact_parser)]))
         .skip(spaces())
         .skip(eof())
         .parse(s)
@@ -198,10 +227,17 @@ mod tests {
             _ => assert!(false)
         }
     }
-
+    
     #[test]
     fn test_open_parser_no_args() {
         let input = ".open";
+        let err = command(&input).expect_err("Expected an error");
+        assert_eq!(err.to_string(), "Missing required argument");
+    }
+
+    #[test]
+    fn test_open_parser_no_args_trailing_whitespace() {
+        let input = ".open ";
         let err = command(&input).expect_err("Expected an error");
         assert_eq!(err.to_string(), "Missing required argument");
     }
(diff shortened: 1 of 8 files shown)


tools/cli/src/mentat_cli/command_parser.rs
```diff
@@ -224,6 +265,112 @@ mod tests {
...
+            Command::Err(message) => {
+                assert_eq!(message, format!("Invalid command {:?}", input));
+            },
+            _ => assert!(false)
+        }
+    }
+
+    #[test]
+    fn test_transact_parser_complete_edn() {
+        let input = ".t [:db/add \"s\" :db/ident :foo/uuid]";
```
ncalexan: A transaction is a vector-of-vectors or a vector-of-maps, in order to handle multiple assertions.  So
```rust
.t [[:db/add "t" ...]
     [:db/add "r" ...]]
```

tools/cli/src/mentat_cli/command_parser.rs
```diff
@@ -40,8 +42,9 @@ impl Command {
 }
 
 pub fn command(s: &str) -> Command {
+    let skip_spaces = skip_many(space());
```
ncalexan: This is the `spaces()` parser from `combine::char`.  I've found that `let p = || ...` and then `p()` is more natural than cloning the parser.

tools/cli/src/mentat_cli/command_parser.rs
```diff
@@ -62,18 +65,44 @@ pub fn command(s: &str) -> Command {
...
         ([&mut try(help_parser),
           &mut try(open_parser),
-          &mut try(close_parser),]))
+          &mut try(close_parser),
+          &mut try(query_parser),
+          &mut try(transact_parser),
+        ]))
     .skip(eof())
     .parse(s)
-    .map(|x| x.0)
```
ncalexan: Eh?  This seems like unnecessary whitespace.

tools/cli/src/mentat_cli/command_parser.rs
```diff
@@ -62,18 +65,44 @@ pub fn command(s: &str) -> Command {
...
-    skip_many(space())
+    let opening_brace_parser = try(string("["))
+                    .or(try(string("{")));
+    
+    let edn_arg_parser = skip_spaces.clone()
+                    .and(opening_brace_parser.clone()
+                    .and(many::<Vec<_>, _>(try(any()))));
+
+    let query_parser = string(QUERY_COMMAND)
+                    .and(edn_arg_parser.clone())
```
ncalexan: See my other commit with `.skip` and `.then` for avoiding these indexing operations.

tools/cli/src/mentat_cli/command_parser.rs
```diff
@@ -62,18 +65,44 @@ pub fn command(s: &str) -> Command {
...
+                    .and(skip_spaces.clone())
                     .map( |_| Command::Close );
 
-    skip_many(space())
+    let opening_brace_parser = try(string("["))
+                    .or(try(string("{")));
+    
+    let edn_arg_parser = skip_spaces.clone()
+                    .and(opening_brace_parser.clone()
+                    .and(many::<Vec<_>, _>(try(any()))));
```
ncalexan: Can you explain why you need to `try` here?  `any` will either consume and succeed or not consume and fail; there shouldn't be consume and fail.

tools/cli/src/mentat_cli/command_parser.rs
```diff
@@ -8,18 +8,20 @@
...
-use combine::{any, eof, many, many1, sep_by, skip_many, token, Parser};
-use combine::combinator::{choice, try};
+use combine::{any, eof, many, many1, skip_many, token, Parser};
+use combine::combinator::{choice, skip, try};
 use combine::char::{space, string};
 
 pub static HELP_COMMAND: &'static str = &"help";
 pub static OPEN_COMMAND: &'static str = &"open";
 pub static CLOSE_COMMAND: &'static str = &"close";
+pub static QUERY_COMMAND: &'static str = &"q";
```
ncalexan: Can we have synonyms, ".query" for ".q", ".transact" for ".t"?

tools/cli/src/mentat_cli/command_parser.rs
fluffyemily: I am not sure why, but if I leave the `try` out, I get no output at all from the `many`.

tools/cli/src/mentat_cli/repl.rs
```diff
@@ -62,7 +56,8 @@ impl Repl {
         let mut input = InputReader::new();
 
         loop {
-            let res = input.read_input(if more.is_some() { MORE_PROMPT } else { DEFAULT_PROMPT });
+            let res = input.read_input(more.clone());
```
ncalexan: I'm surprised you need this `clone`, since you're joining strings in `read_input`, not re-using them.

tools/cli/src/mentat_cli/input.rs
```diff
@@ -80,7 +86,19 @@ impl InputReader {
...
 
-        let cmd = try!(command(&self.buffer));
+        let cmd = match in_process_cmd {
+            Some(Command::Query(args)) => {
+                Command::Query(args + " " + &line)
+            },
+            Some(Command::Transact(args)) => {
+                Command::Transact(args + " " + &line)
+            },
+            _ => {
```
ncalexan: Prefer to explicitly match `Some(...)` and `None` here, so that the compiler will tell you when you need to handle cases.  I feel like there's an abstraction around continuable commands vs. non-continuable commands missing here, but it's your choice how to express this.

tools/cli/src/mentat_cli/input.rs
```diff
@@ -80,7 +86,19 @@ impl InputReader {
...
+                try!(command(&self.buffer))
```
ncalexan: Prefer `...?` to `try!(...)`; it's the modern expression in Rust.  Are you really only doing this evaluation for the error result?  You should consume the result of any `Result`, using `let _ = ...` if needed; the compiler should be warning you that you're dropping a result on the floor here.

tools/cli/src/mentat_cli/command_parser.rs
```diff
@@ -90,34 +93,33 @@ pub fn command(s: &str) -> Result<Command, cli::Error> {
...
-                        let content: String = args.1.iter().collect();
-                        Ok(Command::Query(format!("{}{}", args.0, content)))
-                    });
-
-    let transact_parser = string(TRANSACT_COMMAND)
-                    .and(edn_arg_parser.clone())
+    let edn_arg_parser = || spaces()
+                    .with(try(string("["))
+                        .or(try(string("{")))
+                        .then(|d| parser(move |input| {
```
ncalexan: If I understand correctly, you're trying to parse any line prefixed by optional space and either `[` or `{`.  I think you can do that with:
```rust
spaces()
    .with(look_ahead(string("[").or(string("{")))
        .with(many(any()))
```
Using `then` and a special parser like you do is ingenious but definitely not a natural expression using `combine` :)  You might also do this as
```rust
spaces().then(many(any())).and_then(|line| {
    if line.starts_with("[") || line.starts_with("{") {
        Ok(...)
    } else {
        Err(...)  
    }
})
```
but then you need to know a lot more about error handling in `combine`.  Stick to the defined combinators for easiest use.

tools/cli/src/mentat_cli/command_parser.rs
```diff
@@ -28,15 +27,19 @@ use combine::combinator::{
...
 
 use edn;
 
 pub static HELP_COMMAND: &'static str = &"help";
 pub static OPEN_COMMAND: &'static str = &"open";
 pub static CLOSE_COMMAND: &'static str = &"close";
-pub static QUERY_COMMAND: &'static str = &"q";
-pub static TRANSACT_COMMAND: &'static str = &"t";
+pub static QUERY_COMMAND: &'static str = &"query";
+pub static ALT_QUERY_COMMAND: &'static str = &"q";
```
ncalexan: If you label these `SHORT_` and `LONG_`, your `try` clauses will read more clearly.

tools/cli/src/mentat_cli/command_parser.rs
```diff
@@ -207,6 +236,13 @@ mod tests {
     }
 
     #[test]
+    fn test_open_parser_no_args_trailing_whitespace() {
+        let input = ".open ";
+        let err = command(&input).expect_err("Expected an error");
```
ncalexan: I see that `expect_err` is in 1.17.0 stable, so we can use it -- but we need to bump https://github.com/mozilla/mentat/blob/8a8fcedd1c9d57ed6673f398b84bf72aa9f0bd5f/build/version.rs#L19 as a Pre: part to stop folks failing here (as I just did locally!).

tools/cli/src/mentat_cli/repl.rs
```diff
@@ -52,21 +52,22 @@ impl Repl {
 
     /// Runs the REPL interactively.
     pub fn run(&mut self) {
-        let mut more: Option<Command> = None;
+        let mut more = false;
```
ncalexan: There is a compiler warning with this patch sequence that reveals that `more` is only assigned to, not read.  This is vestigial?

tools/cli/src/mentat_cli/command_parser.rs
```diff
@@ -90,34 +93,33 @@ pub fn command(s: &str) -> Result<Command, cli::Error> {
 
     let close_parser = string(CLOSE_COMMAND)
                     .with(arguments())
+                    .skip(spaces())
+                    .skip(eof())
                     .map(|args| {
                         if args.len() > 0 {
                             return Err(cli::ErrorKind::CommandParse(format!("Unrecognized argument {:?}", args[0])).into());
```
ncalexan: There's a handy `bail!` macro specifically to make this type of early return more pleasant.

tools/cli/src/mentat_cli/repl.rs
fluffyemily: I thought I had removed that here, but it turns out it was done elsewhere. Removing...

tools/cli/src/mentat_cli/command_parser.rs
```diff
@@ -110,16 +112,21 @@ pub fn command(s: &str) -> Result<Command, cli::Error> {
                             Ok((d.to_string() + input, Consumed::Empty(input)))
                         })));
 
-    let query_parser = try(string(QUERY_COMMAND)).or(try(string(ALT_QUERY_COMMAND)))
+    let edn_arg_parser = || spaces()
```
ncalexan: I don't see any movement or discussion of https://github.com/mozilla/mentat/pull/465#discussion_r118084725 -- did you try and fail to improve this, or do you particularly like the expression as is?

Improving this would avoid `.into_iter().collect()` two places below, which is a strong sign that your types are wrong: IIUC, you're collecting `Vec<String>` (where each string is one character!) into a `String`.  That's going to inefficient.

tools/cli/src/mentat_cli/command_parser.rs
```diff
@@ -110,16 +112,21 @@ pub fn command(s: &str) -> Result<Command, cli::Error> {
...
                         .with(edn_arg_parser())
                         .map(|x| {
-                            Ok(Command::Query(x))
+                            Ok(Command::Query(x.into_iter().collect()))
                         });
 
-    let transact_parser = try(string(TRANSACT_COMMAND)).or(try(string(ALT_TRANSACT_COMMAND)))
+    let transact_parser = try(string(LONG_TRANSACT_COMMAND)).or(try(string(SHORT_TRANSACT_COMMAND)))
                     .with(edn_arg_parser())
                     .map( |x| {
```
ncalexan: nit: no leading space before the argument: `.map(|x| { ...`.

tools/cli/src/mentat_cli/command_parser.rs
fluffyemily: really? I could've sworn I totally rewrote that. 

Looking at it, I rewrote it blow here but didn't delete the old one. I think things might've gone wiggy in the rebasing.

Issue #455, opened 2017-05-09
Accept mentat commands passed into CLI as args


fluffyemily:
Once the CLI has started and the required DB (or temp db) has been started, execute the command immediately and output the result before allowing further input.

Issue #454, opened 2017-05-09
Executed accepted mentat commands


fluffyemily:
If a mentat command has been accepted and is ready for executing, pass the command to mentat db for execution as either transaction or query.

* Output result of execution.
* Output parse errors for incorrectly parsing commands.
* Output execution errors from mentat generated during execution.

PR #466, opened 2017-05-22
Send queries and transactions to mentat and output the results

fluffyemily:
move outputting query and transaction results out of store and into repl

https://github.com/mozilla/mentat/issues/454 & https://github.com/mozilla/mentat/issues/455

depends on https://github.com/mozilla/mentat/pull/465

### DIFF
--- a/tools/cli/src/mentat_cli/lib.rs
+++ b/tools/cli/src/mentat_cli/lib.rs
@@ -42,8 +42,8 @@ pub fn run() -> i32 {
 
     opts.optopt("d", "", "The path to a database to open", "DATABASE");
     opts.optflag("h", "help", "Print this help message and exit");
-    opts.optopt("q", "query", "Execute a query on startup. Queries are executed after any transacts.", "QUERY");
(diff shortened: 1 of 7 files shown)


tools/cli/src/mentat_cli/lib.rs
```diff
@@ -42,6 +42,8 @@ pub fn run() -> i32 {
 
     opts.optopt("d", "", "The path to a database to open", "DATABASE");
     opts.optflag("h", "help", "Print this help message and exit");
+    opts.optopt("q", "query", "Execute a query on startup. Queries are executed after any transacts.", "QUERY");
```
ncalexan: Is this better than having folks pipe from a file or heredoc?  Is there a reason to not accept `-t ... -q ...` in any order and just evaluate them in order?  I see that you're building a set of initial commands without regard to the type of the command anyway.

fluffyemily: Yeah, I was kinda lazy about figuring out how to read the `getopts` flags in sequence. I should probably not be lazy.

Issue #456, opened 2017-05-09
Execute mentat commands from a file


fluffyemily:
* Accept file path as argument to CLI tool. 
* Implement `.read` command that takes a filepath as an argument
* open and read file.
* execute all commands inside file after required DB has opened.
* Do not accept further input until all commands have finished executing.
* Stop execution of commands at first error generated by a read command.

Question: Should any executed transactions from the file be retracted if an error occurs?

rnewman:
Fixed in #553 for transact.

PR #467, opened 2017-05-26
Read & execute transactions from file

fluffyemily:
https://github.com/mozilla/mentat/issues/456

Depends on https://github.com/mozilla/mentat/issues/454.

Tests to come...

### DIFF
--- a/db/src/schema.rs
+++ b/db/src/schema.rs
@@ -245,6 +245,7 @@ impl SchemaTypeChecking for Schema {
                 // Ref coerces a little: we interpret some things depending on the schema as a Ref.
                 (&ValueType::Ref, TypedValue::Long(x)) => Ok(TypedValue::Ref(x)),
                 (&ValueType::Ref, TypedValue::Keyword(ref x)) => self.require_entid(&x).map(|entid| TypedValue::Ref(entid)),
+                (&ValueType::Instant, TypedValue::Instant(x)) => Ok(TypedValue::Instant(x)),
                 // Otherwise, we have a type mismatch.
                 (value_type, _) => bail!(ErrorKind::BadEDNValuePair(value.clone(), value_type.clone())),
             }
--- a/query-projector/src/lib.rs
+++ b/query-projector/src/lib.rs
@@ -116,6 +116,29 @@ impl QueryResults {
             &FindRel(_)    => Box::new(|| QueryResults::Rel(vec![])),
         }
     }
+
+    pub fn to_pretty(&self) -> Result<String> {
+        use QueryResults::*;
+        match self {
+            &Scalar(Some(ref val)) => { 
+                Ok(val.value_type().to_edn_value().to_pretty(120).unwrap_or(String::new()) )
+            },
+            &Tuple(Some(ref vals)) => { 
+                Ok(vals.iter().map(|val| format!("{}\t", val.value_type().to_edn_value().to_pretty(120).unwrap_or(String::new()))).collect::<Vec<String>>().into_iter().collect::<String>())
+            },
+            &Coll(ref vals) => { 
+                Ok(vals.iter().map(|val| format!("{}\n", val.value_type().to_edn_value().to_pretty(120).unwrap_or(String::new()))).collect::<Vec<String>>().into_iter().collect::<String>())
+            },
+            &Rel(ref valsvec) => { 
+                let mut output = String::new();
+                for vals in valsvec {
+                    output.push_str(&format!("{}\n", &vals.iter().map(|val| format!("{}\t", val.value_type().to_edn_value().to_pretty(120).unwrap_or(String::new()))).collect::<Vec<String>>().into_iter().collect::<String>()));
+                }
+                Ok(output)
+            },
+            _ => Ok("No results found.".to_string())
+        }
+    }
 }
 
 type Index = i32;            // See rusqlite::RowIndex.
--- a/tools/cli/src/mentat_cli/command_parser.rs
+++ b/tools/cli/src/mentat_cli/command_parser.rs
@@ -59,16 +59,17 @@ impl Command {
     /// false is returned if the command is not considered valid. 
     /// Defaults to true for all commands except Query and Transact.
     /// TODO: for query and transact commands, they will be considered complete if a parsable EDN has been entered as an argument
-    pub fn is_complete(&self) -> bool {
+    pub fn is_complete(&self) -> (bool, Option<edn::ParseError>) {
         match self {
             &Command::Query(ref args) |
             &Command::Transact(ref args) => {
-                edn::parse::value(&args).is_ok()
+                let r = edn::parse::value(&args);
+                (r.is_ok(), r.err())
             },
             &Command::Help(_) |
             &Command::Open(_) |
             &Command::Close |
-            &Command::Read(_) => true
+            &Command::Read(_) => (true, None)
         }
     }
 
@@ -155,7 +156,8 @@ pub fn command(s: &str) -> Result<Command, cli::Error> {
                     .with(arguments())
                     .map(|args| {
                         // strip quotes from file paths.
-                        let mut sargs = Vec::with_capacity(args.len());
+                        // not sure how to map this and still throw the error so doing it the old fashioned way
+                        let mut files = Vec::with_capacity(args.len());
                         for arg in args.iter() {
                             let start_char = arg.chars().nth(0);
                             match start_char {
@@ -163,19 +165,20 @@ pub fn command(s: &str) -> Result<Command, cli::Error> {
                                 Some('\'') => { 
                                     let separator = start_char.unwrap();
                                     if arg.ends_with(separator) {
-                                        sargs.push(arg.split(separator).collect::<Vec<&str>>().into_iter().collect());
+                                        files.push(arg.split(separator).collect::<Vec<&str>>().into_iter().collect());
                                     } else {
                                         return Err(cli::ErrorKind::CommandParse(format!("Unrecognized argument {}", arg)).into());
                                     }
                                 },
-                                _ => sargs.push(arg.clone()),
+                                _ => files.push(arg.clone()),
                             }
                         }
-                        println!("args: {:?}", sargs);
+
+                        // check that we have at least one argument
                         if args.len() == 0 {
                             return Err(cli::ErrorKind::CommandParse("Missing required argument".to_string()).into());
                         }
-                        Ok(Command::Read(sargs.clone()))
+                        Ok(Command::Read(files.clone()))
                     });
 
     spaces()
--- a/tools/cli/src/mentat_cli/command_parser.rs
+++ b/tools/cli/src/mentat_cli/command_parser.rs
@@ -42,6 +42,7 @@ pub static LONG_QUERY_COMMAND: &'static str = &"query";
 pub static SHORT_QUERY_COMMAND: &'static str = &"q";
 pub static LONG_TRANSACT_COMMAND: &'static str = &"transact";
 pub static SHORT_TRANSACT_COMMAND: &'static str = &"t";
+pub static READ_COMMAND: &'static str = &"read";
 
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub enum Command {
@@ -49,6 +50,7 @@ pub enum Command {
     Query(String),
     Help(Vec<String>),
     Open(String),
+    Read(Vec<String>),
     Close,
 }
 
@@ -65,7 +67,8 @@ impl Command {
             },
             &Command::Help(_) |
             &Command::Open(_) |
-            &Command::Close => true
+            &Command::Close |
+            &Command::Read(_) => true
         }
     }
 
@@ -85,6 +88,9 @@ impl Command {
             }
             &Command::Close => {
                 format!(".{}", CLOSE_COMMAND)
+            },
+            &Command::Read(ref args) => {
+                format!(".{} {:?}", READ_COMMAND, args)
             },
         }
     }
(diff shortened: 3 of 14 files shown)


tools/cli/src/mentat_cli/command_parser.rs
```diff
@@ -59,16 +59,17 @@ impl Command {
     /// false is returned if the command is not considered valid. 
     /// Defaults to true for all commands except Query and Transact.
     /// TODO: for query and transact commands, they will be considered complete if a parsable EDN has been entered as an argument
-    pub fn is_complete(&self) -> bool {
+    pub fn is_complete(&self) -> (bool, Option<edn::ParseError>) {
```
ncalexan: I think I understand why you're not using `Result<bool, edn::ParseError>` here.

tools/cli/src/mentat_cli/command_parser.rs
```diff
@@ -157,27 +158,29 @@ pub fn command(s: &str) -> Result<Command, cli::Error> {
...
                                         return Err(cli::ErrorKind::CommandParse(format!("Unrecognized argument {}", arg)).into());
                                     }
                                 },
-                                _ => sargs.push(arg.clone()),
+                                _ => files.push(arg.clone()),
                             }
                         }
-                        println!("args: {:?}", sargs);
+
+                        // check that we have at least one argument
```
ncalexan: nit: full sentence comments throughout.

tools/cli/src/mentat_cli/command_parser.rs
```diff
@@ -157,27 +158,29 @@ pub fn command(s: &str) -> Result<Command, cli::Error> {
                     .with(arguments())
                     .map(|args| {
                         // strip quotes from file paths.
-                        let mut sargs = Vec::with_capacity(args.len());
+                        // not sure how to map this and still throw the error so doing it the old fashioned way
+                        let mut files = Vec::with_capacity(args.len());
                         for arg in args.iter() {
```
ncalexan: This is usually written `for arg in &args`, since `(&args).into_iter()` is equivalent to `args.iter()`.

tools/cli/src/mentat_cli/input.rs
```diff
@@ -102,13 +102,20 @@ impl InputReader {
...
+                        bail!(err)
+                    }
+                }
             }
         };
 
         match cmd {
             Command::Query(_) |
-            Command::Transact(_) if !cmd.is_complete() => {
+            Command::Transact(_) if !cmd.is_complete().0 => {
```
ncalexan: Your `let (command, _) = ...` below is much more readable.

tools/cli/src/mentat_cli/repl.rs
```diff
@@ -198,10 +207,51 @@ impl Repl {
...
+            }
+        }
+    }
+
+    fn read_file(&mut self, file: String) -> Result<(), cli::Error> {
+        let path = Path::new(&file);
+        let display = path.display();
+
+        let f = match File::open(&path) {
+            Err(err) => bail!(cli::ErrorKind::FileError(display.to_string(), err.to_string())),
```
ncalexan: Don't turn everything to strings.  It's hard to pattern match strings; it's possible (although not always easy) to pattern match structured data.  I understand that this can be difficult with `error-chain`, but if that's so, ask for help!

tools/cli/src/mentat_cli/repl.rs
```diff
@@ -198,10 +207,51 @@ impl Repl {
...
+            Ok(file) => file,
+        };
+
+        let mut buffer = String::new();
+        let mut cmd_err: Option<edn::ParseError> = None;
+
+        let file = BufReader::new(&f);
```
ncalexan: You have the whole file, correct?  Why not just ask for the input you really want: a vector of transactions, which is really a vector-of-vectors-of-assertions.  Then you can just use EDN to parse the whole thing and not monkey with this incremental parsing at all.  That is, accept:
```edn
[[[:db/add ...]
   [:db/remove]]
 [{:db/add ...}
  {:db/add ...}]]
```

db/src/schema.rs
```diff
@@ -245,6 +245,7 @@ impl SchemaTypeChecking for Schema {
                 // Ref coerces a little: we interpret some things depending on the schema as a Ref.
                 (&ValueType::Ref, TypedValue::Long(x)) => Ok(TypedValue::Ref(x)),
                 (&ValueType::Ref, TypedValue::Keyword(ref x)) => self.require_entid(&x).map(|entid| TypedValue::Ref(entid)),
+                (&ValueType::Instant, TypedValue::Instant(x)) => Ok(TypedValue::Instant(x)),
```
ncalexan: Huh, didn't realize this would be so smooth.  Awesome!

Consider rewriting this `match` to fail the next time we add ` ValueType` -- since this was silently busted for a while.

ncalexan: Also, this commit seems to have other stuff in it, not related to `:db.type/inst`.

tools/cli/src/mentat_cli/repl.rs
fluffyemily: Do you mean that, if we have two separate transactions in a single file, i.e.

```
[[:db/add ...]
   [:db/remove]]
[{:db/add ...}
  {:db/add ...}]
```, I should wrap them all in a single vector and execute as a single transaction?

Comment (ncalexan) [inline tools/cli/src/mentat_cli/repl.rs]: No, I think you should make the user wrap the transactions in an outer vector.  Or make your parser accept `vector().of_exactly(many1(transaction())).or(transaction())` so that it accepts a vector of transactions or a single transaction.

File: tools/cli/src/mentat_cli/repl.rs [hunk 8]
```diff
@@ -43,6 +44,7 @@ lazy_static! {
         map.insert(ALT_QUERY_COMMAND, "Shortcut for `.query`. Execute a query against the current open database.");
         map.insert(TRANSACT_COMMAND, "Execute a transact against the current open database.");
         map.insert(ALT_TRANSACT_COMMAND, "Shortcut for `.transact`. Execute a transact against the current open database.");
+        map.insert(READ_COMMAND, "Read in the file provided in argument. Transact each edn in turn.");
```
fluffyemily: yes

query-projector/src/lib.rs
```diff
@@ -116,6 +116,29 @@ impl QueryResults {
             &FindRel(_)    => Box::new(|| QueryResults::Rel(vec![])),
         }
     }
+
+    pub fn to_pretty(&self) -> Result<String> {
+        use QueryResults::*;
+        match self {
+            &Scalar(Some(ref val)) => { 
+                Ok(val.value_type().to_edn_value().to_pretty(120).unwrap_or(String::new()) )
```
ncalexan: You don't want to ignore the errors; you want to funnel them up into the outer `Result`.    You want something like
```rust
val.value_type().to_edn_value().to_pretty(120)
```
possibly with a `.map_err()` to make the `Result` types compatible.

query-projector/src/lib.rs
```diff
@@ -116,6 +116,29 @@ impl QueryResults {
             &FindRel(_)    => Box::new(|| QueryResults::Rel(vec![])),
         }
     }
+
+    pub fn to_pretty(&self) -> Result<String> {
```
ncalexan: @rnewman not sure you want `QueryResults` to know how to pretty-print itself in this crate.  Speak or hold your piece.  (I don't care if it's here: we already reference EDN.)

Although... we probably want to move the pretty-printing to an optional feature, since it's not something we want to ship in Firefox itself.  So maybe this will have to go in a different crate?

query-projector/src/lib.rs
```diff
@@ -116,6 +116,29 @@ impl QueryResults {
...
+            &Tuple(Some(ref vals)) => { 
+                Ok(vals.iter().map(|val| format!("{}\t", val.value_type().to_edn_value().to_pretty(120).unwrap_or(String::new()))).collect::<Vec<String>>().into_iter().collect::<String>())
+            },
+            &Coll(ref vals) => { 
+                Ok(vals.iter().map(|val| format!("{}\n", val.value_type().to_edn_value().to_pretty(120).unwrap_or(String::new()))).collect::<Vec<String>>().into_iter().collect::<String>())
+            },
+            &Rel(ref valsvec) => { 
+                let mut output = String::new();
+                for vals in valsvec {
+                    output.push_str(&format!("{}\n", &vals.iter().map(|val| format!("{}\t", val.value_type().to_edn_value().to_pretty(120).unwrap_or(String::new()))).collect::<Vec<String>>().into_iter().collect::<String>()));
```
ncalexan: My point was that if you actually had EDN -- a vector of vectors for `Rel` -- then you just pretty print the whole thing!  Now, it sucks to make an owned EDN struct just for printing, but this is just for debug output, right?  This will give far more consistent output.

If you really don't want to produce EDN<, i might be better to do this using `pretty.rs` directly.

In any case, this really doesn't make sense: your tabs _will not line up_ if the leading column doesn't fit within a single tabstop.  If you want to do this, do it properly with the tab writer crate (used already in `db`).

tools/cli/src/mentat_cli/input.rs
```diff
@@ -71,6 +78,44 @@ impl InputReader {
         self.reader.is_some()
     }
 
+    pub fn read_file(&mut self, filename: &String) -> Result<Vec<InputResult>, cli::Error> {
+        let path = Path::new(&filename);
+        let display = path.display();
+
+        let f = match File::open(&path) {
+            Err(err) => bail!(cli::ErrorKind::FileError(display.to_string(), err.to_string())),
```
ncalexan: You define `FileError`, so make it own a `Path` and a `std::io::Error`, and don't convert to strings everywhere!

tools/cli/src/mentat_cli/input.rs
```diff
@@ -71,6 +78,44 @@ impl InputReader {
...
+                buffer.clear();
+                cmd_err = None;
+            } else {
+                cmd_err = err;
+            }
+
+        }
+
+        if let Some(err) = cmd_err {
+            bail!(err);
```
ncalexan: `bail!` is not lexically scoped; so you can `bail!` just like you can `return`.  There's no need to save this potential error and re-throw it.

tools/cli/src/mentat_cli/input.rs
```diff
@@ -71,6 +78,44 @@ impl InputReader {
...
+        let file = BufReader::new(&f);
+        let mut cmd_err: Option<edn::ParseError> = None;
+
+        let mut results: Vec<InputResult> = Vec::new();
+        for line in file.lines() {
+            let l = line.unwrap();
+            println!("{}", l);
+            buffer.push_str(&l);
+            let cmd = Command::Transact(buffer.to_string());
+            let (is_complete, err) = cmd.is_complete();
```
ncalexan: Further to the `bail!` below, and to the comments I made on the previous review: if you had `is_complete -> Result<bool, SomeErrorType>` then you can just say `is_complete()?` and the early return will be taken care of.  (You might need to set up `error-chain` to define the lifts from parser error to cli error, but that's what it's there for.)  In general, manually managing and propagating errors is a sign things are not converging to an idiomatic expression.

tools/cli/src/mentat_cli/repl.rs
```diff
@@ -76,17 +66,27 @@ impl Repl {
...
         loop {
-            let res = input.read_input();
+            let res = if unexecuted_commands.len() > 0 { Ok(unexecuted_commands.remove(0)) } else { input.read_input() };
 
             match res {
+                Ok(MetaCommand(Command::Read(files))) => {
+                    for file in files {
+                        let file_res = input.read_file(&file);
+                        if let Ok(commands) = file_res {
+                            unexecuted_commands.append(&mut commands.clone());
```
ncalexan: I can't imagine you need an `&mut` here, since you are cloning already.

tools/cli/src/mentat_cli/repl.rs
```diff
@@ -76,17 +66,27 @@ impl Repl {
...
         let mut input = InputReader::new();
 
+        let mut unexecuted_commands: Vec<InputResult> = Vec::new();
+
         if let Some(cmds) = startup_commands {
-            for command in cmds.iter() {
-                println!("{}", command.output());
-                self.handle_command(command.clone());
-            }
+            unexecuted_commands.append(&mut cmds.into_iter().map(|c| MetaCommand(c.clone())).collect());
```
ncalexan: `.append(iter.collect())` is equivalent to `.extend(iter)`, I think.  Also, `into_iter()` takes ownership of each of the elements, so you don't need the inner `clone`: just
```rust
unexecuted_commands.extend(cmds.into_iter().map(MetaCommand))
```
should do it, I think.  (Please correct me if I'm wrong: I'll learn something!)

tools/cli/src/mentat_cli/input.rs
```diff
@@ -71,6 +78,44 @@ impl InputReader {
...
+        };
+
+        let mut buffer = String::new();
+
+        let file = BufReader::new(&f);
+        let mut cmd_err: Option<edn::ParseError> = None;
+
+        let mut results: Vec<InputResult> = Vec::new();
+        for line in file.lines() {
+            let l = line.unwrap();
```
rnewman: `file.read_to_string(&mut buffer)` dumps the whole file into `buffer`. No need for all of this line-munging stuff.

tools/cli/src/mentat_cli/input.rs
rnewman: … and try to use `?` (with `map_err` if necessary) instead of `match` on errors.

fluffyemily:
Sorry, I pushed these latest as a WIP and wasn't really ready for a re-review as it was not complete. Am in the process of rebasing everything to make it all clearer.

rnewman:
I think most of this stuff landed or has been obsoleted, so let's close this.

PR #469, opened 2017-06-02
Ground (reworked)

rnewman:
I'm rebasing and reworking #423 prior to moving on to FTS.

### DIFF
--- a/query-algebrizer/Cargo.toml
+++ b/query-algebrizer/Cargo.toml
@@ -16,3 +16,6 @@ path = "../query"
 # Only for tests.
 [dev-dependencies.mentat_query_parser]
 path = "../query-parser"
+
+[dev-dependencies]
+maplit = "0.1"
--- a/query-algebrizer/Cargo.toml
+++ b/query-algebrizer/Cargo.toml
@@ -18,4 +18,5 @@ path = "../query"
 path = "../query-parser"
 
 [dev-dependencies]
+itertools = "0.5"
 maplit = "0.1"
--- a/query-algebrizer/src/clauses/where_fn.rs
+++ b/query-algebrizer/src/clauses/where_fn.rs
@@ -62,7 +62,7 @@ impl ConjoiningClauses {
     /// The conversion depends on, and can fail because of:
     /// - Existing known types of a variable to which this arg will be bound.
     /// - Existing bindings of a variable `FnArg`.
-    fn typed_value_from_arg<'s>(&self, schema: &'s Schema, arg: FnArg, known_types: ValueTypeSet) -> Result<TypedValue> {
+    fn typed_value_from_arg<'s>(&self, schema: &'s Schema, arg: FnArg, known_types: &ValueTypeSet) -> Result<TypedValue> {
         if known_types.is_empty() {
             bail!(ErrorKind::InvalidGroundConstant);
         }
@@ -190,7 +190,7 @@ impl ConjoiningClauses {
 
     /// Constrain the CC to associate the given var with the given ground argument.
     fn apply_ground_var<'s>(&mut self, schema: &'s Schema, var: Variable, arg: FnArg) -> Result<()> {
-        let value = self.typed_value_from_arg(schema, arg, self.known_type_set(&var))?;
+        let value = self.typed_value_from_arg(schema, arg, &self.known_type_set(&var))?;
 
         if let Some(existing) = self.bound_value(&var) {
             if existing != value {
@@ -297,7 +297,7 @@ impl ConjoiningClauses {
                 let values = children.into_iter()
                                         .map(|arg| {
                                             // We need to get conversion errors out.
-                                            let r = self.typed_value_from_arg(schema, arg, known_types.clone());
+                                            let r = self.typed_value_from_arg(schema, arg, &known_types);
                                             if let Ok(ref tv) = r {
                                                 if accumulated_types.insert(tv.value_type()) &&
                                                    !accumulated_types.is_unit() {
@@ -364,10 +364,10 @@ impl ConjoiningClauses {
                             // Convert each item in the row.
                             let r: Result<Vec<TypedValue>>;
                             r = cols.into_iter()
-                                    .zip(template.iter().cloned())        // Cloned so we don't mix borrow and move in filter_map.
-                                    .filter_map(|(col, (keep, known_types))| {
-                                        if keep {
-                                            Some(self.typed_value_from_arg(schema, col, known_types))
+                                    .zip(template.iter())
+                                    .filter_map(|(col, pair)| {
+                                        if pair.0 {
+                                            Some(self.typed_value_from_arg(schema, col, &pair.1))
                                         } else {
                                             None
                                         }
--- a/query-algebrizer/src/clauses/where_fn.rs
+++ b/query-algebrizer/src/clauses/where_fn.rs
@@ -62,7 +62,7 @@ impl ConjoiningClauses {
     /// The conversion depends on, and can fail because of:
     /// - Existing known types of a variable to which this arg will be bound.
     /// - Existing bindings of a variable `FnArg`.
-    fn typed_value_from_arg<'s>(&self, schema: &'s Schema, arg: FnArg, known_types: &ValueTypeSet) -> Result<TypedValue> {
+    fn typed_value_from_arg<'s>(&self, schema: &'s Schema, arg: FnArg, known_types: ValueTypeSet) -> Result<TypedValue> {
         if known_types.is_empty() {
             bail!(ErrorKind::InvalidGroundConstant);
         }
@@ -190,7 +190,7 @@ impl ConjoiningClauses {
 
     /// Constrain the CC to associate the given var with the given ground argument.
     fn apply_ground_var<'s>(&mut self, schema: &'s Schema, var: Variable, arg: FnArg) -> Result<()> {
-        let value = self.typed_value_from_arg(schema, arg, &self.known_type_set(&var))?;
+        let value = self.typed_value_from_arg(schema, arg, self.known_type_set(&var))?;
 
         if let Some(existing) = self.bound_value(&var) {
             if existing != value {
@@ -297,7 +297,7 @@ impl ConjoiningClauses {
                 let values = children.into_iter()
                                         .map(|arg| {
                                             // We need to get conversion errors out.
-                                            let r = self.typed_value_from_arg(schema, arg, &known_types);
+                                            let r = self.typed_value_from_arg(schema, arg, known_types);
                                             if let Ok(ref tv) = r {
                                                 if accumulated_types.insert(tv.value_type()) &&
                                                    !accumulated_types.is_unit() {
@@ -367,7 +367,7 @@ impl ConjoiningClauses {
                                     .zip(template.iter())
                                     .filter_map(|(col, pair)| {
                                         if pair.0 {
-                                            Some(self.typed_value_from_arg(schema, col, &pair.1))
+                                            Some(self.typed_value_from_arg(schema, col, pair.1))
                                         } else {
                                             None
                                         }
--- a/query-algebrizer/src/clauses/where_fn.rs
+++ b/query-algebrizer/src/clauses/where_fn.rs
@@ -50,18 +50,21 @@ use types::{
 
 macro_rules! coerce_to_typed_value {
     ($var: ident, $val: ident, $types: expr, $type: path, $constructor: path) => { {
-        if !$types.contains($type) {
-            Impossible(EmptyBecause::TypeMismatch($var.clone(), $types, ValueTypeSet::of_one($type)))
-        } else {
-            Ok($constructor($val).into())
-        }
+        Ok(if !$types.contains($type) {
+               Impossible(EmptyBecause::TypeMismatch {
+                   var: $var.clone(),
+                   existing: $types,
+                   desired: ValueTypeSet::of_one($type),
+               })
+           } else {
+               Val($constructor($val).into())
+           })
     } }
 }
 
-enum ValueConversionResult {
-    Ok(TypedValue),
+enum ValueConversion {
+    Val(TypedValue),
     Impossible(EmptyBecause),
-    Err(ErrorKind),
 }
 
 /// Conversion of FnArgs to TypedValues.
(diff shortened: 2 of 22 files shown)


query-algebrizer/src/types.rs
```diff
@@ -55,7 +55,7 @@ pub enum ComputedTable {
     },
     NamedValues {
         names: Vec<Variable>,
-        values: Vec<Vec<TypedValue>>,
+        values: Vec<TypedValue>,
```
ncalexan: You'll use `names.len()` as the column count?

query-algebrizer/Cargo.toml
```diff
@@ -6,6 +6,7 @@ workspace = ".."
 [dependencies]
 enum-set = { git = "https://github.com/rnewman/enum-set" }
 error-chain = "0.8.1"
+maplit = "0.1"
```
ncalexan: Can this be a build-only dependency?  Having investigated the path to landing in Firefox more, I think we'll need to stop using some of our dependencies pretty soon to constrain our code size.

query/src/lib.rs
```diff
@@ -35,6 +35,7 @@ extern crate mentat_core;
 
 use std::collections::{
     BTreeSet,
+    HashSet,
```
ncalexan: Fold this down, especially 'cuz the earlier commit did the wrong thing with placeholders (I think).

query/src/lib.rs
```diff
@@ -578,19 +579,39 @@ impl Binding {
...
             &Binding::BindScalar(_) | &Binding::BindColl(_) => true,
             &Binding::BindRel(ref vars) | &Binding::BindTuple(ref vars) => {
-                let acc = BTreeSet::default();
+                let mut acc = HashSet::<Variable>::new();
                 for var in vars {
-                    if !acc.insert(var) {
-                        // It's invalid if there was an equal var already present in the set --
-                        // i.e., we have a duplicate var.
-                        return false;
+                    if let &VariableOrPlaceholder::Variable(ref var) = var {
```
ncalexan: Consider `match`... although I guess `VariableOrPlaceholder` shouldn't grow arms :)

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -106,37 +106,37 @@ impl<F> FnArgInto<F> for FnArg where F: Fn(Variable) -> Result<TypedValue> {
...
+
+    {     // Scoped borrow of matrix.
+        let first_row = &matrix[0..expected_width];
+        column_types = first_row.iter().map(|x| x.value_type()).collect();
 
-    for row in matrix.iter() {
-        for (column_type, value) in column_types.iter().zip(row.into_iter()) {
-            if *column_type != value.value_type() {
-                bail!(ErrorKind::InvalidGroundConstant)
+        for row in matrix.chunks(expected_width) {
```
ncalexan: What does this do when `expected_width` is 0?  I think that's possible, and this might fail.  A needed test case, I think.

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -308,17 +318,17 @@ mod testing {
         let good = vec![vec![TypedValue::Long(10), TypedValue::Boolean(true)],
                         vec![TypedValue::Long(11), TypedValue::Boolean(false)]];
         assert_eq!(reduced_rows_and_column_types(good.clone(), &vec![true, true]).expect("to reduce rows"),
-                   (good.clone(),
+                   (good.iter().flatten().cloned().collect(),
```
ncalexan: Better to put in the values yourself rather than add the dependency (even just for development), I think.

rnewman:
You were right that this was half-done — I added a bunch more commits just now, but haven't yet got all of the tests passing.

My changes now integrate more tightly with bound values and with type tracking. Lots more tests to come.

query-algebrizer/Cargo.toml
rnewman: I think it's macro-only, and it's only 198 LOC. I'm not sure if a macro-only crate can be marked as dev-only… let's find out!

ncalexan: Aye, so we could copy-paste its code into our project.

rnewman: Yes, it can. Good spot!

rnewman:
@ncalexan: take a look at the last 10 or 12 small commits, in conjunction with the new tests in 9c197f4. This isn't perfect — we need to check type codes to be correct — but it's probably good enough to fold and land so I can move on to FTS.

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -351,32 +398,57 @@ impl ConjoiningClauses {
...
                         },
                         _ => bail!(ErrorKind::InvalidGroundConstant),
                     }
                 }
 
+                // Do we have rows? If not, the CC cannot succeed.
+                if matrix.is_empty() {
+                    // We will either have bailed or will have accumulated *something* into the matrix,
+                    // so we can safely unwrap here.
+                    self.mark_known_empty(skipped_all.expect("we skipped for a reason"));
```
ncalexan: nit: "we skipped all rows for a reason".  The error message probably won't include a stack, so let's make the messages distinct.

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -283,24 +308,45 @@ impl ConjoiningClauses {
...
+                if children.is_empty() {
+                    bail!(ErrorKind::InvalidGroundConstant);
+                }
+
                 // Turn a collection of arguments into a Vec of `TypedValue`s of the same type.
                 let known_types = self.known_type_set(&var);
                 // Check that every value has the same type.
                 let mut accumulated_types = ValueTypeSet::default();
+                let mut skip: Option<EmptyBecause> = None;
                 let values = children.into_iter()
```
ncalexan: This comment might have gotten put in the wrong place; I'll try to make sure it's in the correct place as well.  Go Github!

query-algebrizer/tests/ground.rs
```diff
@@ -0,0 +1,286 @@
...
+    add_attribute(&mut schema, 69, Attribute {
+        value_type: ValueType::Long,
+        multival: false,
+        ..Default::default()
+    });
+    schema
+}
+
+fn bails(schema: &Schema, input: &str) -> Error {
+    let parsed = parse_find_string(input).expect("parse failed");
```
ncalexan: nit: I feel like these should be `.expect("FOO to not have BAR")`.

query-algebrizer/tests/ground.rs
```diff
@@ -0,0 +1,286 @@
...
+fn test_ground_coll_fails_if_all_impossible() {
+    let q = r#"[:find ?x :where [?x :foo/knows ?p] [(ground [5.1 5.2]) [?p ...]]]"#;
+    let schema = prepopulated_schema();
+    let cc = alg(&schema, &q);
+    assert!(cc.empty_because.is_some());
+}
+
+#[test]
+fn test_ground_rel_skips_impossible() {
+    let q = r#"[:find ?x :where [?x :foo/knows ?p] [(ground [[8 "foo"] [5 7] [9.95 9] [11 12]]) [[?x ?p]]]]"#;
```
ncalexan: This is pretty nifty.  Not necessary if you force homogeneous columns, but nifty none-the-less.

query-algebrizer/tests/ground.rs
```diff
@@ -0,0 +1,286 @@
...
+    let schema = prepopulated_schema();
+    let cc = alg(&schema, &q);
+    assert!(cc.empty_because.is_none());
+    assert_eq!(cc.bound_value(&Variable::from_valid_name("?x")), Some(TypedValue::Ref(8)));
+    assert_eq!(cc.bound_value(&Variable::from_valid_name("?p")), Some(TypedValue::Ref(3)));
+}
+
+#[test]
+fn test_ground_rel_placeholders() {
+    let q = r#"[:find ?x :where [?x :foo/knows ?p] [(ground [[8 "foo" 3] [5 false 7] [5 9.95 9]]) [[?x _ ?p]]]]"#;
```
ncalexan: Can you ensure we reject all placeholders?  Not sure I got that earlier.

query-algebrizer/tests/ground.rs
```diff
@@ -0,0 +1,286 @@
...
+        },
+        _ => {
+            panic!();
+        },
+    }
+}
+
+#[test]
+fn test_ground_rel_heterogeneous_types() {
+    let q = r#"[:find ?x :where [?x _ ?v] [(ground [[false][5]]) [[?v]]]]"#;
```
ncalexan: nit: space between `[false]` and `[5]`.

query-translator/tests/translate.rs
```diff
@@ -672,12 +672,15 @@ fn test_compound_with_ground() {
 
 #[test]
 fn test_unbound_attribute_with_ground() {
-    // TODO: this needs to expand the type code. #475.
-    let query = r#"[:find ?x :where [?x _ ?v] (not [(ground 5) ?v])]"#;
+    let query = r#"[:find ?x ?v :where [?x _ ?v] (not [(ground 17) ?v])]"#;
```
ncalexan: Can you assert this does the right thing if it's `?e` in the entity place, so that the ground is interpreted as a ref?  Or verify that it _doesn't_ do the right thing (the ambiguous choice we made earlier, perhaps)?

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -42,28 +41,39 @@ use super::QualifiedAlias;
...
         if !$types.contains($type) {
-            bail!(ErrorKind::InvalidGroundConstant);
+            Impossible(EmptyBecause::TypeMismatch($var.clone(), $types, ValueTypeSet::of_one($type)))
+        } else {
+            Ok($constructor($val).into())
         }
     } }
 }
 
+enum ValueConversionResult {
```
rnewman: >  It "abuses" the `bail!` macro to return `ValueConversionResult::Err` instead of `std::result::Result::Err!` This surprised me greatly, and is (to my eye) a weakness in the Rust macro system.

Hah, it's a surprise to me, too! I hadn't even noticed that I'd left `bail!` in there, and that it was working.

I had originally implemented this with additional error cases, but that got really hairy: everywhere one might use `?` would end up as a `match` to turn some kinds of errors into `Ok` after processing.

I'll see if making this a `Result` is reasonable, and if not I'll take your other suggestion.

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -140,24 +152,23 @@ impl ConjoiningClauses {
 
             // These are all straightforward.
             FnArg::Constant(NonIntegerConstant::Boolean(x)) => {
-                check_type_set!(known_types, ValueType::Boolean);
-                Ok(TypedValue::Boolean(x).into())
+                if !known_types.contains(ValueType::Boolean) {
```
rnewman: Because that's the one I used as a template for the macro 😆

query-algebrizer/tests/ground.rs
rnewman: I still do require homogeneous columns.

These two features are overlapping: we quietly eliminate rows that can't generate bindings, for whatever reason (out-of-range value, type mismatch against the rest of the query, ident doesn't resolve). If we're left with a column that has more than one possible type — e.g., if there's no type imposed by the rest of the query — then we bail.

We bail because we haven't implemented this feature: we have no way of representing a set of bindings of different types for a single variable. We quietly fail when a query is valid but can't return results in _this_ store with _this_ schema.

query-translator/tests/translate.rs
rnewman: You mean "if it's `?v` in the entity place", @ncalexan?

rnewman:
```
#[test]
fn test_unbound_attribute_with_ground_entity() {
    let query = r#"[:find ?x ?v :where [?x _ ?v] (not [(ground 17) ?x])]"#;
    let schema = prepopulated_schema();
    let SQLQuery { sql, .. } = translate(&schema, query);
    assert_eq!(sql, "SELECT DISTINCT `all_datoms00`.e AS `?x`, \
                                     `all_datoms00`.v AS `?v`, \
                                     `all_datoms00`.value_type_tag AS `?v_value_type_tag` \
                     FROM `all_datoms` AS `all_datoms00` \
                     WHERE NOT EXISTS (SELECT 1 WHERE `all_datoms00`.e = 17)");
}
```

added and passes.

query-algebrizer/src/types.rs
```diff
@@ -440,6 +443,10 @@ impl Debug for EmptyBecause {
     fn fmt(&self, f: &mut Formatter) -> ::std::fmt::Result {
         use self::EmptyBecause::*;
         match self {
+            &ConflictingBindings(ref var, ref existing, ref desired) => {
+                write!(f, "Var {:?} can't be bound to both {:?} and {:?}",
```
rnewman: Rephrased to 

```
                write!(f, "Var {:?} can't be {:?} because it's already bound to {:?}",
```

ncalexan:
> You mean "if it's ?v in the entity place", @ncalexan?

Something like that:
```rust
let query = r#"[:find ?x ?e :where [?e _ ?x] (not [(ground 17) ?v])]"#; 
```

query-algebrizer/src/clauses/where_fn.rs
rnewman: This was reasonable!

rnewman:
```
let query = r#"[:find ?x ?e :where [?e _ ?x] (not [(ground 17) ?v])]"#; 
```

That's actually an invalid query, because the `not` refers to a variable not present in the query. I added:

```

#[test]
fn test_ground_nonexistent_variable_invalid() {
    let q = r#"[:find ?x ?e :where [?e _ ?x] (not [(ground 17) ?v])]"#; 
    let schema = prepopulated_schema();
    let e = bails(&schema, &q);
    match e {
        Error(ErrorKind::UnboundVariable(PlainSymbol(v)), _) => {
            assert_eq!(v, "?v".to_string());
        },
        _ => {
            panic!();
        },
    }
}
```

— happy to add more tests if you'd like.

ncalexan:
> let query = r#"[:find ?x ?e :where [?e _ ?x] (not [(ground 17) ?v])]"#;

Type -- I wanted `(not [(ground 17) ?e])`

rnewman:
Ah, yes, I already have that:

```
#[test]
fn test_unbound_attribute_with_ground_entity() {
    let query = r#"[:find ?x ?v :where [?x _ ?v] (not [(ground 17) ?x])]"#;
    let schema = prepopulated_schema();
    let SQLQuery { sql, .. } = translate(&schema, query);
    assert_eq!(sql, "SELECT DISTINCT `all_datoms00`.e AS `?x`, \
                                     `all_datoms00`.v AS `?v`, \
                                     `all_datoms00`.value_type_tag AS `?v_value_type_tag` \
                     FROM `all_datoms` AS `all_datoms00` \
                     WHERE NOT EXISTS (SELECT 1 WHERE `all_datoms00`.e = 17)");
}
```

rnewman:
8ec24f0

PR #470, opened 2017-06-06
Handle :attribute/_reverse in transactor. Fixes #187.

### LINKED ISSUE #187: [tx] Handle _reverse attribute assertions (back references)
An assertion of the form
```edn
[[:db/add IDENT :db/_ref-attr ENTID]]
```
is equivalent to the assertion
```edn
[[:db/add ENTID :db/ref-attr IDENT]]
```
That is, this allows to add back-references.  It's particularly useful using the map shorthand of #180: for example, one can add and install a new attribute into the `:db.part/db` using:
```edn
[{:db/id #id-literal[-1]
    :db/ident :my/ident
    :db/valueType :db.type/long
    :db.install/_attribute :db.part/db}]
```

ncalexan:
See commit comments.  I've intentionally not accepted reverse notation with nested vectors and maps; it's difficult to handle with our current parse-then-apply model.  I think it's a reasonable trade-off, and the error message is clear.

### DIFF
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -2170,6 +2170,27 @@ mod tests {
                         "{\"s\" 65542
                           \"t\" 65543}");
 
+        // Check that we can use the same attribute in both forward and backward form in the same
+        // transaction.
+        let report = assert_transact!(conn, "[[:db/add 888 :test/dangling 889]
+                                              [:db/add 888 :test/_dangling 889]]");
+        assert_matches!(conn.last_transaction(),
+                        "[[888 :test/dangling 889 ?tx true]
+                          [889 :test/dangling 888 ?tx true]
+                          [?tx :db/txInstant ?ms ?tx true]]");
+        assert_matches!(tempids(&report),
+                        "{}");
+
+        // Check that we can use the same attribute in both forward and backward form in the same
+        // transaction in map notation.
+        let report = assert_transact!(conn, "[{:db/id 998 :test/dangling 999 :test/_dangling 999}]");
+        assert_matches!(conn.last_transaction(),
+                        "[[998 :test/dangling 999 ?tx true]
+                          [999 :test/dangling 998 ?tx true]
+                          [?tx :db/txInstant ?ms ?tx true]]");
+        assert_matches!(tempids(&report),
+                        "{}");
+
         // Verify that we can't explode direct reverse notation with nested value maps.
         assert_transact!(conn,
                          "[[:db/add \"t\" :test/_dangling {:test/many 11}]]",
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -2087,4 +2087,127 @@ mod tests {
         // dangling, if we give a :db/id explicitly.
         assert_transact!(conn, "[{:test/dangling {:db/id \"t\" :test/many 11}}]");
     }
+
+    #[test]
+    fn test_explode_reversed_notation() {
+        let mut conn = TestConn::default();
+
+        // Start by installing a few attributes.
+        assert_transact!(conn, "[[:db/add 111 :db/ident :test/many]
+                                 [:db/add 111 :db/valueType :db.type/long]
+                                 [:db/add 111 :db/cardinality :db.cardinality/many]
+                                 [:db/add 222 :db/ident :test/component]
+                                 [:db/add 222 :db/isComponent true]
+                                 [:db/add 222 :db/valueType :db.type/ref]
+                                 [:db/add 333 :db/ident :test/unique]
+                                 [:db/add 333 :db/unique :db.unique/identity]
+                                 [:db/add 333 :db/index true]
+                                 [:db/add 333 :db/valueType :db.type/long]
+                                 [:db/add 444 :db/ident :test/dangling]
+                                 [:db/add 444 :db/valueType :db.type/ref]]");
+
+        // Check that we can explode direct reversed notation, entids.
+        let report = assert_transact!(conn, "[[:db/add 100 :test/_dangling 200]]");
+        assert_matches!(conn.last_transaction(),
+                        "[[200 :test/dangling 100 ?tx true]
+                          [?tx :db/txInstant ?ms ?tx true]]");
+        assert_matches!(tempids(&report),
+                        "{}");
+
+        // Check that we can explode direct reversed notation, idents.
+        let report = assert_transact!(conn, "[[:db/add :test/many :test/_dangling :test/unique]]");
+        assert_matches!(conn.last_transaction(),
+                        "[[333 :test/dangling :test/many ?tx true]
+                          [?tx :db/txInstant ?ms ?tx true]]");
+        assert_matches!(tempids(&report),
+                        "{}");
+
+        // Check that we can explode direct reversed notation, tempids.
+        let report = assert_transact!(conn, "[[:db/add \"s\" :test/_dangling \"t\"]]");
+        assert_matches!(conn.last_transaction(),
+                        "[[65537 :test/dangling 65536 ?tx true]
+                          [?tx :db/txInstant ?ms ?tx true]]");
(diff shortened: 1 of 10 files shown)


db/src/tx.rs
```diff
@@ -245,157 +348,133 @@ impl<'conn, 'a> Tx<'conn, 'a> {
...
-                                });
-                            }
+                                for (inner_a, inner_v) in map_notation {
+                                    if inner_a.is_backward() {
+                                        // We definitely have a reference.  The reference might be
+                                        // dangling (a bare entid, for example), but we don't yet
+                                        // support nested maps and reverse notation simultaneously
+                                        // (i.e., we don't accept {:reverse/_attribute {:nested map}})
+                                        // so we don't need to check that the nested map reference isn't
+                                        // danglign.
```
rnewman: typo

db/src/tx.rs
```diff
@@ -245,157 +348,133 @@ impl<'conn, 'a> Tx<'conn, 'a> {
...
+                                        dangling = false;
+
+                                        let reversed_e = in_process.entity_v_into_term_e(inner_v, &inner_a)?;
+                                        let (reversed_a, _reversed_attribute) = in_process.entity_a_into_term_a(inner_a.to_reversed().unwrap())?;
```
rnewman: You reverse `inner_a` immediately inside `entity_v_into_term_e`, and again here, and you `unwrap` here reliant on `is_backward` above.

I think you might be better off having a function `unreverse` alongside `is_backward` and `to_reversed` that returns  `Option<Entid>`. Use that with `if let` instead of `is_backward`. That gives you the reversed ident immediately, does the work only once, and gets rid of the `unwrap` here.

db/src/tx.rs
```diff
@@ -245,157 +348,133 @@ impl<'conn, 'a> Tx<'conn, 'a> {
...
-                        entmod::AtomOrLookupRefOrVectorOrMapNotation::LookupRef(lookup_ref) => {
-                            if attribute.value_type != ValueType::Ref {
-                                bail!(ErrorKind::NotYetImplemented(format!("Cannot resolve value lookup ref for attribute {} that is not :db/valueType :db.type/ref", a)))
-                            }
-
-                            Either::Right(LookupRefOrTempId::LookupRef(intern_lookup_ref(&mut lookup_refs, lookup_ref)?))
-                        },
+                    if a.is_backward() {
+                        let reversed_e = in_process.entity_v_into_term_e(v, &a)?;
+                        let (reversed_a, _reversed_attribute) = in_process.entity_a_into_term_a(a.to_reversed().unwrap())?;
```
rnewman: Same point here about reversing. I think if you switch these two `is_backward`/`to_reversed` pairs, you can get rid of `to_reversed`.

db/src/db.rs
```diff
@@ -2087,4 +2087,127 @@ mod tests {
...
+                                 [:db/add 222 :db/valueType :db.type/ref]
+                                 [:db/add 333 :db/ident :test/unique]
+                                 [:db/add 333 :db/unique :db.unique/identity]
+                                 [:db/add 333 :db/index true]
+                                 [:db/add 333 :db/valueType :db.type/long]
+                                 [:db/add 444 :db/ident :test/dangling]
+                                 [:db/add 444 :db/valueType :db.type/ref]]");
+
+        // Check that we can explode direct reversed notation, entids.
+        let report = assert_transact!(conn, "[[:db/add 100 :test/_dangling 200]]");
```
rnewman: For completeness, can we have an example of reversed and non-reversed use of an attribute in the same tx?

```
[[:db/add 100 :test/_dangling 200]
 [:db/add 200 :test/dangling 101]]
```

or something like that.

db/src/tx.rs
```diff
@@ -51,6 +51,7 @@ use std::collections::{
     BTreeSet,
     VecDeque,
 };
+use std::rc::Rc;
```
rnewman: Newline before.

ncalexan: I haven't been doing this elsewhere, including in this file, so I'm going to keep it this way.  Happy to change everywhere if we're committed.

I have been putting lines between *sections*, i.e., between `std::`, crates in this workspace, and modules in this crate.

db/src/tx.rs
```diff
@@ -191,46 +194,146 @@ impl<'conn, 'a> Tx<'conn, 'a> {
...
+                }
             }
 
-            let lr_typed_value: TypedValue = self.schema.to_typed_value(&lookup_ref.v, &lr_attribute)?;
-            Ok(lookup_refs.intern((lr_a, lr_typed_value)))
-        };
+            fn intern_lookup_ref(&mut self, lookup_ref: &entmod::LookupRef) -> Result<LookupRef> {
+                let lr_a: i64 = match lookup_ref.a {
+                    entmod::Entid::Entid(ref a) => *a,
+                    entmod::Entid::Ident(ref a) => self.schema.require_entid(&a)?,
```
ncalexan: Yeah, sorry.  I think it'll be better for you to rebase on this than vice-versa.

db/src/tx.rs
ncalexan: Fixed!

db/src/db.rs
ncalexan: Done, in direct and map notation.

db/src/tx.rs
ncalexan: 've done a version of this, and simplified `.._into_term_a` at the same time.

ncalexan:
> Is it possible to additionally handle the simple cases — perhaps just [:db/add a b c] — in the parser?
> 
> My reasoning is that we can reject obviously invalid input, like [:db/add 123 :foo/_bar 9.95], much earlier in the process, just by looking at the bare EDN — without reference to the schema, without parsing the rest of the input, and before taking locks, hopping into queues, etc.

I did the entire thing in the parsing layer, and decided against it.  The hard thing is that we split the tempid space into internal (allocated by Mentat) and external (strings allocated by the user), and there's no easy way to represent internal tempids in the `edn::Value` space.  That adds a lot of book-keeping, roughly equivalent to the dynamic approach but bulkier, which I didn't like.  And the parsing layer produces worse errors than the dynamic transacting layer (paradoxically).
 
Anyway, I cannibalized by original parsing approach just for the simple direct notation, as you say.  Sadly, it's hard to get the error messages to be informative... but it's definitely more efficient.  See what you think about the last commit.

tx-parser/src/lib.rs
```diff
@@ -106,18 +121,23 @@ def_matches_namespaced_keyword!(Tx, literal_db_retract, "db", "retract");
...
-                op: op,
-                e: e,
-                a: a,
-                v: v,
-            }
-        }))
+          try((Tx::entid_or_lookup_ref_or_temp_id(),
+               Tx::forward_entid(),
+               Tx::atom_or_lookup_ref_or_vector()))
+          .or(try((Tx::atom_or_lookup_ref_or_vector(),
```
rnewman: I'm suspicious of the cost of `try`, especially when the two patterns here are so similar, and the disambiguation is often done by the second term — in the success case of a backward attribute you always parse the entity twice.

The code's already written, but I would probably have done this by matching the most general pattern and then filtering. Are my instincts wrong?

ncalexan: I think you're insticts are correct.  I thought for a short time about this, and I don't know how to achieve better with `combine`; and the first line of the `combine` README says it's LL(1) unless you opt-in to more lookahead with `try`, so this is the blessed way.

The best way might be to parse `:db/{add,retract} edn::Value :attribute/{forward,backward} edn::Value` and then to parse the two `edn::Value` pieces into the correct generalities.  Otherwise you don't get the restriction on the entid location that you want in the parser, which is how we reject mal-formed input early.

My expectation is that `try` is really pretty fast in this situation, especially 'cuz backward entities are rare.

PR #471, opened 2017-06-07
Dedupe SQL arguments.

rnewman:
This isn't perfect — we still need to clone in a couple of cases — but it avoids us passing duplicate strings down into SQLite whenever the same value is mentioned more than once in a query.

Suggestions for improvement most welcome!

### DIFF
--- a/sql/src/lib.rs
+++ b/sql/src/lib.rs
@@ -17,11 +17,14 @@ extern crate mentat_core;
 
 use std::rc::Rc;
 
+use std::collections::HashMap;
+
 use ordered_float::OrderedFloat;
 
 use mentat_core::{
     ToMicros,
     TypedValue,
+    Uuid,
 };
 
 pub use rusqlite::types::Value;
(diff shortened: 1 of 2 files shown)


sql/src/lib.rs
```diff
@@ -190,9 +219,26 @@ impl QueryBuilder for SQLiteQueryBuilder {
     }
 
     fn finish(self) -> SQLQuery {
+        // We collected string and UUID arguments into separate maps so that we could
+        // dedupe them. Now we need to turn them into rusqlite Values.
+        let mut args = self.args;
+        let string_args = self.string_args.into_iter().map(|(val, arg)| {
+             (arg, Rc::new(rusqlite::types::Value::Text(val.as_ref().clone())))
+        });
+        let uuid_args = self.uuid_args.into_iter().map(|(val, arg)| {
```
ncalexan: Might as well make this `bytes_args` and have the caller handle how to get bytes from their data.  We're likely to store more than just UUIDs as bytes.

rnewman:
> The only change to the approach I might suggest is to name the arguments by the hash of the values, but I think the list of cons is longer than the list of pros.

I thought about this, but decided that (a) having heterogeneous values would be more challenging, (b) we want to keep the value anyway, and (c) hashes can collide.

rnewman:
634b7a8

PR #472, opened 2017-06-07
Ground rebased

ncalexan:
@rnewman I couldn't tell what the overall picture was with your work on top of my original work, so I rebased into this simpler sequence.  Can you adopt this?  It's just re-ordering and coalescing your change sequence, no additional changes.

### DIFF
--- a/query-algebrizer/src/clauses/where_fn.rs
+++ b/query-algebrizer/src/clauses/where_fn.rs
@@ -62,7 +62,7 @@ impl ConjoiningClauses {
     /// The conversion depends on, and can fail because of:
     /// - Existing known types of a variable to which this arg will be bound.
     /// - Existing bindings of a variable `FnArg`.
-    fn typed_value_from_arg<'s>(&self, schema: &'s Schema, arg: FnArg, known_types: ValueTypeSet) -> Result<TypedValue> {
+    fn typed_value_from_arg<'s>(&self, schema: &'s Schema, arg: FnArg, known_types: &ValueTypeSet) -> Result<TypedValue> {
         if known_types.is_empty() {
             bail!(ErrorKind::InvalidGroundConstant);
         }
@@ -190,7 +190,7 @@ impl ConjoiningClauses {
 
     /// Constrain the CC to associate the given var with the given ground argument.
     fn apply_ground_var<'s>(&mut self, schema: &'s Schema, var: Variable, arg: FnArg) -> Result<()> {
-        let value = self.typed_value_from_arg(schema, arg, self.known_type_set(&var))?;
+        let value = self.typed_value_from_arg(schema, arg, &self.known_type_set(&var))?;
 
         if let Some(existing) = self.bound_value(&var) {
             if existing != value {
@@ -297,7 +297,7 @@ impl ConjoiningClauses {
                 let values = children.into_iter()
                                         .map(|arg| {
                                             // We need to get conversion errors out.
-                                            let r = self.typed_value_from_arg(schema, arg, known_types.clone());
+                                            let r = self.typed_value_from_arg(schema, arg, &known_types);
                                             if let Ok(ref tv) = r {
                                                 if accumulated_types.insert(tv.value_type()) &&
                                                    !accumulated_types.is_unit() {
@@ -364,10 +364,10 @@ impl ConjoiningClauses {
                             // Convert each item in the row.
                             let r: Result<Vec<TypedValue>>;
                             r = cols.into_iter()
-                                    .zip(template.iter().cloned())        // Cloned so we don't mix borrow and move in filter_map.
-                                    .filter_map(|(col, (keep, known_types))| {
-                                        if keep {
-                                            Some(self.typed_value_from_arg(schema, col, known_types))
+                                    .zip(template.iter())
+                                    .filter_map(|(col, pair)| {
+                                        if pair.0 {
+                                            Some(self.typed_value_from_arg(schema, col, &pair.1))
                                         } else {
                                             None
                                         }
--- a/query-algebrizer/src/clauses/where_fn.rs
+++ b/query-algebrizer/src/clauses/where_fn.rs
@@ -1,0 +1,486 @@
+// Copyright 2016 Mozilla
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+use std::rc::Rc;
+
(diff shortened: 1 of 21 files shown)


query-algebrizer/src/errors.rs
```diff
@@ -42,6 +48,21 @@ error_chain! {
...
+            display("invalid binding for {}: {:?}.", function, binding_error)
+        }
+
+        GroundBindingsMismatch {
+            description("mismatched bindings in ground")
+            display("mismatched bindings in ground")
+        }
+
+        InvalidGroundConstant {
+            // TODO: flesh this out.
```
ncalexan: File follow-up to enrich all of these error types.  I have the same process to come in the transactor, boo :(

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -0,0 +1,486 @@
...
+    } }
+}
+
+/// Conversion of FnArgs to TypedValues.
+impl ConjoiningClauses {
+    /// Convert the provided `FnArg` to a `TypedValue`.
+    /// The conversion depends on, and can fail because of:
+    /// - Existing known types of a variable to which this arg will be bound.
+    /// - Existing bindings of a variable `FnArg`.
+    fn typed_value_from_arg<'s>(&self, schema: &'s Schema, arg: FnArg, known_types: ValueTypeSet) -> Result<TypedValue> {
```
ncalexan: `known_types` is read-only, I think, so it can be taken by reference.

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -0,0 +1,486 @@
...
+        if known_types.is_empty() {
+            bail!(ErrorKind::InvalidGroundConstant);
+        }
+
+        match arg {
+            // This is potentially ambiguous.
```
ncalexan: Prefer explicit over "this", perhaps: "Numeric values are potentially ambiguous: they might be long or ref values."

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -0,0 +1,486 @@
...
+            FnArg::EntidOrInteger(x) => {
+                match (ValueType::Ref.accommodates_integer(x),
+                       known_types.contains(ValueType::Ref),
+                       known_types.contains(ValueType::Long)) {
+                    (true, true, true) => {
+                        // Ambiguous: this arg could be an entid or a long.
+                        // We default to long.
```
ncalexan: This I don't understand. What's the complete solution rather than defaulting? Do we need to narrow type sets?  What impact might this have?

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -0,0 +1,486 @@
...
+                if expected_width == 0 {
+                    // They can't all be placeholders.
+                    bail!(ErrorKind::InvalidGroundConstant);
+                }
+
+                // Accumulate values into `matrix` and types into `a_t_f_c`.
+                // This representation of a rectangular matrix is more efficient than one composed
+                // of N separate vectors.
+                let mut matrix = Vec::with_capacity(expected_width * expected_rows);
+                let mut accumulated_types_for_columns: Vec<ValueTypeSet> =
```
ncalexan: There's a version of `vec!` that takes an element and a count; it should be both clearer and more efficient than this.

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -0,0 +1,486 @@
...
+                    match row {
+                        FnArg::Vector(cols) => {
+                            // Make sure that every row is the same length.
+                            if cols.len() != full_width {
+                                bail!(ErrorKind::InvalidGroundConstant);
+                            }
+
+                            // Convert each item in the row.
+                            let r: Result<Vec<TypedValue>>;
+                            r = cols.into_iter()
```
ncalexan: This expression is awkward.  I think you'd do better to use the turbofish and `?`directly here, rather than staging it.

query-algebrizer/src/clauses/where_fn.rs
```diff
@@ -0,0 +1,486 @@
...
+                                    .zip(template.iter().cloned())        // Cloned so we don't mix borrow and move in filter_map.
```
ncalexan: It shouldn't be necessary to clone.  I'll try to improve this locally, since it looks like you tried without success already.

ncalexan: See https://github.com/mozilla/mentat/pull/472/commits/a347bca3f7454a7a4e00b5d87b1bd19d258e0fbe.

rnewman:
@ncalexan, what was the last commit in the series when you rebased? My last commit in my branch is a5f02c07f30e3ac4e934539d88956a55046f0ae3.

query-algebrizer/src/clauses/where_fn.rs
rnewman: There are four stages we can apply.

One is during algebrizing, which I've implemented here: if we know a type is impossible (-1 can't be an entid; `:nope` isn't an ident), then we can narrow down to an individual type, and there's no ambiguity.

The second is also during algebrizing: if we know that `?x` will later be used as an attribute, then we don't need to treat it as a keyword. This can be confusing: after all, one could write a query like:

```
[:find … :where [(ground :foo/bar) ?x] [?a ?x 15] [?a :some/keyword ?x]]
```

— there `?x` can be the same EDN value used in two different places as two different "purpose types" (ref and keyword).

This is part-implemented: it uses the current state of the CC. We would need to do more than one pass through the input patterns in order to determine types more generally.

The third is to defer this decision until the latest possible moment. That's not really feasible: there's no "last", because we would need to iteratively converge.

The last is to defer resolution of user input until the query is underway: to generate the appropriate mess of `UNION` expressions to allow the value to pass as either type.

Rather than doing that, I've followed your approach: if you enter `[(ground 10) ?foo]`, then you'll get a `long`… unless somewhere else on the query unambiguously requires `?foo` to be an entity.

This particular bit of code says: the value is in range of being a ref, and nothing else in the query has yet told us that it's one or the other, so we just pick one.

query-algebrizer/src/clauses/where_fn.rs
rnewman: Yes, but it's an `EnumSet`, which generates way faster code when copied than when passed by reference.

query-algebrizer/src/clauses/where_fn.rs
rnewman: I recall I did it this way because the compiler wasn't able to infer the type correctly when I smashed it all together. I'll try again.

rnewman:
I moved this back to #469.

query-algebrizer/src/errors.rs
rnewman: #473.

query-algebrizer/src/clauses/where_fn.rs
ncalexan: Ah, I see.  You'll need to roll this back from my follow-up comment and throw in a clone, perhaps.

rnewman: Already did!

PR #477, opened 2017-06-12
Fulltext search

rnewman:
This is currently a rebase and cleanup of #402. It now reflects changes I made while revving `ground`, so it builds and passes tests.

I did a fair bit of rebasing to take the opportunity to split up `where_fn.rs`; that refactoring commit contains no other changes, unless I screwed something up.

This isn't quite ready yet: I replaced a bunch of `bail!` with `unimplemented!`, because I want users to be able to unify inputs with fulltext-bound variables, and otherwise have `fulltext` clauses behave consistently with pattern clauses.


rnewman:
I think this is ready for review, @ncalexan. I kept my work separate; see the commits _after_ "implement fulltext".

Significant changes:

- I fixed constant floats.
- I allowed implicit placeholders in bindings.
- I reused a bunch of existing pattern code.
- I made bound places unify.

query-algebrizer/src/clauses/fulltext.rs
```diff
@@ -0,0 +1,310 @@
...
+            FnArg::IdentOrKeyword(i) => schema.get_entid(&i),
+            // Must be an entid.
+            FnArg::EntidOrInteger(e) => Some(e),
+            _ => None,
+        };
+
+        let a = a.ok_or(ErrorKind::InvalidArgument(where_fn.operator.clone(), "attribute".into(), 1))?;
+        let attribute = schema.attribute_for_entid(a).cloned().ok_or(ErrorKind::InvalidArgument(where_fn.operator.clone(), "attribute".into(), 1))?;
+
+        let fulltext_values = DatomsTable::FulltextValues;
```
ncalexan: nit: is there value to these copies?  I doubt it.

query-algebrizer/src/clauses/fulltext.rs
```diff
@@ -0,0 +1,310 @@
...
+        }
+
+        if let VariableOrPlaceholder::Variable(ref var) = bindings[3] {
+            if self.bound_value(&var).is_some() || self.input_variables.contains(&var) {
+                // It's OK for this to collide: the two must unify.
+                unimplemented!();
+            }
+            self.constrain_var_to_type(var.clone(), ValueType::Double);
+
+            // Right now we don't support binding a column to a constant value.  Therefore, we
```
ncalexan: You have now done the work for `ground` to produce exactly the constant value, all the way out of the projector.  Can you translate this to a `(ground ?score 0.0)` rather than injecting a computed table into the mix?

query-algebrizer/src/clauses/mod.rs
```diff
@@ -338,7 +338,25 @@ impl ConjoiningClauses {
 impl ConjoiningClauses {
     /// Be careful with this. It'll overwrite existing bindings.
     pub fn bind_value(&mut self, var: &Variable, value: TypedValue) {
-        self.constrain_var_to_type(var.clone(), value.value_type());
+        let vt = value.value_type();
```
ncalexan: nit: `vt` doesn't have value here.

query-algebrizer/src/clauses/mod.rs
```diff
@@ -338,7 +338,25 @@ impl ConjoiningClauses {
...
+        // If so, generate a constraint against the primary column.
+        if let Some(vec) = self.column_bindings.get(var) {
+            if let Some(col) = vec.first() {
+                self.wheres.add_intersection(ColumnConstraint::Equals(col.clone(), QueryValue::TypedValue(value.clone())));
+            }
+        }
+
+        // Are we also trying to figure out the type of the value when the query runs?
+        // If so, constrain that!
+        if let Some(table) = self.extracted_types.get(&var)
```
ncalexan: Just a style point: I find the `.map(... clone())` odd -- I'd generally say `if let Some(ref table) ... table.clone()`.

sql/src/lib.rs
```diff
@@ -160,7 +160,14 @@ impl QueryBuilder for SQLiteQueryBuilder {
             &Ref(entid) => self.push_sql(entid.to_string().as_str()),
             &Boolean(v) => self.push_sql(if v { "1" } else { "0" }),
             &Long(v) => self.push_sql(v.to_string().as_str()),
-            &Double(OrderedFloat(v)) => self.push_sql(v.to_string().as_str()),
+            &Double(OrderedFloat(v)) => {
+                // Rust's floats print without a trailing '.' in some cases.
```
ncalexan: TIL.  Is there a potential loss of resolution?  That is, I'm wondering if `a.bcdefEx` and `abcdef.0` might not represent different IEEE floats?  I believe the resolution for very large and very small numbers is different, and "naive" parsing changes very large to very small (with high resolution) floats, but I have really not thought this through.

Would https://github.com/rust-lang/rust/issues/30967#issuecomment-172315537 (`{:.1}`) address the loss of resolution?

query-algebrizer/src/clauses/fulltext.rs
```diff
@@ -102,15 +95,27 @@ impl ConjoiningClauses {
...
         let a = match args.next().unwrap() {
             FnArg::IdentOrKeyword(i) => schema.get_entid(&i),
             // Must be an entid.
             FnArg::EntidOrInteger(e) => Some(e),
+            FnArg::Variable(v) => {
+                // If it's already bound, then let's expand the variable.
+                self.bound_value(&v).and_then(|tv| {
+                    if let TypedValue::Ref(entid) = tv {
+                        Some(entid)
+                    } else {
```
ncalexan: You might want to differentiate a non-`ValueType::Ref` from the errors below; your call.

query-algebrizer/src/clauses/fulltext.rs
```diff
@@ -152,64 +169,46 @@ impl ConjoiningClauses {
...
-                values: vec![TypedValue::Double(0.0.into())],
-            };
-            let table = self.computed_tables.push_computed(named_values);
-            let alias = self.next_alias_for_table(table);
-
-            // Stitch the computed table into column_bindings, so we get cross-linking.
-            self.bind_column_to_var(schema, alias.clone(), VariableColumn::Variable(var.clone()), var.clone());
-
-            self.from.push(SourceAlias(table, alias.clone()));
+            self.bind_value(var, TypedValue::Double(0.0.into()));
```
ncalexan: Ah, very nice.  I never found the expression for this idea while working on these patches.

query-translator/tests/translate.rs
```diff
@@ -741,28 +741,91 @@ fn test_fulltext() {
...
+#[test]
+fn test_fulltext_inputs() {
+    let schema = prepopulated_typed_schema(ValueType::String);
+
+    // Bind ?entity. We expect the output to collide.
+    let query = r#"[:find ?val
+                    :in ?entity
+                    :where [(fulltext $ :foo/bar "hello") [[?entity ?val _ _]]]]"#;
+    let SQLQuery { sql, args } = translate(&schema, query);
+    assert_eq!(sql, "SELECT DISTINCT `fulltext_values00`.text AS `?val` \
```
ncalexan: Is this test correct?  This doesn't seem to be any different from the query without `:in ?entity`.   Can you spell out what's changed here, or compare the two (with and without the `:in`) in the test?

query-algebrizer/tests/fulltext.rs
```diff
@@ -0,0 +1,102 @@
...
+fn alg(schema: &Schema, input: &str) -> ConjoiningClauses {
+    let parsed = parse_find_string(input).expect("query input to have parsed");
+    algebrize(schema.into(), parsed).expect("algebrizing to have succeeded").cc
+}
+
+#[test]
+fn test_apply_fulltext() {
+    let schema = prepopulated_schema();
+
+    // If you use a non-FTS attribute, we will short-circuit.
```
ncalexan: Not a coding error, like you had for a missing attribute?  I think I prefer to favor the static consume" with great error messages.

tests/query.rs
```diff
@@ -254,3 +254,40 @@ fn test_instants_and_uuids() {
...
+        [:db/add "s" :db/valueType :db.type/string]
+        [:db/add "s" :db/fulltext true]
+        [:db/add "s" :db/cardinality :db.cardinality/many]
+    ]"#).unwrap();
+    let v = conn.transact(&mut c, r#"[
+        [:db/add "v" :foo/fts "hello darkness my old friend"]
+        [:db/add "v" :foo/fts "I've come to talk with you again"]
+    ]"#).unwrap().tempids.get("v").cloned().expect("v was mapped");
+
+    let r = conn.q_once(&mut c,
```
ncalexan: A helper that turns `QueryResults` into `EDN` wouldn't go amiss here...

tests/query.rs
```diff
@@ -290,4 +299,60 @@ fn test_fulltext() {
...
+    match r {
+        Err(Error(ErrorKind::QueryError(mentat_query_algebrizer::ErrorKind::InvalidArgument(PlainSymbol(s), ty, i)), _)) => {
+            assert_eq!(s, "fulltext");
+            assert_eq!(ty, "string");
+            assert_eq!(i, 2);
+        },
+        _ => panic!("Expected query to fail."),
+    }
+
+    // If it's bound, and the right type, it'll work!
```
ncalexan: Nifty!

query-algebrizer/src/clauses/fulltext.rs
rnewman: This is removed in a later commit: we no longer use a computed table for score.

sql/src/lib.rs
rnewman: I did some experimentation in a Rust playground for this.

Using `{:.1}` doesn't behave as you'd expect:

https://is.gd/8QaoD5

```
fn main() {
    let i: f64 = 123.0f64;                     // Integer.
    let v: f64 = 9999.00123f64;
    let x: f64 = 9.99900123e3f64;
    let z: f64 = 1.000123f64;
    println!(":: {:e} vs {:.1} vs {}", i, i, i);
    println!(":: {:e} vs {:.1} vs {}", x, x, x);
    println!(":: {:e} vs {:.1} vs {}", v, v, v);
    println!(":: {:e} vs {:.1} vs {}", z, z, z);
        
    let s = "9.99900123e3";
    let parsed = s.parse::<f64>().unwrap();
    println!("Parsed to {:e}, {:.1}, {}", parsed, parsed, parsed);
    println!("Same: {}, {}", parsed == v, parsed == x);
}
```

=>

```
:: 1.23e2 vs 123.0 vs 123
:: 9.99900123e3 vs 9999.0 vs 9999.00123
:: 9.99900123e3 vs 9999.0 vs 9999.00123
:: 1.000123e0 vs 1.0 vs 1.000123
Parsed to 9.99900123e3, 9999.0, 9999.00123
Same: true, true
```

As you can see, the only solution that round-trips, never prints as an integer, and always preserves precision is `{:e}`.

query-translator/tests/translate.rs
rnewman: The later version of the test expands this into three or four tests: `q_once` is the bit that validates whether you supplied enough inputs.

query-algebrizer/tests/fulltext.rs
rnewman: I'm trying to draw a line between detecting coding errors and allowing queries to adapt to different schema. Most queries should not produce an error when run: they're valid Datalog, they just don't match the data in the store.

I imagine a future in which we provide a logging or reporting channel that includes things like "this query is known to return no results because attribute A isn't present in the schema".

rnewman:
dd39f6df5b90e3e0a0ee2a20608a09839e442830

Issue #385, opened 2017-03-22
[query] Allow later constraints to affect earlier patterns' table choices


rnewman:
We naïvely alias tables as patterns are processed: as soon as we see

```edn
[?x _ ?y]
```

we decide to query `all_datoms`, because the attribute is unknown, and we alias that to something like `all_datoms00`.

This is expedient: we can process a collection of patterns mutably, one by one, and at the end be almost ready to produce SQL.

However, if the next pattern is:

```edn
[(< ?y 10)]
```

we _know_ that for this CC to produce results `?y` must be numeric. Armed with that knowledge, we know we can use `datoms` instead of `all_datoms`, which will yield a more efficient query.

As noted in https://github.com/mozilla/mentat/pull/374#issuecomment-288278950, in order to do this we need to do one of three things:

- Delay table choice until we're ready to finish the CC. This can require full descent: the top-level CC might know that `?y` is numeric because both arms of a nested `or-join` are numeric. That gets a little complicated if we wish to do this right.
- Do a multi-pass algebrizing step, where we collect some type knowledge before accumulating constraints and aliases. I'm not sure how feasible this is yet.
- Re-alias as we go.

rnewman:
Note that this issue isn't massively important: most queries will have bound attributes, which unambiguously determines the type of the value.

rnewman:
We recently made some progress on this, pre-processing input patterns prior to application. It doesn't cover the example above, but it's better than it was.

PR #478, opened 2017-06-14
Type tag expansion

fluffyemily:
https://github.com/mozilla/mentat/issues/293

This works. It is horrible though. Am working on seeing if I can make it better. But it does work.
N.B. Commits will get squashed  before merge.

### DIFF
--- a/query-algebrizer/src/clauses/mod.rs
+++ b/query-algebrizer/src/clauses/mod.rs
@@ -293,7 +293,7 @@ impl ConjoiningClauses {
                     value_bindings: values,
                     ..Default::default()
                 };
-
+                
                 // Pre-fill our type mappings with the types of the input bindings.
                 cc.known_types
                   .extend(types.iter()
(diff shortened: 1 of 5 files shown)


query-translator/tests/translate.rs
```diff
@@ -304,7 +304,7 @@ fn test_numeric_not_equals_known_attribute() {
     let schema = prepopulated_typed_schema(ValueType::Long);
     let query = r#"[:find ?x . :where [?x :foo/bar ?y] [(!= ?y 12)]]"#;
     let SQLQuery { sql, args } = translate(&schema, query);
-    assert_eq!(sql, "SELECT `datoms00`.e AS `?x` FROM `datoms` AS `datoms00` WHERE `datoms00`.a = 99 AND `datoms00`.v <> 12 LIMIT 1");
+    assert_eq!(sql, "SELECT `datoms00`.e AS `?x` FROM `datoms` AS `datoms00` WHERE `datoms00`.a = 99 AND `datoms00`.v <> 12 AND `datoms00`.value_type_tag = 5 LIMIT 1");
```
rnewman: Your code is doing too much work.

We know that `?y` is in the range of `:foo/bar` — all possible bindings of `?y` have a single type, known in advance. We don't need to restrict the type tag: indeed, we've already used the known type to simplify the query.

query-algebrizer/src/clauses/mod.rs
```diff
@@ -825,7 +826,139 @@ impl ConjoiningClauses {
...
+        for (v, value_type_set) in self.known_types.iter() {
+            if self.column_bindings.contains_key(v) {
+                let cols = self.column_bindings.get(v).unwrap().clone();
+                for col in cols {
+                    match col.1 {
+                        Column::Fixed(DatomsColumn::Value) => { 
+                            // attempt to figure out exactly which type is used
+                            // TODO: But only if the attribute is a variable or not present and therefore of unknown type
+                            let mut value_type: Option<ValueType> = None;
+                            for constraint in self.wheres.0.iter() {
```
rnewman: There has to be a better way of doing this. The CC is our own intermediate representation: if we're having to crawl over it to figure out what's happening, then it means we're not tracking something we should be!

query-algebrizer/src/clauses/mod.rs
```diff
@@ -825,7 +826,139 @@ impl ConjoiningClauses {
...
+            for col in cols {
+                if col.0 != tag.0 { 
+                    let value = QueryValue::Column(QualifiedAlias(col.0.clone(), Column::Fixed(DatomsColumn::ValueTypeTag)));
+                    let alias = tag.clone();
+                    self.wheres.0.push(ColumnConstraintOrAlternation::Constraint(ColumnConstraint::Equals(alias, value)));
+                }
+            }
+        }
+
+        let mut computed_types: BTreeMap<Variable, TypedValue> = BTreeMap::default();
```
rnewman: Why `TypedValue` and not `ValueType`?

query-algebrizer/src/clauses/mod.rs
```diff
@@ -825,7 +826,139 @@ impl ConjoiningClauses {
...
+                        computed_types.insert(item.clone(), values[i].clone());
+                    }
+                },
+                _ => (),
+            }
+        }
+
+        let mut empty_reason: Option<EmptyBecause> = None;
+        let mut wheres_to_add = vec![];
+        for (v, value_type_set) in self.known_types.iter() {
```
rnewman: Any variable for which `value_type_set.is_unit()`, or any variable for which all types in the `value_type_set` map to a different SQLite storage type (so strings and doubles are safe to compare, but doubles = longs, and instants = longs), is safe: simply comparing the value is adequate, right?

query-algebrizer/src/clauses/mod.rs
fluffyemily: Ok, the thing I am struggling to figure out is:

In the case of 

```
                [:find ?x ?v
                 :where
                    [?x _ ?v]
                    [(< ?v 10)]]
```

We generate this cc:

```
ConjoiningClauses { 
    empty_because: None, 
    from: [SourceAlias(AllDatoms, all_datoms00)], 
    computed_tables: [], 
    wheres: [Constraint(all_datoms00.v < value(Long(10)))], 
    column_bindings: {var(?v): [all_datoms00.v], var(?x): [all_datoms00.e]},
    input_variables: {}, 
    value_bindings: {}, 
    known_types: {var(?v): ValueTypeSet({Long, Double}), var(?x): ValueTypeSet({Ref})}, 
    extracted_types: {var(?v): all_datoms00.value_type_tag} 
}
```

Without crawling over the `wheres`, how do I figure out that I need to add the type tag for a Long and not a Double?

rnewman: Strictly speaking you don't: `0.5` is a perfectly valid binding for `?v` here. `<` constrains its arguments to be _numeric_, hence `ValueTypeSet({Long, Double})`.

What you _do_ need to do here is ensure that the generated SQL ensures that `all_datoms00.value_type_tag = 5`, which is the code shared by both `Long` and `Double`.

The thing you're looking for is a pattern that could generate a broader range of types than we eventually will accept in the query.

`[?x _ ?v]` can bind `?v` to `ValueTypeSet::any()`.

But we know _later_ that `?v` must be `Long` or `Double`.

There are two ways to tackle this.

The inefficient way: for _every_ column binding, generate a type constraint. This is redundant and expensive.

The better way: figure out which `QualifiedAlias`es will contribute overly broad results and generate a constraint only for those.

You don't really have that information at this point in processing the CC: it's part of applying the pattern. I suggest that we collect it: it's an association between the var, the column, and the types we can infer from the pattern considered alone.

fluffyemily: Question:

In this example:

```
[:find ?x 
 :where [?x ?y true]
        [?z ?y ?x]]
```

we get the cc:

```
ConjoiningClauses { 
    empty_because: None, 
    from: [
        SourceAlias(Datoms, datoms00), 
        SourceAlias(Datoms, datoms01)
    ], 
    computed_tables: [], 
    wheres: [
        Constraint(datoms00.v = value(Boolean(true))), 
        Constraint("datoms00".value_type_tag = Boolean), 
        Constraint(datoms00.e = datoms01.v), 
        Constraint(datoms00.a = datoms01.a)
    ], 
    column_bindings: {
        var(?x): [datoms00.e, datoms01.v], 
        var(?y): [datoms00.a, datoms01.a], 
        var(?z): [datoms01.e]
    }, 
    input_variables: {}, 
    value_bindings: {}, 
    known_types: {
        var(?x): ValueTypeSet({Ref}), 
        var(?y): ValueTypeSet({Ref}), 
        var(?z): ValueTypeSet({Ref})
    }, 
    extracted_types: {} 
}
```

`var(?y)` has a known type of `Ref` but has a `bool` assigned to it. This should cause a `TypeMismatch` `EmptyBecause` to be attached. 

Why is it that the known types for `var(?y)` does not contain `Boolean`? I would have thought we could determine that, but we only determine the `Ref` type. 

Is this a bug? 

If the known types for `var(?y)` included `Boolean` then that would make everything a lot easier in this case...

rnewman: There are three variables in that query.

- `?y` is the attribute, so it's an entity (`ValueType::Ref`).
- `?x` is the 'e' of the first pattern, so it's also a ref.
- `?z` is the 'e' of the second pattern, so it's also a ref.

Processing the rest of the pattern, we see that `?y` is used again in the second pattern in the same way — still ref. `?x` is used again as the 'v' of the second pattern, which doesn't change anything.

We conclude that all of `?x`, `?y`, and `?z` must be refs, and we don't need to project a type from the SQL tables.

The thing we need to conclude at this point, for this issue, is that `datoms01`, constrained by `datoms01.v = datoms00.e AND datoms01.a = datoms00.a`, can yield bindings for `?x` that are not refs, but share a value-space representation with refs: `1` could match `true`, `1970-01-01…00.001`, etc.

As such we need to additionally constrain by type.

query-algebrizer/src/clauses/mod.rs
```diff
@@ -806,26 +807,24 @@ impl ConjoiningClauses {
...
-    /// ```
-    ///
-    /// where `?y` must simultaneously be a ref-typed attribute and a boolean-typed attribute. This
-    /// function deduces that and calls `self.mark_known_empty`. #293.
+    /// For each bound var. discover which ones have 
     pub fn expand_type_tags(&mut self) {
-        // TODO.
+        let mut vars_to_remove = Vec::new();
+        for (v, tag) in self.extracted_types.iter() {
+            if let Some(cols) = self.column_bindings.get(&v) {
```
rnewman: I don't think this is correct.

What we'd like to do is find SQL patterns for which the attribute place is insufficient to determine an unambiguous SQL representation of the value place, and impose additional restrictions.

What you've done here is found variables for which the type isn't known (which is better done by walking `column_bindings` and looking at `self.known_type_set(&var)` — `extracted_types` isn't reliable for this purpose), generating cross-links to other column-uses of the variable.

Sometimes those two things are the same, and sometimes they're not.

We don't need to generate tag comparisons if the potential types don't share a SQL representation. We sometimes _do_ need to generate type code constraints when the type of a place is later known precisely.

For example,

```
[:find …
 [?x :foo/bar ?y]
 [?z _ ?x]]
```

— `?x` is known for sure to have `ValueType::Ref`, but that query translates to

```
SELECT …
FROM datoms AS datoms00, all_datoms AS all_datoms01
WHERE datoms00.a = 99 AND datoms00.e = all_datoms00.v
```

which is incorrect. This is even true when there's no other pattern:

```
[:find …
 [?x _ ?y]
 [(< ?y 10)]]
```

should constrain `?y` to numeric values, type code `5`. (#385 would cover dropping back to `datoms` from `all_datoms` in this case, too.)

My suggestion is this:

As each pattern is applied, we know _within the scope of the particular contributing table_ which value types might be returned: `?e` `?a` and `?tx` will always be refs, and `?v` will, in the naive case, always be either `ValueTypeSet::one_of(range_of_a)` or `ValueTypeSet::any()`. You can imagine a future in which we know more precisely which values might be returned, given what we know about the entity and tx!

At the end of algebrizing, we know one more thing: what types a variable _ends up as_, and where else it is used.

We need to generate a constraint on all columns for which the first type set is broader than the second.

(Sometimes this might be a little redundant, but not often.)

In the unit-type case — where we later find out that a variable is bound to values of a single type — that'll take care of cross-linking, too. Otherwise, we'll restrict the primary and cross-link to the secondaries.

I don't see how you can implement this ticket without tracking the output types of each pattern as you go.

rnewman:
@fluffyemily: I wrote some failing tests to demonstrate.

c185e2fa024bde19ffbfa5d81190fed4d015f17a

PR #480, opened 2017-06-16
Fix microsecond conversion

rnewman:
There are one million microseconds in a second, not one hundred thousand.

Hooray for round-trip tests — they don't catch this kind of bug!

### DIFF
--- a/edn/src/types.rs
+++ b/edn/src/types.rs
@@ -563,7 +563,7 @@ pub trait FromMicros {
 
 impl FromMicros for DateTime<UTC> {
     fn from_micros(ts: i64) -> Self {
-        UTC.timestamp(ts / 100_000, ((ts % 100_000).abs() as u32) * 1_000)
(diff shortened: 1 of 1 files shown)


edn/src/types.rs
```diff
@@ -563,7 +563,7 @@ pub trait FromMicros {
 
 impl FromMicros for DateTime<UTC> {
     fn from_micros(ts: i64) -> Self {
-        UTC.timestamp(ts / 100_000, ((ts % 100_000).abs() as u32) * 1_000)
+        UTC.timestamp(ts / 1_000_000, ((ts % 1_000_000).abs() as u32) * 1_000)
```
fluffyemily: Is it worth defining a constant for `1_000_000` that we can use here so if we need to implement `FromMicros` for another type we are less likely to make the same mistake?

rnewman: Probably not — my reasoning is: we're likely to wrap `DateTime` rather than directly implement `FromMicros`, and if we _do_ do it again we'll probably forget to use the constant!

hashtag-realism hashtag-bugs

PR #481, opened 2017-06-16
Allow instants to pass through schema validation.

rnewman:
We missed a case. Hooray placeholders!


rnewman:
> Channeling @ncalexan's comment when I did this - we should also expand the default case to cover all remaining cases so that we don't miss a case next time!

I thought about doing that… but don't we have a product here of all `ValueType`s and all `TypedValue`s? That's 56 more cases. We could choose to only do `ValueType` _or_ `TypedValue` thoroughly…

rnewman:
@fluffyemily ^ howzat?

PR #482, opened 2017-06-16
Support inequality predicates for instants

### LINKED ISSUE #439: [query] Support range predicates on instants

rnewman:
This issue fixes #439.

#481 and #480 are fixes necessary for this to work correctly.

### DIFF
--- a/query-algebrizer/src/clauses/convert.rs
+++ b/query-algebrizer/src/clauses/convert.rs
@@ -51,6 +51,53 @@ macro_rules! coerce_to_typed_value {
     } }
 }
 
+pub trait ValueTypes {
+    fn potential_types(&self, schema: &Schema) -> Result<ValueTypeSet>;
+}
+
+impl ValueTypes for FnArg {
+    fn potential_types(&self, schema: &Schema) -> Result<ValueTypeSet> {
+        Ok(match self {
+                &FnArg::EntidOrInteger(x) => {
+                    if ValueType::Ref.accommodates_integer(x) {
+                        // TODO: also see if it's a valid entid?
+                        ValueTypeSet::of_longs()
+                    } else {
+                        ValueTypeSet::of_one(ValueType::Long)
+                    }
+                },
+
+                &FnArg::IdentOrKeyword(ref x) => {
+                    if schema.get_entid(x).is_some() {
+                        ValueTypeSet::of_keywords()
+                    } else {
+                        ValueTypeSet::of_one(ValueType::Keyword)
+                    }
+                },
+
+                &FnArg::Variable(_) => {
+                    ValueTypeSet::any()
+                },
+
+                &FnArg::Constant(NonIntegerConstant::BigInteger(_)) => {
+                    // Not yet implemented.
+                    bail!(ErrorKind::UnsupportedArgument)
+                },
+
+                // These don't make sense here. TODO: split FnArg into scalar and non-scalar…
+                &FnArg::Vector(_) |
+                &FnArg::SrcVar(_) => bail!(ErrorKind::UnsupportedArgument),
+
+                // These are all straightforward.
+                &FnArg::Constant(NonIntegerConstant::Boolean(x)) => ValueTypeSet::of_one(ValueType::Boolean),
+                &FnArg::Constant(NonIntegerConstant::Instant(x)) => ValueTypeSet::of_one(ValueType::Instant),
+                &FnArg::Constant(NonIntegerConstant::Uuid(x)) => ValueTypeSet::of_one(ValueType::Uuid),
+                &FnArg::Constant(NonIntegerConstant::Float(x)) => ValueTypeSet::of_one(ValueType::Double),
+                &FnArg::Constant(NonIntegerConstant::Text(_)) => ValueTypeSet::of_one(ValueType::String),
+            })
+    }
+}
+
 pub enum ValueConversion {
     Val(TypedValue),
     Impossible(EmptyBecause),
--- a/query-algebrizer/src/clauses/convert.rs
+++ b/query-algebrizer/src/clauses/convert.rs
@@ -89,10 +89,10 @@ impl ValueTypes for FnArg {
                 &FnArg::SrcVar(_) => bail!(ErrorKind::UnsupportedArgument),
 
                 // These are all straightforward.
-                &FnArg::Constant(NonIntegerConstant::Boolean(x)) => ValueTypeSet::of_one(ValueType::Boolean),
-                &FnArg::Constant(NonIntegerConstant::Instant(x)) => ValueTypeSet::of_one(ValueType::Instant),
-                &FnArg::Constant(NonIntegerConstant::Uuid(x)) => ValueTypeSet::of_one(ValueType::Uuid),
-                &FnArg::Constant(NonIntegerConstant::Float(x)) => ValueTypeSet::of_one(ValueType::Double),
+                &FnArg::Constant(NonIntegerConstant::Boolean(_)) => ValueTypeSet::of_one(ValueType::Boolean),
+                &FnArg::Constant(NonIntegerConstant::Instant(_)) => ValueTypeSet::of_one(ValueType::Instant),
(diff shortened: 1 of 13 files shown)


query-algebrizer/src/types.rs
```diff
@@ -274,20 +274,21 @@ impl From<Order> for OrderBy {
 }
 
 #[derive(Copy, Clone, PartialEq, Eq)]
-/// Define the different numeric inequality operators that we support.
+/// Define the different inequality operators that we support.
 /// Note that we deliberately don't just use "<=" and friends as strings:
 /// Datalog and SQL don't use the same operators (e.g., `<>` and `!=`).
-pub enum NumericComparison {
+/// These are applicable to numbers and instants.
+pub enum Inequality {
```
fluffyemily: Is `Inequality` entirely accurate when there are `LessThanOrEquals` and `GreaterThanOrEquals` as options on the enum?

query-translator/tests/translate.rs
```diff
@@ -888,3 +888,20 @@ fn test_fulltext_inputs() {
                        AND `datoms02`.a = 99");
     assert_eq!(args, vec![make_arg("$v0", "hello"),]);
 }
+
+#[test]
+fn test_instant_range() {
```
fluffyemily: Can we have another test please testing the `self.mark_known_empty(EmptyBecause::NonInstantArgument);` case?

query-algebrizer/src/types.rs
rnewman: Yes; the full technical term is "non-strict inequality".

query-algebrizer/src/clauses/convert.rs
```diff
@@ -51,6 +51,53 @@ macro_rules! coerce_to_typed_value {
...
+                    if ValueType::Ref.accommodates_integer(x) {
+                        // TODO: also see if it's a valid entid?
+                        ValueTypeSet::of_longs()
+                    } else {
+                        ValueTypeSet::of_one(ValueType::Long)
+                    }
+                },
+
+                &FnArg::IdentOrKeyword(ref x) => {
+                    if schema.get_entid(x).is_some() {
```
ncalexan: Maybe a note that interacts with multi-stage algebrizing?  I guess the `&Schema` parameter says that pretty clearly.

query-algebrizer/src/clauses/convert.rs
```diff
@@ -89,10 +89,10 @@ impl ValueTypes for FnArg {
                 &FnArg::SrcVar(_) => bail!(ErrorKind::UnsupportedArgument),
 
                 // These are all straightforward.
-                &FnArg::Constant(NonIntegerConstant::Boolean(x)) => ValueTypeSet::of_one(ValueType::Boolean),
-                &FnArg::Constant(NonIntegerConstant::Instant(x)) => ValueTypeSet::of_one(ValueType::Instant),
-                &FnArg::Constant(NonIntegerConstant::Uuid(x)) => ValueTypeSet::of_one(ValueType::Uuid),
-                &FnArg::Constant(NonIntegerConstant::Float(x)) => ValueTypeSet::of_one(ValueType::Double),
+                &FnArg::Constant(NonIntegerConstant::Boolean(_)) => ValueTypeSet::of_one(ValueType::Boolean),
```
ncalexan: Fold into the previous patch.

query-algebrizer/src/clauses/predicate.rs
```diff
@@ -60,21 +71,36 @@ impl ConjoiningClauses {
...
 
+        // The types we're handling here must be the intersection of the possible types of the arguments,
+        // the known types of any variables, and the types supported by our inequality operators.
+        let left_types = self.potential_types(schema, &left)?;
+        let right_types = self.potential_types(schema, &right)?;
+        let shared_types = left_types.intersection(&right_types)
+                                     .intersection(&comparison.supported_types());
+
+        // We expect the intersection to be Long, Long+Double, Double, or Instant.
+        let left_v;
```
ncalexan: I didn't know Rust supported single-static-assignment with bare `let`.  I wonder if there's a version requirement.

query-algebrizer/src/clauses/predicate.rs
```diff
@@ -60,21 +71,36 @@ impl ConjoiningClauses {
...
+        let right_v;
+        if shared_types == ValueTypeSet::of_one(ValueType::Instant) {
```
ncalexan: Would it not be possible to ask for values that could be long or instants?  Something like:
```edn
[:find ?x ?y :where
 (or
  (and [_ :some/instant ?x]
       [_ :some/other-instant ?y])
  (and [_ :some/long ?x]
       [_ :some/other-long ?y]))]
```

query-algebrizer/src/clauses/resolve.rs
```diff
@@ -65,6 +66,36 @@ impl ConjoiningClauses {
...
+                self.column_bindings
+                    .get(&var)
+                    .and_then(|cols| cols.first().map(|col| QueryValue::Column(col.clone())))
+                    .ok_or_else(|| Error::from_kind(ErrorKind::UnboundVariable(var.name())))
+            },
+            Constant(NonIntegerConstant::Instant(v)) => {
+                Ok(QueryValue::TypedValue(TypedValue::Instant(v)))
+            },
+
+            // TODO: should we allow integers if they seem to be timestamps? It's ambiguous…
```
ncalexan: Generally, I'd prefer Mentat to be strict (with errors rather than empty result sets) in ambiguous cases.

query-algebrizer/tests/predicate.rs
```diff
@@ -0,0 +1,149 @@
...
+fn test_instant_predicates_require_instants() {
+    let schema = prepopulated_schema();
+
+    // You can't use a string for an inequality: this is a straight-up error.
+    let query = r#"[:find ?e
+                    :where
+                    [?e :foo/date ?t]
+                    [(> ?t "2017-06-16T00:56:41.257Z")]]"#;
+    match bails(&schema, query).0 {
+        ErrorKind::InvalidArgument(op, why, idx) => {
```
ncalexan: Not feasible to derive `Eq` or `PartialEq` on `ErrorKind`, eh?  Rats.

query-algebrizer/src/clauses/convert.rs
rnewman: This bit in particular doesn't, except in the manner that _all_ of our queries could fundamentally change behavior if the schema changes!

query-algebrizer/src/clauses/predicate.rs
rnewman: I think what you mean is: can we algebrize a predicate where the types of the values are not known precisely, even if they could be mutually consistent?

I think your example needs a final clause for illustration:

```
[:find ?x ?y :where
 (or
  (and [_ :some/instant ?x]
       [_ :some/other-instant ?y])
  (and [_ :some/long ?x]
       [_ :some/other-long ?y]))
 [(< ?x ?y)]]
```

The answer is: no, at least not yet. We don't maintain possibility trees, only type sets, so the best we know about your query is that `?x` and `?y` can _each_ be _either_ `Instant` or `Long`. That means it's possible — as far as the currently implementation is concerned, anyway — for `?x` to be an `Instant` and `?y` to be a `Long`, which we're not prepared to handle.

The simplest workaround is to push the condition into each branch:


```
[:find ?x ?y :where
 (or
  (and [_ :some/instant ?x]
       [_ :some/other-instant ?y]
       [(< ?x ?y)])
  (and [_ :some/long ?x]
       [_ :some/other-long ?y]
       [(< ?x ?y)]))]
```

which will also likely be more efficient.

The other way of looking at your question is to realize that we have built ourselves this problem: the comparison operators in Clojure, and the comparisons in SQL, are overloaded. If the SQL comparison were `DATE_LESS_THAN` instead of `<`, then you wouldn't even be asking this question!

query-algebrizer/tests/predicate.rs
rnewman: https://github.com/brson/error-chain/issues/95
https://github.com/brson/error-chain/issues/134

We're also waiting for that so we can derive `Sync` and upgrade to the current version of `error_chain`!

rnewman:
eaf3e7f

PR #483, opened 2017-06-19
Simple aggregates

rnewman:
No good errors yet, but I'm happy with the end-to-end tests.

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -18,7 +18,11 @@ extern crate edn;
 
 pub mod values;
 
-use std::collections::BTreeMap;
+use std::collections::{
+    BTreeMap,
+    BTreeSet,
+};
+
 use std::fmt;
 use std::rc::Rc;
 
@@ -63,6 +67,8 @@ pub enum ValueType {
     Keyword,
     Uuid,
 }
+
+pub type ValueTypeTag = i32;
 
 impl ValueType {
     pub fn all_enums() -> EnumSet<ValueType> {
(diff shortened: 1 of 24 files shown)


rnewman:
This still needs work on `GROUP BY` extraction.

rnewman:
Ready for review. This implements the same behavior as Datomic: all variables are used for grouping unless mentioned in an aggregate expression.

Note that I extended `ValueTypeSet` a little as I moved it: it now comes with a handy `SQLValueTypeSet` trait that can answer questions like "do all of these types end up as a single SQL type code?".

Have at, @ncalexan!

query-projector/src/lib.rs
```diff
@@ -183,21 +221,209 @@ fn candidate_type_column(cc: &ConjoiningClauses, var: &Variable) -> (ColumnOrExp
...
+        self.args[0]
+            .as_variable()
+            .and_then(|v| SimpleAggregationOp::for_function(&self.func)
+                              .map(|op| SimpleAggregate { op, var: v.clone(), }))
     }
 }
 
+/// An internal temporary struct to pass between the projection 'walk' and the
+/// resultant projector.
+/// Projection accumulates three things:
```
ncalexan: Looks like the name changed.

query-projector/src/lib.rs
```diff
@@ -183,21 +221,209 @@ fn candidate_type_column(cc: &ConjoiningClauses, var: &Variable) -> (ColumnOrExp
...
+    Min,
+    Sum,
+}
+
+impl SimpleAggregationOp {
+    fn to_sql(&self) -> &'static str {
+        use SimpleAggregationOp::*;
+        match self {
+            &Avg => "avg",
+            &Count => "count",
```
ncalexan: Can we get a test to ensure that "count" isn't "count-distinct"?

query-projector/src/lib.rs
```diff
@@ -243,23 +487,119 @@ fn project_elements<'a, I: IntoIterator<Item = &'a Element>>(
...
+        // If it's a fixed value, we need do nothing further.
+        if query.cc.is_value_bound(&var) {
+            continue;
+        }
+
+        let (column, name) = candidate_column(&query.cc, &var)?;
         cols.push(ProjectedColumn(column, name));
-        if query.cc.known_type(&var).is_none() {
+
+        // We don't care if a column has a single _type_, we care if it has a single type _code_,
```
ncalexan: We have very carefully used "type tag" to be "numeric, Mentat-specific type encoding".  And we've called the `i32` type definition `ValueTypeTag`.  Let's not lose that!  s/code/tag/ everywhere.

query-projector/src/lib.rs
```diff
@@ -243,23 +487,119 @@ fn project_elements<'a, I: IntoIterator<Item = &'a Element>>(
...
+        return Ok(ProjectedElements {
+                      sql_projection: Projection::Columns(cols),
+                      templates,
+                      group_by: None,
+                  });
+    }
+
+    // group_by: (with + variables) - aggregated
+    let mut group_by_vars: BTreeSet<Variable> = query.with.union(&variables).cloned().collect();
+    for var in aggregated {
```
ncalexan: No need to consume `aggregated`; try `&aggegrated`.

query-projector/src/lib.rs
```diff
@@ -243,23 +487,119 @@ fn project_elements<'a, I: IntoIterator<Item = &'a Element>>(
...
+        group_by_vars.remove(&var);
+    }
+
+    // We never need to group by a constant.
```
ncalexan: I guess the number of aggregated vars is similar to the number of bound vars in most instances.  You didn't like the extra allocations in `difference`?

query-projector/src/lib.rs
```diff
@@ -243,23 +487,119 @@ fn project_elements<'a, I: IntoIterator<Item = &'a Element>>(
...
+
+    // Turn this collection of vars into a collection of columns from the query.
+    // Right now we don't allow grouping on anything but a variable bound in the query.
+    // TODO: also group by type tag.
+    let group_by = if group_by_vars.is_empty() {
+        None
+    } else {
+        let mut group_cols = Vec::with_capacity(2 * group_by_vars.len());
+
+        for var in group_by_vars.into_iter() {
```
ncalexan: nit: no need for `into_iter()`.

query-projector/src/lib.rs
```diff
@@ -458,6 +806,10 @@ pub struct CombinedProjection {
 
     /// True if this query requires the SQL query to include DISTINCT.
     pub distinct: bool,
+
+    // An optional list of column names to use as a GROUP BY clause.
+    // Right now these are all `Name`s, and present in the `Projection`.
```
ncalexan: ... but in the future ...?

core/src/lib.rs
```diff
@@ -229,6 +235,178 @@ impl SQLValueType for ValueType {
     }
 }
 
+trait EnumSetExtensions<T: enum_set::CLike + Clone> {
```
rnewman: I expect we'll use them whenever we use an `EnumSet`; they're not related to `ValueType`, but they are useful utilities.

tests/query.rs
```diff
@@ -390,3 +391,275 @@ fn test_instant_range_query() {
...
+        [:db/add "d" :foo/is-vegetarian false]
+        [:db/add "a" :foo/age 14]
+        [:db/add "b" :foo/age 22]
+        [:db/add "c" :foo/age 42]
+        [:db/add "d" :foo/age 28]
+    ]"#).unwrap().tempids;
+
+    // What are the oldest and youngest ages?
+    let r = conn.q_once(&mut c,
+                        r#"[:find [(min ?age) (max ?age)]
```
rnewman: We don't yet support ordering by aggregates. It's relatively straightforward.

query-projector/src/lib.rs
```diff
@@ -183,21 +221,209 @@ fn candidate_type_column(cc: &ConjoiningClauses, var: &Variable) -> (ColumnOrExp
...
+    Max,
+    Min,
+    Sum,
+}
+
+impl SimpleAggregationOp {
+    fn to_sql(&self) -> &'static str {
+        use SimpleAggregationOp::*;
+        match self {
+            &Avg => "avg",
```
rnewman: These aren't keywords; these are functions, which are typically lowercase.

query-projector/src/lib.rs
rnewman: What I really want is in-place set subtraction…

query-projector/src/lib.rs
rnewman: Sadly, it is — needs fixing.

rnewman:
I've landed the 'pre' commits here, and pushed the rest of the work to #498.

PR #485, opened 2017-06-21
Support tx places in queries

fluffyemily:
First pass. 

What this doesn't do is utilise an old use the tx id to fetch results from a previous version. Should that be included in this issue?

### DIFF
--- a/query-algebrizer/src/clauses/mod.rs
+++ b/query-algebrizer/src/clauses/mod.rs
@@ -615,8 +615,10 @@ impl ConjoiningClauses {
     /// present.
     fn constrain_to_tx(&mut self, tx: &PatternNonValuePlace) {
         match *tx {
-            PatternNonValuePlace::Placeholder => (),
-            _ => unimplemented!(),           // TODO: #440.
+            PatternNonValuePlace::Variable(ref v) => self.constrain_var_to_type(v.clone(), ValueType::Ref),
+            PatternNonValuePlace::Placeholder |
+            PatternNonValuePlace::Entid(_) |
+            PatternNonValuePlace::Ident(_) => (),
         }
     }
 
--- a/query-algebrizer/src/clauses/mod.rs
+++ b/query-algebrizer/src/clauses/mod.rs
@@ -611,15 +611,8 @@ impl ConjoiningClauses {
     }
 
     /// Ensure that the given place has the correct types to be a tx-id.
-    /// Right now this is mostly unimplemented: we fail hard if anything but a placeholder is
-    /// present.
     fn constrain_to_tx(&mut self, tx: &PatternNonValuePlace) {
-        match *tx {
-            PatternNonValuePlace::Variable(ref v) => self.constrain_var_to_type(v.clone(), ValueType::Ref),
-            PatternNonValuePlace::Placeholder |
-            PatternNonValuePlace::Entid(_) |
-            PatternNonValuePlace::Ident(_) => (),
-        }
+        self.constrain_to_ref(tx);
     }
 
     /// Ensure that the given place can be an entity, and is congruent with existing types.
--- a/query-algebrizer/src/clauses/pattern.rs
+++ b/query-algebrizer/src/clauses/pattern.rs
@@ -247,9 +247,6 @@ impl ConjoiningClauses {
                 self.bind_column_to_var(schema, col.clone(), DatomsColumn::Tx, v.clone());
             },
             PatternNonValuePlace::Entid(entid) => {
-                // TODO: we want to check whether the tx-id is within range for the database's tx partition. 
-                // (That applies after ident lookup, too.)
-                // Possible solution: https://github.com/mozilla/mentat/tree/fluffyemily/tx-id-check
                 self.constrain_column_to_entity(col.clone(), DatomsColumn::Tx, entid);
             },
             PatternNonValuePlace::Ident(ref ident) => {
--- a/query-algebrizer/src/clauses/pattern.rs
+++ b/query-algebrizer/src/clauses/pattern.rs
@@ -241,14 +241,15 @@ impl ConjoiningClauses {
             },
         }
 
-        // TODO pattern.tx -> like handling entity place 
-        // carefully read Datomic query doc to check for correct handling
         match pattern.tx {
             PatternNonValuePlace::Placeholder => (),
             PatternNonValuePlace::Variable(ref v) => {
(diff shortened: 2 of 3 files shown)


rnewman:
> What this doesn't do is utilise an old use the tx id to fetch results from a previous version. Should that be included in this issue?

No — regular patterns, without a specified database, query just what's currently true.

query-algebrizer/src/clauses/mod.rs
```diff
@@ -615,8 +615,10 @@ impl ConjoiningClauses {
     /// present.
```
rnewman: This comment needs to be changed.

query-algebrizer/src/clauses/mod.rs
```diff
@@ -615,8 +615,10 @@ impl ConjoiningClauses {
...
     fn constrain_to_tx(&mut self, tx: &PatternNonValuePlace) {
         match *tx {
-            PatternNonValuePlace::Placeholder => (),
-            _ => unimplemented!(),           // TODO: #440.
+            PatternNonValuePlace::Variable(ref v) => self.constrain_var_to_type(v.clone(), ValueType::Ref),
+            PatternNonValuePlace::Placeholder |
+            PatternNonValuePlace::Entid(_) |
+            PatternNonValuePlace::Ident(_) => (),
```
rnewman: Consider making this function simply pass through to `constrain_to_ref`.

query-algebrizer/src/clauses/pattern.rs
```diff
@@ -238,8 +238,28 @@ impl ConjoiningClauses {
...
 
+        // TODO pattern.tx -> like handling entity place 
+        // carefully read Datomic query doc to check for correct handling
+        match pattern.tx {
+            PatternNonValuePlace::Placeholder => (),
+            PatternNonValuePlace::Variable(ref v) => {
+                self.bind_column_to_var(schema, col.clone(), DatomsColumn::Tx, v.clone());
+            },
+            PatternNonValuePlace::Entid(entid) => {
+                self.constrain_column_to_entity(col.clone(), DatomsColumn::Tx, entid);
```
rnewman: TODO: we want to check whether the tx-id is within range for the database's `tx` partition. (That applies after ident lookup, too.)

tests/query.rs
```diff
@@ -258,6 +256,59 @@ fn test_instants_and_uuids() {
...
+    let mut c = new_connection("").expect("Couldn't open conn.");
+    let mut conn = Conn::connect(&mut c).expect("Couldn't open DB.");
+    conn.transact(&mut c, r#"[
+        [:db/add "s" :db/ident :foo/uuid]
+        [:db/add "s" :db/valueType :db.type/uuid]
+        [:db/add "s" :db/cardinality :db.cardinality/one]
+    ]"#).unwrap();
+    let t = conn.transact(&mut c, r#"[
+        [:db/add "u" :foo/uuid #uuid "cf62d552-6569-4d1b-b667-04703041dfc4"]
+    ]"#).unwrap();
```
rnewman: Transact *again* here, so we know we're not just accidentally getting the last tx!

tests/query.rs
```diff
@@ -258,6 +256,59 @@ fn test_instants_and_uuids() {
...
+    let r = conn.q_once(&mut c,
+                        r#"[:find ?tx
```
rnewman: You can use `:find ?tx .` — the scalar form — to simplify your results checking.

tests/query.rs
```diff
@@ -258,6 +256,59 @@ fn test_instants_and_uuids() {
...
+    ]"#).unwrap();
+    let t = conn.transact(&mut c, r#"[
+        [:db/add "u" :foo/uuid #uuid "cf62d552-6569-4d1b-b667-04703041dfc4"]
+    ]"#).unwrap();
+    let tx = (Variable::from_valid_name("?tx"), TypedValue::Ref(t.tx_id));
+    let inputs = QueryInputs::with_value_sequence(vec![tx]);
+    let r = conn.q_once(&mut c,
+                        r#"[:find ?uuid 
+                            :in ?tx
+                            :where [?x :foo/uuid ?uuid ?tx]]"#, inputs);
```
rnewman: Again, adding a couple more separate transactions with UUIDs would make this test stronger.

tests/query.rs
```diff
@@ -258,6 +256,59 @@ fn test_instants_and_uuids() {
...
+    match r {
+        Result::Ok(QueryResults::Rel(ref v)) => {
+            assert_eq!(*v, vec![
+                vec![TypedValue::Uuid(Uuid::from_str("cf62d552-6569-4d1b-b667-04703041dfc4").unwrap()),]
```
rnewman: Use `.expect("valid UUID")` instead of `.unwrap()`. (And similar throughout.)

query-algebrizer/src/clauses/pattern.rs
```diff
@@ -238,8 +238,29 @@ impl ConjoiningClauses {
...
 
+        match pattern.tx {
+            PatternNonValuePlace::Placeholder => (),
+            PatternNonValuePlace::Variable(ref v) => {
+                self.bind_column_to_var(schema, col.clone(), DatomsColumn::Tx, v.clone());
+            },
+            PatternNonValuePlace::Entid(entid) => {
+                // TODO: we want to check whether the tx-id is within range for the database's tx partition. 
+                // (That applies after ident lookup, too.)
+                // Possible solution: https://github.com/mozilla/mentat/tree/fluffyemily/tx-id-check
```
fluffyemily: I had a go at achieving this:
https://github.com/mozilla/mentat/compare/fluffyemily/tx-places...fluffyemily/tx-id-check

It feels messy but I couldn't find another way of getting access to the `tx` partition map inside the query callstack.

rnewman: Yeah, that's pretty messy! 😁 

The only way I think it would be less messy is for the CC to have a handle to the DB, not the schema.

Thanks for giving it a shot.

But two things cause me to disagree with myself:

- Not everything is worth checking before executing the query.
- Algebrizing with respect to a schema means that only schema changes invalidate the output. Algebrizing with respect to the partition map means that any entity allocation invalidates the output! That's probably acceptable for `q_once`, but not useful for prepared queries.

rnewman:
Looks great. Undo my comment from the first round of review, and ship it 😁

Issue #458, opened 2017-05-09
Show schema of current db


fluffyemily:
Displays the schema for the database.

* Implement `.schema` command.
* Output is the pretty printed transaction statement that would be used to create this schema.

PR #487, opened 2017-06-22
Show schema of current DB

fluffyemily:
https://github.com/mozilla/mentat/issues/458

### DIFF
--- a/tools/cli/src/mentat_cli/command_parser.rs
+++ b/tools/cli/src/mentat_cli/command_parser.rs
@@ -37,6 +37,7 @@ pub static OPEN_COMMAND: &'static str = &"open";
 pub static CLOSE_COMMAND: &'static str = &"close";
 pub static LONG_QUERY_COMMAND: &'static str = &"query";
 pub static SHORT_QUERY_COMMAND: &'static str = &"q";
+pub static SCHEMA_COMMAND: &'static str = &"schema";
 pub static LONG_TRANSACT_COMMAND: &'static str = &"transact";
 pub static SHORT_TRANSACT_COMMAND: &'static str = &"t";
 pub static LONG_EXIT_COMMAND: &'static str = &"exit";
@@ -49,6 +50,7 @@ pub enum Command {
     Help(Vec<String>),
     Open(String),
     Query(String),
+    Schema,
     Transact(String),
 }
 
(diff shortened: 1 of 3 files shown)


tools/cli/src/mentat_cli/command_parser.rs
```diff
@@ -129,6 +135,15 @@ pub fn command(s: &str) -> Result<Command, cli::Error> {
                         }
                         Ok(Command::Close)
                     });
+                    
+    let schema_parser = string(SCHEMA_COMMAND)
```
fluffyemily: I tried to generalise `no_arg_parser` to take a command as an arg and then use that to determine how the command was parsed, but I couldn't get the lifetimes right. I think that was due to `CLOSE_COMMAND`, `SCHEMA_COMMAND` and `EXIT_COMMAND` all being static strings, which I then also couldn't use in a `match`. I think there is probably some way to get around it that I just haven't figured out yet. Need to spend some more time with the rust book I think. Will file a follow up bug for it.

fluffyemily: https://github.com/mozilla/mentat/issues/488

rnewman: I think this might be an unnecessary rabbit hole.

Take a look at https://docs.rs/combine/2.3.2/combine/fn.not_followed_by.html instead?

tools/cli/src/mentat_cli/command_parser.rs
```diff
@@ -312,8 +328,7 @@ mod tests {
 
     #[test]
     fn test_exit_parser_with_args() {
-        let input = ".exit arg1";
-        let err = command(&input).expect_err("Expected an error");
+        let input = ".exit arg1";let err = command(&input).expect_err("Expected an error");
```
rnewman: Put that newline back!

tools/cli/src/mentat_cli/repl.rs
```diff
@@ -112,6 +114,10 @@ impl Repl {
             },
             Command::Close => self.close(),
             Command::Query(query) => self.execute_query(query),
+            Command::Schema => {
+                let edn = self.store.fetch_schema();
+                println!("{}", edn.to_pretty(120).unwrap());
```
rnewman: Use `expect` instead of `unwrap` — or, better, handle the error!

PR #489, opened 2017-06-23
Atomic multi-tx

rnewman:
This is the basic primitive needed to implement atomic schema migration: the ability to establish a transaction and run queries and transact within it, rolling back or committing in its entirety when done.

This is a long way from being ready — I need to introduce some kind of equivalent of `Conn` that I can query within this, and perhaps maintains mutable state rather than invoking `transact` directly — but I wanted to get to a POC first.

### DIFF
--- a/src/conn.rs
+++ b/src/conn.rs
@@ -30,6 +30,8 @@ use mentat_db::{
     TxReport,
 };
 
+use mentat_tx;
+
 use mentat_tx_parser;
 
 use errors::*;
@@ -81,6 +83,74 @@ pub struct Conn {
     // the schema changes. #315.
 }
 
+pub struct InProgress<'a, 'c> {
+    transaction: rusqlite::Transaction<'c>,
+    mutex: &'a Mutex<Metadata>,
+    generation: u64,
+    partition_map: PartitionMap,
+    schema: Schema,
+}
+
+impl<'a, 'c> InProgress<'a, 'c> {
+    fn transact_entities<I>(mut self, entities: I) -> Result<(InProgress<'a, 'c>, TxReport)> where I: IntoIterator<Item=mentat_tx::entities::Entity> {
+        let (report, next_partition_map, next_schema) = transact(&self.transaction, self.partition_map, &self.schema, &self.schema, entities)?;
+        self.partition_map = next_partition_map;
+        if let Some(schema) = next_schema {
+            self.schema = schema;
+        }
+        Ok((self, report))
+    }
+
+    /// Query the Mentat store, using the given connection and the current metadata.
+    pub fn q_once<T>(&self,
+                     query: &str,
+                     inputs: T) -> Result<QueryResults>
+        where T: Into<Option<QueryInputs>>
+        {
+
+        q_once(&*(self.transaction),
+               &self.schema,
+               query,
+               inputs)
+    }
+
+    fn transact(self, transaction: &str) -> Result<(InProgress<'a, 'c>, TxReport)> {
+        let assertion_vector = edn::parse::value(transaction)?;
+        let entities = mentat_tx_parser::Tx::parse(&assertion_vector)?;
+        self.transact_entities(entities)
+    }
+
+    fn rollback(self) -> Result<()> {
+        self.transaction.rollback().map_err(|e| e.into())
+    }
+
+    fn commit(self) -> Result<()> {
+        {
+            // The mutex is taken during this block.
+            let mut metadata = self.mutex.lock().unwrap();
+
+            if self.generation != metadata.generation {
+                // Somebody else wrote!
+                // Retrying is tracked by https://github.com/mozilla/mentat/issues/357.
+                // This should not occur -- an attempt to take a competing IMMEDIATE transaction
+                // will fail with `SQLITE_BUSY`, causing this function to abort.
+                bail!("Lost the transact() race!");
+            }
+
+            // Commit the SQLite transaction while we hold the mutex.
+            self.transaction.commit()?;
+
+            metadata.generation += 1;
+            metadata.partition_map = self.partition_map;
+            if self.schema != *(metadata.schema) {
+                metadata.schema = Arc::new(self.schema);
+            }
+        }
+
+        Ok(())
+    }
+}
+
 impl Conn {
     // Intentionally not public.
     fn new(partition_map: PartitionMap, schema: Schema) -> Conn {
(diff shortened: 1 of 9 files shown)


rnewman:
@ncalexan, @fluffyemily: could you take a look at this? We'll need this at some point for syncing, I expect.

rnewman:
And you might find this interesting, @grigoryk.

src/conn.rs
```diff
@@ -83,22 +84,28 @@ pub struct Conn {
...
+/// Call `commit` to commit your changes, or `rollback` to discard them.
+/// A transaction is held open until you do so.
+/// Your changes will be implicitly dropped along with this struct.
 pub struct InProgress<'a, 'c> {
     transaction: rusqlite::Transaction<'c>,
     mutex: &'a Mutex<Metadata>,
     generation: u64,
     partition_map: PartitionMap,
     schema: Schema,
+    last_report: Option<Rc<TxReport>>,
```
ncalexan: Why not grow a `Vec<Rc<TxReport>>` or similar?

src/conn.rs
```diff
@@ -114,40 +121,43 @@ impl<'a, 'c> InProgress<'a, 'c> {
...
-                metadata.schema = Arc::new(self.schema);
-            }
+        metadata.generation += 1;
+        metadata.partition_map = self.partition_map;
+        if self.schema != *(metadata.schema) {
+            metadata.schema = Arc::new(self.schema);
         }
 
-        Ok(())
+        Ok(self.last_report)
```
ncalexan: This feels wrong -- I think we should give back complete information (a vector of reports), and in the case of `rollback`, an empty `vec![]`.

src/conn.rs
```diff
@@ -342,11 +353,14 @@ mod tests {
         let t2 = "[{:db.schema/attribute \"three\", :db/ident :a/keyword1}]";
 
         let in_progress = conn.begin_transaction(&mut sqlite).expect("begun successfully");
-        let (in_progress, report) = in_progress.transact(t).expect("transacted successfully");
-        let one = report.tempids.get("one").cloned();
-        let (in_progress, report) = in_progress.transact(t2).expect("transacted again");
-        let mut three = report.tempids.get("three").cloned();
-        in_progress.commit().expect("Commit succeeded.");
+        let in_progress = in_progress.transact(t).expect("transacted successfully");
+        let one = in_progress.last_report().unwrap().tempids.get("one").cloned();
```
ncalexan: There are helpers to assert the shape of `tempids`, IIRC.  And it would be good to ensure you're really getting new entity IDs (i.e., that the metadata updating is correct).

src/conn.rs
```diff
@@ -234,50 +234,17 @@ impl Conn {
     pub fn transact(&mut self,
                     sqlite: &mut rusqlite::Connection,
                     transaction: &str) -> Result<TxReport> {
-
+        // Parse outside the SQL transaction. This is a choice…
```
ncalexan: Explicate trade-offs, please.

src/conn.rs
```diff
@@ -234,50 +234,17 @@ impl Conn {
...
-            }
-        }
+        let report = self.begin_transaction(sqlite)?
+                         .transact_entities(entities)?
+                         .commit()?
+                         .expect("we always get a report");
 
-        Ok(report)
+        // Our Rc never leaked, so this unwrap should always succeed.
+        Ok(Rc::try_unwrap(report).expect("solo Rc"))
```
ncalexan: Do we have an impedance here?  I'm not sure why we're `Rc` wrapping these reports at all in this series, but I didn't think about it much.

rnewman: `Rc` was left over and no longer needed — I removed it. That turns this into `Ok(report)` again!

src/conn.rs
rnewman: That's a good question. I suppose simplicity: you'll only need these reports if you want to resolve a tempid, in which case you should do so immediately after transacting on the `InProgress`. If we return a sequence, then you'll need to figure out which report you need.

On the other hand, perhaps we will need this anyway when syncing (tracking parts), so this is open to change. I'll add a comment.

src/conn.rs
rnewman: OTOH, doing it this way leaves `None` for no writes, and otherwise behaves just like you had written a series of linear independent `transact`s. I see your point, but I'd rather leave that work until we need it.

rnewman:
Everything done apart from the Vec of reports, which we can do later if we need it. @ncalexan ?

rnewman:
@ncalexan to explain your transaction question:

- `IMMEDIATE` means 'start the transaction now, but don't exclude readers'. It prevents other connections from taking immediate or exclusive transactions. This is appropriate for our writes and `InProgress`: it means we are ready to write whenever we want to, and nobody else can start a transaction that's not `DEFERRED`, but we don't need exclusivity yet.
- `EXCLUSIVE` means just that. This is appropriate when bootstrapping: if anything will go wrong with preparing to write, it will happen right here, and we don't expect other writers _or readers_ yet.

There is an argument for always using `EXCLUSIVE`, but I think it's a little cheaper for read-only transactions and some amount of read-concurrency to stick with `IMMEDIATE`.

ncalexan:
> @ncalexan to explain your transaction question:
> 
>     IMMEDIATE means 'start the transaction now, but don't exclude readers'. It prevents other connections from taking immediate or exclusive transactions. This is appropriate for our writes and InProgress: it means we are ready to write whenever we want to, and nobody else can start a transaction that's not DEFERRED, but we don't need exclusivity yet.
>     EXCLUSIVE means just that. This is appropriate when bootstrapping: if anything will go wrong with preparing to write, it will happen right here, and we don't expect other writers or readers yet.
> 
> There is an argument for always using EXCLUSIVE, but I think it's a little cheaper for read-only transactions and some amount of read-concurrency to stick with IMMEDIATE.

OK, this is helpful.  Fold this into the code in some way, please.

ncalexan:
> Everything done apart from the Vec of reports, which we can do later if we need it. @ncalexan ?

I see this landed, which is fine.  But I still think that there's a weirdness with this API.  How do you extract all of the IDs allocated during a sequence of `InProgress` transactions?  Right now, you can't (and I mean you really can't, even in consumer code, since you can't even know the first transaction ID allocated by an `InProgress` transaction).  I think that's a bad design.

rnewman:
>  How do you extract all of the IDs allocated during a sequence of InProgress transactions?

There are several responses to this.

1. You don't, just as in SQLite you can't do

```
BEGIN TRANSACTION;
INSERT INTO …;
UPDATE …;
INSERT INTO …;
END TRANSACTION;
```

and get a single count of inserted and updated rows, or the `lastUpdateRowid`. The same applies here: within the in-progress transaction, each time you act you find out what happened, but you don't get some kind of aggregate report at the end. It would be fair for `InProgress` to return nothing at all, but we return the last thing for convenience, because it includes the final-state schema etc.

2. The output includes a final parts table, so you know what happened — look at the previous parts table vs the output if you want to see how many identifiers were added.

Perhaps you'd like some kind of `InProgressResult` type that aggregates some data from each write?

3. Calling code should be looking at the result of each internal-transact that they invoke. The results (in terms of ID allocations and tx IDs) are stable.

4. It doesn't make sense to aggregate tempID mappings, because they can overlap.


In short, though: `InProgress` is exactly like just doing `transact` and `q_once` on the raw connection, except that your operations all succeed or fail or rollback together.

Does that change anything for you, @ncalexan?

rnewman:
Added a comment in b7fb44a.

Issue #459, opened 2017-05-09
Show attributes of current database


fluffyemily:
Displays the attributes for the database.

* Implement `.attributes` command.
* Output is the pretty printed transaction statement that would be used to create each attribute..

PR #490, opened 2017-06-29
show attributes of current database

fluffyemily:
https://github.com/mozilla/mentat/issues/459

### DIFF
--- a/tools/cli/src/mentat_cli/store.rs
+++ b/tools/cli/src/mentat_cli/store.rs
@@ -62,7 +62,24 @@ impl Store {
         Ok(self.conn.transact(&mut self.handle, &transaction)?)
     }
 
+    // the schema is the entire schema of the store including structure used to describe the store.
     pub fn fetch_schema(&self) -> edn::Value {
         self.conn.current_schema().to_edn_value()
+    }
+
+    // the attributes are the specific attributes added to the schema for this particular store.
+    pub fn fetch_attributes(&self) -> edn::Value {
+        let schema = self.conn.current_schema();
+
+        edn::Value::Vector((&schema.schema_map).iter()
+            .filter_map(|(entid, attribute)| {
+                if let Some(ident) = schema.get_ident(*entid) {
+                    if !ident.namespace.starts_with("db") {
+                        return Some(attribute.to_edn_value(Some(ident.clone())));
+                    }
+                }
+                return None;
+            })
+            .collect())
     }
 }
--- a/tools/cli/src/mentat_cli/command_parser.rs
+++ b/tools/cli/src/mentat_cli/command_parser.rs
@@ -128,46 +128,26 @@ pub fn command(s: &str) -> Result<Command, cli::Error> {
                         }
                         Ok(Command::Open(args[0].clone()))
                     });
-    
-    let no_arg_parser = || arguments()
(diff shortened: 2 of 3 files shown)


tools/cli/src/mentat_cli/store.rs
```diff
@@ -62,7 +62,24 @@ impl Store {
...
     }
+
+    // the attributes are the specific attributes added to the schema for this particular store.
+    pub fn fetch_attributes(&self) -> edn::Value {
+        let schema = self.conn.current_schema();
+
+        edn::Value::Vector((&schema.schema_map).iter()
+            .filter_map(|(entid, attribute)| {
+                if let Some(ident) = schema.get_ident(*entid) {
+                    if !ident.namespace.starts_with("db") {
```
rnewman: I think we want an `is_builtin` predicate instead of doing this.

A more thorough approach is to check whether the entid of the ident is in the `db` part or the `user` part, no?

tools/cli/src/mentat_cli/store.rs
```diff
@@ -62,7 +62,24 @@ impl Store {
...
+                        return Some(attribute.to_edn_value(Some(ident.clone())));
+                    }
+                }
+                return None;
```
rnewman: Equivalent to just

```
    None
```

tools/cli/src/mentat_cli/store.rs
```diff
@@ -62,7 +62,24 @@ impl Store {
...
         self.conn.current_schema().to_edn_value()
     }
+
+    // the attributes are the specific attributes added to the schema for this particular store.
+    pub fn fetch_attributes(&self) -> edn::Value {
+        let schema = self.conn.current_schema();
+
+        edn::Value::Vector((&schema.schema_map).iter()
+            .filter_map(|(entid, attribute)| {
+                if let Some(ident) = schema.get_ident(*entid) {
```
rnewman: One advantage of `match`:

```
match schema.get_ident(*entid) {
    Some(ident) if ident.is_builtin() => {
        Some(attribute.to_edn_value(Some(ident.clone())))
    },
    _ => None,
}
```

rnewman:
@fluffyemily is there anything to salvage from this PR?

sc13-bioinf:
It would be great to have an edn dump of the current schema available in the android API. I am happy to try and make a PR for that. I guess that you have some code to convert the output of current_schema to edn in the PR?

rnewman:
@sc13-bioinf: `Conn.current_schema().to_edn_value()` does what you want.

sc13-bioinf:
Is there a way to convert that into something similar to the contents of cities.edn?
I seem to get something like:

Keyword(Keyword(NamespaceableName { namespace: Some("db"), name: "ident" })): Keyword(Keyword(NamespaceableName { namespace: Some("community"), name: "name" }))

sc13-bioinf:
I had a go at adding this to the Android bindings. Not sure how correct / safe this is as I am new to rust + JNA.

https://github.com/mozilla/mentat/compare/master...sc13-bioinf:android-current-schema

If you approve then I can submit a PR or make some changes.

rnewman:
@sc13-bioinf you should submit a PR, rather than trying to have folks discuss it here. Thanks!

PR #492, opened 2017-10-10
Update some dependencies.

rnewman:
I decided to fork `error-chain` to restore `Send + Sync`, so we're now up to date to v0.11.

The current `rustc` warns if you use `///` somewhere it's not expecting a docstring, so I fixed some of those, too.

Our date library decided to rename `UTC` to `Utc`. *sigh*

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ rustc_version = "0.1.7"
 
 [dependencies]
 chrono = "0.4"
-error-chain = "0.8.1"
+error-chain = { git = "https://github.com/rnewman/error-chain", branch = "rnewman/sync" }
 time = "0.1"
 
 [dependencies.rusqlite]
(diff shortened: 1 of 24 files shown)


Cargo.toml
```diff
@@ -17,12 +17,12 @@ members = []
 rustc_version = "0.1.7"
 
 [dependencies]
-chrono = "0.3"
-error-chain = "0.8.1"
+chrono = "0.4"
+error-chain = { git = "https://github.com/rnewman/error-chain", branch = "rnewman/sync" }
```
fluffyemily: Are the related changes going to be pushed back to `rust-lang-nursery/error-chain`? I wouldn't want to be maintaining a fork for too long.

rnewman: Not directly. They added `Sync` in 0.8.1, then they removed it in 0.9, with figuring out how to allow consumers to opt in to different kinds of errors being a bit TBD. This is a very simple fork, at least.

PR #496, opened 2017-11-29
Update cargo import for uuid to enable the V4 features for creating n…

fluffyemily:
…ew Uuids

I want to use `mentat_core::Uuid` in the storage prototype, but I want to access some features that are only enabled with the V4 features. This patch enables V4 features for the current version.

### DIFF
--- a/edn/Cargo.toml
+++ b/edn/Cargo.toml
@@ -16,7 +16,7 @@ itertools = "0.7"
 num = "0.1"
 ordered-float = "0.5"
 pretty = "0.2"
-uuid = "0.5"
+uuid = { version = "0.5", features = ["v4"] }
 
 [build-dependencies]
 peg = "0.5"


rnewman:
RFC 4122 advises that "distributed applications generating UUIDs at a variety of hosts must be willing to rely on the random number source at all hosts. If this is not feasible, the namespace variant should be used."

@fluffyemily, what's missing without this? Is there a reason not to use a type 1 UUID?

fluffyemily:
@rnewman I guess I was using the `new_v4` function already. Even `new_v1` requires the enabling of the V1 crate, but I suppose `new(UuidVersion::Mac)` will do?

rnewman:
Yeah, `Mac` should be fine — it'll ensure global uniqueness. The only reason to use a V4 UUID is if you would rather have randomness than leak some device identifiers in order to get uniqueness…

PR #500, opened 2017-12-05
Fix issue whereby when an incorrect command was entered, the buffer w…

fluffyemily:
…asn't cleared and the next command was appended on the end of the incorrect one

### DIFF
--- a/tools/cli/src/mentat_cli/input.rs
+++ b/tools/cli/src/mentat_cli/input.rs
@@ -17,7 +17,7 @@ use self::InputResult::*;
 
 use command_parser::{
     Command, 
-    command
+    command,
 };
 
 use errors as cli;
@@ -55,8 +55,8 @@ impl InputReader {
             Ok(mut r) => {
                 r.set_word_break_chars(" \t\n!\"#$%&'()*+,-./:;<=>?@[\\]^`");
                 Some(r)
-            }
-            Err(_) => None
+            },
+            Err(_) => None,
         };
 
         InputReader{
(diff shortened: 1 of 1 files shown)


tools/cli/src/mentat_cli/input.rs
```diff
@@ -96,16 +96,23 @@ impl InputReader {
...
-                try!(command(&self.buffer))
+                let res = command(&self.buffer);
+                res
             }
         };
 
+        if cmd.is_err() {
+            self.buffer.clear();
+            self.in_process_cmd = None;
+            return Err(cmd.err().unwrap());
```
rnewman: Just `return cmd`?

tools/cli/src/mentat_cli/input.rs
```diff
@@ -96,16 +96,23 @@ impl InputReader {
...
+        }
+        let cmd = cmd.ok().unwrap();
```
rnewman: The Rusty way to do this, I think, is to make the `match` richer and have it encompass each of these chunks:

```
match cmd {
    e @ Err(_) => {
        self.buffer.clear();
        self.in_process_cmd = None;
        return e;
    },
    Ok(command) => {
        …
    },
}
```

tools/cli/src/mentat_cli/input.rs
rnewman: Never mind, wrong value type.

rnewman:
I did my own review comments!

PR #502, opened 2017-12-06
Sync metadata schema and SyncMetadataClient

grigoryk:


tolstoy/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+
+[dependencies]
+futures = "0.1"
+hyper = "0.11"
+tokio-core = "0.1"
+serde = "1.0"
+serde_json = "1.0"
+serde_derive = "1.0"
+lazy_static = "0.2"
+uuid = { version = "0.5", features = ["serde", "v4"] }
```
rnewman: We should make this `uuid` feature change in every Cargo file.

tolstoy/src/lib.rs
```diff
@@ -0,0 +1,42 @@
...
+
+error_chain! {
+    types {
+        Error, ErrorKind, ResultExt, Result;
+    }
+
+    foreign_links {
+        IOError(std::io::Error);
+        UriError(hyper::error::UriError);
+        HttpError(hyper::Error);
```
rnewman: Does `hyper::Error` contain `hyper::error::UriError`?

tolstoy/src/lib.rs
```diff
@@ -0,0 +1,42 @@
...
+        SqlError(rusqlite::Error);
+    }
+}
+
+pub type TolstoyResult = Result<()>;
```
rnewman: Not sure this adds value.

tolstoy/src/metadata.rs
```diff
@@ -0,0 +1,96 @@
...
+
+use TolstoyResult;
+use uuid;
+use schema;
+use rusqlite;
+use Result;
+
+trait SyncMetadataClientInterface {
+    fn get_remote_head(&self) -> Result<uuid::Uuid>;
+    fn set_remote_head(&mut self, uuid: uuid::Uuid) -> TolstoyResult;
```
rnewman: … indeed, this would be clearer as `Result<()>`.

rnewman: If you don't consume the `Uuid`, you might choose to take it as a ref.

tolstoy/src/metadata.rs
```diff
@@ -0,0 +1,96 @@
...
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+use TolstoyResult;
+use uuid;
+use schema;
+use rusqlite;
+use Result;
+
+trait SyncMetadataClientInterface {
```
rnewman: Traits are typically named via adjectives; in this case you might choose `HeadTrackable` or `MetadataStore`.

tolstoy/src/metadata.rs
```diff
@@ -0,0 +1,96 @@
...
+    fn get_remote_head(&self) -> Result<uuid::Uuid>;
```
rnewman: Just `remote_head` is fine.

tolstoy/src/metadata.rs
```diff
@@ -0,0 +1,96 @@
...
+
+impl SyncMetadataClientInterface for SyncMetadataClient {
+    fn get_remote_head(&self) -> Result<uuid::Uuid> {
+        let uuid_query_res = self.conn.query_row(
+            "SELECT value FROM tolstoy_metadata WHERE key = ?",
+            &[&schema::REMOTE_HEAD_KEY], |r| {
+                let raw_uuid: Vec<u8> = r.get(0);
+                match uuid::Uuid::from_bytes(raw_uuid.as_slice()) {
+                    Ok(uuid) => uuid,
+                    Err(e) => panic!("Couldn't parse UUID: {}", e)
```
rnewman: Best not to panic. And this method returns a result! So:

```
let bytes: Vec<u8> = r.get(0);
uuid::Uuid::from_bytes(bytes.as_slice()).map_err(|e| e.into())
```

should suffice to translate the error into your own chain, or perhaps:

```
let bytes: Vec<u8> = r.get(0);
uuid::Uuid::from_bytes(bytes.as_slice())?
```

tolstoy/src/metadata.rs
```diff
@@ -0,0 +1,96 @@
...
+                }
+            }
+        );
+
+        match uuid_query_res {
+            Ok(uuid) => Ok(uuid),
+            Err(e) => panic!("Could not query for REMOTE_HEAD_KEY: {}", e)
```
rnewman: Same here. Don't panic.

tolstoy/src/metadata.rs
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    fn set_remote_head(&mut self, uuid: uuid::Uuid) -> TolstoyResult {
+        let tx = self.conn.transaction()?;
+        let uuid_bytes = uuid.as_bytes().to_vec();
+        tx.execute("UPDATE tolstoy_metadata SET value = ? WHERE key = ?", &[&uuid_bytes, &schema::REMOTE_HEAD_KEY])?;
+
+        match tx.commit() {
+            Ok(()) => Ok(()),
+            Err(e) => panic!(e)
```
rnewman: Same.

tolstoy/src/metadata.rs
```diff
@@ -0,0 +1,96 @@
...
+    use super::*;
+
+    #[test]
+    fn test_get_remote_head_default() {
+        let conn = schema::tests::setup_conn();
+        let metadata_client: SyncMetadataClient = SyncMetadataClient::new(conn);
+        let head_res = metadata_client.get_remote_head();
+        match head_res {
+            Ok(uuid) => assert_eq!(uuid::Uuid::nil(), uuid),
+            Err(e) => panic!("Query error: {}", e)
```
rnewman: You can also:

```
assert_eq!(uuid::Uuid::nil(), metadata_client.get_remote_head().expect("fetch succeeded"));
```

tolstoy/src/schema.rs
```diff
@@ -0,0 +1,126 @@
...
+use TolstoyResult;
+
+pub static REMOTE_HEAD_KEY: &str = r#"remote_head"#;
+
+lazy_static! {
+    /// SQL statements to be executed, in order, to create the Tolstoy SQL schema (version 1).
+    #[cfg_attr(rustfmt, rustfmt_skip)]
+    static ref V1_STATEMENTS: Vec<&'static str> = { vec![
+        r#"CREATE TABLE IF NOT EXISTS tolstoy_tu (tx INTEGER PRIMARY KEY, uuid BLOB NOT NULL UNIQUE) WITHOUT ROWID"#,
+        r#"CREATE TABLE IF NOT EXISTS tolstoy_metadata (key BLOB NOT NULL UNIQUE, value BLOB NOT NULL)"#,
```
rnewman: Can `key` be `PRIMARY KEY` and this be `WITHOUT ROWID`?

tolstoy/src/schema.rs
```diff
@@ -0,0 +1,126 @@
...
+
+use rusqlite;
+use TolstoyResult;
+
+pub static REMOTE_HEAD_KEY: &str = r#"remote_head"#;
+
+lazy_static! {
+    /// SQL statements to be executed, in order, to create the Tolstoy SQL schema (version 1).
+    #[cfg_attr(rustfmt, rustfmt_skip)]
+    static ref V1_STATEMENTS: Vec<&'static str> = { vec![
```
rnewman: `V1` is a bit misleading — we're flying blind here.

tolstoy/src/schema.rs
```diff
@@ -0,0 +1,126 @@
...
+            PRAGMA foreign_keys=ON;
+        ").expect("success");
+
+        conn
+    }
+
+    pub fn setup_conn() -> rusqlite::Connection {
+        let mut conn = setup_conn_bare();
+        match ensure_current_version(&mut conn) {
+            Err(e) => panic!("Couldn't setup conn: {}", e),
```
rnewman: Again, you can probably short-circuit here by using `expect`.

tolstoy/Cargo.toml
grigoryk: Made this change in `edn/Cargo.toml` as well, but IIUC it's a preemptive one.

grigoryk:
@rnewman addressed your feedback.

tolstoy/Cargo.toml
rnewman: I should really understand how feature stacking works in Cargo; what's the behavior with and without the change?

grigoryk: We re-export Uuid from `edn` and use it across the codebase. I'm not sure what happens in case of feature... clashing, but it's best to either be consistent throughout or just centralize and re-use `edn::Uuid` here as well, to avoid confusion.

grigoryk:
@rnewman last push removed an explicit 'uuid' dependency from tolstoy, which should have been the only outstanding issue here.

rnewman:
c61bc79

Issue #505, opened 2017-12-15
(6410) SQLITE_IOERR_GETTEMPPATH thrown on first write on Android


grigoryk:
For reference: `The SQLITE_IOERR_GETTEMPPATH error code is an extended error code for SQLITE_IOERR indicating that the VFS is unable to determine a suitable directory in which to place temporary files.`

This implies that SQLite being used is compiled with `SQLITE_TEMP_STORE=1` build flag, and somehow in this particular environment it's failing to determine a suitable location on disk for temp files.

From Firefox's build config, on Android SQLite is forced to use memory for temp store: https://dxr.mozilla.org/mozilla-central/source/db/sqlite3/src/moz.build#90
"On Android there's no tmp partition, so always use a MEMORY temp store" - via SQLITE_TEMP_STORE=3 flag.

We can force temp files to be stored in memory via `PRAGMA temp_store=2`.

grigoryk:
Following Firefox's config, it's using memory store on Android (no other choice) and on 64bit platforms.

> Try to use a MEMORY temp store when possible. That allows for better
> performance and doesn't suffer from a full separate tmp partition.
> Exclude 32bit platforms due to address space fragmentation issues.

Proposed patch in #506 forces memory storage indiscriminately. Ideally I'd like to follow what Firefox is doing and exclude 32bit platforms, but I'm not sure how to actually achieve that.

@ncalexan , thoughts?

ncalexan:
The linked code isn't testing for 32-bit on Android at all. For reference, from https://sqlite.org/tempfiles.html:

>  The SQLITE_TEMP_STORE compile-time parameter is a #define whose value is an integer between 0 and 3, inclusive. The meaning of the SQLITE_TEMP_STORE compile-time parameter is as follows:
    0. Temporary files are always stored on disk regardless of the setting of the temp_store pragma.
    1. Temporary files are stored on disk by default but this can be overridden by the temp_store pragma.
    2. Temporary files are stored in memory by default but this can be overridden by the temp_store pragma.
    3. Temporary files are always stored in memory regardless of the setting of the temp_store pragma.

 I think we should go with 2 everywhere for Mentat: default to memory, let consumers override.

rnewman:
I'm happy defaulting to memory storage for now (#506), but keeping this issue open to alter our SQLite build options or provide a configuration parameter: when used with Firefox's SQLite we'll get the former for free, and #506 doesn't allow consumers to override.

PR #506, opened 2017-12-15
Force SQLite temp files to be stored in memory (#505) r=rnewman

grigoryk:
See Issue #505

### DIFF
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -69,9 +69,12 @@ pub fn new_connection<T>(uri: T) -> rusqlite::Result<rusqlite::Connection> where
         _ => rusqlite::Connection::open(uri)?,
     };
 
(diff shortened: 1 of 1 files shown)


db/src/db.rs
```diff
@@ -69,12 +69,16 @@ pub fn new_connection<T>(uri: T) -> rusqlite::Result<rusqlite::Connection> where
         _ => rusqlite::Connection::open(uri)?,
     };
 
+    /// temp_store=2 is present to force SQLite to store temp files in memory.
+    /// Some of the platforms we support do not have a tmp partition (e.g. Android)
+    /// necessary to store temp files on disk.
```
rnewman: Please also refer to issue #505, and mention that it has pointers to Firefox's configuration.

We rely here on consumers being built using an appropriate set of flags; it's probably worth mentioning that.

PR #521, opened 2018-01-18
CLI quality-of-life fixes

thomcc:
These have been bugging me while working on type hinting, and are easy to fix so I just did them. None of them should be particularly controversial, hopefully.

Bumping the version fixes a bunch of issues like pasting into the CLI not actually working right until you type a character after the paste, etc (or was I the only one who has that happen to them?). It also makes emacs-style movement keys work.

Arguably, I think `:`, `/`, and `?` should be removed from the the word separators also, but I think it doesn't matter very much, and I can see the argument the other way too.

### DIFF
--- a/tools/cli/src/mentat_cli/input.rs
+++ b/tools/cli/src/mentat_cli/input.rs
@@ -84,13 +84,15 @@ impl InputReader {
             None => return Ok(Eof),
         };
 
+        if !self.buffer.is_empty() {
+            self.buffer.push('\n');
+        }
+
         self.buffer.push_str(&line);
 
         if self.buffer.is_empty() {
             return Ok(Empty);
         }
-
-        self.add_history(&line);
 
         // if we have a command in process (i.e. an incomplete query or transaction),
         // then we already know which type of command it is and so we don't need to parse the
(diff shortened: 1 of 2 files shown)


tools/cli/src/mentat_cli/input.rs
```diff
@@ -118,13 +123,17 @@ impl InputReader {
                         Ok(More)
                     },
                     _ => {
+                        let history = self.buffer.clone();
+                        self.add_history(&history);
```
rnewman: Just make `self.add_history` take a `String`, and pass `self.buffer.clone()` here.

tools/cli/src/mentat_cli/input.rs
```diff
@@ -118,13 +123,17 @@ impl InputReader {
...
                         self.buffer.clear();
                         self.in_process_cmd = None;
                         Ok(InputResult::MetaCommand(cmd))
                     }
                 }
             },
             Err(e) => {
+                let history = self.buffer.clone();
+                self.add_history(&history);
```
rnewman: Same.

Issue #474, opened 2017-06-08
[query] Add type-hinting functions


rnewman:
It would be convenient to have a way to constrain the types of variables explicitly:

```
[:find ?x :where
 [(uuid ?v)]
 [?x _ ?v]]
```

This hinting/constraining would be useful for queries like the above — find all entities associated with any UUID — but also allows for `ground` (and potentially other functions) to accept ambiguous EDN input. We don't have a good way right now to distinguish between `10` the `Long` and `10` the `entid`.

These functions are very straightforward to implement.

rnewman:
Beginner's Guide®:

0. Make a new file, alongside `query-algebrizer/src/ground.rs`, called `hints.rs` that implements a function much like `apply_ground`; perhaps call it `apply_type_hint`. Leave the body of the function empty. Add the file to `query-algebrizer/src/lib.rs`, just as for `ground.rs`. Don't forget to `git add` it.
1. In `query-algebrizer/src/clauses/where_fn.rs`, you'll see a `match` block that dispatches on function name. Look at `core/src/lib.rs`'s definition of `ValueType` to see the type names for which we want to define hints. Add `match` arms to match.
2. Implement the body of `apply_type_hint`. It should take the type, get a variable argument, and call `self.constrain_var_to_type` on the variable. That function is defined in `query-algebrizer/src/clauses/mod.rs`.
3. Write tests, just like `ground`'s tests — you'll have some that test algebrizing, like `query-algebrizer/tests/ground.rs`, and some that work with real data, in the top-level `tests/query.rs`. You should be able to create a database, add two attributes of different types, add values for each attribute, and then run a query like:

```
[:find ?v
 :where
 [(instant ?v)]
 [_ _ ?v]]
```

to get only values of attributes that are `:db/valueType :db.type/instant`.

This should be equivalent to the query

```
[:find ?v
 :where
 [?a :db/valueType :db.type/instant]
 [_ ?a ?v]]
```

rnewman:
Fixed in 98502eb6.

PR #523, opened 2018-01-18
Add a CLI command to output the SQL we would execute for a query.

thomcc:
Not having this has been bugging me while working on #474, and since I was already poking around in the CLI code...

Anyway, I can see arguments against it (it's an implementation detail), but it seems generally useful enough to be worth having, especially while we're working on mentat. 

Usage examples:
```
mentat=> .qsql [:find [?v ...] :where [65543 _ ?v]]
SELECT DISTINCT `all_datoms00`.v AS `?v`, `all_datoms00`.value_type_tag AS `?v_value_type_tag` FROM `all_datoms` AS `all_datoms00` WHERE `all_datoms00`.e = 65543

mentat=> .qsql [:find ?eid :where [?eid :t/name "foo"]]
SELECT DISTINCT `datoms00`.e AS `?eid` FROM `datoms` AS `datoms00` WHERE `datoms00`.a = 65536 AND `datoms00`.v = $v0
Bindings:
  $v0 = Text("foo")
```

Having an equivalent for transact seems useful eventually (maybe? I don't really know, to be honest), hence naming it `.qsql` and not the more natural `.sql`.

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -11,6 +11,8 @@
 use rusqlite;
 use rusqlite::types::ToSql;
 
+use std::rc::Rc;
+
 use mentat_core::{
     Entid,
     Schema,
@@ -20,6 +22,7 @@ use mentat_core::{
 use mentat_query_algebrizer::{
     AlgebraicQuery,
     algebrize_with_inputs,
+    EmptyBecause,
 };
 
 pub use mentat_query_algebrizer::{
@@ -90,6 +93,45 @@ impl IntoResult for QueryExecutionResult {
     }
 }
 
+/// A struct describing information about how Mentat would execute a query.
+pub enum QueryExplanation {
+    /// A query known in advance to be empty, and why we believe that.
+    KnownEmpty(EmptyBecause),
+    /// A query that takes actual work to execute.
+    ExecutionPlan {
+        /// The translated query and any bindings.
+        query: SQLQuery,
+        /// The output of SQLite's `EXPLAIN QUERY PLAN`.
+        steps: Vec<QueryPlanStep>,
+    },
+}
+
+/// A single row in the output of SQLite's `EXPLAIN QUERY PLAN`.
+/// See https://www.sqlite.org/eqp.html for an explanation of each field.
+pub struct QueryPlanStep {
+    pub select_id: i32,
+    pub order: i32,
+    pub from: i32,
+    pub detail: String,
+}
+
+fn algebrize_query<'schema, T>
+(schema: &'schema Schema,
+ query: FindQuery,
+ inputs: T) -> Result<AlgebraicQuery>
+    where T: Into<Option<QueryInputs>>
+{
+    let algebrized = algebrize_with_inputs(schema, query, 0, inputs.into().unwrap_or(QueryInputs::default()))?;
+    let unbound = algebrized.unbound_variables();
+    // Because we are running once, we can check that all of our `:in` variables are bound at this point.
+    // If they aren't, the user has made an error -- perhaps writing the wrong variable in `:in`, or
+    // not binding in the `QueryInput`.
+    if !unbound.is_empty() {
+        bail!(ErrorKind::UnboundVariables(unbound.into_iter().map(|v| v.to_string()).collect()));
+    }
+    Ok(algebrized)
+}
+
 fn fetch_values<'sqlite, 'schema>
 (sqlite: &'sqlite rusqlite::Connection,
  schema: &'schema Schema,
@@ -111,7 +153,7 @@ fn fetch_values<'sqlite, 'schema>
     let query = FindQuery::simple(spec,
                                   vec![WhereClause::Pattern(pattern)]);
 
-    let algebrized = algebrize_with_inputs(schema, query, 0, QueryInputs::default())?;
+    let algebrized = algebrize_query(schema, query, None)?;
 
     run_algebrized_query(sqlite, algebrized)
 }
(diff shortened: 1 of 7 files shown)


thomcc:
As suggested by @rnewman from slack, this now outputs the query plan as well, and has been renamed to `.explain_query` (short form is `.eq`).

Examples:

Complex query, no bindings:
```
mentat=> .eq [:find [?v ...] :where [65543 _ ?v]]
SQL:
  SELECT DISTINCT `all_datoms00`.v AS `?v`, `all_datoms00`.value_type_tag AS `?v_value_type_tag` FROM `all_datoms` AS `all_datoms00` WHERE `all_datoms00`.e = 65543
Plan: select id | order | from | detail
  2|0|0|SEARCH TABLE datoms USING INDEX idx_datoms_eavt (e=?)
  3|0|0|SEARCH TABLE datoms USING INDEX idx_datoms_eavt (e=?)
  3|1|1|SCAN TABLE fulltext_values VIRTUAL TABLE INDEX 1:
  1|0|0|COMPOUND SUBQUERIES 2 AND 3 (UNION ALL)
  0|0|0|SCAN SUBQUERY 1 AS all_datoms00
  0|0|0|USE TEMP B-TREE FOR DISTINCT
```

Simpler query, but one that requires bindings.
```
mentat=> .eq [:find ?eid :where [?eid :t/name "foo"]]
SQL:
  SELECT DISTINCT `datoms00`.e AS `?eid` FROM `datoms` AS `datoms00` WHERE `datoms00`.a = 65536 AND `datoms00`.v = $v0
  Bindings:
    $v0 = Text("foo")
Plan: select id | order | from | detail
  0|0|0|SEARCH TABLE datoms AS datoms00 USING COVERING INDEX idx_datoms_aevt (a=?)
```

src/query.rs
```diff
@@ -90,6 +90,22 @@ impl IntoResult for QueryExecutionResult {
     }
 }
 
+/// A struct describing information about how mentat would execute a query.
```
rnewman: Mentat

src/query.rs
```diff
@@ -214,3 +230,50 @@ pub fn q_once<'sqlite, 'schema, 'query, T>
...
+            detail: row.get(3)
+        }
+    };
+
+    let mut rows = if query.args.is_empty() {
+        statement.query_map(&[], from_row)?
+    } else {
+        let refs: Vec<(&str, &ToSql)> =
+            query.args.iter()
+                .map(|&(ref k, ref v)| (k.as_str(), v.as_ref() as &ToSql))
```
rnewman: Stack your dots. Or, better, factor this logic out of `run_algebrized_query`! Everything up to `query_to_select` is shared, and so is this logic for running some SQL.

You'll need to use `query` and `query_named` instead of the `_map_` variant, but that's OK, I think.

src/query.rs
```diff
@@ -214,3 +230,50 @@ pub fn q_once<'sqlite, 'schema, 'query, T>
...
+                .collect();
+        statement.query_map_named(refs.as_slice(), from_row)?
+    };
+    let mut steps = Vec::new();
```
rnewman: I believe `vec![]` is more common.

src/query.rs
```diff
@@ -214,3 +230,50 @@ pub fn q_once<'sqlite, 'schema, 'query, T>
...
+    while let Some(step) = rows.next() {
```
rnewman: One of the reasons for using `query_map` is that it returns a `MappedRows`, which implements `Iterator`. Rather than this manual collection, you can just do `let steps: Vec<QueryPlanStep> = rows.collect();`.

src/query.rs
thomcc: I think putting it into `run_algebrized_query` will result in a lot of conditionals in that function to support both cases. I could probably pull the common parts into something separate though.

src/query.rs
thomcc: Each item is a result (note the `?` on line 276), which is why I did it this way. I'll see if I can avoid the explicit iteration without making it too complex though.

src/query.rs
```diff
@@ -245,41 +287,22 @@ pub fn q_explain<'sqlite, 'schema, 'query, T>
...
-    let parsed = parse_find_string(query)?;
-    let algebrized = algebrize_with_inputs(schema, parsed, 0, inputs.into().unwrap_or(QueryInputs::default()))?;
-
-    let unbound = algebrized.unbound_variables();
-    if !unbound.is_empty() {
-        bail!(ErrorKind::UnboundVariables(unbound.into_iter().map(|v| v.to_string()).collect()));
-    }
+    let algebrized = algebrize_query_str(schema, query, inputs)?;
 
     let empty_because = algebrized.cc.empty_because.clone();
```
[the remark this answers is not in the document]
thomcc: The string we generate for empty queries always has `LIMIT 0` but it's not always `SELECT 1 limit 0`, and the thing it selects seems like it could be generally useful for debugging. (Maybe not though, it's not like we're going to actually execute that string...)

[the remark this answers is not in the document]
rnewman: Aye, it's probably not useful; the different shape of the SELECT query will just be to match the expected caller.

src/query.rs
```diff
@@ -291,8 +291,9 @@ pub fn q_explain<'sqlite, 'schema, 'query, T>
         where T: Into<Option<QueryInputs>>
 {
     let algebrized = algebrize_query_str(schema, query, inputs)?;
-
-    let empty_because = algebrized.cc.empty_because.clone();
+    if let Some(empty_because) = algebrized.cc.empty_because {
```
rnewman: Might be neater to use the public API:

```
if algebrized.is_known_empty() {
    return Ok(QueryExplanation::KnownEmpty(algebrized.cc.empty_because.unwrap()));
}

File: tools/cli/src/mentat_cli/repl.rs [hunk 7]
```diff
@@ -207,11 +207,15 @@ impl Repl {
     }
 
     pub fn explain_query(&self, query: String) {
-        let explanation = match self.store.explain_query(query) {
-            Result::Ok(vals) => vals,
+        let (query, steps) = match self.store.explain_query(query) {
```
rnewman: If you put the match outside, you can avoid the `return`s and flatten the function some:

```
match self.store.explain_query(query) {
    Result::Err(err) => println!(…),
    Result::Ok(QueryExplanation::KnownEmpty(because)) => println!(…),
    Result::Ok(QueryExplanation::ExecutionPlan { query, steps }) => {
        …
    }
```

thomcc:
(Hopefully I'm not overstepping my bounds by merging this myself)

Issue #520, opened 2018-01-17
Inequality predicates seem to match nonsensical types


thomcc:
It seems like the types returned by the inequality predicates (`>`, `>=`, `<`, `<=`) aren't limited to what is vaguely sensible.

To reproduce, you can copy/paste the following into the mentat CLI:

```
.t[{:db/ident :t/name :db/valueType :db.type/string :db/cardinality :db.cardinality/one}
   {:db/ident :t/weight :db/valueType :db.type/double :db/cardinality :db.cardinality/one}
   {:db/ident :t/key :db/valueType :db.type/keyword :db/cardinality :db.cardinality/one}
   {:db/ident :t/count :db/valueType :db.type/long :db/cardinality :db.cardinality/one}
   {:db/ident :t/id :db/valueType :db.type/uuid :db/cardinality :db.cardinality/one}
   {:db/ident :t/date :db/valueType :db.type/instant :db/cardinality :db.cardinality/one}
   {:db/ident :t/ref :db/valueType :db.type/ref :db/cardinality :db.cardinality/one}]
.t[{:t/name "foo" :t/weight 0.3 :t/key :foo/bar :t/count 12 :t/id #uuid "ff62d552-6569-4d1b-b667-04703041dfc4" :t/date #inst "2018-01-01T11:00:00.000Z" }]
.t[{:t/name "bar" :t/weight 0.5 :t/key :foo/bar :t/count 99 :t/id #uuid "ff62d552-1234-1234-b667-047030414321" :t/date #inst "2017-01-01T11:00:00.000Z" }]
.t[[:db/add 65543 :t/ref 65544]]
.q[:find [?v ...] :where [65543 _ ?v] [(> ?v 0)]]
```

Which will output (a bunch of TxReports, followed by):

```
"foo"
0.3
:foo/bar
12
ff62d552-6569-4d1b-b667-04703041dfc4
2018-01-01 11:00:00 UTC
65544
```

Note that `[(> ?v 0)]` matches strings, keywords, and uuids, in addition to the types that seem intentional that it matches (e.g. doubles, longs, instants, and refs -- I'm assuming these are intentional, since a similarly broad `<` will return those, but not the strings/keywords/uuids).

In fact, it *always* seems to include the nonsense types, no matter how large I make the second argument to `>` (I've tried `1.79e308` and `9223372036854775807`, at least). 

[...]

PR #526, opened 2018-01-18
Implement type hinting predicates for queries

thomcc:
I think the name "hint" is a misnomer (it's more of a type check for an existing item).

If it needs more tests let me know.

### DIFF
--- a/query-algebrizer/src/clauses/inputs.rs
+++ b/query-algebrizer/src/clauses/inputs.rs
@@ -30,7 +30,6 @@ use errors::{
 /// the bindings that will be used at execution time.
 /// When built correctly, `types` is guaranteed to contain the types of `values` -- use
 /// `QueryInputs::new` or `QueryInputs::with_values` to construct an instance.
-#[derive(Clone)]
 pub struct QueryInputs {
     // These should be crate-private.
     pub types: BTreeMap<Variable, ValueType>,
--- a/query-algebrizer/src/clauses/inputs.rs
+++ b/query-algebrizer/src/clauses/inputs.rs
@@ -30,6 +30,7 @@ use errors::{
 /// the bindings that will be used at execution time.
 /// When built correctly, `types` is guaranteed to contain the types of `values` -- use
 /// `QueryInputs::new` or `QueryInputs::with_values` to construct an instance.
+#[derive(Clone)]
 pub struct QueryInputs {
     // These should be crate-private.
     pub types: BTreeMap<Variable, ValueType>,
--- a/query-algebrizer/src/clauses/mod.rs
+++ b/query-algebrizer/src/clauses/mod.rs
@@ -83,8 +83,6 @@ use validate::{
     validate_or_join,
 };
 
-use self::predicate::parse_type_predicate;
-
 pub use self::inputs::QueryInputs;
 
 // We do this a lot for errors.
@@ -969,22 +967,18 @@ impl ConjoiningClauses {
 
 impl ConjoiningClauses {
     pub fn apply_clauses(&mut self, schema: &Schema, where_clauses: Vec<WhereClause>) -> Result<()> {
-        let mut deferred = Vec::with_capacity(where_clauses.len());
         // We apply (top level) type predicates first as an optimization.
-        for c in where_clauses {
-            match &c {
-                &WhereClause::Pred(ref p) => {
-                    if let Some(ty) = parse_type_predicate(p.operator.0.as_str()) {
-                        self.apply_type_requirement(p, ty)?;
-                    }
-                },
-                _ => {}
-            };
-            deferred.push(c);
+        for clause in where_clauses.iter() {
+            if let &WhereClause::TypeAnnotation(ref anno) = clause {
+                self.apply_type_anno(anno)?;
+            }
         }
         // Then we apply everything else.
-        for c in deferred {
-            self.apply_clause(schema, c)?;
+        for clause in where_clauses {
+            if let &WhereClause::TypeAnnotation(_) = &clause {
+                continue;
+            }
+            self.apply_clause(schema, clause)?;
         }
         Ok(())
     }
@@ -1009,6 +1003,9 @@ impl ConjoiningClauses {
             WhereClause::NotJoin(n) => {
                 validate_not_join(&n)?;
                 self.apply_not_join(schema, n)
+            },
+            WhereClause::TypeAnnotation(anno) => {
+                self.apply_type_anno(&anno)
             },
             _ => unimplemented!(),
         }
--- a/query-algebrizer/src/clauses/mod.rs
+++ b/query-algebrizer/src/clauses/mod.rs
@@ -218,7 +218,7 @@ pub struct ConjoiningClauses {
     pub extracted_types: BTreeMap<Variable, QualifiedAlias>,
 
     /// Map of variables to the set of type requirements we have for them.
-    required_types: BTreeMap<Variable, ValueType>,
+    required_types: BTreeMap<Variable, ValueTypeSet>,
 }
 
 impl PartialEq for ConjoiningClauses {
@@ -551,17 +551,27 @@ impl ConjoiningClauses {
         }
     }
 
-    pub fn add_type_requirement(&mut self, var: Variable, ty: ValueType) {
-        if let Some(existing) = self.required_types.insert(var.clone(), ty) {
-            // If we already have a required type for `var`, we're empty.
-            if existing != ty {
-                self.mark_known_empty(EmptyBecause::TypeMismatch {
-                    var: var.clone(),
-                    existing: ValueTypeSet::of_one(existing),
-                    desired: ValueTypeSet::of_one(ty)
-                });
-            }
+    /// Require that `var` be one of the types in `types`. If any existing
+    /// type requirements exist for `var`, the requirement after this
+    /// function returns will be the intersection of the requested types and
+    /// the type requirements in place prior to calling `add_type_requirement`.
+    ///
+    /// If the intersection will leave the variable so that it cannot be any
+    /// type, we'll call mark_known_empty.
+    pub fn add_type_requirement(&mut self, var: Variable, types: ValueTypeSet) {
+        let existing = self.required_types.get(&var).cloned().unwrap_or(ValueTypeSet::any());
+
+        // We have an existing requirement. The new requirement will be
+        // the intersection, but we'll mark_known_empty if that's empty.
+        let intersection = types.intersection(&existing);
+        if intersection.is_empty() {
+            self.mark_known_empty(EmptyBecause::TypeMismatch {
+                var: var.clone(),
+                existing: existing,
+                desired: types,
+            });
         }
+        self.required_types.insert(var, intersection);
     }
 
     /// Like `constrain_var_to_type` but in reverse: this expands the set of types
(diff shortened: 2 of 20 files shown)


query-algebrizer/src/clauses/hints.rs
```diff
@@ -0,0 +1,58 @@
...
+    ErrorKind,
+    Result,
+};
+
+use types::{
+    EmptyBecause,
+};
+
+impl ConjoiningClauses {
+
```
rnewman: Nit: kill a few of these surplus newlines, here and elsewhere.

query-algebrizer/src/clauses/hints.rs
```diff
@@ -0,0 +1,58 @@
...
+    pub fn apply_type_hint(&mut self, pred: Predicate, ty: ValueType) -> Result<()> {
+        if pred.args.len() != 1 {
+            bail!(ErrorKind::InvalidNumberOfArguments(pred.operator.clone(), pred.args.len(), 1));
+        }
+
+        if let FnArg::Variable(ref v) = pred.args[0] {
+            if let Some(existing) = self.type_hints.insert(v.clone(), ty) {
```
rnewman: I agree with your naming concern, and I think I would call this `required_types` instead of `type_hints`.

rnewman: Refactoring request: make this a separate method that takes a type and a variable. Leave the arg parsing to this function. We will use this functionality elsewhere (e.g., to fix #520).

rnewman: Change request: make `type_hints` (aka `required_types`) a `ValueTypeSet`, not a `ValueType`. That way we can impose both numeric types.

query-algebrizer/src/clauses/not.rs
```diff
@@ -53,13 +53,14 @@ impl ConjoiningClauses {
             template.apply_clause(&schema, clause)?;
         }
 
+        template.expand_column_bindings();
+        template.prune_extracted_types();
+        template.process_required_types()?;
+
         if template.is_known_empty() {
```
rnewman: Do this check before and after each step, no?

query-algebrizer/src/clauses/predicate.rs
```diff
@@ -34,11 +34,26 @@ use types::{
...
+        _ => None
+    }
+}
+
 /// Application of predicates.
 impl ConjoiningClauses {
     /// There are several kinds of predicates in our Datalog:
     /// - A limited set of binary comparison operators: < > <= >= !=.
     ///   These are converted into SQLite binary comparisons and some type constraints.
+    /// - A set of type requirements constraining their argument to be a specific ValueType
```
rnewman: Closing period.

query-algebrizer/src/types.rs
```diff
@@ -334,11 +334,25 @@ pub enum ColumnConstraint {
...
+        value: TableAlias,
+        value_types: ValueTypeSet,
+        strict: bool,
+    },
     NotExists(ComputedTable),
     Matches(QualifiedAlias, QueryValue),
 }
 
+impl ColumnConstraint {
+    pub fn has_type(value: TableAlias, value_type: ValueType) -> ColumnConstraint {
```
rnewman: TODO: expand this to a `ValueTypeSet`.

query-algebrizer/tests/utils/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+    algebrize_with_inputs,
+    ConjoiningClauses,
+    Error,
+    QueryInputs,
+};
+
+// Common utility functions used in multiple test files. Note: Import this with
+// `pub mod utils` (not `mod utils`), or you'll get spurious unused function
+// warnings when functions exist in this file but are only used by modules that
+// don't import with `pub` (yes, this is annoying).
```
rnewman: Better to add `#![allow(unused_code)]` underneath the block comment in this file, and not use `pub mod`.

query-sql/src/lib.rs
```diff
@@ -345,11 +371,16 @@ impl QueryFragment for Constraint {
                     out.push_sql("0");
                     return Ok(())
                 }
-                out.push_sql("(");
+                let wrap_parens = constraints.len() > 1;
```
rnewman: Skip the conditional; always output parens. It's safe and it avoids two branches.

query-translator/tests/translate.rs
```diff
@@ -286,6 +286,50 @@ fn test_unknown_ident() {
...
+                            typeof(`all_datoms00`.v) = 'integer')");
+
+    assert_eq!(args, vec![]);
+}
+
+#[test]
+fn test_type_required_double() {
+    let schema = Schema::default();
+
+    let query = r#"[:find ?x :where [?x _ ?e] [(double ?e)]]"#;
```
rnewman: Here's a question for you. If we rephrase this as

`[:find ?x :where [(double ?e)] [?x _ ?e]]`

do we still use `all_datoms` rather than `datoms`? If so, please file a bug for that.

query-algebrizer/src/types.rs
thomcc: This is intended to be a helper for the most common case, e.g. the single type case where loose checks are desired.

query-algebrizer/src/clauses/mod.rs
```diff
@@ -844,7 +876,43 @@ impl ConjoiningClauses {
...
+                }
+                if already_known.is_subset(types) {
+                    // We're certain the value is already constrained enough.
+                    continue;
+                }
+            }
 
+            let qa = self.extracted_types
+                .get(&var)
+                .ok_or_else(|| Error::from_kind(ErrorKind::UnboundVariable(var.name())))?;
```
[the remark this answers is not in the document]
thomcc: This doesn't work without additional type hinting, which seems to require an intermediate variable.

query-algebrizer/src/clauses/pattern.rs
```diff
@@ -444,7 +444,7 @@ mod testing {
         // TODO: implement expand_type_tags.
         assert_eq!(cc.wheres, vec![
                    ColumnConstraint::Equals(d0_v, QueryValue::TypedValue(TypedValue::Boolean(true))),
-                   ColumnConstraint::HasType("datoms00".to_string(), ValueType::Boolean),
+                   ColumnConstraint::has_type("datoms00".to_string(), ValueType::Boolean),
```
rnewman: Maybe call this `has_unit_type`?

query-translator/tests/translate.rs
thomcc: Just so that it's recorded, this is no longer an issue as of b0d96e598a6f29e6dd8701279860e423d895ae72 (commit 4), which ensures that we use top level type hints when determining which table to use, even if they appear late.

query-algebrizer/src/clauses/mod.rs
```diff
@@ -873,6 +968,26 @@ impl ConjoiningClauses {
 }
 
 impl ConjoiningClauses {
+    pub fn apply_clauses(&mut self, schema: &Schema, where_clauses: Vec<WhereClause>) -> Result<()> {
+        let mut deferred = Vec::with_capacity(where_clauses.len());
+        // We apply (top level) type predicates first as an optimization.
+        for c in where_clauses {
```
rnewman: If you're going to collect into `deferred`, then use `into_iter` here to consume the input `Vec`… and then you can even use a filter and `collect` to do it.

But I have a better suggestion. Can we just add a `WhereClause::TypeAnnotation` case to make your check efficient, and go back to walking once by reference and once by value? Maybe a `good-first-bug` follow-up, 'cos it adds a bit more work to something that already works.

query-algebrizer/src/clauses/predicate.rs
```diff
@@ -34,11 +34,26 @@ use types::{
     Inequality,
 };
 
+pub fn parse_type_predicate(s: &str) -> Option<ValueType> {
```
rnewman: In my suggestion, this would go into the query parser, I think.

query-algebrizer/src/types.rs
```diff
@@ -334,11 +334,25 @@ pub enum ColumnConstraint {
         left: QueryValue,
         right: QueryValue,
     },
-    HasType(TableAlias, ValueType),
+    HasTypes {
+        value: TableAlias,
+        value_types: ValueTypeSet,
+        strict: bool,
```
rnewman: I think `strict` is a confusing name. Perhaps `check_value`?

query-sql/src/lib.rs
```diff
@@ -105,9 +105,22 @@ pub enum Constraint {
...
         subquery: TableOrSubquery,
+    },
+    TypeCheck {
+        value: ColumnOrExpression,
+        datatype: SQLDatatype
     }
 }
 
+/// Type safe representation of the possible return values from `typeof`
+pub enum SQLDatatype {
```
rnewman: s/Datatype/TypeAffinity

query-translator/src/translate.rs
```diff
@@ -157,10 +159,53 @@ impl ToConstraint for ColumnConstraint {
                     right: right.into(),
                 }
             },
+            HasTypes { value: table, value_types, strict: strict_requested } => {
+                // If strict mode checking is on, and need to check exactly 1
+                // (not both or neither) of ValueType::Double and ValueType::Long
+                // we emit a Constraint::TypeCheck.
+                let num_numeric_checks = (value_types.contains(ValueType::Double) as i32) +
+                                         (value_types.contains(ValueType::Long) as i32);
+                let strict = strict_requested && num_numeric_checks == 1;
```
rnewman: You're doing the numeric checks even if `!strict_requested`!

So:

```
let strict = strict_requested &&
             value_types.is_unit() &&
             …;
```

rnewman: This bothers me a little, because you're imposing here some knowledge about the precise SQL encoding that we use for types, and that logic is mostly elsewhere. If we add, say, a geospatial floating-point representation — different type tag but `Double` internally — is something going to go wrong here? Can we make this more abstract?

Is the abstract thing you're asking: do we need to check the type affinity of the value to determine whether the value matches our requested value type set?

I think you're saying that for {Double, Integer} we do not need to check the type affinity, because only Double and Integer share a type tag, and if they're both present then the value types correspond to the entire value space.

And so the inverse, which you're doing here, is: are all of the _types_ that are represented by the type tags that are derived from our `value_types` present in `value_types`?

We use `SQLValueTypeSet` to hide all the type-tag-related stuff, so imagine something like:

```
use mentat_core::SQLValueTypeSet;
…

// Ridiculously expensive, but doesn't include its own type tag manipulation.
fn requires_type_affinity_check(value_types: ValueTypeSet) -> bool {
    let mut type_set = ValueTypeSet::none();
    for tag in value_types.iter().map(|vt| vt.value_type_tag()) {
        // from_tag doesn't exist.
        type_set = type_set.union(SQLValueTypeSet::from_tag(tag));
    }
    !type_set.is_strict_subset(value_types)
}
```
    
I'm not suggesting you write exactly that code, but I'd like you to think about what's happening here, and move whatever relates to type tags into `SQLValueType` and friends. This function here should just check a predicate.

query-translator/src/translate.rs
```diff
@@ -157,10 +159,53 @@ impl ToConstraint for ColumnConstraint {
...
+                    // represented using the same value type tag (we choose to
+                    // remove the check for ValueType::Double, but it's an
+                    // arbitrary choice)
+                    value_types.difference(&ValueTypeSet::of_one(ValueType::Double))
+                } else {
+                    value_types
+                };
+                let constraints = types.into_iter().map(|ty| {
+                    let type_column = QualifiedAlias::new(table.clone(), DatomsColumn::ValueTypeTag).to_column();
+                    let loose = Constraint::equal(type_column, ColumnOrExpression::Integer(ty.value_type_tag()));
```
rnewman: `check_type_tag`.

query-translator/src/translate.rs
```diff
@@ -157,10 +159,53 @@ impl ToConstraint for ColumnConstraint {
...
+                        loose
+                    } else {
+                        let val_column = QualifiedAlias::new(table.clone(), DatomsColumn::Value).to_column();
+                        // We're handling strict equality for a numeric type, so we need to emit
+                        // a `typeof(col) = '(real|integer)'` too. This should happen a maximum of
+                        // once per `HasTypes`
+                        Constraint::And {
+                            constraints: vec![
+                                loose,
+                                Constraint::TypeCheck {
```
rnewman: This one would be called `check_type_affinity`.

query-translator/src/translate.rs
thomcc: > If we add, say, a geospatial floating-point representation — different type tag but Double internally — is something going to go wrong here?

No, I think that should work. If you added arbitrary precision integer type that had tag = 5 but was stored as a blob, it would be broken though. (Admittedly, you'll loose free comparisons, so I don't think this would be a good idea, but then again, I don't really think using the same tag for Double and Long is a good idea either).

> We use SQLValueTypeSet to hide all the type-tag-related stuff, so imagine something like:

That's still not enough to get to the point where you could add new types that are only discriminated by their SQLite type affinity without modifying translate.rs. Note that a single predicate can never be enough, since you need to know which affinity to emit the check for.

What you end up needing is something like what I've done in the 2nd of my new commits. I'm not sure it's what you had in mind, I'm not a big fan of it since it's more complex (at least conceptually) and less efficient to support cases that can't happen given the way the types are currently set up. I also don't love exporting SQLTypeAffinity from core, but it seems hard to avoid.

query-algebrizer/src/clauses/mod.rs
thomcc: I just went ahead and did this because it bothered me less than the alternative. It also didn't take nearly as long as (for example) the translator changes.

query-algebrizer/src/clauses/inputs.rs
```diff
@@ -30,6 +30,7 @@ use errors::{
 /// the bindings that will be used at execution time.
 /// When built correctly, `types` is guaranteed to contain the types of `values` -- use
 /// `QueryInputs::new` or `QueryInputs::with_values` to construct an instance.
+#[derive(Clone)]
```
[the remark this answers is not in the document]
thomcc: This was needed for a test (and that test fails to compile without it). I don't think I agree with your reasoning here, `clone()` is already a known not-cheap action (hence it not being done automatically, like in c++), but it's also your call and I don't care that much so I just changed the test.

rnewman:
98502eb6

Nice work, @thomcc!

PR #527, opened 2018-01-20
Automatically remove trailing whitespace from text files.

thomcc:
This is a very large pet peeve of mine, but I know I'm not alone. Unless we decide not to land this because our stance is that 'We don't care', which, well, hopefully it isn't.

Anyway this was done with a crude shell line I listed in the commit message, after this lands (assuming it's not surprisingly controversial) you can configure your editor to delete it, or use `git diff --check master` to detect it, or something.

I thought about trying to make Travis detect it (adding `! git --no-pager grep -I -n -e ' $'` to `before_install` would *probably* do it, but it would probably also leave a very unhelpful error message), either way, we can probably just catch it in reviews.


rnewman:
`git diff -w master..` shows no changes, so I'm fine with this.

PR #528, opened 2018-01-20
Add a way to abort the current command on the CLI

thomcc:
Currently if you enter a multiline command with invalid syntax on a previous line, you're basically screwed and need to abort the CLI. For example, if you type `.q [:find)<ret>`, you can never get out of the 'awaiting more input' state, since what's currently in the buffer will never parse as valid EDN.

Another use case is "I'm 5 brackets parentheses deep in a query, don't want to finish it, and would rather not be forced to figure out the correct order to close them".

This patch adds support for allowing Ctrl+C (or whatever else generates SIGINT for them) to clear the current command. If the user hits Ctrl+C when there's no current command, the program quits. Behavior vaguely similar to this is present in many different kinds of REPLs.


tools/cli/src/mentat_cli/input.rs
```diff
@@ -80,8 +96,16 @@ impl InputReader {
...
         let line = match self.read_line(prompt) {
-            Some(s) => s,
-            None => return Ok(Eof),
+            UserAction::TextInput(s) => s,
+            UserAction::Interrupt if self.in_process_cmd.is_some() => {
+                self.in_process_cmd = None;
+                self.buffer.clear();
+                // Move to the next line, so that our next prompt isn't on top
+                // of the previous.
+                println!("");
```
rnewman: You can just `println!()`.

PR #529, opened 2018-01-22
Expose Uuid from the top level crate

fluffyemily:
So we don't have to import a subcrate when using mentat elsewhere


src/lib.rs
```diff
@@ -46,6 +46,7 @@ pub fn get_connection() -> Connection {
 pub use mentat_core::{
     TypedValue,
     ValueType,
+    Uuid,
```
rnewman: Switch these lines: sorted order.

PR #530, opened 2018-01-22
Allow customers to assert facts about the current transaction.

### LINKED ISSUE #225: [tx] Allow [:db/tx :db/txInstant INSTANT] in transaction input
Datomic (and the Clojure implementation of Mentat) allow to specify the `:db/txInstant` that the transaction will be applied with by including a datom like
```edn
[[:db/add :db/tx :db/txInstant INSTANT]]
```
(Note that the `:db/tx` is a Mentat addition, although I think Datomic added an equivalent thing recently.)  This requires `INSTANT` to be strictly after the last transaction timestamp and strictly before the current transactor timestamp.

Adding this allows to import data with meaningful timestamps, for example.

eoger:
Fixes #225

### DIFF
--- a/db/src/tx.rs
+++ b/db/src/tx.rs
@@ -141,7 +141,7 @@ pub struct Tx<'conn, 'a> {
     tx_id: Entid,
 
     /// The timestamp when the transaction began to be committed.
-    tx_instant: DateTime<Utc>,
+    tx_instant: Option<DateTime<Utc>>,
 }
 
 impl<'conn, 'a> Tx<'conn, 'a> {
@@ -150,15 +150,14 @@ impl<'conn, 'a> Tx<'conn, 'a> {
         partition_map: PartitionMap,
         schema_for_mutation: &'a Schema,
         schema: &'a Schema,
-        tx_id: Entid,
-        tx_instant: DateTime<Utc>) -> Tx<'conn, 'a> {
+        tx_id: Entid) -> Tx<'conn, 'a> {
         Tx {
             store: store,
             partition_map: partition_map,
             schema_for_mutation: Cow::Borrowed(schema_for_mutation),
             schema: schema,
             tx_id: tx_id,
-            tx_instant: tx_instant,
+            tx_instant: None,
         }
     }
 
@@ -206,16 +205,18 @@ impl<'conn, 'a> Tx<'conn, 'a> {
             partition_map: &'a PartitionMap,
             schema: &'a Schema,
             mentat_id_count: i64,
+            tx_id: KnownEntid,
             temp_ids: InternSet<TempId>,
             lookup_refs: InternSet<AVPair>,
         }
 
         impl<'a> InProcess<'a> {
-            fn with_schema_and_partition_map(schema: &'a Schema, partition_map: &'a PartitionMap) -> InProcess<'a> {
+            fn with_schema_and_partition_map(schema: &'a Schema, partition_map: &'a PartitionMap, tx_id: KnownEntid) -> InProcess<'a> {
                 InProcess {
                     partition_map,
                     schema,
                     mentat_id_count: 0,
+                    tx_id,
                     temp_ids: InternSet::new(),
                     lookup_refs: InternSet::new(),
                 }
@@ -267,6 +268,11 @@ impl<'conn, 'a> Tx<'conn, 'a> {
                             entmod::Entid::Ident(ref e) => self.ensure_ident_exists(&e)?,
                         };
                         Ok(Either::Left(e))
+                    },
+
+                    // Special case: current tx ID.
+                    entmod::EntidOrLookupRefOrTempId::TempId(TempId::Tx) => {
+                        Ok(Either::Left(self.tx_id))
                     },
 
                     entmod::EntidOrLookupRefOrTempId::TempId(e) => {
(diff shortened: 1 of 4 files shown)


db/src/tx.rs
```diff
@@ -616,6 +624,17 @@ impl<'conn, 'a> Tx<'conn, 'a> {
                         tx_might_update_metadata = true;
                     }
 
+                    if e == KnownEntid(self.tx_id) && a == entids::DB_TX_INSTANT {
+                        tx_instant_set = true;
```
rnewman: Consider whether it is an error to have:

```
[:db/tx :db/txInstant #inst "2017-01-01T10:00:00Z"]
[:db/tx :db/txInstant #inst "2017-01-01T22:00:00Z"]
```

db/src/tx.rs
```diff
@@ -616,6 +624,17 @@ impl<'conn, 'a> Tx<'conn, 'a> {
...
+                        if let TypedValue::Instant(instant) = v {
+                            self.tx_instant = instant;
+                        } else {
+                            panic!("This type error should have been caught earlier.");
+                        }
+                        // TODO: INSTANT to be strictly after the last transaction
+                        // timestamp and strictly before the current transactor timestamp.
```
rnewman: We might want to weaken this condition — Datomic can assume correct clocks, but embedded clients cannot.

And we will need a way to bypass this, too, for syncing.

db/src/tx.rs
```diff
@@ -628,13 +647,14 @@ impl<'conn, 'a> Tx<'conn, 'a> {
...
-                          entids::DB_TX_INSTANT,
-                          self.schema.require_attribute_for_entid(entids::DB_TX_INSTANT).unwrap(),
-                          TypedValue::Instant(self.tx_instant),
-                          true));
+        // Transact [:db/add :db/txInstant NOW :db/tx] if it doesn't exist.
+        if !tx_instant_set {
+            non_fts_one.push((self.tx_id,
+                              entids::DB_TX_INSTANT,
+                              self.schema.require_attribute_for_entid(entids::DB_TX_INSTANT).unwrap(),
+                              TypedValue::Instant(self.tx_instant),
```
rnewman: So this would be:

```
self.tx_instant.map_or_else(TypedValue::current_instant, TypedValue::Instant)
```

tx/src/entities.rs
```diff
@@ -23,20 +23,21 @@ use self::edn::symbols::NamespacedKeyword;
...
-            TempId::Internal(_) => None,
+            _ => None,
         }
     }
 
     pub fn into_internal(self) -> Option<i64> {
         match self {
             TempId::Internal(x) => Some(x),
-            TempId::External(_) => None,
+            _ => None,
```
rnewman: The trick you used to find these locations — add the `enum` case, look for the compile errors — stops working if you use `_`. Add the explicit `TempId::Tx` case here instead:

```
TempId::Internal(x) => Some(x),
TempId::Tx |
TempId::External(_) => None,
```

eoger:
Thank you @rnewman, I have amended my patch.
Here is what I came up with regarding multiple values passed to txInstant:

```
[:db/add :db/tx :db/txInstant A]
[:db/add :db/tx :db/txInstant B]
// The last :db/add value is always picked up. Here txInstant = B.
```

```
[:db/add :db/tx :db/txInstant A]
[:db/retract :db/tx :db/txInstant A]
// If the last assertion about txInstant is a :db/retract and it corresponds to the last :db/add value, the system picks up a value. Here txInstant = ::now()

[:db/add :db/tx :db/txInstant A]
[:db/retract :db/tx :db/txInstant B]
// Here, the system would give txInstant = A
```

rnewman:
N.B., retractions need to pass through; they conceptually apply before the additions.

After working on #533 I think duplicate specifications within a single input should be an error; feel free to define a better error type ("duplicate datom") if one doesn't exist.

eoger:
Updated, thanks.
I removed most of the code that was in that `for term in final_terms` loop since #533 does the duplicate-detection heavy-lifting now.

db/src/tx.rs
```diff
@@ -633,13 +654,16 @@ impl<'conn, 'a> Tx<'conn, 'a> {
...
         }
 
-        // Transact [:db/add :db/txInstant NOW :db/tx].
-        // TODO: allow this to be present in the transaction data.
-        non_fts_one.push((self.tx_id,
-                          entids::DB_TX_INSTANT,
-                          self.schema.require_attribute_for_entid(entids::DB_TX_INSTANT).unwrap(),
-                          TypedValue::Instant(self.tx_instant),
-                          true));
+        tx_instant = self.tx_instant.unwrap_or(::now());
```
rnewman: `unwrap_or_else(TypeValue::current_instant)`.

db/src/tx.rs
```diff
@@ -633,13 +654,16 @@ impl<'conn, 'a> Tx<'conn, 'a> {
...
+
+        // Transact [:db/add :db/txInstant NOW :db/tx] if it doesn't exist.
+        if self.tx_instant == None {
```
rnewman: You always set `tx_instant` right above this. I don't think you need this conditional check (or the comment).

rnewman: Yeah, but the point is: we've filtered out all txInstant-setting datoms from the input. Every tx is supposed to have one. So `tx_instant` (the local var) is the value we want, and we should always add it to the output. Right?

rnewman: Oh, you stripped out the `continue`. Got it.

rnewman:
3bf74593

Issue #532, opened 2018-01-23
[tx] Interpret transact input with set semantics, like Datomic


rnewman:
Consider simple input like this:

```
[:db/add "foo" :thing/unique "x"]
[:db/add "foo" :thing/v 92]
[:db/add "bar" :thing/unique "x"]
[:db/add "bar" :thing/s true]
```

"foo" and "bar" are intended to be temporary IDs for the same entity. However, if that entity does not already exist in the store — that is, if a search with `:thing/unique "x"` returns no results — a new entid will be allocated _for each_, and insertion will fail.

```
.t [{:db/ident :thing/unique :db/valueType :db.type/string :db/cardinality :db.cardinality/one :db/unique :db.unique/identity :db/index true}]
.t [{:db/ident :thing/v :db/valueType :db.type/long :db/cardinality :db.cardinality/many}]
.t [{:db/ident :thing/s :db/valueType :db.type/boolean :db/cardinality :db.cardinality/one}]
.t [[:db/add "foo" :thing/unique "x"]
[:db/add "foo" :thing/v 92]
[:db/add "bar" :thing/unique "x"]
[:db/add "bar" :thing/s true]]
```

```
Error(MentatError(DbError(Msg("Could not update datoms: failed to add datoms not already present"))), State { next_error: Some(SqliteFailure(Error { code: ConstraintViolation, extended_code: 2067 }, Some("UNIQUE constraint failed: datoms.a, datoms.value_type_tag, datoms.v"))), backtrace: None }).
```

Fixing this means deducing equivalence between a graph of related tempids, so that only a single entid is allocated for equivalent entity placeholders. This is what the transactor does when it handles upserts, so it might be best to fix it there, perhaps by inserting collections of datoms grouped by tempid-entity in order that other tempid searches can succeed.

ncalexan:
I thought deeply about this transaction this morning, and I conclude that the behaviour is correct.  (The error message, of course, could use improvement.)  Suppose that you _knew_ that there was an existing entity with `:thing/unique "x"`, so that you could replace the tempids with `(lookup-ref :thing/unique "x")` without altering the meaning of the transaction.  Then you'd have:
```
[:db/add (lookup-ref :thing/unique "x") :thing/unique "x"]
[:db/add (lookup-ref :thing/unique "x") :thing/v 92]
[:db/add (lookup-ref :thing/unique "x") :thing/unique "x"]
[:db/add (lookup-ref :thing/unique "x") :thing/s true]
```
[...]

PR #533, opened 2018-01-23
Don't allow violation of cardinality-one restrictions within a single transact.

rnewman:
This requires maintaining a unique index so that we can `INSERT OR REPLACE`. The last value will win. We could also do `INSERT OR IGNORE`, check the number of changed rows, and throw.

This only works if the lookup ref succeeds. If it doesn't, then we're into #532.

Thoughts, @ncalexan?

### DIFF
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -743,6 +743,10 @@ impl MentatStoring for rusqlite::Connection {
                value_type_tag0 SMALLINT NOT NULL,
                added0 TINYINT NOT NULL,
                flags0 TINYINT NOT NULL)"#,
+
+            // We create this unique index so that it's impossible to violate a cardinality constraint
+            // within a transaction.
+            r#"CREATE UNIQUE INDEX IF NOT EXISTS temp.inexact_searches_unique ON inexact_searches (e0, a0) WHERE added0 = 1"#,
             r#"DROP TABLE IF EXISTS temp.search_results"#,
             // TODO: don't encode search_type as a STRING.  This is explicit and much easier to read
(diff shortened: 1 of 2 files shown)


db/src/db.rs
```diff
@@ -2269,4 +2273,39 @@ mod tests {
...
+            [:db/add 112 :db/valueType :db.type/string]
+            [:db/add 112 :db/cardinality :db.cardinality/one]
+            [:db/add 112 :db/unique :db.unique/identity]
+        ]"#);
+
+        assert_transact!(conn, r#"[
+            [:db/add "foo" :test/unique "x"]
+        ]"#);
+
+        // You can try to assert two values for the same entity and attribute,
```
ncalexan: Does this agree with Datomic?  With DataScript?

rnewman: Datomic raises an error:

```
":db.error/datoms-conflict Two datoms in the same transaction conflict\n{:d1 [17592186045418 :test/one 123 13194139534313 true],\n :d2 [17592186045418 :test/one 124 13194139534313 true]}\n"
```

rnewman:
We can't easily detect duplicates within a tx boundary: we can use `INSERT` and have an opaque error about uniqueness constraints being violated, which we wrap to `Could not insert non-fts one statements into temporary search table!`; use `INSERT OR IGNORE` and silently take the first value; or `INSERT OR REPLACE` and silently take the last (based on some permutation of the input order, I assume).

I think I'd prefer the opaque error, which we can improve in the future. I'll file that now.

rnewman:
ef9f2d9 with a map notation test.

PR #536, opened 2018-01-25
Implement `q_prepare` with pre-bound variables

kitcambridge:
Here's a first cut for feedback, without any algebrizer changes, so we can only bind variables at preparation time, not at execution time.

It's not clear to me if I need the logic to check if the schema is current: it looks like `begin_read` borrows the connection, and, since a `PreparedQuery` can't outlive its `InProgress` (I don't think...), we shouldn't be able to transact changes to the schema. In that case, I can revert the changes to pass around a `Conn` instead of the metadata.

### DIFF
--- a/src/conn.rs
+++ b/src/conn.rs
@@ -65,7 +65,9 @@ use errors::*;
 use query::{
     lookup_value_for_attribute,
     lookup_values_for_attribute,
+    PreparedResult,
     q_once,
+    q_prepare,
     q_explain,
     QueryExplanation,
     QueryInputs,
@@ -138,6 +140,8 @@ pub trait Queryable {
         where T: Into<Option<QueryInputs>>;
     fn q_once<T>(&self, query: &str, inputs: T) -> Result<QueryOutput>
         where T: Into<Option<QueryInputs>>;
+    fn q_prepare<T>(&self, query: &str, inputs: T) -> PreparedResult
+        where T: Into<Option<QueryInputs>>;
     fn lookup_values_for_attribute<E>(&self, entity: E, attribute: &edn::NamespacedKeyword) -> Result<Vec<TypedValue>>
         where E: Into<Entid>;
     fn lookup_value_for_attribute<E>(&self, entity: E, attribute: &edn::NamespacedKeyword) -> Result<Option<TypedValue>>
@@ -167,6 +171,11 @@ impl<'a, 'c> Queryable for InProgressRead<'a, 'c> {
         self.0.q_once(query, inputs)
     }
 
+    fn q_prepare<T>(&self, query: &str, inputs: T) -> PreparedResult
+        where T: Into<Option<QueryInputs>> {
+        self.0.q_prepare(query, inputs)
+    }
+
     fn q_explain<T>(&self, query: &str, inputs: T) -> Result<QueryExplanation>
         where T: Into<Option<QueryInputs>> {
         self.0.q_explain(query, inputs)
@@ -191,6 +200,15 @@ impl<'a, 'c> Queryable for InProgress<'a, 'c> {
                &self.schema,
                query,
                inputs)
+    }
+
+    fn q_prepare<T>(&self, query: &str, inputs: T) -> PreparedResult
+        where T: Into<Option<QueryInputs>> {
+
+        q_prepare(&*(self.transaction),
+                  &self.schema,
+                  query,
+                  inputs)
     }
 
     fn q_explain<T>(&self, query: &str, inputs: T) -> Result<QueryExplanation>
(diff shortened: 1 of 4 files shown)


src/query.rs
```diff
@@ -69,6 +77,46 @@ use errors::{
 };
 
 pub type QueryExecutionResult = Result<QueryResults>;
+pub type PreparedResult<'conn, 'sqlite> = Result<PreparedQuery<'conn, 'sqlite>>;
```
rnewman: `PreparedQueryResult`, I think.

src/query.rs
```diff
@@ -69,6 +77,46 @@ use errors::{
...
+        args: Vec<(String, Rc<rusqlite::types::Value>)>,
+        projector: Box<Projector>,
+    },
+}
+
+impl<'conn, 'sqlite> PreparedQuery<'conn, 'sqlite> {
+    pub fn run<T>(&mut self, _inputs: T) -> QueryExecutionResult where T: Into<Option<QueryInputs>> {
+        match self {
+            &mut PreparedQuery::Empty { ref conn, ref schema, ref find_spec } => {
+                let current_schema = &*conn.current_schema();
```
rnewman: I don't believe you need the schema checks right now.

At present you're scoping a `PreparedQuery` to an `InProgress(Read)`, which is isolated.

In the future `PreparedQuery` (or a related kind of prepared query!) instances might be scoped to a `Conn`, in which case they can be invalidated when the schema changes… but I think we can reduce the coupling required by that check. (See discussion below.)

Your choice whether you want to do the narrow-scoped version, the broad-scoped version, or both.

src/query.rs
```diff
@@ -289,6 +337,43 @@ pub fn q_once<'sqlite, 'schema, 'query, T>
     run_algebrized_query(sqlite, algebrized)
 }
 
+pub fn q_prepare<'conn, 'sqlite, 'schema, 'query, T>
+(conn: &'conn Conn,
```
rnewman: You don't actually use the `conn` here except to push it into the output. And it's not actually used by the output except to check the current schema. So my suspicion is that you shouldn't take `conn` as an argument: you might add a concept of invalidation as a wrapper around this, but it's not an essential part of a prepared statement — e.g., prepare and use within an `InProgress` or `InProgressRead` doesn't need invalidation checks.

src/conn.rs
```diff
@@ -119,7 +123,7 @@ pub trait Queryable {
 /// Your changes will be implicitly dropped along with this struct.
 pub struct InProgress<'a, 'c> {
     transaction: rusqlite::Transaction<'c>,
-    mutex: &'a Mutex<Metadata>,
+    conn: &'a Conn,
```
rnewman: Let's keep this as `mutex`.

src/query.rs
kitcambridge: I thought about doing the broad-scoped version, but it's tricky to scope a `PreparedQuery` to a `Conn`, I think because `Conn` doesn't currently manage the `rusqlite::Connection`. I tried this out with a lifetime annotation like `PreparedQuery<'conn, 'sqlite>`, but that confuses the borrow checker: either the `conn` or the `statement` doesn't live long enough.

src/query.rs
kitcambridge: Oh, that makes sense. So, similar to how `InProgressRead` is a wrapper around `InProgress` now, we'd have a `LongLivedPreparedQuery` or something wrap a `PreparedQuery` and ensure the schema hasn't changed?

src/query.rs
rnewman: Remember we discussed coupling the metadata/conn and the SQLite connection?

Right now `InProgress` and `InProgressRead` are a pairing of SQLite and metadata, with the side-effect of also managing a transaction. That is: they manage a `rusqlite::Transaction`, not a `rusqlite::Connection`.

I think there's another concept here, which is the pairing of a `rusqlite::Connection` and metadata. That might, as you say, be `Conn`, or a sibling of it.

I think the narrow-scoped version here is fine for now; we can generalize later.

src/query.rs
rnewman: Yes, or an inversion of that where a `PreparedQuery` (or a related type) is actually _owned_ by a `Conn` or similar (`QueryCache`?), and you can only _borrow_ it:

```
let prepared: &PreparedQuery = conn.prepare_query("…")?;
prepared.run(…);
…
let prepared = conn.prepare_query("…")?;    // Gives us a borrow to the same statement.
…
```

src/conn.rs
```diff
@@ -135,6 +139,11 @@ impl<'a, 'c> Queryable for InProgressRead<'a, 'c> {
         self.0.q_once(query, inputs)
     }
 
+    fn q_prepare<T>(&self, query: &str, inputs: T) -> PreparedResult
```
rnewman: How are you getting away without specifying the lifetime on the `PreparedResult`? Magic!

I wonder if the compiler successfully prevents you from leaking the prepared query outside the scope of the ongoing transaction… it should.

src/query.rs
```diff
@@ -289,6 +321,38 @@ pub fn q_once<'sqlite, 'schema, 'query, T>
...
+(sqlite: &'sqlite rusqlite::Connection,
+ schema: &'schema Schema,
+ query: &'query str,
+ inputs: T) -> PreparedResult<'sqlite>
+        where T: Into<Option<QueryInputs>>
+{
+    let algebrized = algebrize_query_str(schema, query, inputs)?;
+
+    let unbound = algebrized.unbound_variables();
+    if !unbound.is_empty() {
```
rnewman: Leave a TODO here, please!

src/conn.rs
kitcambridge: I had to specify an explicit lifetime after rebasing. 😄

PR #537, opened 2018-01-30
Simplify keywords; add `open`; re-export symbols

rnewman:
You can see how much cleaner this makes the vocabulary tests and other end-to-end examples.


rnewman:
I seem to recall that @kitcambridge's work around pairing connections and Conn is relevant to `open`; we should smush those things together at some point.

rnewman:
Whoever's free, please take a look!

tests/vocabulary.rs
```diff
@@ -117,14 +120,34 @@ fn test_real_world() {
...
+    // ::new is helpful. ::default is not.
+    assert!(vocabulary::AttributeBuilder::default()
+                  .value_type(ValueType::String)
+                  .multival(true)
+                  .fulltext(true)
+                  .build()
+                  .validate(|| "Foo".to_string())
+                  .is_err());
+
+    assert!(!vocabulary::AttributeBuilder::new()
```
fluffyemily: Could this be
```
assert!(vocabulary::AttributeBuilder::new()
    .value_type(ValueType::String)
    ...
    .is_ok());
```

instead? It's much clearer. I missed the negation at the beginning first time I read it.

rnewman: Yes! Excellent idea.

tools/cli/src/mentat_cli/store.rs
```diff
@@ -8,66 +8,68 @@
...
 
 use mentat::query::QueryResults;
 
-use mentat::conn::Conn;
 use mentat_db::types::TxReport;
 
-pub struct Store {
-    handle: rusqlite::Connection,
-    conn: Conn,
+pub struct OpenStore {
```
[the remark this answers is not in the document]
rnewman: Done. I flattened it into `Repl`.

rnewman:
2614f498beecd6744a9bf8bc47c539a8c85e998b

Issue #544, opened 2018-02-11
Cargo build failed


rleungx:
failed to run custom build command for `libsqlite3-sys v0.8.1`

--- stdout
TARGET = Some("x86_64-apple-darwin")
OPT_LEVEL = Some("0")
TARGET = Some("x86_64-apple-darwin")
HOST = Some("x86_64-apple-darwin")
TARGET = Some("x86_64-apple-darwin")
TARGET = Some("x86_64-apple-darwin")
HOST = Some("x86_64-apple-darwin")
CC_x86_64-apple-darwin = None
CC_x86_64_apple_darwin = None
HOST_CC = None
CC = None
HOST = Some("x86_64-apple-darwin")
TARGET = Some("x86_64-apple-darwin")
HOST = Some("x86_64-apple-darwin")
CFLAGS_x86_64-apple-darwin = None
CFLAGS_x86_64_apple_darwin = None
HOST_CFLAGS = None
CFLAGS = None
DEBUG = Some("true")
running: "cc" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-m64" "-DSQLITE_CORE" "-DSQLITE_DEFAULT_FOREIGN_KEYS=1" "-DSQLITE_ENABLE_API_ARMOR" "-DSQLITE_ENABLE_COLUMN_METADATA" "-DSQLITE_ENABLE_DBSTAT_VTAB" "-DSQLITE_ENABLE_FTS3" "-DSQLITE_ENABLE_FTS3_PARENTHESIS" "-DSQLITE_ENABLE_FTS5" "-DSQLITE_ENABLE_JSON1" "-DSQLITE_ENABLE_LOAD_EXTENSION=1" "-DSQLITE_ENABLE_MEMORY_MANAGEMENT" "-DSQLITE_ENABLE_RTREE" "-DSQLITE_ENABLE_STAT2" "-DSQLITE_ENABLE_STAT4" "-DSQLITE_HAVE_ISNAN" "-DSQLITE_SOUNDEX" "-DSQLITE_THREADSAFE=1" "-DSQLITE_USE_URI" "-DHAVE_USLEEP=1" "-Wall" "-Wextra" "-o" "../mentat/target/debug/build/libsqlite3-sys-a1f16be914e07476/out/sqlite3/sqlite3.o" "-c" "sqlite3/sqlite3.c"
[...]

PR #546, opened 2018-02-12
Allow control over rusqlite's bundling of SQLite

rnewman:
This is everything I think we should do for #544.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,8 +12,12 @@ authors = [
   "Thom Chiovoloni <tchiovoloni@mozilla.com>",
 ]
 name = "mentat"
-version = "0.6.0"
+version = "0.6.1"
 build = "build/version.rs"
+
+[features]
+default = ["bundled_sqlite3"]
+bundled_sqlite3 = ["rusqlite/bundled"]
 
 [workspace]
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -6,11 +6,19 @@ authors = [
...
+  "Kit Cambridge <kit@yakshaving.ninja>",
+  "Edouard Oger <eoger@fastmail.com>",
+  "Thom Chiovoloni <tchiovoloni@mozilla.com>",
 ]
 name = "mentat"
-version = "0.6.0"
+version = "0.6.1"
 build = "build/version.rs"
 
+[features]
```
fluffyemily: I expect that you have checked that this block doesn't need to be replicated across the other `Cargo.toml` files? Currently this seems to be the only crate that enables bundled, whereas before each individual crate did.

rnewman:
Yeah, the feature is implicitly shared — rusqlite appears several times in our dep tree, and only one set of flags is used because the versions are all the same. I checked for gcc appearing as a dev dependency.

rnewman:
2ac7a1b1dec5b7a905d6d6cb1e03483b5efe357a

PR #553, opened 2018-02-14
Thread caching into the query engine

rnewman:
This is a bunch of prep work, then a small amount of actual caching.

This:

- Splits out a trait for the cache.
- Pulls the trait down into `core`.
- Combines `Schema` and cache references into a trait object, `Known`.
- Alters the paths into the algebrizer where we previously passed `&Schema`, instead passing `Known`.
- In one spot, does the work to turn `[:find ?v :where [29 1 ?v]]` into `SELECT $v0 LIMIT 1` instead of `SELECT datoms00.v FROM datoms datoms00 WHERE datoms00.a = 1 AND datoms00.e = 29`.

It turns out that for a trivial database, it's quicker to retrieve the value from disk rather than take it as input, but (a) it should make a difference on bigger queries and data, and (b) the culmination of this work has us _not run SQL at all_.

Ignore the "Debug logging" commit; it's there so you can see log output when you run

```
cargo test --release test_cache_usage -- --nocapture
```

A small step after this is to (a) allow flipping cache on/off in the CLI, and (b) allow caching attributes from the CLI. That'll let us demo caching improvements interactively.


query-algebrizer/src/clauses/pattern.rs
```diff
@@ -263,15 +264,72 @@ impl ConjoiningClauses {
...
+                                    }
+                                    _ => {},    // Nothing we can do.
+                                }
+                            }
+                            _ => {},      // TODO: check constant values against cache.
+                        }
+                    }
+                }
+            },
+            _ => {},
```
fluffyemily: If I channel @ncalexan then we should be enumerating all of our options here, but probably not worth it.

src/conn.rs
```diff
@@ -527,6 +591,7 @@ impl Conn {
             partition_map: current_partition_map,
             schema: (*current_schema).clone(),
             cache: self.attribute_cache.write().unwrap(),
+            use_caching: true,
```
fluffyemily: It feels slightly clumsy to `begin_read`, get an `InProgress` with caching on by default and then remember to switch it off. It feels more ergonomic to enable caching as an arg to `begin_read` and pass it through to the creation of the `InProgress`.

query-algebrizer/src/clauses/mod.rs
```diff
@@ -146,6 +156,37 @@ impl<K: Clone + Ord, V: Clone> Intersection<K> for BTreeMap<K, V> {
     }
 }
 
+type VariableBindings = BTreeMap<Variable, TypedValue>;
+
+trait CheckEmpty {
+    fn insert_or_check(&mut self, key: Variable, value: TypedValue) -> Option<EmptyBecause>;
+}
+
+impl CheckEmpty for VariableBindings {
```
fluffyemily: I can't find anywhere this is actually being used. Is it left over code?

query/src/lib.rs
```diff
@@ -673,6 +673,14 @@ pub struct Pattern {
 }
 
 impl Pattern {
+    pub fn has_default_source(&self) -> bool {
```
fluffyemily: This is also not used anywhere that I can see.

query-algebrizer/src/clauses/pattern.rs
```diff
@@ -244,34 +245,388 @@ impl ConjoiningClauses {
...
-        // For now we only support the default source.
-        match pattern.source {
-            Some(SrcVar::DefaultSrc) | None => (),
-            _ => unimplemented!(),
+    // TODO: generalize.
+    // TODO: use constant values -- extract transformation code from apply_pattern_clause_for_alias.
+    // TODO: loop over all patterns until no more cache values apply?
+    fn attempt_cache_lookup(&mut self, known: Known, pattern: &EvolvedPattern) -> bool {
+        // Precondition: default source. If it's not default, don't call this.
+        assert!(pattern.source == SrcVar::DefaultSrc);
```
fluffyemily: `pattern.has_default_source()`?

query-algebrizer/src/clauses/pattern.rs
```diff
@@ -244,34 +245,388 @@ impl ConjoiningClauses {
...
+        use self::PlaceOrEmpty::*;
+        match self.make_evolved_pattern(known, pattern) {
+            Empty(e) => self.mark_known_empty(e),
+            Place(p) => self.apply_pattern(known, p),
         };
+    }
+
+    pub fn apply_pattern(&mut self, known: Known, pattern: EvolvedPattern) {
+        // For now we only support the default source.
+        if pattern.source != SrcVar::DefaultSrc {
```
fluffyemily: `if pattern.has_default_source()`?

PR #554, opened 2018-02-14
publically expose KnownEntid

fluffyemily:
I need this to use the EntityBuilder from outside of Mentat.


fluffyemily:
`TermBuilder::add` takes `KnownEntid` as the `e` arg and on update i I already have that arg as an `Entid` rather than fetching it as a `temp_id` so I need to convert it to a `KnownEntid` first.

Issue #555, opened 2018-02-14
Expose Sync at the top level


grigoryk:
External consumers need to be able to trigger syncs!

grigoryk:
Closed by  #557

PR #557, opened 2018-02-15
Revised uploader flow (battle-tested); CLI support for sync

grigoryk:
Fixed https://github.com/mozilla/mentat/issues/556, https://github.com/mozilla/mentat/issues/555

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ chrono = "0.4"
 error-chain = { git = "https://github.com/rnewman/error-chain", branch = "rnewman/sync" }
 lazy_static = "0.2"
 time = "0.1"
+uuid = "0.5"
 
 [dependencies.rusqlite]
 version = "0.12"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,10 +6,18 @@ authors = [
   "Jordan Santell <jsantell@mozilla.com>",
   "Joe Walker <jwalker@mozilla.com>",
   "Emily Toop <etoop@mozilla.com>",
+  "Grisha Kruglov <grigory@kruglov.ca>",
+  "Kit Cambridge <kit@yakshaving.ninja>",
+  "Edouard Oger <eoger@fastmail.com>",
+  "Thom Chiovoloni <tchiovoloni@mozilla.com>",
 ]
 name = "mentat"
-version = "0.6.0"
+version = "0.6.1"
 build = "build/version.rs"
+
+[features]
+default = ["bundled_sqlite3"]
+bundled_sqlite3 = ["rusqlite/bundled"]
 
 [workspace]
 members = ["tools/cli"]
@@ -27,7 +35,7 @@ uuid = "0.5"
 [dependencies.rusqlite]
 version = "0.12"
 # System sqlite might be very old.
-features = ["bundled", "limits"]
+features = ["limits"]
 
 [dependencies.edn]
 path = "edn"
--- a/src/conn.rs
+++ b/src/conn.rs
@@ -48,6 +48,10 @@ use mentat_tx::entities::TempId;
 
 use mentat_tx_parser;
 
+use mentat_tolstoy::Syncer;
+
+use uuid::Uuid;
+
 use errors::*;
 use query::{
     lookup_value_for_attribute,
@@ -131,6 +135,10 @@ pub trait Queryable {
         where E: Into<Entid>;
     fn lookup_value_for_attribute<E>(&self, entity: E, attribute: &edn::NamespacedKeyword) -> Result<Option<TypedValue>>
         where E: Into<Entid>;
+}
+
+pub trait Syncable {
+    fn sync(&mut self, server_uri: &String, user_uuid: &String) -> Result<()>;
 }
 
 /// Represents an in-progress, not yet committed, set of changes to the store.
@@ -378,6 +386,13 @@ impl Queryable for Store {
     fn lookup_value_for_attribute<E>(&self, entity: E, attribute: &edn::NamespacedKeyword) -> Result<Option<TypedValue>>
         where E: Into<Entid> {
         self.conn.lookup_value_for_attribute(&self.sqlite, entity.into(), attribute)
+    }
+}
+
+impl Syncable for Store {
+    fn sync(&mut self, server_uri: &String, user_uuid: &String) -> Result<()> {
+        let uuid = Uuid::parse_str(&user_uuid)?;
+        Ok(Syncer::flow(&mut self.sqlite, server_uri, &uuid)?)
     }
 }
 
--- a/src/conn.rs
+++ b/src/conn.rs
@@ -10,7 +10,7 @@
 
 #![allow(dead_code)]
 
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Mutex, RwLock, RwLockReadGuard, RwLockWriteGuard};
 
 use rusqlite;
 use rusqlite::{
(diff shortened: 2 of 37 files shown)


grigoryk:
Once we got past some server issues with @rfk, this revisions syncs successfully against either an empty or an unchanged prototype server.

Cargo.toml
```diff
@@ -30,6 +30,7 @@ chrono = "0.4"
 error-chain = { git = "https://github.com/rnewman/error-chain", branch = "rnewman/sync" }
 lazy_static = "0.2"
 time = "0.1"
+uuid = "0.5"
```
rnewman: I feel like syncing should be a feature, default-on (take a look at how we configure rusqlite's "bundled" feature). Follow-up is fine.

src/conn.rs
```diff
@@ -148,6 +151,10 @@ pub trait Queryable {
         where E: Into<Entid>;
 }
 
+pub trait Syncable {
```
rnewman: Put this, and the implementation of it, in `src/sync.rs`. Makes feature-flagging much easier — the `Syncable` trait would simply never exist, 'cos the `mod` import wouldn't happen.

src/conn.rs
```diff
@@ -418,6 +425,13 @@ impl Queryable for Store {
     }
 }
 
+impl Syncable for Store {
+    fn sync(&mut self, server_uri: &String, user_uuid: &String) -> Result<()> {
+        let uuid = Uuid::parse_str(&user_uuid)?;
```
rnewman: Can we just make `user_uuid` an opaque `user_token`, and make the parsing (and the error!) part of Tolstoy? Or, realistically, just take a `base_uri` here instead, which includes the user ID? Seems simpler…

src/errors.rs
```diff
@@ -35,6 +38,7 @@ error_chain! {
     foreign_links {
         EdnParseError(edn::ParseError);
         Rusqlite(rusqlite::Error);
+        UuidParseError(uuid::ParseError);
```
rnewman: Pushing the UUID errors down into being a Tolstoy error would allow us to omit this.

tolstoy/src/syncer.rs
```diff
@@ -82,10 +100,14 @@ impl<'c> TxReceiver for UploadingTxReceiver<'c> {
         // TODO separate bits of network work should be combined into single 'future'
 
         // Upload all chunks.
-        for datom in d {
+        for datom in datoms {
             let datom_uuid = Uuid::new_v4();
             tx_chunks.push(datom_uuid);
-            self.remote_client.put_chunk(&datom_uuid, serde_cbor::to_vec(&datom)?)?
+            d(&format!("putting chunk: {:?}, {:?}", &datom_uuid, &datom));
+            // TODO switch over to CBOR once we're past debugging stuff.
```
rnewman: Interesting to note: Amazon Ion has a single flag to flip to switch between a CBOR-ish and a JSON-ish representation; no other code changes. Pretty neat.

tolstoy/src/syncer.rs
```diff
@@ -155,20 +155,70 @@ impl Syncer {
...
+        let mut db_tx = sqlite.transaction()?;
+
+        let remote_head = remote_client.get_head()?;
+        d(&format!("remote head {:?}", remote_head));
+
+        let locally_known_remote_head = SyncMetadataClient::remote_head(&db_tx)?;
+        d(&format!("local head {:?}", locally_known_remote_head));
+
+        // Check if the server is empty - populate it.
+        if remote_head == Uuid::nil() {
```
rnewman: I'm expecting to see, at some point, a little state enum pop up here. After all, we're encapsulating a bunch of potential states:

- Both sides are at the same head.
- The remote server is empty, and we are at head H.
- The remote server is not empty, and we are empty.
- The remote server is at head HO and we are at HN, where HN descends from HO.
- … and vice versa.
- The remote server is at head HR, we are at head HL, and we have a shared parent HS.

Breaking this into an enum allows us to separate the "where are we?" diagnostic stage (fetch the local head, fetch the remote head, figure out parents…) from the "now what do we do?" step, which can then be neatly laid out in a `match` on the enum.

Follow-up is fine.

tolstoy/src/syncer.rs
```diff
@@ -155,20 +155,70 @@ impl Syncer {
...
+
+        Ok(())
+    }
+
+    pub fn flow(sqlite: &mut rusqlite::Connection, server_uri: &String, user_uuid: &Uuid) -> Result<()> {
+        d(&format!("sync flowing"));
 
-            // Commit everything: tx->uuid mappings and the new HEAD. We're synced!
-            db_tx.commit()?;
+        ensure_current_version(sqlite)?;
```
rnewman: Future: `e_c_v` expects to be called during open or create, so it establishes its own transaction. You can probably  omit it here, but if you don't you should figure out how to share the transaction.

tolstoy/src/tx_mapper.rs
```diff
@@ -46,6 +46,24 @@ impl TxMapper {
...
+    pub fn get_tx_for_uuid(db_tx: &rusqlite::Transaction, uuid: &Uuid) -> Result<Option<Entid>> {
+        let mut stmt = db_tx.prepare_cached(
+            "SELECT tx FROM tolstoy_tu WHERE uuid = ?"
+        )?;
+
+        let uuid_bytes = uuid.as_bytes().to_vec();
+        let results = stmt.query_map(&[&uuid_bytes], |r| r.get(0))?;
+
+        let mut txs = vec![];
+        txs.extend(results);
```
rnewman: Lines 57-64 can be replaced by:

```
let tx = results.next();
if results.next().is_some() {
    bail!(…);     // Two or more!
}
Ok(tx)
```

tools/cli/src/mentat_cli/command_parser.rs
```diff
@@ -189,14 +208,15 @@ pub fn command(s: &str) -> Result<Command, cli::Error> {
                         });
     spaces()
     .skip(token('.'))
-    .with(choice::<[&mut Parser<Input = _, Output = Result<Command, cli::Error>>; 8], _>
+    .with(choice::<[&mut Parser<Input = _, Output = Result<Command, cli::Error>>; 9], _>
```
rnewman: You'll need to rebase on top of my CLI work — this will change to 10.

tools/cli/src/mentat_cli/command_parser.rs
```diff
@@ -152,6 +158,19 @@ pub fn command(s: &str) -> Result<Command, cli::Error> {
...
                     });
 
+    let sync_parser = string(SYNC_COMMAND)
+                    .with(spaces())
+                    .with(arguments())
+                    .map(|args| {
+                        if args.len() < 1 {
+                            bail!(cli::ErrorKind::CommandParse("Missing required argument".to_string()));
+                        }
+                        if args.len() > 2 {
```
rnewman: I think this can probably be simplified — just have the parser only take one argument?

tools/cli/src/mentat_cli/command_parser.rs
```diff
@@ -287,6 +307,19 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_sync_parser_path_arg() {
+        let input = ".sync https://example.com/api/ 316ea470-ce35-4adf-9c61-e0de6e289c59";
```
rnewman: I presume terrible things happen if we try to sync to two places at the same time, no? Follow-up issue.

PR #560, opened 2018-02-15
Add retract_kw

fluffyemily:
Turns out we needed `retract_kw` too.


src/entity_builder.rs
```diff
@@ -259,6 +259,19 @@ impl<'a, 'c> InProgressBuilder<'a, 'c> {
...
+    }
+
+    pub fn retract_kw<E, V>(&mut self, e: E, a: &NamespacedKeyword, v: V) -> Result<()>
+    where E: IntoThing<KnownEntidOr<TempIdHandle>>,
+          V: IntoThing<TypedValueOr<TempIdHandle>> {
+        let (attribute, value) = self.extract_kw_value(a, v)?;
+        self.retract(e, attribute, value)
+    }
+
+    fn extract_kw_value<V>(&mut self, a: &NamespacedKeyword, v: V) -> Result<(KnownEntid, TypedValueOr<TempIdHandle>)>
```
rnewman: Make this helper non-generic: call `into` in the call sites.

PR #561, opened 2018-02-15
Provide an API for creating truly empty stores

rnewman:
This is necessary for 'fresh-start' syncing: create an empty vessel into which the entire remote transaction log will be poured.

Use `Store::create_empty` with a path, or `Conn::empty` with a SQLite connection handle.

@grigoryk let me know what you think.


grigoryk:
I've pushed a commit to match semantics of what I'm trying to do a bit better - open CLI, open empty database, sync against a populated server.

Issue #514, opened 2018-01-12
[sync] Local transaction write client


grigoryk:
Tolstoy needs an ability to insert a bunch of remote transactions as provided by the downloader (https://github.com/mozilla/mentat/issues/511) into the local transactions table.

This includes:
- maintaining a txid->UUID mapping in `tolstoy_tu`
- updating the tx partition accordingly

This might also include performing raw assertions of datoms derived from the transactions (https://github.com/mozilla/mentat/issues/512).

rnewman:
The unstated assumption here: given a client at state A, including identifiers and datoms, where A was agreed with a server and (implicitly) other clients, then a tx T1 applied to A by another client _should_ safely apply to the client without any changes.

Issue #513, opened 2018-01-12
[sync] "Download & assert to unchanged client" flow


grigoryk:
Given...
- an unchanged local client
- a downloading http client (https://github.com/mozilla/mentat/issues/511)
- ability to assert datoms after inserting transactions (https://github.com/mozilla/mentat/issues/512)

... implement the following flow:
- Download any transactions after currently known `remote_head`
- Re-partition as necessary
- Insert transactions, generate and persist a UUID->txid map in `tolstoy_tu`, and assert datoms
- Advance `remote_head` via `SyncMetadataClient`

Issue #511, opened 2018-01-12
HTTP client for downloading remote transactions


grigoryk:
Client will be written around the API described in https://github.com/rfk/mentat-sync-prototype/ 

General flow:
```
    Using `SyncMetadataClient`, get the currently known remote_head.
    Get list of new transactions via GET /transactions?from={remote_head}
    For each transaction:
        Get list of chunks via GET /transactions/{trn}
        For each chunk:
            Get chunk contents via GET /chunks/{chunk}
    Return a list of transactions w/ chunks
```

PR #563, opened 2018-02-16
Basic sync support

grigoryk:
Fast-forward syncing with support for "merging" the bootstrap transaction.

Issues more or less resolved with this: #513, #514, #511

### DIFF
--- a/src/conn.rs
+++ b/src/conn.rs
@@ -118,10 +118,6 @@ pub struct Conn {
     pub(crate) tx_observer_service: Mutex<TxObservationService>,
 }
 
-pub trait Syncable {
-    fn sync(&mut self, server_uri: &String, user_uuid: &String) -> Result<()>;
-}
-
 impl Conn {
     // Intentionally not public.
     fn new(partition_map: PartitionMap, schema: Schema) -> Conn {
--- a/src/conn.rs
+++ b/src/conn.rs
@@ -131,17 +131,6 @@ impl Conn {
         }
     }
 
-    /// Prepare the provided SQLite handle for use as a Mentat store. Creates tables but
-    /// _does not_ write the bootstrap schema. This constructor should only be used by
-    /// consumers that expect to populate raw transaction data themselves.
-
-    pub(crate) fn empty(sqlite: &mut rusqlite::Connection) -> Result<Conn> {
-        let (tx, db) = db::create_empty_current_version(sqlite)?;
-        tx.commit()?;
-        Ok(Conn::new(db.partition_map, db.schema))
-    }
-
-
     pub fn connect(sqlite: &mut rusqlite::Connection) -> Result<Conn> {
         let db = db::ensure_current_version(sqlite)?;
         Ok(Conn::new(db.partition_map, db.schema))
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -178,13 +178,25 @@ pub mod query_builder;
 pub mod store;
 pub mod vocabulary;
 
+#[cfg(feature = "syncable")]
+mod sync;
+
+#[cfg(feature = "syncable")]
+pub use sync::{
+    Syncable,
+};
+
+#[cfg(feature = "syncable")]
+pub use mentat_tolstoy::{
+    SyncReport,
+};
+
 pub use query_builder::{
     QueryBuilder,
 };
 
 pub use conn::{
     Conn,
-    Syncable,
 };
 
 pub use mentat_transaction::{
--- a/src/sync.rs
+++ b/src/sync.rs
@@ -1,0 +1,46 @@
+// Copyright 2018 Mozilla
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+use uuid::Uuid;
+
+use mentat_transaction::{
+    InProgress,
+};
+
+use errors::{
+    Result,
+};
+
+use mentat_tolstoy::{
+    Syncer,
+    RemoteClient,
+    SyncReport,
+};
+
+pub trait Syncable {
+    fn sync(&mut self, server_uri: &String, user_uuid: &String) -> Result<SyncReport>;
+}
+
+impl<'a, 'c> Syncable for InProgress<'a, 'c> {
+    fn sync(&mut self, server_uri: &String, user_uuid: &String) -> Result<SyncReport> {
+        // Syncer behaves as if it's part of InProgress.
+        // This split into a separate crate is segment synchronization functionality
+        // in a single crate which can be easily disabled by consumers,
+        // and to separate concerns.
+        // But for all intents and purposes, Syncer operates over a "mentat transaction",
+        // which is exactly what InProgress represents.
+        let mut remote_client = RemoteClient::new(
+            server_uri.to_string(),
+            Uuid::parse_str(&user_uuid)?
+        );
+        Syncer::sync(self, &mut remote_client)
+            .map_err(|e| e.into())
+    }
+}
--- a/tolstoy/README.md
+++ b/tolstoy/README.md
@@ -1,0 +1,138 @@
+# Tolstoy, a Mentat Sync implementation
+
+## Current state
+This work is partially a proof-of-concept, partially an alpha implementation of how a generic sync might operate on top of a log-oriented storage layer a la Mentat.
+
+## Overview
+### Very briefly
+Tolstoy will synchronize a local Mentat database against a remote server, modifying local state if necessary, and uploading changes to the server if necessary. Schema additions are allowed (adding vocabulary). Schema mutations are currently not implemented (changing vocabulary). Mentat's core schema must be the same on all participating clients (i.e. core schema alterations are unsupported).
+
+**Basic example:**
+
+Client 1 knows about `name` and `age` of a person.
+```
+[
+    {
+        :db/ident :person/name
+        :db/valueType :db.type/string
+        :db/cardinality :db.cardinality/one
+    }
+    {
+        :db/ident :person/age
+        :db/valueType :db.type/long
+        :db/cardinality :db.cardinality/one
+    }
+    {:person/name "Grisha" :person/age 30}
+]
+```
+
+Client 2 doesn't know about `age`, but knows about `ssn`:
+```
+[
+    {
+        :db/ident :person/name
+        :db/valueType :db.type/string
+        :db/cardinality :db.cardinality/one
+    }
+    {
+        :db/ident :person/ssn
+        :db/valueType :db.type/long
+        :db/cardinality :db.cardinality/one
+        :db/unique :db.unique/identity
+        :db/index true
+    }
+    {:person/name "Grisha" :person/ssn 123}
+]
+```
+Sync Client 1, then Client 2, then Client 1 again.
+
+Entity `Grisha` will be "duplicated", since `:person/name` is not defined as unique.
+```
+[
+    [:person/name :db/ident :person/name]
+    [:person/name :db/valueType :db.type/string]
+    [:person/name :db/cardinality :db.cardinality/one]
+    
+    [:person/age :db/ident :person/age]
+    [:person/age :db/valueType :db.type/long]
+    [:person/age :db/cardinality :db.cardinality/one]
+    
+    [:person/ssn :db/ident :person/ssn]
+    [:person/ssn :db/valueType :db.type/long]
+    [:person/ssn :db/cardinality :db.cardinality/one]
+    [:person/ssn :db/unique :db.unique/identity]
+    [:person/ssn :db/index true]
+    
+    [?grisha_one :person/name "Grisha"]
+    [?grisha_one :person/age 30]
+
+    [?grisha_two :person/name "Grisha"]
+    [?grisha_two :person/ssn 123]
+]
+```
+
+If, instead, `:person/name` was defined as `:db/unique :db.unique/identity`, then our final state will be:
+```
+[
+    [...schema datoms...]
+
+    [?united_grisha :person/name "Grisha"]
+    [?united_grisha :person/age 30]
+    [?united_grisha :person/ssn 123]
+]
+```
+
+Note that in above examples, symbols such as `?grisha_one` are meant to expand to some internal entitiy id (e.g. 65536).
+
+### Try it via the CLI
+In the Mentat CLI, a `.sync` operation exposes Tolstoy's functionality. Basic usage: `.sync http://path-to-server account-uuid`. Authentication, etc., is not implemented.
+
+### In more detail...
+Syncing is defined in terms of coming to an agreement between local and remote states. A local state is what's currently present on the current instance. A remote state is what's currently present on a server.
+
+Mentat is a log-oriented store, and so "local" and "remote" are really just two transaction logs.
+
+Internally, Tolstoy tracks the "locally known remote HEAD" and the "last-synced local transaction", which gives us three basic primitives:
+- a shared root, a state which is at the root of both local and remote logs
+- incoming changes - what remote changed on top of the shared root
+- local changes of the shared root.
+
+Thus, four kinds of positive-case things might occur during a sync:
+- a no-op - there are no incoming changes and local didn't change
+- a local fast-forward - there are remote changes, but no local changes
+- a remote fast-forward - there are local changes, but no remote changes
+- a merge - there are both local and remote changes.
+
+The first three cases are "trivial" - we either do nothing, or we download and transact remote transactions, or we upload local transactions and advance the remote HEAD.
+
+The merge case is a little more complicated. During a merging sync, a kind of a rebase is performed:
+1. local transactions are moved off of the main timeline
+2. remote transactions are transacted on top of the shared root
+3. local transactions are transacted
+
+Generally, intuition about the transactor's behaviour applies to reasoning about Tolstoy's sync as well. If a transaction "makes sense", it will be applied.
+
+Remote transactions are applied "as-is", with an exception of the `txInstance` - it must be preserved, and so the datom describing it is re-written prior to application to use the `(transaction-tx)` transaction function.
+
+Local transactions are rewritten to use tempids instead of their entids if they are assertions, and `(lookup-ref a v)` form in cases of retractions - but only if `lookup-ref` is guaranteed to succeed, otherwise retractions are dropped on the floor. Cases where local retractions are dropped:
+- we're retracting an entitiy which isn't `:db/unique`
+- we're retracting an entitiy which was already retracted by one of the `remote` transactions.
+
+### Sync report
+A sync operation produces either a single or multiple sync reports.
+
+A single report - internally referred to as an atomic sync report - indicates that sync was able to finish within a single local database transaction.
+
+Alternatively a non-atomic report is produced. It's a series of regular atomic reports. This indicates that sync required multiple "passes" to complete - e.g. a merge first, then remote fast-forward - and each step was performed within a separate local database transaction.
+
+## Explicitly not supported - will abort with a NotYetImplemented
+This alpha implementation doesn't support some cases, but it recognizes them and gracefully aborts (leaving local and remote states untouched):
+- Syncing against a Mentat instance which uses a different core schema version.
+- Syncing with schema mutations. Schema additions are fine, but transactions which change a set of attributes that define a user-defined `:db/ident` will cause sync to abort.
+
+## Misc operational properties
+- All sync operations happen in a context of an `InProgress` - an internal Mentat transaction representation. If sync succeeds, all necessary operations are comitted to the underlying database in a single SQLite transaction. Similarly, an aborting sync will simply drop an uncomitted transaction.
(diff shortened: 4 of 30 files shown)


src/conn.rs
```diff
@@ -451,9 +458,56 @@ impl Queryable for Store {
...
+
+            // TODO this should be done _after_ any parts below are transacted!!
+            // Commits any changes Syncer made (tu mappings during an upload!)
+            db_tx.commit()?;
+        }
+        let mut last_tx_entid = None;
+        let mut last_tx_uuid = None;
+        // Giant hack: we're expecting 'parts' to not actually contain metadata
+        // about any transactions - just their datoms! we generate metadata ourselves
+        // while transacting, thus loosing when transactions were made originally /o\
```
rnewman: Yep, for that you'll need to go a little lower, and also renumber (#494).

src/conn.rs
```diff
@@ -451,9 +458,56 @@ impl Queryable for Store {
...
+        match tx_list {
+            Some(list) => {
+                for tx in list {
+                    let in_progress = self.begin_transaction()?;
```
rnewman: I think you should lift this out of the `for` loop — you can transact multiple times against an `in_progress`, and they'll all share a single SQLite transaction.

src/conn.rs
```diff
@@ -451,9 +458,56 @@ impl Queryable for Store {
...
 
 impl Syncable for Store {
+    // TODO refactor this out of there, somehow!
     fn sync(&mut self, server_uri: &String, user_uuid: &String) -> Result<()> {
         let uuid = Uuid::parse_str(&user_uuid)?;
-        Ok(Syncer::flow(&mut self.sqlite, server_uri, &uuid)?)
+        let mut tx_list = None;
+        {
+            let mut db_tx = self.sqlite.transaction()?;
+            if let Some(list) = Syncer::flow(&mut db_tx, server_uri, &uuid)? {
```
rnewman: If you implement `flow` on `InProgress`, you get the transaction for free… the only tricky bit is that you'd ideally like to not do any work (including establishing an exclusive transaction!) if there's no data to write.

src/conn.rs
```diff
@@ -451,9 +458,56 @@ impl Queryable for Store {
...
+                tx_list = Some(list);
+            }
+
+            // TODO this should be done _after_ any parts below are transacted!!
+            // Commits any changes Syncer made (tu mappings during an upload!)
+            db_tx.commit()?;
+        }
+        let mut last_tx_entid = None;
+        let mut last_tx_uuid = None;
```
rnewman: Use a pair to indicate that these two things must be set together.

```
let mut last_tx = None;
…
last_tx = Some((uuid, entid));
```

tolstoy/src/syncer.rs
```diff
@@ -169,6 +171,13 @@ impl<'c> TxReceiver for UploadingTxReceiver<'c> {
     }
 }
 
+// For returning out of the downloader as an ordered list.
+#[derive(Debug)]
+pub struct Tx {
+    pub tx: Uuid,
+    pub parts: Vec<TxPart>
```
rnewman: Nit: trailing comma, please.

rnewman (commented):
Keep going!

src/conn.rs
```diff
@@ -157,6 +154,12 @@ impl Store {
             sqlite: connection,
         })
     }
+
+    pub fn fast_forward_user_partition(&mut self, new_head: Entid) -> Result<()> {
```
rnewman: What I expect to see here is essentially a compare-and-set: if my local partition map is _this_, then expand it to _this one_.

Doing that work requires writing to the DB, so it shouldn't be on `Store`, it should be on `InProgress`; if the DB write fails the metadata will be discarded.

And you want to advance all parts — typically `user` and `tx` — at the same time, atomically, right? Remember that the data itself will refer to tx IDs — it's not enough to only work with the user partition!

src/conn.rs
```diff
@@ -157,6 +154,12 @@ impl Store {
...
+        let mut metadata = self.conn.metadata.lock().unwrap();
+        metadata.partition_map.expand_up_to(":db.part/user", new_head);
+        db::update_partition_map(&mut self.sqlite, &metadata.partition_map).map_err(|e| e.into())
```
rnewman: I would rather see the metadata change as a side-effect of updating the DB. The `db::` code is where we'll be able to do compare-and-set, and we already have code to update the metadata after a `transact`.

rnewman: Relatedly: it might be worth thinking about what you're doing here as a special case of renumbering.

Renumbering is the process of taking a space of new remote identifiers, moving any conflicting local identifiers out of the way, and making space locally for the remote identifiers. If there are no conflicting local identifiers, then the second step involves no work.

This is a nice unification because it reduces the number of different code paths…

src/lib.rs
```diff
@@ -95,6 +97,13 @@ pub mod conn;
 pub mod query;
 pub mod entity_builder;
 
+#[cfg(feature = "syncable")]
+pub mod sync;
+
+pub fn get_name() -> String {
```
rnewman: I'm  pretty sure I deleted this… it was just there for testing, long ago.

src/sync.rs
```diff
@@ -0,0 +1,133 @@
+// Copyright 2016 Mozilla
```
rnewman: 2018

src/sync.rs
```diff
@@ -0,0 +1,133 @@
...
+        let mut last_tx_entid = None;
+        let mut last_tx_uuid = None;
+
+        // During fast-forwarding, we will insert datoms with known entids
+        // which, by definition, fall outside of our user partition.
+        // Once we've done with insertion, we need to ensure that user
+        // partition's next allocation will not overlap with just-inserted datoms.
+        // To allow for "holes" in the user partition (due to data excision),
+        // we track the highest incoming entid we saw, and expand our
+        // local partition to match.
```
rnewman: Again, I think this process would benefit from splitting out renumbering.

1. Download the new remote parts table.
2. Renumber locally, which also makes space locally for remotely allocated identifiers. (In the fast-forward case no work is done here.)

Now we get into the state machine. For fast-forwarding:

3. Apply remote datoms (which now don't exceed the parts table!).

Make sense?

src/errors.rs
```diff
@@ -15,6 +15,7 @@ use std; // To refer to std::result::Result.
 use std::collections::BTreeSet;
 
 use rusqlite;
+use uuid;
```
ncalexan: Lift this commit out and land it, since it's mechanical.

tolstoy/src/metadata.rs
```diff
@@ -41,6 +46,14 @@ impl SyncMetadataClient {
         }
         Ok(())
     }
+
+    pub fn get_partitions(tx: &rusqlite::Transaction) -> Result<PartitionMap> {
```
ncalexan: We have https://github.com/mozilla/mentat/blob/5fe4f12d32379415e36eb80877cbeec6fd443ffa/db/src/db.rs#L443 just waiting to be generalized; let's use it.

tolstoy/src/schema.rs
```diff
@@ -29,6 +37,12 @@ pub fn ensure_current_version(tx: &mut rusqlite::Transaction) -> Result<()> {
         tx.execute(statement, &[])?;
     }
 
+    // Initial partition information is what we'd see at bootstrap, and is used during first sync.
+    // For now, only information about the user partition is tracked.
+    // We rely on the transactor to advance the tx partition.
+    // Support for migrating the db partition forward (that is, bootstrap schema) is TBD.
+    tx.execute("INSERT OR IGNORE INTO tolstoy_parts VALUES (?, ?, ?)", &[&PARTITION_USER, &USER0, &USER0])?;
```
ncalexan: We also have https://github.com/mozilla/mentat/blob/5fe4f12d32379415e36eb80877cbeec6fd443ffa/db/src/bootstrap.rs#L88.

tolstoy/src/schema.rs
```diff
@@ -98,6 +122,12 @@ pub mod tests {
             }
         }
 
+        let new_idx = USER0 + 1;
```
ncalexan: There's a function for writing the partition map in general, too.

tolstoy/src/schema.rs
```diff
@@ -80,6 +96,14 @@ pub mod tests {
             },
             (_, _) => { panic!("Wrong number of results."); },
         }
+
+        let partitions = SyncMetadataClient::get_partitions(&tx).unwrap();
```
ncalexan: In general, we can't be `:db.part/user`-centric, 'cuz `:db.part/tx` advances too.  I think this will get simpler as you generalize, not more complicated.

tolstoy/src/syncer.rs
```diff
@@ -143,8 +147,27 @@ impl<'c> TxReceiver for UploadingTxReceiver<'c> {
...
+        // Figure our the "high water-mark" for the user partition.
+        let mut largest_e = USER0;
+        for datom in &datoms {
+            if datom.e > largest_e {
+                largest_e = datom.e;
+            }
+        }
+
+        // Annotate first datom in the series with the user partition information.
+        // TODO this is obviously wrong - we want to read partition info without
```
ncalexan: It feels like this is at the wrong place in the stack.  You want to change the flow from `Vec<Tx>` to `(PartitionMap, Vec<Tx>)`, but what you've done is change it to `Vec<(Option<PartitionMap>, Tx)>` and an assumption about option.

It feels to me like we need some notion of a *chunk* that's more general than what we have now; some chunks are transaction data, yes; some are transactions themselves; but some are also partition maps.

I think you're probably aware of this, and tomorrow we can talk through the difficulty of changing the underlying model more broadly.

db/src/renumber.rs
```diff
@@ -300,10 +300,16 @@ mod tests {
                             [268435458 :db/txInstant ?ms 268435458 true]]"#);
 
         // N.b., after renumbering, the on-disk data is not reflected in the `TestConn`!
-        let mut db = renumber(&local.sqlite,
+        let mut db;
```
ncalexan: I think something has gotten confused.  Everything here is going to take place inside an encapsulating SQL transaction; we shouldn't be managing transactions in this helper function at all.  It's up to consumers to manage isolation, etc; and `sync` (like `transact`) will do so.

Reason to yourself: suppose renumbering succeeds but something later fails -- how will I roll back renumbering?  If you commit, you can't.  (Without using nested transactions or checkpoints or other complicated things, which I really don't think we need.)

ncalexan (commented):
We'll talk more tomorrow, @grigoryk.  Thanks for pushing the spear on this!

tolstoy/src/syncer.rs
grigoryk: I'm viewing this particular change as purely temporary - I really don't like having partition information one step removed from the transaction - and following the path of least resistance for now. The goal is to move this out of chunks entirely, and lift it up into transactions. They should be annotated with partitionmaps, and the only reason that's not the way it is now is lack of server support.

Currently, I don't think there's any _other_ need to generalize a chunk, and doing so seems like a distraction.

My higher level view of this - let's settle on the overall flow first - how data generally moves around, what do we need to move around, etc - then re-shape the server and clients to match. In the meantime, iterate quickly and maintain some kind of momentum.

"Nothing is as permanent as the temporary", but I hope that won't apply :dancing_men:

grigoryk:
Rebased on top of master & the timelines work in #783, folded history together as it's not really valuable and needs to be retold.

ncalexan (approved):
I see what you've done -- you've inserted the Tolstoy create in between the top-level crate and all the other crates.

But there's another approach: push the details of Mentat transactions (most of `Conn`) down closer to the `db` crate.  That is, the `db` crate basically "is" the `transaction` crate  right now; we just do a bunch of work gluing together bits of `db` at top-level right now.

I don't know what the right approach is.  My mental model was that the `db` crate would end up owning more of the metadata-handling functionality over time.  But I'm not really against this approach.  Have you thought through both?  Are you confident that this split is the one we should pursue?

One thing that's not clear to me is that there will ever be a _second_ Tolstoy like feature built on Mentat.  That is, a second feature that is _inside_ Mentat but wants the high-level query and transaction interfaces.  But as I type that, I realize I'm being short-sighted: **of course** there's going to be a feature like that, I just don't know what it is yet.

Give me your perspective, and then roll on.  (I've reviewed up to the "split" Pre:.)

grigoryk:
On the `db` crate - yeah, I briefly thought about it and I decided that I don't have a good enough sense yet for how to optimally organize this stuff, but splitting high level metadata away from the top level _is_ needed in some form.

My main goal was to separate that stuff out of the top-level crate, to enable nicer patterns for the internal consumers (well, just tolstoy for now). If we decide that it's better to keep shuffling this stuff a bit more to find it a better home, at least it _should_ be easier from this point on, and the consumers building on top of the `transaction` crate won't be affected significantly - likely just a few import changes.

You're right that what's in the `transaction` right now is a thin layer on top of `db`, doing some bookkeeping one layer above it. I'm also not a fan of having `Store` and `Conn` at the top level, either - it makes writing tests that want to use a real `Conn` awkward, forcing them to be at the top level, away from the functionality that's being tested. The `TestConn` surrogate is a smell of that. So, there's more re-shuffling to be done! My hope is that little steps like this will inform what the ideal world looks like.

db/src/lib.rs
```diff
@@ -48,6 +48,7 @@ mod bootstrap;
 pub mod entids;
 pub mod internal_types;    // pub because we need them for building entities programmatically.
 mod metadata;
+mod renumber;
```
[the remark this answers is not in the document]
grigoryk: I don't actually use it currently, no. Or rather, it's _used_ in a trivial case of fast-forwarding local as a way of expanding a local partition, but that operation is a no-op with the "derived partition maps" changes.

grigoryk:
Wow, this has been open for 7 months. This is a good example of how "not to PR". Anyway...

The current state is as follows.

Part 1 is very non-controversial. It clears up the code base, introduces bunch of internal types and solidifies the general flow of things. As for mechanics of a sync beyond fast-forward scenarios, only a "no-op rebase of local on top of remote" is supported. That is, if a merge generates any assertions/retractions, we bail out.

A merge is essentially: "rewind local transactions, transact the remote transactions, then transact local. succeed if local transactions had no effect, bail out otherwise".

Part 2 takes things further, and generally allows anything one can throw at a transactor (sans implemention bugs and some TODOs). Except for schema alterations. It's fine to add new vocabulary (including at the same time, on both "remote" and "local", as long as it's defined the same way), but we bail out if we see changes to existing vocabulary. This is done largely so that we can address syncing of schema alterations by themselves.

A merge is essentially: "rewind local transactions, transact the remote transactions, then ensure local may be transacted, perform some modifications, transact them, and request a follow-up sync if that operation produced any new datoms."

Certain choices were made around handling of entids and retractions, which are non-optimal in some cases but produce understandable results - and are natural given the current approach.

General approach is to let the transactor handle merges, letting the schema dictate how that happens. A concrete manifestation of this:
- :db/unique attributes will upsert. Asserting entities against non-unique attributes will naturally produce duplicates.

During a rebase, local entids are replaced with tempids. Retractions pose a challenge: a retraction won't allocate, so if a tempid mentioned in the retraction doesn't resolve, transaction will fail. If a retraction datom is by itself, its tempids will never resolve. Current approach is to replace tempids with lookup-refs whenever we see retractions that use tempids that we recognize won't resolve. Additionally, a lookup-ref mentioned in a retraction must resolve, and so we drop on the floor any retractions that fail to meet that criteria.

While all of this has certain limitations, it has very nice general property: an intuition about transactor's behaviour applies directly to sync's behaviour. Additionally, expanding what transactor accepts and supports will expand sync's capabilities with little additional effort. For example, if we ever add compound uniqueness, above sync mechanics should "just work" as you'd expect them.

grigoryk:
Currently tests are failing due to a bug in the transactor: https://github.com/mozilla/mentat/issues/818

tolstoy/README.md
```diff
@@ -0,0 +1,135 @@
+# Tolstoy, a Mentat Sync implementation
+
+## Current state
+This work is partially a proof-of-concept, partially an alpha implementation of how a generic sync might operate on top of a log-oriented storage layer a la Mentat.
+
+## Overview
+### Very briefly
+Tolstoy will synchronize a local Mentat database against a remote server, modifying local state if necessary, and uploading changes to the server if necessary. Schema additions are allowed (adding vocabulary). Schema mutations are currently not implemented (changing vocabulary). Mentat's core schema must be the same.
```
ncalexan: nit: "[must be the same] on all participating clients (i.e., core schema alterations are unsupported)."

tolstoy/README.md
```diff
@@ -0,0 +1,135 @@
...
+    [?united_grisha :person/ssn 123]
+]
+```
+
+Note that in above examples, symbols such as `?grisha_one` are meant to expand to some internal entitiy id (e.g. 65536).
+
+### In more detail...
+Syncing is defined in terms of coming to an agreement between local and remote states. A local state is what's currently present on the current instance. A remote state is what's currently present on a server.
+
+We're synchronizing transaction logs are synchronized, and so we may think about the primitive operations in context of smushing together two logs - local and remote.
```
ncalexan: nit: this sentence got caught in an edit.

tolstoy/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+Internally, Tolstoy tracks the "locally known remote HEAD" and the "last-synced local transaction", which gives us three basic primitives:
+- a shared root, a state which is at the root of both local and remote
+- incoming changes - what remote changed on top of the shared root
+- local changes on top of shared root.
```
ncalexan: nit: "[of] the [shared root]".

tolstoy/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+Thus, four kinds of positive-case things might occur during a sync:
+- a no-op - there are no incoming changes and local didn't change
+- a local fast-forward - there are remote changes, but no local changes
+- a remote fast-forward - there are local changes, but no remote changes
+- a merge - there are both local and remote changes.
+
+First three cases are "trivial" - we either do nothing, or we download and transact remote transactions, or we upload local transactions and advance remote HEAD.
```
ncalexan: nit: "The [first three] ...".

ncalexan (approved):

tolstoy/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+The merge case is a little more complicated. During a merging sync, a kind of a rebase is performed:
+1. local transactions are moved off of the main timeline
+2. remote transactions are transacted on top of the shared root
+3. local transactions are transacted
+
+Generally, intuition about transactor's behaviour applies to reasoning about Tolstoy's sync as well. If a transaction "makes sense", it will be applied.
```
ncalexan: nit: "[about] the [transactor's]"

tolstoy/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+Remote transactions are applied "as-is", with an exception of the `txInstance` - it must be preserved, and so the datom describing it is re-written prior to application to use the `(transaction-tx)` transaction function.
+
+Local transactions are rewritten to use tempids instead of their entids if they are additions, and `(lookup-ref a v)` form in cases of retractions - but only if `lookup-ref` is guaranteed to succeed, otherwise retractions are dropped on the floor. Cases where local retractions are dropped:
```
ncalexan: s/additions/assertions/

tolstoy/README.md
```diff
@@ -0,0 +1,138 @@
...
+- we're retracting an entitiy which isn't `db/unique`
```
ncalexan: `:db/unique` (include colon)

tolstoy/README.md
```diff
@@ -0,0 +1,138 @@
...
+- we're retracting an entitiy which was already retracted by `remote`.
```
ncalexan: by the remote client

tolstoy/README.md
```diff
@@ -0,0 +1,138 @@
...
+A sync operation produces either a single or multiple sync reports.
+
+A single report - internally referred to as an atomic sync report - indicates that sync was able to finish within a single local database transaction.
+
+Alternatively a non-atomic report is produced. It's a series of regular atomic reports. This indicates that sync required multiple "passes" to complete - e.g. a merge first, then remote fast-forward - and each step was performed within a separate local database transaction.
+
+## Explicitly not supported - will abort with a NotYetImplemented
+This alpha implementation doesn't support some cases, but it recognizes them and gracefully aborts (leaving local and remote states untouched):
+- Syncing against a Mentat instance which uses a different core schema version.
+- Syncing with schema mutations. Schema additions are fine, but transactions which change a set of attributes that define a user-defined `db/ident` will cause sync to abort.
```
ncalexan: nit: `:db/ident` (throughout -- keep the colons).

tolstoy/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Misc operational properties
+- All sync operations happen in a context of an `InProgress` - an internal Mentat transaction representation. If sync succeeds, all necessary operations are comitted to the underlying database in a single SQLite transaction. Similarly, an aborting sync will simply drop an uncomitted transaction.
+- "Follow-up" syncing is currently supported in a most basic way: if there are local changes arising from a merge operation, they are comitted to the local store, and a full sync is requested which is expected to fast-forward remote state in an optimal case, and if we lost the race to the server - to merge the local "merged state" with further remote changes.
```
ncalexan: "in a basic manner"

tolstoy/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Server
+Tolstoy operates against an instance of [Mentat Sync Prototype Server](https://github.com/rfk/mentat-sync-prototype). That repository defines a transaction-oriented API, which is all that Tolstoy expects of the server.
```
ncalexan: Can you include a hash, so that we know where this left off?

ncalexan (approved):
Looking good.  Some small changes, then bombs away!

tolstoy/README.md
grigoryk: Oh man, it did indeed.

PR #566, opened 2018-02-22
Update cache during transact

rnewman:
This PR makes our cache part of `Metadata`, making it clone-on-write, and updates it during assertion.

We have to consider that the cache needs to be updated within a transaction, but might need to be rolled back at any point. That means we can't rely on usual Cow-ish semantics, as with `Arc::make_mut` — we use that trick here because the `Conn` keeps a ref, so as soon as we write we will definitely clone, rather than having `get_mut` succeed. We must clone at some point; we do so at the level of each unique `(a, e)`.

### DIFF
--- a/db/src/cache.rs
+++ b/db/src/cache.rs
@@ -1,4 +1,4 @@
-// Copyright 2016 Mozilla
+// Copyright 2018 Mozilla
 //
 // Licensed under the Apache License, Version 2.0 (the "License"); you may not use
 // this file except in compliance with the License. You may obtain a copy of the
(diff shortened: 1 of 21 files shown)


rnewman:
OK, this is ready for review. I commented the heck out of `db/src/cache.rs`; I hope that makes it clear how this works.

rnewman:
```
Fully unit-bound. Returning constant projector.
Prepared cache execution took 3µs
```

Woo!

rnewman:
FTS now handled (with interned strings, at least within a generation!).

Builds on top of #577 and #576; it'll look a lot shorter when those land.

rnewman:
I added two commits on the end of this PR.

The first enables LTO for our release builds. This does cross-crate optimization, which is important in our world!

The second takes advantage of visibility specifiers to simplify some internal APIs. In short: scattering `pub(crate)` everywhere and deleting anything that seems dead.

fluffyemily:
I guess I should be pleased that much of the last rebase got rid of my unsubmitted review comments. Confused me for a while though.

db/src/cache.rs
```diff
@@ -557,8 +1084,361 @@ impl SQLiteAttributeCache {
...
+            },
+        }
+    }
+
+    fn done(&mut self, schema: &Schema) -> Result<()> {
+        // Oh, I wish we had impl trait. Without it we have a six-line type signature if we
+        // try to break this out as a helper function.
+        let collected_retractions = mem::replace(&mut self.collected_retractions, Default::default());
+        let collected_assertions = mem::replace(&mut self.collected_assertions, Default::default());
+        let mut do_twice = once(collected_retractions).chain(once(collected_assertions))
```
fluffyemily: nit: I hate this var name. It should describe what the var contains, not how many times you should call it.

db/src/watcher.rs
```diff
@@ -0,0 +1,155 @@
...
+    fn datom(&mut self, _op: OpType, _e: Entid, _a: Entid, _v: &TypedValue) {
+    }
+
+    fn done(&mut self, _schema: &Schema) -> Result<()> {
+        Ok(())
+    }
+}
+
+
+/*
```
fluffyemily: commented code kept around for a reason?

db/src/cache.rs
```diff
@@ -557,8 +1084,361 @@ impl SQLiteAttributeCache {
...
         self.inner.value_pairs(schema, attribute)
     }
 }
+
+/// We maintain a diff on top of the `inner` -- existing -- cache.
+/// That involves tracking unregisterings and registerings.
+#[derive(Debug, Default)]
+pub struct InProgressSQLiteAttributeCache {
+    inner: Arc<AttributeCaches>,
+    pub in_progress: AttributeCaches,
```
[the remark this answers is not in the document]
rnewman: I'm going to call this 'overlay', because that's what it is.

db/src/cache.rs
```diff
@@ -557,8 +1084,361 @@ impl SQLiteAttributeCache {
...
+        }
+
+        Self::repopulate(&mut self.in_progress, schema, sqlite, a)
+    }
+
+    fn repopulate(caches: &mut AttributeCaches,
+                  schema: &Schema,
+                  sqlite: &rusqlite::Connection,
+                  attribute: Entid) -> Result<()> {
+        let sql = "SELECT a, e, v, value_type_tag FROM datoms WHERE a = ? ORDER BY a ASC, e ASC";
```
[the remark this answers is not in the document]
rnewman: I'll move this down to `AttributeCaches`, actually.

db/src/cache.rs
rnewman: The name is intended to point out that it contains an iterator of two processed things, simply because we have no good way of describing the types (see the comment right above); we use this pattern quite often in the core because simple chaining of iterators this way doesn't allocate. I'll change the name to something else.

PR #574, opened 2018-03-01
Create Mentat FFI and expose observers

fluffyemily:
Expose TxObservers over FFI, forming the basis for an FFI crate for mentat.


ffi/src/android.rs
```diff
@@ -0,0 +1,11 @@
+// TODO just use https://github.com/tomaka/android-rs-glue somehow?
+
+use std::os::raw::c_char;
+use std::os::raw::c_int;
+
+// Logging
+pub const ANDROID_LOG_DEBUG: i32 = 3;
+pub const ANDROID_LOG_INFO: i32 = 4;
+pub const ANDROID_LOG_WARN: i32 = 5;
+pub const ANDROID_LOG_ERROR: i32 = 6;
```
rnewman: I presume there's a good reason for not using an enum with associated values here?

```
pub enum LogLevel {
    Debug = 3,
}
```

… that, I think, will make the values `isize`. Do you really need these to be `i32`?

ffi/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+};
+
+use utils::log;
+
+#[repr(C)]
+#[derive(Debug, Clone)]
+pub struct ExternTxReport {
+    pub txid: i64,
+    pub changes: Box<[i64]>,
+    pub changes_len: usize
```
rnewman: Trailing commas, please.

ffi/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+                ExternResult {
+                    error: string_to_c_char(e.description().into())
+                }
+            }
+        }
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn new_store(uri: *const c_char) -> *mut Store {
```
rnewman: `store_open` might be a better name.

Worth documenting here that opening the same store twice at the same time is verboten, and that the caller is responsible for calling `store_destroy`.

ffi/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn store_destroy(store: *mut Store) {
+    let _ = Box::from_raw(store);
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn store_register_observer(store: *mut Store,
+                                                   key: *const c_char,
```
rnewman: Nit: alignment.

ffi/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+                                                   attributes: *const i64,
+                                                   attributes_len: usize,
+                                                   callback: extern fn(key: *const c_char, reports: &ExternTxReportList)) {
+    let store = &mut*store;
```
rnewman: Nit: there should probably be a space in there somewhere, no? Or am I just not used to the oddities of writing C in Rust?

ffi/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+    let mut attribute_set = BTreeSet::new();
+    let slice = slice::from_raw_parts(attributes, attributes_len);
+    for i in 0..attributes_len {
+        let attr = slice[i].into();
+        attribute_set.insert(attr);
```
rnewman: You should be able to avoid this explicit loop, 'cos you've made a slice, which has a length.

```
attribute_set.extend(slice.iter().map(|x| x.into()));
```

(or `into_iter`) should work.

ffi/src/utils.rs
```diff
@@ -0,0 +1,56 @@
+// Copyright 2016 Mozilla
```
rnewman: 2018

src/conn.rs
```diff
@@ -528,6 +528,16 @@ impl Store {
...
     }
+
+    pub fn register_observer(&mut self, key: String, observer: Arc<TxObserver>) {
+        self.conn.register_observer(key, observer);
+    }
+
+    pub fn unregister_observer(&mut self, key: &String) {
+        println!("Store: unregistering observer for key {:?}", key);
+        self.conn.unregister_observer(key);
+        println!("Store: unregistered observer for key {:?}", key);
```
rnewman: Do you need to keep the debug logging?

src/conn.rs
```diff
@@ -818,7 +828,9 @@ impl Conn {
     }
 
     pub fn unregister_observer(&mut self, key: &String) {
+        println!("Conn: unregistering observer for key {:?}", key);
```
rnewman: And here.

ffi/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+                }
+            }
+        }
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn new_store(uri: *const c_char) -> *mut Store {
+    let uri = c_char_to_string(uri);
+    let store = Store::open(&uri).expect("expected a store");
```
rnewman: I suspect you want some kind of error reporting here. This will do for now.

ffi/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+
+#[no_mangle]
+pub unsafe extern "C" fn store_destroy(store: *mut Store) {
+    let _ = Box::from_raw(store);
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn store_register_observer(store: *mut Store,
+                                                   key: *const c_char,
+                                                   attributes: *const i64,
```
rnewman: Think about whether it's clearer to explicitly use `Entid` here instead of `i64`.

ffi/src/android.rs
fluffyemily: I copied these direct over from storage protos without paying much attention. I'll update.

ffi/src/lib.rs
fluffyemily: There doesn't seem to be any written conventions around this, but examples I have seen on blog posts etc. generally go with no space.

PR #577, opened 2018-03-02
CLI improvements

rnewman:
- Print nanoseconds.
- `cargo cli` is release by default (because it often is used for timing). Use `cargo debugcli` to get a debug version.
- Time query runtime, not including printing results!
- Make our release builds really release: no debug symbols!
- Expose prepared queries in the CLI. We can't hold on to them, but we can separate the execution time and preparation time, and verify that behavior is the same as non-prepared.

### DIFF
--- a/tools/cli/src/mentat_cli/command_parser.rs
+++ b/tools/cli/src/mentat_cli/command_parser.rs
@@ -9,6 +9,7 @@
 // specific language governing permissions and limitations under the License.
 
 use combine::{
+    Parser,
     any,
     eof,
     look_ahead,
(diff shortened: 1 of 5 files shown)


rnewman:
Before these changes, a release build:
`mentat_cli: 11626392`

With these changes:
`mentat_cli: 9646304`

Saving 2MB in debug symbols!

tools/cli/src/mentat_cli/repl.rs
```diff
@@ -128,13 +158,14 @@ fn format_time(duration: Duration) {
 }
 
 /// Executes input and maintains state of persistent items.
-pub struct Repl {
+pub struct Repl<'a> {
     path: String,
     store: Store,
     timer_on: bool,
+    prepared: BTreeMap<String, PreparedQuery<'a>>,
```
grigoryk: You don't seem to be using this map?

tools/cli/src/mentat_cli/repl.rs
```diff
@@ -200,51 +232,99 @@ impl Repl {
...
+                self.toggle_timer(on);
+            },
+            Command::Transact(transaction) => {
+                self.execute_transact(transaction);
+            },
         }
 
-        let end = PreciseTime::now();
+        let end = end.unwrap_or_else(PreciseTime::now);
         if should_time {
```
grigoryk: nit: `should_time` is probably better named as `should_print_timings`? Timing happens regardless of this bool.

tools/cli/src/mentat_cli/repl.rs
```diff
@@ -200,51 +232,99 @@ impl Repl {
...
+                self.store
+                    .q_prepare(query.as_str(), None)
+                    .and_then(|mut p| {
+                        let prepare_end = PreciseTime::now();
+                        if should_time {
+                            eprint_out("Prepare time");
+                            eprint!(": ");
+                            format_time(start.to(prepare_end));
+                        }
+                        // This is a hack.
```
grigoryk: Right, because timing assumes 1-to-1 mapping between command & "thing to be timed", whereas here it's a 1-to-many.

tools/cli/src/mentat_cli/command_parser.rs
```diff
@@ -9,65 +9,71 @@
 // specific language governing permissions and limitations under the License.
 
 use combine::{
+    Parser,
```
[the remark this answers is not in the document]
rnewman: Caps come before lowercase.

tools/cli/src/mentat_cli/repl.rs
rnewman: Ah yes, I removed that.

PR #584, opened 2018-03-06
Simple aggregates again 

rnewman:
Obsoletes #498. Rebased on top of current master.

Still to do:

- [ ] TODOs for constant queries. This will involve implementing known-value aggregation: e.g., `(count ?x)` where `?x` is bound ➡️ `1`.
- [x] Porting forward [the conflicting tests from `tests/query.rs`](https://github.com/mozilla/mentat/pull/498/files#diff-71478f97b1dc3363dff408ae0a5d2e58).
- [x] Figuring out what to do with [these commented-out tests](https://github.com/mozilla/mentat/pull/498/commits/3eb898566b954da1988c20d721d87c04f6d0a596#diff-25c2dbbb356ae34ae4d3a085cc48f740)
- [x] Fixing the distinct/non-distinct case: we need to always generate a projecting subquery and aggregate in an enclosing query. `:with` controls the projection in the inner query. The simplest case is where the outer projection list is the same as the inner, and neither features aggregation — that's what we already do! — and aggregation means slightly untangling that.

### DIFF
--- a/query-algebrizer/src/clauses/predicate.rs
+++ b/query-algebrizer/src/clauses/predicate.rs
@@ -137,7 +137,7 @@ impl ConjoiningClauses {
         if shared_types == ValueTypeSet::of_one(ValueType::Instant) {
             left_v = self.resolve_instant_argument(&predicate.operator, 0, left)?;
             right_v = self.resolve_instant_argument(&predicate.operator, 1, right)?;
-        } else if !shared_types.is_empty() && shared_types.is_subset(&ValueTypeSet::of_numeric_types()) {
+        } else if !shared_types.is_empty() && shared_types.is_only_numeric() {
             left_v = self.resolve_numeric_argument(&predicate.operator, 0, left)?;
             right_v = self.resolve_numeric_argument(&predicate.operator, 1, right)?;
         } else {
--- a/query-algebrizer/src/clauses/predicate.rs
+++ b/query-algebrizer/src/clauses/predicate.rs
@@ -92,13 +92,13 @@ impl ConjoiningClauses {
         let mut left_types = self.potential_types(known.schema, &left)?
                                  .intersection(&supported_types);
         if left_types.is_empty() {
-            bail!(ErrorKind::InvalidArgument(predicate.operator.clone(), "numeric or instant", 0));
+            bail!(ErrorKind::InvalidArgumentType(predicate.operator.clone(), supported_types, 0));
         }
 
         let mut right_types = self.potential_types(known.schema, &right)?
                                   .intersection(&supported_types);
         if right_types.is_empty() {
-            bail!(ErrorKind::InvalidArgument(predicate.operator.clone(), "numeric or instant", 1));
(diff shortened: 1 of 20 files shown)


rnewman:
This is ready for you to look at, @fluffyemily.

query-algebrizer/src/types.rs
```diff
@@ -283,6 +283,9 @@ pub enum Inequality {
     GreaterThan,
     GreaterThanOrEquals,
     NotEquals,
+
+    // Ref operators.
+    Differ,
```
fluffyemily: can we have a `differ` test?

tests/query.rs
```diff
@@ -882,6 +882,58 @@ fn test_basic_aggregates() {
...
+
+    store.transact(r#"[
+        [:db/add "a" :foo/name "Alice"]
+        [:db/add "b" :foo/name "Beli"]
+        [:db/add "c" :foo/name "Carlos"]
+        [:db/add "d" :foo/name "Diana"]
+
+        ;; Alice danced with Beli.
+        [:db/add "a"  :foo/dance "ab"]
+        [:db/add "b"  :foo/dance "ab"]
```
fluffyemily: what would happen if I did `[:db/add "b" :foo/dance "ba"]` here? Does this rely upon the 'dance' ordering being identical for unpermute to work?

query-algebrizer/src/clauses/predicate.rs
```diff
@@ -134,14 +134,21 @@ impl ConjoiningClauses {
         // We expect the intersection to be Long, Long+Double, Double, or Instant.
         let left_v;
         let right_v;
+        if shared_types.is_empty() {
```
fluffyemily: We already check `shared_types.is_empty()` in line 115. The result of that check, if the set is empty, is to `mark_known_empty` and return `Ok(())`. As we do nothing else between that check and this, this code will _never_ succeed. Do we really need it?

tests/query.rs
rnewman: That would add a second dance for Beli, one in which she doesn't dance with anyone else. The only set of bindings for `?left` and `?right` would be her, and naturally she doesn't unpermute with herself, so that solo dance wouldn't contribute anything to the output.

If you were to add both Alice and Beli to that second dance, the output would be unchanged: the inner query would generate `[Alice, Beli] [Alice, Alice] [Beli, Alice] [Beli, Beli]` _twice_, once for each of the two dances; the `unpermute` step would remove `[x, x]` pairs and one half of the orders; and then the internal `DISTINCT` would remove duplicates.

I've added this to the test.

query-algebrizer/src/clauses/predicate.rs
rnewman: Thanks! Probably a screwup when rebasing a very old branch.

query-algebrizer/src/types.rs
rnewman: Added.

PR #588, opened 2018-03-14
Tests and fixes for aggregates over different or unknown types.

rnewman:


tests/query.rs
```diff
@@ -529,6 +534,133 @@ fn test_lookup() {
     let fetched_many = conn.lookup_value_for_attribute(&c, *entid, &foo_many).unwrap().unwrap();
     assert!(two_longs.contains(&fetched_many));
 }
+
+#[test]
+fn test_aggregates_type_handling() {
```
fluffyemily: Personally I would prefer a separate test function for each thing we are testing rather than one big one, with a descriptive test name.

Issue #518, opened 2018-01-17
Assertion from CLI when querying unbound variable


thomcc:
Specifically [this assert](https://github.com/mozilla/mentat/blob/158910484191dee013b299d130f48078f376119b/query-projector/src/lib.rs#L211) is hit.

The following sequence triggers it in the CLI
```
.t [{:db/ident :test/name :db/valueType :db.type/string :db/cardinality :db.cardinality/one}]
.t [{:test/name "foo"}]
.q [:find ?eid ?v :where [65537 _ ?v]]
```

The log of the queries made, specific error message, and backtrace is available here: https://gist.github.com/thomcc/e19e15341b0aa4b461ceba1a972b5bd8

rnewman:
This is because we're not verifying that the set of projected variables is a subset of the variables mentioned in the query body unioned with bound inputs. 

We can fix this either by extending `run_algebrized_query`, or by adjusting the projector's `candidate_column` function to be fallible.

rnewman:
Closed by #590.

PR #590, opened 2018-03-15
Err, don't panic, on unbound variable (#518)

rnewman:


rnewman:
I don't know why this change makes `test_type_reqs` fail; I'll look into this tomorrow.

rleungx:
This `alias` is `datoms00.v` instead of `datoms00.value_type_tag`.
Actually, you should use`extracted_types`, not `column_bindings` here.

rnewman:
> extracted_types should be used here, not column_bindings.

Yep, that'll do it! Thanks for saving me some time, @rleungx :)

PR #591, opened 2018-03-19
Simplify.

rnewman:
This has a watcher collect txid -> AttributeSet mappings each time a
transact occurs. On commit we retrieve those mappings and hand them over
to the observer service, which filters them and packages them up for
dispatch.

### DIFF
--- a/db/src/tx_observer.rs
+++ b/db/src/tx_observer.rs
@@ -8,20 +8,18 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
-use std::collections::{
-    BTreeMap,
-};
 use std::sync::{
     Arc,
-    Mutex,
     Weak,
 };
+
 use std::sync::mpsc::{
     channel,
     Receiver,
     RecvError,
     Sender,
 };
+
 use std::thread;
 
 use indexmap::{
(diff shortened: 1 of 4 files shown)


src/conn.rs
```diff
@@ -219,6 +219,7 @@ pub struct InProgress<'a, 'c> {
     use_caching: bool,
     tx_ids: AccumulatedTxids,
```
fluffyemily: we don't need this any more

src/conn.rs
```diff
@@ -219,6 +219,7 @@ pub struct InProgress<'a, 'c> {
...
     tx_observer: &'a Mutex<TxObservationService>,
```
fluffyemily: OK, this makes sense. I got into my lifetime problems by trying to keep the `TxObservationService` inside the watcher. Removing it makes everything a lot simpler.

db/src/tx_observer.rs
```diff
@@ -174,42 +154,28 @@ impl Drop for TxObservationService {
     }
 }
 
-pub struct InProgressObserverTransactWatcher<'a> {
-    collected_datoms: AttributeSet,
-    observer_service: &'a Mutex<TxObservationService>,
-    active: bool
+pub struct InProgressObserverTransactWatcher {
+    collected_attributes: AttributeSet,
```
fluffyemily: I called it `collected_datoms` mostly because the next stage will be to collect more than just attributes for more advanced observer filtering and there didn't seem much point renaming it, but this does make it more descriptive for the immediacy.

PR #598, opened 2018-03-23
WIP Query builder

fluffyemily:


fluffyemily:
OK. I approached this with the thought that we needed to entirely replace EDN strings with a programmatic API as we have done with `EntityBuilder` and `Vocabulary`, but if that is not required that does simplify this a lot.

rnewman:
> OK. I approached this with the thought that we needed to entirely replace EDN strings with a programmatic API as we have done with `EntityBuilder` and `Vocabulary`, but if that is not required that does simplify this a lot.

I think users will be relatively comfortable relatively quickly with EDN for querying, and I expect them to use the CLI, so we don't get to escape too much from it anyway.

The reason we have an entity builder is that composing EDN strings from known data structures just to push through transact is both ugly and inefficient; the same applies to queries only to the bits of a query that aren't fixed, like inputs and limits.

The reason we have a vocabulary builder is fourfold: 

1. We want compiler support/autocomplete/etc., so that it's impossible to ship code that builds an invalid attribute.
2. Vocabulary check/transact is in the hot path for getting to a store you can query; if we can save 100µs on parsing, we should! (Queries in a hot path can be prepared to avoid doing extra work; vocabs are too early for that.)
3. There is no Datomic-inherited representation of vocabulary, so there's less of a win from making it EDN-only.
4. We expect vocabulary definitions to include pre/post upgrade functions, which takes us entirely out of the realm of doing this in strings.

core/src/lib.rs
```diff
@@ -329,6 +329,91 @@ impl From<f64> for TypedValue {
     }
 }
 
+pub trait TypedValueConverter {
+    fn as_entid(self) -> Option<Entid>;
```
rnewman: All of these are `into`s, not `as`. I think `as` or `to` are what we want here:

```
fn to_entid(&self) -> Option<Entid>;
fn as_kw(&self) -> Option<&NamespacedKeyword>;
```

core/src/lib.rs
```diff
@@ -329,6 +329,91 @@ impl From<f64> for TypedValue {
...
+    fn as_kw(self) -> Option<Rc<NamespacedKeyword>>;
```
rnewman: … but if this is for the FFI, you probably don't want to be leaking `Rc`, and perhaps not even `NamespacedKeyword`…

core/src/lib.rs
rnewman: Further: what you've defined here is effectively `TryFrom` (and see `IntoThing` elsewhere in Mentat). You might consider whether you really want a generic trait here: `AsOption<T>`, perhaps?

src/query_builder.rs
```diff
@@ -0,0 +1,737 @@
...
+        Ok(QueryBuilder { sql: sql.into(), values: BTreeMap::new(), types: BTreeMap::new(), store })
+    }
+
+    pub fn bind_value<T>(mut self, var: &str, value: T) -> Self where T: Into<TypedValue> {
+        self.values.insert(Variable::from_valid_name(var), value.into());
+        self
+    }
+
+    pub fn bind_ref_from_kw(mut self, var: &str, value: NamespacedKeyword) -> Result<Self> {
+        let entid = self.store.conn().current_schema().get_entid(&value).ok_or(ErrorKind::UnknownAttribute(value.to_string()))?;
```
rnewman: I'm a bit leery of having to fetch the current schema for each attempt to bind; not only is it not free, but the schema might change between bindings!

Is there a reason why you wouldn't buffer these and resolve them in `execute`?

src/query_builder.rs
```diff
@@ -0,0 +1,737 @@
...
+        self.values.insert(Variable::from_valid_name(var), TypedValue::Ref(entid.into()));
+        Ok(self)
+    }
+
+    pub fn bind_ref<T>(mut self, var: &str, value: T) -> Self where T: Into<Entid> {
+       self.values.insert(Variable::from_valid_name(var), TypedValue::Ref(value.into()));
+       self
+    }
+
+    pub fn bind_type(mut self, var: &str, value_type: ValueType) -> Self {
```
rnewman: Perhaps `declare_type`?

src/query_builder.rs
```diff
@@ -0,0 +1,737 @@
...
+    fn as_entid(mut self) -> Option<Entid> {
+        self.0.take().map_or(None, |t| t.as_entid())
+    }
+
+    fn as_kw(mut self) -> Option<Rc<NamespacedKeyword>> {
+        self.0.take().map_or(None, |t| t.as_kw())
+    }
+
+    fn as_boolean(mut self) -> Option<bool> {
+        self.0.take().map_or(None, |t| t.as_boolean())
```
rnewman: By using `take` you make it impossible to do this:

```
let result = query.execute_scalar()?;
if let Some(yes) = result.as_boolean() {
  …
} else if let Some(long) = result.as_long() {
  …
}
```

Again, this comes back to `as`/`into`/`to` distinctions. These `Copy` types can all be `to_boolean` and use things like `self.0.map(|t| t.to_boolean())`. The non-copy types probably want `as` and `into` versions.

src/query_builder.rs
```diff
@@ -0,0 +1,737 @@
...
+
+    fn as_uuid_string(mut self) -> Option<String> {
+        self.0.take().map_or(None, |t| t.as_uuid_string())
+    }
+}
+
+pub struct CollResult(pub Vec<TypedValue>);
+
+impl CollResult {
+    fn get_row(&self, row: usize) -> Option<TypedValue> {
```
rnewman: `item` and `index`, not `row`?

src/query_builder.rs
```diff
@@ -0,0 +1,737 @@
...
+        self.0.get(row).map(|v| v.to_owned())
```
rnewman: This is a bad pattern to establish. Imagine a consumer that does:

```
let results = query.execute_coll();
let count = result.len();
for let i = 0..count {
    let s = results.col_as_uuid_string(i).unwrap();
    println!(s);
}
```

This `get_row` call will clone the UUID in order to copy the value out of the collection, and then `col_as_uuid_string` will allocate a new string.

We really want this query interface to be efficient, and that means not cloning unnecessarily. I think that means making this result interface simpler.

src/query_builder.rs
```diff
@@ -0,0 +1,737 @@
...
+            }
+        }
+        TupleResult {map}
+    }
+
+    pub fn iter(&self) -> indexmap::map::Iter<String, TypedValue> {
+        self.map.iter()
+    }
+
+    fn get_value_for_var(&self, var: &str) -> Option<TypedValue> {
```
rnewman: Hm, interesting. This is not a good fit for aggregates. Does it 'feel' better in use than indexed lookups?

src/query_builder.rs
```diff
@@ -0,0 +1,737 @@
...
+        self.map.get(var).map(|v| v.to_owned())
+    }
+
+    fn get_value_for_column(&self, col: usize) -> Option<TypedValue> {
+        self.map.get_index(col).map(|(_k, v)| v.to_owned())
```
rnewman: Again, I'd like for us to not routinely clone.

src/query_builder.rs
```diff
@@ -0,0 +1,737 @@
...
+    type Item = (String, TypedValue);
+    type IntoIter = indexmap::map::IntoIter<String, TypedValue>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.map.into_iter()
+    }
+}
+
+pub struct RelResult {
+    rows: Vec<TupleResult>,
```
rnewman: This is expensive. You're heap-allocating a new map for each row, and doing so eagerly on creation, even if the caller only wants to get the first three rows. (It's bad enough that our rel results are `Vec<Vec<TypedValue>>`; they should really be `Vec<TypedValue>` with a width.)

Does `RelResult` buy us anything over using the raw result?

src/query_builder.rs
```diff
@@ -0,0 +1,737 @@
...
+        let l_yes = report.tempids.get("l").expect("found it").clone();
+
+        #[derive(Debug, PartialEq)]
+        struct Res {
+            entid: i64,
+            boolean: bool,
+            long_val: i64,
+        };
+
+        let results = QueryBuilder::query(&mut store, r#"[:find [?v ?i]
```
rnewman: This makes me think that `store.query(…)` would be a big help.

rnewman:
What's the status of this, @fluffyemily ?

fluffyemily:
I thought I had closed it already. Closed now in favour of https://github.com/mozilla/mentat/pull/602

PR #602, opened 2018-03-29
FFI Query API

fluffyemily:
Store API for FFI. This handles Querying and fetching single value attributes by value.

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -329,6 +329,78 @@ impl From<f64> for TypedValue {
     }
 }
 
+impl TypedValue {
+    pub fn into_entid(self) -> Option<Entid> {
+        match self {
+            TypedValue::Ref(v) => Some(v),
+            _ => None,
+        }
+    }
+
+    pub fn into_kw(self) -> Option<Rc<NamespacedKeyword>> {
+        match self {
+            TypedValue::Keyword(v) => Some(v),
+            _ => None,
+        }
+    }
+
+    pub fn into_boolean(self) -> Option<bool> {
+        match self {
+            TypedValue::Boolean(v) => Some(v),
+            _ => None,
+        }
+    }
+
+    pub fn into_long(self) -> Option<i64> {
+        match self {
+            TypedValue::Long(v) => Some(v),
+            _ => None,
+        }
+    }
+
+    pub fn into_double(self) -> Option<f64> {
+        match self {
+            TypedValue::Double(v) => Some(v.into_inner()),
+            _ => None,
+        }
+    }
+
+    pub fn into_instant(self) -> Option<DateTime<Utc>> {
+        match self {
+            TypedValue::Instant(v) => Some(v),
+            _ => None,
+        }
+    }
+
+    pub fn into_timestamp(self) -> Option<i64> {
+        match self {
+            TypedValue::Instant(v) => Some(v.timestamp()),
+            _ => None,
+        }
+    }
+
+    pub fn into_string(self) -> Option<Rc<String>> {
+        match self {
+            TypedValue::String(v) => Some(v),
+            _ => None,
+        }
+    }
+
+    pub fn into_uuid(self) -> Option<Uuid> {
+        match self {
+            TypedValue::Uuid(v) => Some(v),
+            _ => None,
+        }
+    }
+
+    pub fn into_uuid_string(self) -> Option<String> {
+        match self {
+            TypedValue::Uuid(v) => Some(v.hyphenated().to_string()),
+            _ => None,
+        }
+    }
+}
+
 /// Type safe representation of the possible return values from SQLite's `typeof`
 #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialOrd, PartialEq)]
 pub enum SQLTypeAffinity {
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -57,6 +57,9 @@ pub use utils::strings::{
     string_to_c_char,
 };
 
+pub type TypedValueIterator = vec::IntoIter<TypedValue>;
+pub type TypedValueListIterator = vec::IntoIter<Vec<TypedValue>>;
+
 #[repr(C)]
 #[derive(Debug, Clone)]
 pub struct ExternTxReport {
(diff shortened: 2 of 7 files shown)


ffi/src/utils.rs
```diff
@@ -9,27 +9,40 @@
...
+    pub fn kw_from_string(mut keyword_string: String) -> NamespacedKeyword {
+        let attr_name = keyword_string.split_off(1);
+        let parts: Vec<&str> = attr_name.split("/").collect();
+        NamespacedKeyword::new(parts[0], parts[1])
     }
 
-    pub fn str_to_c_char(r_string: &str) -> *mut c_char {
-        string_to_c_char(r_string.to_string())
+    pub fn c_char_from_rc(rc_string: Rc<String>) -> *mut c_char {
+        if let Some(str_ptr) = unsafe { Rc::into_raw(rc_string).as_ref() } {
```
rnewman: Does this work instead?

```
rc.try_unwrap()                    // Get the value out of the `Rc` if we're the only owner.
  .map(CString::new)               // Make a CString.
  .map(|cs| cs.into_raw())         // Turn it into a raw pointer if successful.
  .unwrap_or_else(std::ptr::null_mut())   // Otherwise fail.
```

src/conn.rs
```diff
@@ -865,6 +870,15 @@ impl Conn {
     pub fn unregister_observer(&mut self, key: &String) {
         self.tx_observer_service.lock().unwrap().deregister(key);
     }
+
+    pub fn add_value_for_attribute<T>(&mut self, sqlite: &mut rusqlite::Connection, entid: T, attribute: NamespacedKeyword, value: TypedValue) -> Result<()> where T: Into<KnownEntid> {
```
rnewman: You probably want to take `&NamespacedKeyword` for `attribute`.

src/query_builder.rs
```diff
@@ -0,0 +1,384 @@
...
+    pub fn bind_type(mut self, var: &str, value_type: ValueType) -> Self {
+        self.types.insert(Variable::from_valid_name(var), value_type);
+        self
+    }
+
+    pub fn execute(&mut self) -> Result<QueryOutput> {
+        let values = ::std::mem::replace(&mut self.values, Default::default());
+        let types = ::std::mem::replace(&mut self.types, Default::default());
+        let query_inputs = QueryInputs::new(types, values)?;
+        let read = self.store.begin_read()?;
```
rnewman: If you make this function `execute(mut self)` and do

```
let (sql, values, types, store) = (self.sql, self.values, self.types, self.store);
let inputs = QueryInputs::new(types, values)?;
let read = store.begin_read()?;
read.q_once(&sql, inputs)
```

can we avoid the `mem::replace`s?

src/query_builder.rs
```diff
@@ -0,0 +1,384 @@
...
+    }
+
+    pub fn execute_scalar(mut self) -> Result<Option<TypedValue>> {
+        let results = self.execute()?;
+        results.into_scalar().map_err(|e| e.into())
+    }
+
+    pub fn execute_coll(mut self) -> Result<Vec<TypedValue>> {
+        let results = self.execute()?;
+        results.into_coll().map_err(|e| e.into())
```
rnewman: I think you probably want to import the `IntoResult` trait and use `into_coll_result` and friends.

core/src/lib.rs
```diff
@@ -329,6 +329,84 @@ impl From<f64> for TypedValue {
     }
 }
 
+impl From<i64> for TypedValue {
+    fn from(value: i64) -> TypedValue {
+        TypedValue::Long(value)
```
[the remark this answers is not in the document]
rnewman: I should add to this: because `Entid` is a type alias, not a newtype, the compiler can't help. We do have a `KnownEntid` newtype, but `Entid` still exists (for good reason), and so we have multiple distinct uses of what the compiler sees as a single type.

ffi/src/utils.rs
fluffyemily: nope. `try_unwrap` works if you are calling from Rust, but once you invoke this from Swift, `try_unwrap` fails. This was the _only_ way I could get the `String` out of the `Rc`, but I also figured there must be a better way.

src/query_builder.rs
fluffyemily: Not if we want to maintain the non-consuming builder pattern and therefore avoid all of the boxing and unboxing in the FFI. Either, we're consuming and have an box-heavy FFI layer, in which case this is possible. Or we are non-consuming, have a smaller FFI layer that deals with references and use `mem::replace`. As far as I can tell, anyway.

src/query_builder.rs
```diff
@@ -0,0 +1,389 @@
...
+
+        #[derive(Debug, PartialEq)]
+        struct Res {
+            entid: i64,
+            boolean: bool,
+            long_val: i64,
+        };
+
+        let mut results: Vec<Res> = QueryBuilder::new(&mut store, r#"[:find ?x ?v ?i
+                                              :where [?x :foo/boolean ?v]
```
rnewman: Indenting for all of these test queries, plz.

src/conn.rs
```diff
@@ -865,6 +870,18 @@ impl Conn {
     pub fn unregister_observer(&mut self, key: &String) {
         self.tx_observer_service.lock().unwrap().deregister(key);
     }
+
+    // TODO: expose the entity builder over FFI and remove the need for this function entirely
+    // It's really only here in order to keep the FFI layer as thin as possible.
+    // Once the entity builder is exposed, we can perform all of these functions over FFI from the client.
+    pub fn add_value_for_attribute<T>(&mut self, sqlite: &mut rusqlite::Connection, entid: T, attribute: NamespacedKeyword, value: TypedValue) -> Result<()> where T: Into<KnownEntid> {
```
rnewman: Just call this `assert_datom`, 'cos that's what it is.

src/conn.rs
```diff
@@ -578,6 +579,10 @@ impl Store {
     pub fn unregister_observer(&mut self, key: &String) {
         self.conn.unregister_observer(key);
     }
+
+    pub fn add_value_for_attribute<T>(&mut self, entid: T, attribute: NamespacedKeyword, value: TypedValue) -> Result<()> where T: Into<KnownEntid> {
```
rnewman: `assert_datom`, same.

ffi/src/lib.rs
```diff
@@ -178,3 +555,131 @@ pub unsafe extern "C" fn store_sync(store: *mut Store, user_uuid: *const c_char,
...
+#[no_mangle]
+pub unsafe extern "C" fn store_destroy(obj: *mut Store) {
+    if !obj.is_null() {
+        let _ = Box::from_raw(obj);
+    }
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn typed_value_destroy(obj: *mut TypedValue) {
+    if !obj.is_null() {
```
rnewman:
```
macro_rules! define_destructor {
    ( $name:ident, $t:ty ) => {{
        #[no_mangle]
        pub unsafe extern "C" fn $name(obj: *mut $t) {
            if !obj.is_null() { let _ = Box::from_raw(obj); }
        }
    }}
}
```

?

core/src/lib.rs
```diff
@@ -329,6 +329,78 @@ impl From<f64> for TypedValue {
     }
 }
 
+impl TypedValue {
+    pub fn into_entid(self) -> Option<Entid> {
```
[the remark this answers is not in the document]
fluffyemily: https://github.com/mozilla/mentat/issues/628

src/query_builder.rs
fluffyemily: Fucking VSCode keeps realigning them, but only in tests. Grrr.

ffi/src/lib.rs
fluffyemily: ooh, I didn't know you could do that.

fluffyemily: Oh, weird. Adding that function means that _no_ FFI functions can be found from Swift. I guess another lesson - no macros in FFI!

Issue #603, opened 2018-03-30
Don't spawn a CommandExecutor when there are no observers


rnewman:

rnewman:
Fixed in #604.

PR #604, opened 2018-03-30
Don't create a CommandExecutor if there are no observers. (#603) 

rnewman:


rnewman:
Review ping, @fluffyemily

Issue #379, opened 2017-03-20
[tx] Handling retracting schema characteristics, like `:db/cardinality` and `:db/unique`.


ncalexan:
After #370 (fixing #294 and #295), we still need to implement handling `[:db/retract :attribute :db/* ...]` of schema characteristics.  This might also address `[:db/retractEntity :attribute]` if it comes after https://github.com/mozilla/mentat/issues/378.

The tricky thing to ensure here is that we fall back to the default correctly, and that the "schema" materialized view does not get out of sync with the in-memory state.

rnewman:
Fixed!

PR #626, opened 2018-04-03
Allow retraction of some schema attributes. (#379)

rnewman:
The error messages are a bit ugly, but I expect users won't be using this interface directly.


rnewman:
https://github.com/mozilla/mentat/commit/27dde378e0149eff22b6450e771bd064eae9902c

PR #627, opened 2018-04-04
Vocabulary migration

rnewman:


rnewman:
@fluffyemily, over to you.

Nick already reviewed #379, so you can skip over that commit if you like.

I left in the projector refactor commit in this stack, but it's just splitting into files, so I suggest you skip that, too.

326ed05 is the meat of this; I hope the docstrings in `src` and the thorough example in `test` give you enough to work with.

src/vocabulary.rs
```diff
@@ -120,19 +128,62 @@ pub type Version = u32;
...
+            version: version,
+            attributes: attributes.into(),
+            pre: Definition::no_op,
+            post: Definition::no_op,
+        }
+    }
+
+    /// Called with an in-progress transaction and the previous vocabulary version
+    /// if the definition's version is later than that of the vocabulary in the store.
+    fn pre(&self, ip: &mut InProgress, from: &Vocabulary) -> Result<()> {
```
fluffyemily: In the interests of good usage documentation, perhaps we could show an example usage here and for `post`? I know we have some in the tests, but external users won't necessarily have access to those and these make more sense when seeing how they are used.

rnewman:
I added more documentation, cleaned up the existing stuff, and flattened the stack. Here it goes!

PR #632, opened 2018-04-09
Extract test macros.

ncalexan:
@rnewman can spend a minute trying to make this compile?  I vaguely remember there's magic around the order of imports, etc, but it's pretty frustrating.

My motivation is to use the handy `db` testing apparatus in the `tolstoy` crate (or a new crate), so that I can write low-level tests for synchronization without having to build the `db` crate frequently (which is very slow since it's large).


rnewman:
Let me know when you want me to look at this, @ncalexan.

ncalexan:
@grigoryk landed this in https://github.com/mozilla/mentat/pull/787.

PR #633, opened 2018-04-09
Fix (the ?foo)

rnewman:
I incorrectly failed to exclude corresponding-projection vars from grouping. This fixes that error, and also makes error checking more thorough.


rnewman:
> Is it possible to include a test (example) where the distinction is clear?

The test included in this PR shows the distinction — one query returns two results, and the other returns one. The only difference is `the`: without it we group by that value, getting the maximum visit date for each title; with it we don't group by value, instead getting the title that corresponds to the maximum visit date.

Would it help for me to add that verbiage to the test?

rnewman:
Review comments addressed, @ncalexan.

PR #635, opened 2018-04-12
Closes #634 - Fix variables in predicates

### LINKED ISSUE #634: Variable not bound when passing to predicate
`[:find [?e ?d] :in ?now :where [?e :foo/instant ?d] [(< ?d ?now)]]` fails with the error: `Error(QueryError(UnboundVariable(PlainSymbol("?now")))`. This is because we are failing to bind variables when calling `resolve_x_argument` in `resolve.rs` from `predicate.rs`.

fluffyemily:


query-algebrizer/src/clauses/resolve.rs
```diff
@@ -45,11 +45,18 @@ impl ConjoiningClauses {
...
             FnArg::Variable(var) => {
-                self.constrain_var_to_numeric(var.clone());
-                self.column_bindings
-                    .get(&var)
-                    .and_then(|cols| cols.first().map(|col| QueryValue::Column(col.clone())))
-                    .ok_or_else(|| Error::from_kind(ErrorKind::UnboundVariable(var.name())))
+                match self.bound_value(&var) {
+                    // The type is already known if it's a bound variable….
+                    Some(TypedValue::Long(v)) => Ok(QueryValue::TypedValue(TypedValue::Long(v))),
+                    Some(TypedValue::Double(v)) => Ok(QueryValue::TypedValue(TypedValue::Double(v))),
```
rnewman: This isn't quite right. If it's already bound to the wrong kind of value, we'll still fall through to the column binding and get an `UnboundVariable` error.

I think you want something like:

```
Some(v) => {
    bail!(ErrorKind::InputTypeDisagreement(var.name.clone(), ValueType::Long, v.value_type()));
},
```

and similar on line 85.

rnewman: I also think you want to add `ValueType::is_numeric` to `core`, and here do:

```
Some(v) if v.value_type().is_numeric() => Ok(QueryValue::TypedValue(v)),
Some(v) => { bail!(…); },
```

— we will at some point add `BigInteger` support, which would introduce a bug here.

fluffyemily: Hmm, I don't think I can do this. The guard statement claims ownership so it's not available to the case body. If I make v a reference,  the guard works but I cannot claim ownership to return the value inside the body without cloning it.

Also looks like there is ongoing discussion about this https://github.com/rust-lang/rfcs/issues/2036

I will try and do something else but similar. This is where I miss Swifts chaning if-lets and guards so I could do:

```
guard let v = self.bound_value(&var),
          v.value_type().is_numeric() else { ... }
```

query-algebrizer/src/clauses/resolve.rs
```diff
@@ -73,11 +82,16 @@ impl ConjoiningClauses {
...
         match arg {
             FnArg::Variable(var) => {
-                self.constrain_var_to_type(var.clone(), ValueType::Instant);
-                self.column_bindings
-                    .get(&var)
-                    .and_then(|cols| cols.first().map(|col| QueryValue::Column(col.clone())))
-                    .ok_or_else(|| Error::from_kind(ErrorKind::UnboundVariable(var.name())))
+                match self.bound_value(&var) {
+                    Some(TypedValue::Instant(v)) => Ok(QueryValue::TypedValue(TypedValue::Instant(v))),
+                    _ => {
```
rnewman: Same review comment as before:

> This isn't quite right. If it's already bound to the wrong kind of value, we'll still fall through to the column binding and get an `UnboundVariable` error.

You need a similar `bail` line here as there is on line 53.

rnewman:
@fluffyemily don't forget to wrap this up…

fluffyemily:
Finally. Not sure why I forgot about this.

PR #636, opened 2018-04-12
Add iOS Framework

fluffyemily:
An iOS framework wrapping Mentat's API over FFI.

How to build for iOS

```
cd ffi
cargo lipo --release
```

When the build is complete, open `sdks/swift/Mentat/Mentat.xcproj` in Xcode, do a force clean (Cmd + alt + Shift + k).

To run tests, ensure that the `MentatDebug` target is selected.

@rnewman can you please check something for me? About 3/4 of the way through developing this, something weird happened with `libmentat_ffi.a`. It stopped bundling `sqlite` up when running `cargo lipo`, however if I build Toodle with the current mentat and linked that, everything was fine. Can you please see if you have the same problem, or whether it's just my machine being weird?

### DIFF
--- a/ffi/Cargo.toml
+++ b/ffi/Cargo.toml
@@ -4,11 +4,12 @@ version = "0.0.1"
 authors = ["Emily Toop <etoop@mozilla.com>"]
 
 [lib]
-name = "mentat"
-crate-type = ["staticlib", "cdylib"]
+name = "mentat_ffi"
+crate-type = ["lib", "staticlib", "cdylib"]
 
 [dependencies]
-libc = "0.2"
+libc = "0.2.32"
 
 [dependencies.mentat]
-path = ".."
+path = "../"
+
--- a/ffi/Cargo.toml
+++ b/ffi/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mentat_ffi"
-version = "0.1.0"
+version = "0.0.1"
 authors = ["Emily Toop <etoop@mozilla.com>"]
 
 [lib]
--- a/ffi/Cargo.toml
+++ b/ffi/Cargo.toml
@@ -3,6 +3,10 @@ name = "mentat_ffi"
 version = "0.1.0"
 authors = ["Emily Toop <etoop@mozilla.com>"]
 
+[lib]
+name = "mentat"
+crate-type = ["staticlib", "cdylib"]
+
 [dependencies]
 libc = "0.2"
 
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -137,6 +137,19 @@ pub extern "C" fn store_open(uri: *const c_char) -> *mut Store {
 
 // TODO: begin_transaction
 
+#[no_mangle]
+pub unsafe extern "C" fn store_transact(store: *mut Store, transaction: *const c_char) -> *mut ExternResult {
+    let store = &mut*store;
+    let transaction = c_char_to_string(transaction);
+    let result = store.begin_transaction().map(|mut in_progress| {
+        in_progress.transact(&transaction).map(|tx_report| {
+            in_progress.commit()
+                       .map(|_| tx_report)
+        })
+    });
+    Box::into_raw(Box::new(result.into()))
+}
+
 // TODO: cache
 
 // TODO: q_once
@@ -468,12 +481,6 @@ pub unsafe extern "C" fn value_at_index_as_uuid(values: *mut Vec<TypedValue>, in
     let value = result.get(index as usize).expect("No value at index");
     string_to_c_char(value.clone().into_uuid_string().expect("Typed value cannot be coerced into a Uuid"))
 }
-
-// TODO: q_prepare
-
-// TODO: q_explain
-
-// TODO: lookup_values_for_attribute
 
 #[no_mangle]
 pub unsafe extern "C" fn store_value_for_attribute(store: *mut Store, entid: i64, attribute: *const c_char) ->  *mut ExternResult {
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -42,6 +42,7 @@ pub use mentat::{
     Syncable,
     TypedValue,
     TxObserver,
+    TxReport,
     Uuid,
     ValueType,
     Variable,
@@ -62,16 +63,23 @@ pub type TypedValueListIterator = vec::IntoIter<Vec<TypedValue>>;
 
 #[repr(C)]
 #[derive(Debug, Clone)]
-pub struct ExternTxReport {
+pub struct ExternTempId {
+    pub key: *const c_char,
+    pub value: i64,
+}
+
+#[repr(C)]
+#[derive(Debug, Clone)]
+pub struct TransactionChange {
     pub txid: Entid,
-    pub changes: Box<[Entid]>,
     pub changes_len: usize,
+    pub changes: Box<[Entid]>,
 }
 
 #[repr(C)]
 #[derive(Debug)]
-pub struct ExternTxReportList {
-    pub reports: Box<[ExternTxReport]>,
+pub struct TxChangeList {
+    pub reports: Box<[TransactionChange]>,
     pub len: usize,
 }
 
@@ -141,13 +149,36 @@ pub extern "C" fn store_open(uri: *const c_char) -> *mut Store {
 pub unsafe extern "C" fn store_transact(store: *mut Store, transaction: *const c_char) -> *mut ExternResult {
     let store = &mut*store;
     let transaction = c_char_to_string(transaction);
-    let result = store.begin_transaction().map(|mut in_progress| {
-        in_progress.transact(&transaction).map(|tx_report| {
+    let result = store.begin_transaction().and_then(|mut in_progress| {
+        in_progress.transact(&transaction).and_then(|tx_report| {
             in_progress.commit()
                        .map(|_| tx_report)
         })
     });
     Box::into_raw(Box::new(result.into()))
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn tx_report_get_entid(tx_report: *mut TxReport) -> i64 {
+    let tx_report = &*tx_report;
+    tx_report.tx_id
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn tx_report_get_tx_instant(tx_report: *mut TxReport) -> i64 {
+    let tx_report = &*tx_report;
+    tx_report.tx_instant.timestamp()
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn tx_report_entity_for_temp_id(tx_report: *mut TxReport, tempid: *const c_char) -> *mut i64 {
+    let tx_report = &*tx_report;
+    let key = c_char_to_string(tempid);
+    if let Some(entid) = tx_report.tempids.get(&key) {
+        Box::into_raw(Box::new(entid.clone()))
+    } else {
+        std::ptr::null_mut()
+    }
 }
 
 // TODO: cache
@@ -159,14 +190,6 @@ pub unsafe extern "C" fn store_query<'a>(store: *mut Store, query: *const c_char
     let store = &mut*store;
     let query_builder = QueryBuilder::new(store, query);
     Box::into_raw(Box::new(query_builder))
-}
-
-#[no_mangle]
-pub unsafe extern "C" fn query_builder_bind_int(query_builder: *mut QueryBuilder, var: *const c_char, value: c_int) {
-    let var = c_char_to_string(var);
-    let query_builder = &mut*query_builder;
-    let value = value as i32;
-    query_builder.bind_value(&var, value);
 }
 
 #[no_mangle]
@@ -294,9 +317,9 @@ pub unsafe extern "C" fn typed_value_as_kw(typed_value: *mut TypedValue) ->  *co
 
 //as_boolean
 #[no_mangle]
-pub unsafe extern "C" fn typed_value_as_boolean(typed_value: *mut TypedValue) ->  bool {
+pub unsafe extern "C" fn typed_value_as_boolean(typed_value: *mut TypedValue) -> i32 {
     let typed_value = Box::from_raw(typed_value);
-    typed_value.into_boolean().expect("Typed value cannot be coerced into a Boolean")
+    if typed_value.into_boolean().expect("Typed value cannot be coerced into a Boolean") { 1 } else { 0 }
 }
 
 //as_double
@@ -310,7 +333,8 @@ pub unsafe extern "C" fn typed_value_as_double(typed_value: *mut TypedValue) -> 
 #[no_mangle]
 pub unsafe extern "C" fn typed_value_as_timestamp(typed_value: *mut TypedValue) ->  i64 {
     let typed_value = Box::from_raw(typed_value);
-    let val = typed_value.into_timestamp().expect("Typed value cannot be coerced into a Timestamp");
+    let t = typed_value.value_type();
+    let val = typed_value.into_timestamp().expect(&format!("Typed value of type {:?} cannot be coerced into a Timestamp", t));
     val
 }
 
@@ -326,6 +350,13 @@ pub unsafe extern "C" fn typed_value_as_string(typed_value: *mut TypedValue) -> 
 pub unsafe extern "C" fn typed_value_as_uuid(typed_value: *mut TypedValue) ->  *const c_char {
     let typed_value = Box::from_raw(typed_value);
     string_to_c_char(typed_value.into_uuid_string().expect("Typed value cannot be coerced into a Uuid"))
+}
+
+//value_type
+#[no_mangle]
+pub unsafe extern "C" fn typed_value_value_type(typed_value: *mut TypedValue) ->  ValueType {
+    let typed_value = &*typed_value;
+    typed_value.value_type()
 }
 
 #[no_mangle]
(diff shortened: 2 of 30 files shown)


ffi/Cargo.toml
```diff
@@ -1,10 +1,15 @@
...
+version = "0.0.1"
 authors = ["Emily Toop <etoop@mozilla.com>"]
 
+[lib]
+name = "mentat_ffi"
+crate-type = ["lib", "staticlib", "cdylib"]
+
 [dependencies]
-libc = "0.2"
+libc = "0.2.32"
```
rnewman: What's the reason for pegging to such a precise release?

ffi/src/lib.rs
```diff
@@ -297,7 +333,8 @@ pub unsafe extern "C" fn typed_value_as_double(typed_value: *mut TypedValue) ->
 #[no_mangle]
 pub unsafe extern "C" fn typed_value_as_timestamp(typed_value: *mut TypedValue) ->  i64 {
     let typed_value = Box::from_raw(typed_value);
-    let val = typed_value.into_timestamp().expect("Typed value cannot be coerced into a Timestamp");
+    let t = typed_value.value_type();
+    let val = typed_value.into_timestamp().expect(&format!("Typed value of type {:?} cannot be coerced into a Timestamp", t));
```
rnewman: All of these strings end up in our binary. I'm curious whether we can reduce that impact.

ffi/src/lib.rs
```diff
@@ -383,7 +427,7 @@ pub unsafe extern "C" fn values_iter_next_as_double(iter: *mut TypedValueIterato
 #[no_mangle]
 pub unsafe extern "C" fn values_iter_next_as_timestamp(iter: *mut TypedValueIterator) ->  *const i64 {
     let iter = &mut *iter;
-    iter.next().map_or(std::ptr::null_mut(), |v| v.into_timestamp().expect("Typed value cannot be coerced into a Timestamp") as *const i64)
+    iter.next().map_or(std::ptr::null_mut(), |v| { let t = v.value_type(); v.into_timestamp().expect(&format!("Typed value of type {:?} cannot be coerced into a Timestamp", t)) as *const i64 })
```
rnewman: Can we get some more newlines?

query-algebrizer/src/clauses/resolve.rs
```diff
@@ -45,11 +45,18 @@ impl ConjoiningClauses {
         use self::FnArg::*;
         match arg {
             FnArg::Variable(var) => {
-                self.constrain_var_to_numeric(var.clone());
-                self.column_bindings
-                    .get(&var)
-                    .and_then(|cols| cols.first().map(|col| QueryValue::Column(col.clone())))
-                    .ok_or_else(|| Error::from_kind(ErrorKind::UnboundVariable(var.name())))
+                match self.bound_value(&var) {
```
rnewman: I think this change snuck into this PR…

sdks/swift/Mentat/Mentat.xcodeproj/xcuserdata/emilytoop.xcuserdatad/xcschemes/xcschememanagement.plist
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
rnewman: I am pretty sure we shouldn't check in `emilytoop.xcuserdatad`.

sdks/swift/Mentat/Mentat/Core/TypedValue.swift
```diff
@@ -0,0 +1,75 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
rnewman: Let's not force our users to conform to both the MPL2.0 and Apache. Switch all of these to Apache.

sdks/swift/Mentat/Mentat/Core/TypedValue.swift
```diff
@@ -0,0 +1,75 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+import Mentatlib
```
rnewman: Can we get a better name for this? It's not clear why we have two "Mentat lib"s.

Perhaps `libmentat` for the native code?

sdks/swift/Mentat/Mentat/Core/TypedValue.swift
```diff
@@ -0,0 +1,75 @@
...
+
+    var valueType: ValueType {
+        return typed_value_value_type(self.raw!)
+    }
+
+    func asLong() -> Int64 {
+        defer {
+            self.raw = nil
+        }
+        return typed_value_as_long(self.raw!)
```
rnewman: So if I call

```swift
logger.info("My value is \(tv.asLong())")
SomeAPI.getThing(tv.asLong())
```

my code is going to crash, no?

sdks/swift/Mentat/Mentat/Core/TypedValue.swift
```diff
@@ -0,0 +1,75 @@
...
+            self.raw = nil
+        }
+        return typed_value_as_entid(self.raw!)
+    }
+
+    func asKeyword() -> String {
+        defer {
+            self.raw = nil
+        }
+        return String(cString: typed_value_as_kw(self.raw!))
```
rnewman: So we have Rust copying a Rust `String` to be a C string, and then Swift copying the C string to be a Swift `String`. Please file an issue to not do that; it's a performance papercut for things like keywords, but for larger strings it'll be a potential crasher. It should be possible for us to directly turn a Rust `String` into a Swift `String`, copying no more than once.

sdks/swift/Mentat/Mentat/Core/TypedValue.swift
```diff
@@ -0,0 +1,75 @@
...
+            self.raw = nil
+        }
+        return String(cString: typed_value_as_string(self.raw!))
+    }
+
+    func asUUID() -> UUID? {
+        defer {
+            self.raw = nil
+        }
+        return UUID(uuidString: String(cString: typed_value_as_uuid(self.raw!)))
```
rnewman: File an issue to use `uuid_t` instead of round-tripping through strings. The native representation of a UUID is bytes, not a string.

sdks/swift/Mentat/Mentat/Mentat.swift
```diff
@@ -0,0 +1,188 @@
...
+            throw MentatError(message: "Unspecified Error")
+        }
+        return TxReport(raw: success)
+    }
+
+    func entidForAttribute(attribute: String) -> Int64 {
+        return Int64(store_entid_for_attribute(self.raw, attribute))
+    }
+
+    func sync_now() -> Bool {
```
rnewman: Let's comment this out or delete it until it's ready.

sdks/swift/Mentat/Mentat/Mentat.swift
```diff
@@ -0,0 +1,188 @@
...
+    func query(query: String) -> Query {
+        return Query(raw: store_query(self.raw, query))
+    }
+
+    func value(forAttribute attribute: String, ofEntity entid: Int64) -> TypedValue? {
+        let result = store_value_for_attribute(self.raw, entid, attribute).pointee
+        guard let success = result.ok else {
+            if let error = result.err {
+                let str = String(cString: error)
+                print("Error: \(str)")
```
rnewman: Just make this method throw?

sdks/swift/Mentat/Mentat/Mentat.swift
```diff
@@ -0,0 +1,188 @@
...
+        return TypedValue(raw: success)
+    }
+
+    func set(date value: Date, forAttribute attribute: String, onEntity entid: Int64) throws {
+        let result = store_set_timestamp_for_attribute_on_entid(self.intoRaw(), entid, attribute, value.toMicroseconds())
+        guard let err = result.pointee.err else {
+            return
+        }
+
+        throw QueryError.executionFailed(message: String(cString: err))
```
rnewman: `QueryError` -> `TransactError`?

sdks/swift/Mentat/Mentat/Mentat.swift
```diff
@@ -0,0 +1,188 @@
...
+    func set(string value: String, forAttribute attribute: String, onEntity entid: Int64) throws {
+        let result = store_set_string_for_attribute_on_entid(self.intoRaw(), entid, attribute, value)
+        guard let err = result.pointee.err else {
+            return
+        }
+
+        throw QueryError.executionFailed(message: String(cString: err))
+    }
+
+    func set(uuid value: UUID, forAttribute attribute: String, onEntity entid: Int64) throws {
```
rnewman: I'm not a big fan of these simple setter methods. I think they'll encourage code like:

```
store.set(double: 99.0, forAttribute: ":my/score", onEntity: 65536)?;
store.set(double: 128.5, forAttribute: ":my/height", onEntity: 65536)?;
```

which will run _two independent transactions_, dramatically increasing the amount of data written (one timestamp and tx metadata for each!) and thrashing the disk.

I'd even go so far as to say: expose `InProgress` and move these methods there, or delete them altogether.

sdks/swift/Mentat/Mentat/Query/Query.swift
```diff
@@ -0,0 +1,218 @@
...
+                return
+            }
+            let rows = RelResult(raw: rowsPtr)
+            for row in rows {
+                map(row, nil)
+            }
+        }
+    }
+
+    func execute(callback: @escaping (RelResult?, QueryError?) -> Void) throws {
```
rnewman: Maybe `run` is friendlier than `execute`?

sdks/swift/Mentat/Mentat/Query/Query.swift
```diff
@@ -0,0 +1,218 @@
...
+        return self
+    }
+
+    func executeMap(map: @escaping (TupleResult?, QueryError?) -> Void) throws {
+        guard let r = self.raw else {
+            throw QueryError.builderConsumed
+        }
+
+        DispatchQueue.global(qos: .background).async {
+            let result = query_builder_execute(r)
```
rnewman: So the query builder here is capturing a `Store` instance as its hidden raw pointer? I don't think I understand how this works, because I don't see where the reference to the SQLite connection and the `Store` are managed.

sdks/swift/Mentat/Mentat/Transact/TxReport.swift
```diff
@@ -0,0 +1,30 @@
+//
```
rnewman: Nit: there are these surplus comment lines at the top of some of these files. Please make sure you clean them up when switching licenses.

sdks/swift/Mentat/Mentat/Transact/TxReport.swift
```diff
@@ -0,0 +1,30 @@
...
+
+    public var txId: Int64 {
+        return tx_report_get_entid(self.raw)
+    }
+
+    public var txInstant: Date {
+        return Date(timeIntervalSince1970: TimeInterval(tx_report_get_tx_instant(self.raw)))
+    }
+
+    public func entidForTmpId(tmpId: String) -> Int64? {
```
rnewman: temp, not tmp, throughout. Datomic calls them `tempid`.

sdks/swift/Mentat/MentatTests/MentatTests.swift
```diff
@@ -0,0 +1,836 @@
...
+            let schemaUrl = bundle.url(forResource: "cities", withExtension: "schema", subdirectory: "fixtures")!
+            let schema = try String(contentsOf: schemaUrl)
+            self.schema = schema
+            return schema
+        }
+
+        return schema
+    }
+
+    func readEdn() throws -> String {
```
rnewman: `readSeattleData`?

sdks/swift/Mentat/MentatTests/MentatTests.swift
```diff
@@ -0,0 +1,836 @@
...
+        return report
+    }
+
+    func transactEdn(mentat: Mentat) throws -> TxReport {
+        let edn = try readEdn()
+        let report = try mentat.transact(transaction: edn)
+        return report
+    }
+
+    func newStore() -> Mentat {
```
rnewman: `openAndInitializeStore`?

sdks/swift/Mentat/MentatTests/MentatTests.swift
```diff
@@ -0,0 +1,836 @@
...
+        } catch {
+            assertionFailure(error.localizedDescription)
+        }
+    }
+
+    func testQueryScalar() {
+        let mentat = newStore()
+        let query = "[:find ?n . :in ?name :where [(fulltext $ :community/name ?name) [[?e ?n]]]]"
+        let expect = expectation(description: "Query is executed")
+        try! mentat.query(query: query).bind(varName: "?name", toString: "Wallingford").executeScalar(callback: { (scalarResult, error) in
```
rnewman: What happens if I write:

```
let q1 = mentat.query(query: "[:find ?x :where [?x _ _]]")
let q2 = mentat.query(query: "[:find ?x :where [_ _ ?x]]")
```

or

```
try! mentat.query(…).… {
    try! mentat.query(…)
```

sdks/swift/Mentat/MentatTests/MentatTests.swift
```diff
@@ -0,0 +1,836 @@
...
+            assert(error == nil, "Unexpected error: \(String(describing: error))")
+            guard let result = scalarResult?.asString() else {
+                return assertionFailure("No String value received")
+            }
+            assert(result == "KOMO Communities - Wallingford")
+            expect.fulfill()
+        })
+        waitForExpectations(timeout: 1) { error in
```
rnewman: I don't much like expectations; they were unreliable in Firefox tests. I would prefer to give the caller control over which queue runs a query, so they can kick off a background task and run a bunch of queries in a row… and that allows us to write serial tests, too.

sdks/swift/Mentat/MentatTests/MentatTests.swift
```diff
@@ -0,0 +1,836 @@
...
+        } catch {
+            assertionFailure(error.localizedDescription)
+        }
+    }
+
+    func test2TransactEntities() {
+        do {
+            let mentat = Mentat()
+            let _ = try self.transactSchema(mentat: mentat)
+            let report = try self.transactEdn(mentat: mentat)
```
rnewman: I take it it's currently not possible to transact both of these in the same transaction?

sdks/swift/Mentat/MentatTests/MentatTests.swift
```diff
@@ -0,0 +1,836 @@
...
+        let formatter = DateFormatter()
+        formatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ssZZZZZ"
+        let expectedDate = formatter.date(from: "2017-01-01T11:00:00+00:00")
+
+        try! mentat.query(query: query)
+            .bind(varName: "?e", toReference: aEntid)
+            .executeScalar { (value, error) in
+                assert(error == nil, "Unexpected error: \(String(describing: error))")
+                XCTAssertNotNil(value)
+                assert(value?.asDate() == expectedDate)
```
rnewman: Duplicate this line.

sdks/swift/Mentat/MentatTests/MentatTests.swift
```diff
@@ -0,0 +1,836 @@
...
+        let report = self.populateWithTypesSchema(mentat: mentat)!
+        let aEntid = report.entidForTmpId(tmpId: "a")!
+        let query = "[:find ?v . :in ?e :where [?e :foo/string ?v]]"
+        let expect = expectation(description: "Query is executed")
+        try! mentat.query(query: query)
+            .bind(varName: "?e", toReference: aEntid)
+            .executeScalar { (value, error) in
+                assert(error == nil, "Unexpected error: \(String(describing: error))")
+                XCTAssertNotNil(value)
+                assert(value?.asString() == "The higher we soar the smaller we appear to those who cannot fly.")
```
rnewman: Duplicate this line, too.

ffi/Cargo.toml
fluffyemily: I think that was a result of my trying to find the solve the above mentioned build issue by trying to replicate the versioning in Toodle's Cargo.lock exactly and it accidentally got checked in.

query-algebrizer/src/clauses/resolve.rs
fluffyemily: I did it here originally to get the tests to pass. Will rebase this patch onto the other one instead.

sdks/swift/Mentat/Mentat/Query/Query.swift
fluffyemily: When we create a `Query` we pass in a reference to store (see `Mentat.query`). We return a pointer to the query builder which we store in `Query`. Therefore we have:
`Mentat` holding the pointer to `Store` as it's raw pointer.
`Query` holding the pointer to `QueryBuilder` as it's raw pointer.
`QueryBuilder` holding a reference to `Store` in Rust.

sdks/swift/Mentat/Mentat/Core/TypedValue.swift
fluffyemily: Yes, just as it would if you tried to do that when getting a value twice from a result in `rusqlite` as I tried to replicate that behaviour. I was trying to make the the behaviour as similar as possible to what users might already have used, but clearly that has not worked as you are expecting it to behave differently. I will change these to be non-consuming.

sdks/swift/Mentat/MentatTests/MentatTests.swift
fluffyemily: Addressing the later comment about not threading inside `run` fixes this. Both tests now pass.

fluffyemily:
@rnewman I've made the changes related to your comments on https://github.com/mozilla/mentat/pull/643 so I could make the relevant changes to both the iOS and Android versions. I shall close this PR and move all work over to #643

PR #639, opened 2018-04-12
Introduce RelResult rather than Vec<Vec<TypedValue>>.

rnewman:


rnewman:
I cleaned up and improved this, @fluffyemily .

core/src/types.rs
```diff
@@ -172,6 +186,134 @@ pub enum TypedValue {
     Uuid(Uuid),                        // It's only 128 bits, so this should be acceptable to clone.
 }
 
+/// The values stored in a `StructuredMap` can be:
```
ncalexan: The comment refers to `StructuredMap`, but it's attached to `Binding`.  Huh?

core/src/types.rs
```diff
@@ -172,6 +186,134 @@ pub enum TypedValue {
...
+/// * Vecs of structured values, for multi-valued component attributes or nested expressions.
+/// * Single structured values, for single-valued component attributes or nested expressions.
+/// * Single typed values, for simple attributes.
+///   Note that the `TypedValue` enum is inlined here for convenience: `Binding` is
+///   an expansion of it.
+///
+/// Datomic also supports structured _bindings_; at present Mentat does not, but this type
+/// would also serve that purpose.
+///
+/// Note that maps are not ordered….
```
ncalexan: It seems like you think this is significant.  If it is, explain why.

core/src/types.rs
rnewman: This comment is cleaned up later.

core/src/types.rs
rnewman: … 'cos it means this doesn't implement `Ord`.

rnewman:
Using this PR for the `RelResult` change, then landing the file split as a separate commit, then I'll open and land another PR for `Binding`.

PR #640, opened 2018-04-13
Implement (tx-log ...) function.

ncalexan:
`tx-log` allows to query the transaction log efficiently.

@rnewman, can I get early feedback on this?  I haven't added the underlying `transaction_log` table yet; that's next.

### DIFF
--- a/query-algebrizer/src/clauses/tx_log_api.rs
+++ b/query-algebrizer/src/clauses/tx_log_api.rs
@@ -1,0 +1,394 @@
+// Copyright 2018 Mozilla
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+use mentat_core::{
+    ValueType,
+};
+
+use mentat_query::{
+    Binding,
+    FnArg,
+    SrcVar,
+    VariableOrPlaceholder,
+    WhereFn,
+};
+
+use clauses::{
+    ConjoiningClauses,
+};
+
+use errors::{
+    BindingError,
+    ErrorKind,
+    Result,
+};
+
+use types::{
+    Column,
+    ColumnConstraint,
+    DatomsTable,
+    Inequality,
+    QualifiedAlias,
+    QueryValue,
+    SourceAlias,
+    TransactionsColumn,
+};
+
+use Known;
+
+impl ConjoiningClauses {
+    // Log in Query: tx-ids and tx-data
+    //
+    // The log API includes two convenience functions that are available within query. The tx-ids
+    // function takes arguments similar to txRange above, but returns a collection of transaction
+    // entity ids. You will typically use the collection binding form [?tx …] to capture the
+    // results.
+    //
+    // [(tx-ids ?log ?tx1 ?tx2) [?tx ...]]
+    //
+    // TODO: handle tx1 == 0 (more generally, tx1 < bootstrap::TX0) specially (no after constraint).
+    // TODO: allow tx2 to be missing (no before constraint).
+    // TODO: allow txK arguments to be instants.
+    // TODO: allow arbitrary additional attribute arguments that restrict the tx-ids to those
+    // transactions that impact one of the given attributes.
+    pub(crate) fn apply_tx_ids(&mut self, known: Known, where_fn: WhereFn) -> Result<()> {
+        if where_fn.args.len() != 3 {
+            bail!(ErrorKind::InvalidNumberOfArguments(where_fn.operator.clone(), where_fn.args.len(), 3));
+        }
+
+        if where_fn.binding.is_empty() {
+            // The binding must introduce at least one bound variable.
+            bail!(ErrorKind::InvalidBinding(where_fn.operator.clone(), BindingError::NoBoundVariable));
+        }
+
+        if !where_fn.binding.is_valid() {
+            // The binding must not duplicate bound variables.
+            bail!(ErrorKind::InvalidBinding(where_fn.operator.clone(), BindingError::RepeatedBoundVariable));
+        }
+
+        // We should have exactly one binding. Destructure it now.
+        let tx_var = match where_fn.binding {
+            Binding::BindRel(bindings) => {
+                let bindings_count = bindings.len();
+                if bindings_count != 1 {
+                    bail!(ErrorKind::InvalidBinding(where_fn.operator.clone(),
+                                                    BindingError::InvalidNumberOfBindings {
+                                                        number: bindings_count,
+                                                        expected: 1,
+                                                    }));
+                }
+                match bindings.into_iter().next().unwrap() {
+                    VariableOrPlaceholder::Placeholder => unreachable!("binding.is_empty()!"),
+                    VariableOrPlaceholder::Variable(v) => v,
+                }
+            },
+            Binding::BindColl(v) => v,
+            Binding::BindScalar(_) |
+            Binding::BindTuple(_) => {
+                bail!(ErrorKind::InvalidBinding(where_fn.operator.clone(), BindingError::ExpectedBindRelOrBindColl))
+            },
+        };
+
+        let mut args = where_fn.args.into_iter();
+
+        // TODO: process source variables.
+        match args.next().unwrap() {
+            FnArg::SrcVar(SrcVar::DefaultSrc) => {},
+            _ => bail!(ErrorKind::InvalidArgument(where_fn.operator.clone(), "source variable", 0)),
+        }
+
+        let tx1 = self.resolve_tx_argument(&known.schema, &where_fn.operator, 1, args.next().unwrap())?;
+        let tx2 = self.resolve_tx_argument(&known.schema, &where_fn.operator, 2, args.next().unwrap())?;
+
+        let transactions = self.next_alias_for_table(DatomsTable::Transactions);
+
+        self.from.push(SourceAlias(DatomsTable::Transactions, transactions.clone()));
+
+        // Bound variable must be a ref.
+        self.constrain_var_to_type(tx_var.clone(), ValueType::Ref);
+        if self.is_known_empty() {
+            return Ok(());
+        }
+
+        self.bind_column_to_var(known.schema, transactions.clone(), TransactionsColumn::Tx, tx_var.clone());
+
+        let after_constraint = ColumnConstraint::Inequality {
+            operator: Inequality::LessThanOrEquals,
+            left: tx1,
+            right: QueryValue::Column(QualifiedAlias(transactions.clone(), Column::Transactions(TransactionsColumn::Tx))),
+        };
+        self.wheres.add_intersection(after_constraint);
+
+        let before_constraint = ColumnConstraint::Inequality {
+            operator: Inequality::LessThan,
+            left: QueryValue::Column(QualifiedAlias(transactions.clone(), Column::Transactions(TransactionsColumn::Tx))),
+            right: tx2,
+        };
+        self.wheres.add_intersection(before_constraint);
+
+        Ok(())
+    }
+
+    pub(crate) fn apply_tx_data(&mut self, known: Known, where_fn: WhereFn) -> Result<()> {
+        if where_fn.args.len() != 2 {
+            bail!(ErrorKind::InvalidNumberOfArguments(where_fn.operator.clone(), where_fn.args.len(), 2));
+        }
+
+        if where_fn.binding.is_empty() {
+            // The binding must introduce at least one bound variable.
+            bail!(ErrorKind::InvalidBinding(where_fn.operator.clone(), BindingError::NoBoundVariable));
+        }
+
+        if !where_fn.binding.is_valid() {
+            // The binding must not duplicate bound variables.
+            bail!(ErrorKind::InvalidBinding(where_fn.operator.clone(), BindingError::RepeatedBoundVariable));
+        }
+
+        // We should have at most five bindings. Destructure them now.
+        let bindings = match where_fn.binding {
+            Binding::BindRel(bindings) => {
+                let bindings_count = bindings.len();
+                if bindings_count < 1 || bindings_count > 5 {
+                    bail!(ErrorKind::InvalidBinding(where_fn.operator.clone(),
+                                                    BindingError::InvalidNumberOfBindings {
+                                                        number: bindings.len(),
+                                                        expected: 5,
+                                                    }));
+                }
+                bindings
+            },
+            Binding::BindScalar(_) |
+            Binding::BindTuple(_) |
+            Binding::BindColl(_) => bail!(ErrorKind::InvalidBinding(where_fn.operator.clone(), BindingError::ExpectedBindRel)),
+        };
+        let mut bindings = bindings.into_iter();
+        let b_e = bindings.next().unwrap_or(VariableOrPlaceholder::Placeholder);
+        let b_a = bindings.next().unwrap_or(VariableOrPlaceholder::Placeholder);
+        let b_v = bindings.next().unwrap_or(VariableOrPlaceholder::Placeholder);
+        let b_tx = bindings.next().unwrap_or(VariableOrPlaceholder::Placeholder);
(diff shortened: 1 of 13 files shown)


query-algebrizer/src/clauses/tx_log_api.rs
```diff
@@ -0,0 +1,241 @@
+// Copyright 2016 Mozilla
```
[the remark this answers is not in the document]
ncalexan: Fixed.

query-translator/tests/translate.rs
```diff
@@ -1134,3 +1134,42 @@ fn test_tx_before_and_after() {
...
+                       AND `transaction_log00`.tx < 2000");
+    assert_eq!(args, vec![]);
+
+    let query = r#"[:find (count ?tx) :where [(tx-ids $ 1000 2000) [?tx ...]]]"#;
+    let SQLQuery { sql, args } = translate(&schema, query);
+    assert_eq!(sql, "SELECT count(`?tx`) AS `(count ?tx)` FROM (\
+                     SELECT DISTINCT `transaction_log00`.tx AS `?tx` \
+                     FROM `transaction_log` AS `transaction_log00` \
+                     WHERE 1000 <= `transaction_log00`.tx \
+                       AND `transaction_log00`.tx < 2000)");
```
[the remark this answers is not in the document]
ncalexan: I've made this `tx_log` throughout.  It's not convenient to have aliases distinct from table names at this time, so I didn't do that.

query-translator/tests/translate.rs
```diff
@@ -1134,3 +1134,42 @@ fn test_tx_before_and_after() {
...
+    assert_eq!(args, vec![]);
+
+    // In practice the following query would be inefficient because of the filter on all_datoms.tx,
+    // but that is what (tx-data) is for.
```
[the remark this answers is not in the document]
ncalexan: That ticket is relevant, but this tx-* log API is a way to avoid that complexity entirely.  That is, we push the optimization problem to the consumer rather than trying to do (decidely non-trivial!) optimization in Mentat proper.

query-algebrizer/src/clauses/tx_log_api.rs
```diff
@@ -0,0 +1,241 @@
...
+    // results.
+    //
+    // [(tx-ids ?log ?tx1 ?tx2) [?tx ...]]
+    //
+    // TODO: handle tx1 == 0 specially (no after constraint).
+    // TODO: allow tx2 to be missing (no before constraint).
+    // TODO: allow txK arguments to be instants.
+    // TODO: allow arbitrary additional attribute arguments that restrict the tx-ids to those
+    // transactions that impact one of the given attributes.
+    pub(crate) fn apply_tx_ids(&mut self, known: Known, where_fn: WhereFn) -> Result<()> {
```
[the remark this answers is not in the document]
ncalexan: I'm not against tracking this, but it seems pretty easy to add later, and rather hard to actually do anything other than bail if a tracking flag is set.  That is, it's mighty difficult to capture a meaningful reference to the part of the query that is relevant to the log; and this is likely to only get worse as we do more intensive optimization and in-place rearrangement.

query-algebrizer/src/clauses/tx_log_api.rs
```diff
@@ -0,0 +1,241 @@
...
+use Known;
+
+impl ConjoiningClauses {
+    pub(crate) fn match_one_tx_argument(&self, arg: FnArg) -> Result<Option<i64>> {
+        match arg {
+            FnArg::Variable(v) => {
+                // If it's already bound, then let's expand the variable.
+                match self.bound_value(&v) {
+                    Some(TypedValue::Long(v)) => Ok(Some(v)),
+                    // TODO: handle instants.
```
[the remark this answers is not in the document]
ncalexan: I'm willing to put something about UUIDs in a TODO, but I personally don't think of Tolstoy as part of Mentat at this time.  It's an experiment that might be fruitful and might not.  I don't want to build core parts of Mentat around assumptions made or desired by Tolstoy.

ncalexan:
@rnewman this is ready for initial review.

query-algebrizer/src/clauses/tx_log_api.rs
```diff
@@ -0,0 +1,451 @@
...
+                // If it's already bound, then let's expand the variable.
+                match self.bound_value(&v) {
+                    Some(TypedValue::Ref(v)) => Ok(Either::Left(v)),
+                    // TODO: handle instants.
+                    // Some(TypedValue::Instant(v)) => Some(v),
+                    Some(tv) => {
+                        bail!(ErrorKind::InputTypeDisagreement(v.name().clone(), ValueType::Ref, tv.value_type()));
+                    },
+                    None => {
+                        // Regardless of whether we'll be providing a ref later, or the value
```
rnewman: We normally do this by simply applying the constraint, following this approximate pattern:

```
            FnArg::Variable(var) => {
                self.constrain_var_to_type(var.clone(), ValueType::Ref);
                if let Some(TypedValue::Ref(e)) = self.bound_value(&var) {
                    // Incorrect types will be handled by the constraint, above.
                    Ok(QueryValue::Entid(e))
                } else {
                    self.column_bindings
                        .get(&var)
                        .and_then(|cols| cols.first().map(|col| QueryValue::Column(col.clone())))
                        .ok_or_else(|| Error::from_kind(ErrorKind::UnboundVariable(var.name())))
                }
            },
```

If the variable isn't bound, we'll get `UnboundVariable`. If the type is wrong, the pattern will be marked known-empty.

query-algebrizer/src/clauses/tx_log_api.rs
```diff
@@ -0,0 +1,451 @@
...
+        if self.is_known_empty() {
+            return Ok(());
+        }
+
+        self.bind_column_to_var(known.schema, tx_log_alias.clone(), TxLogColumn::Tx, tx_var.clone());
+
+        let after_constraint = ColumnConstraint::Inequality {
+            operator: Inequality::LessThanOrEquals,
+            left: match tx1 {
+                Either::Left(v) => QueryValue::PrimitiveLong(v),
```
rnewman: This `QueryValue` construction is usually implemented by the `resolve_*` function, which saves you code here.

db/src/db.rs
```diff
@@ -147,6 +147,14 @@ lazy_static! {
         r#"CREATE TABLE transactions (e INTEGER NOT NULL, a SMALLINT NOT NULL, v BLOB NOT NULL, tx INTEGER NOT NULL, added TINYINT NOT NULL DEFAULT 1, value_type_tag SMALLINT NOT NULL)"#,
         r#"CREATE INDEX idx_transactions_tx ON transactions (tx, added)"#,
 
+        // TODO: migration?
+        // Associate transactions to the attributes that they assert or retract.
+        r#"CREATE TABLE tx_log (tx INTEGER NOT NULL, a SMALLINT NOT NULL)"#,
```
rnewman: I don't see anywhere that you use `tx_log.a`, so you're only using this as a way to get a unique, ordered table of tx IDs.

But we already have one of those: we already have an index on `transactions.tx`. Can you remove `tx_log` here, using `transactions` instead, and save this database change for later?

query-algebrizer/src/clauses/tx_log_api.rs
```diff
@@ -0,0 +1,451 @@
...
+                let bindings_count = bindings.len();
+                if bindings_count != 1 {
+                    bail!(ErrorKind::InvalidBinding(where_fn.operator.clone(),
+                                                    BindingError::InvalidNumberOfBindings {
+                                                        number: bindings_count,
+                                                        expected: 1,
+                                                    }));
+                }
+                match bindings.into_iter().next().unwrap() {
+                    VariableOrPlaceholder::Placeholder => unreachable!("binding.is_empty()!"),
```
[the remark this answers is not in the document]
ncalexan: It's not at all obvious, but such a statement fails when we verify that we have at least one binding.  (I originally bailed in this situation.)  We could go belt-and-braces here.

My expectation is that we'll rework all of these function declarations to be more declarative sometime soon: to share argument parsing, to share binding details, etc.  The common threads are starting to become clear as we implement more of them.

query-algebrizer/src/clauses/tx_log_api.rs
```diff
@@ -0,0 +1,451 @@
...
+    QueryValue,
+    SourceAlias,
+    TransactionsColumn,
+    TxLogColumn,
+};
+
+use Known;
+
+impl ConjoiningClauses {
+    pub(crate) fn match_one_tx_argument(&self, operator: &PlainSymbol, index: usize, arg: FnArg) -> Result<Either<i64, QualifiedAlias>> {
```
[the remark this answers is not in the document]
ncalexan: I can do that.

db/src/db.rs
ncalexan: The intention is to follow-up to allow narrowing by `a`, more or less right now :)  I'm just deciding whether to take an (optional) set of attributes, which requires parsing sets, or to make the function variadic and accept an (optionally empty) set of attributes as arguments.  Usually I prefer strongly typed fixed inputs, but we don't support sets as arguments yet.

Do you have a strong preference?  (This is relevant to `(fulltext ...)` as well; there's a TODO about exactly this issue in that code.)

db/src/db.rs
```diff
@@ -667,6 +675,22 @@ fn update_datoms(conn: &rusqlite::Connection, tx: Entid) -> Result<()> {
...
+    let s = r#"
+      INSERT INTO tx_log (tx, a)
+      SELECT DISTINCT tx, a
+      FROM transactions
+      WHERE tx = ?"#;
+
+    let mut stmt = conn.prepare_cached(s)?;
+    stmt.execute(&[&tx])
+        .map(|_c| ())
+        .chain_err(|| "Could not insert tx_log")?;
```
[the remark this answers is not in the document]
ncalexan: But we definitely shouldn't return `Ok()`!

db/src/db.rs
```diff
@@ -147,6 +147,14 @@ lazy_static! {
         r#"CREATE TABLE transactions (e INTEGER NOT NULL, a SMALLINT NOT NULL, v BLOB NOT NULL, tx INTEGER NOT NULL, added TINYINT NOT NULL DEFAULT 1, value_type_tag SMALLINT NOT NULL)"#,
         r#"CREATE INDEX idx_transactions_tx ON transactions (tx, added)"#,
 
+        // TODO: migration?
```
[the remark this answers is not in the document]
ncalexan: Presumably yes, and it's easy to achieve, but I didn't get to it (since we have no stores to maintain in the wild).

db/src/db.rs
rnewman: As I understand it, you're strictly expressing an optimization here: that we can narrow down the set of `tx-id`s by pre-joining the `tx_log` table against a known set of attributes, rather than scanning `transactions`, right?

That rests on two assumptions, I think:

- That such filtering will reduce the space of `tx-id`s considered, often to zero. You're looking ahead to a query like

```
[:find ?vocab :where
 [(tx-data …) [[?e ?a ?v _]]]
 [(ground ?a :db.schema/vocabulary)]]
```

- That the aggregate benefit of doing a number of shorter index walks (one per joined ID) despite the cost and disk fragmentation of maintaining `tx_log`, will outweigh the cost of a longer contiguous scan of `transactions`.

Am I understanding correctly?

db/src/db.rs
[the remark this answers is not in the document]
ncalexan: Roger that.

db/src/db.rs
ncalexan: Yes, but I was looking ahead to expressing that query more directly as
```
[(tx-data ... :db.schema/vocabulary) [[...]]]
```
directly.  That is, baking more of the optimization into the `tx-data` function itself, and less into the general mechanism of joining things together.

I have a hard time believing that walking the tail of transactions to find the things that are relevant to an incremental sync is a good idea.

query-algebrizer/src/clauses/tx_log_api.rs
```diff
@@ -0,0 +1,397 @@
...
+            if self.is_known_empty() {
+                return Ok(());
+            }
+
+            self.bind_column_to_var(known.schema, transactions.clone(), TransactionsColumn::Attribute, var.clone());
+        }
+
+        if let VariableOrPlaceholder::Variable(ref var) = b_v {
+            // We need to extract types for the value place.
+            self.extracted_types.insert(var.clone(), QualifiedAlias(transactions.clone(), Column::Transactions(TransactionsColumn::ValueTypeTag)));
```
rnewman: `bind_column_to_var` takes care of checking whether we need to extract types for a variable and doing so, so long as your alias implements `for_type_tag()`.

https://github.com/mozilla/mentat/blob/master/query-algebrizer/src/clauses/mod.rs#L526

query-algebrizer/src/types.rs
```diff
@@ -91,6 +93,17 @@ pub enum FulltextColumn {
     Text,
 }
 
+/// One of the named columns of our transactions table.
+#[derive(PartialEq, Eq, Clone)]
+pub enum TransactionsColumn {
```
rnewman: btw, I think I'm going to need to move these out of `query_algebrizer` for use from pull. *sigh*

Crate proliferation!

query-translator/tests/translate.rs
```diff
@@ -1134,3 +1134,89 @@ fn test_tx_before_and_after() {
...
+                     `transactions00`.v AS `?v`, \
+                     `transactions00`.value_type_tag AS `?v_value_type_tag`, \
+                     `transactions00`.tx AS `?tx`, \
+                     `transactions00`.added AS `?op` \
+                     FROM `transactions` AS `transactions00` \
+                     WHERE `transactions00`.tx = 1000");
+    assert_eq!(args, vec![]);
+
+    // This is awkward since the transactions table is queried twice, once to list transaction IDs
+    // and a second time to extract data.  It is possible to improve this.
```
rnewman: Could you file an issue, if you haven't already?

query-translator/tests/translate.rs
```diff
@@ -1134,3 +1134,89 @@ fn test_tx_before_and_after() {
...
+                     AND `transactions00`.tx < 2000 \
+                     AND `transactions00`.tx = `all_datoms01`.tx");
+    assert_eq!(args, vec![]);
+}
+
+#[test]
+fn test_tx_data() {
+    let schema = prepopulated_typed_schema(ValueType::Double);
+
+    let query = r#"[:find ?e ?a ?v ?tx ?op :where [(tx-data $ 1000) [[?e ?a ?v ?tx ?op]]]]"#;
```
rnewman: Could I get a sibling test that doesn't bind everything — perhaps `[[_ ?a ?v]]` — to demonstrate that we don't project columns we don't need?

tests/query.rs
```diff
@@ -1339,3 +1341,115 @@ fn test_aggregation_implicit_grouping() {
...
+
+    let tx3 = store.transact(r#"[
+        [:db/add "v" :foo/term "3"]
+    ]"#).expect("tx3 to apply").tx_id;
+
+    fn assert_tx_id_range(store: &Store, after: Entid, before: Entid, expected: Vec<TypedValue>) {
+        // TODO: after https://github.com/mozilla/mentat/issues/641, use q_prepare with inputs bound
+        // at execution time.
+        let r = store.q_once(r#"[:find [?tx ...]
+                             :in ?after ?before
```
rnewman: Indent plz.

tests/query.rs
```diff
@@ -1339,3 +1341,115 @@ fn test_aggregation_implicit_grouping() {
...
+
+    let tx2 = store.transact(r#"[
+        [:db/add "e" :foo/term "2"]
+    ]"#).expect("tx2 to apply");
+
+    fn assert_tx_data(store: &Store, tx: &TxReport, value: TypedValue) {
+        // TODO: after https://github.com/mozilla/mentat/issues/641, use q_prepare with inputs bound
+        // at execution time.
+        let r = store.q_once(r#"[:find ?e ?a-name ?v ?tx ?op
+                             :in ?tx-in
```
rnewman: Throughout.

query-translator/tests/translate.rs
ncalexan: https://github.com/mozilla/mentat/issues/644

query-algebrizer/src/clauses/tx_log_api.rs
ncalexan: I've removed this, and made a note that this is relevant to https://github.com/mozilla/mentat/issues/642.

query-translator/tests/translate.rs
ncalexan: Sure.

rnewman:
LGTM!

Issue #614, opened 2018-04-03
Create Android Framework for Mentat


fluffyemily:
Create an Android framework wrapping Mentat that can be imported to Android Studio with Ant, Maven or Gradle.

Issue #609, opened 2018-04-03
Create iOS Framework for Mentat


fluffyemily:
This involves creating a Swift framework wrapping Mentat's FFI. The framework should be compatible with SwiftPackageManager, Carthage and Cocoapods.

Issue #586, opened 2018-03-13
Use failure instead of error_chain


rnewman:
Thanks to Send/Sync issues, we are stranded on a fork of error_chain. In any case, it’s a painful and error-prone library to use. 

We should port to `failure`, which is the better replacement on the block.

And we should consider flattening out some of our error hierarchy, too.

grigoryk:
Fixed in #730.

PR #643, opened 2018-04-17
iOS and Android (Java) sdk framework

fluffyemily:
iOS and Android frameworks for Mentat

Still to be done: 
- Exceptions in Android (we don't have them)
- Tests for Observers in both iOS and Android

https://github.com/mozilla/mentat/issues/609
https://github.com/mozilla/mentat/issues/614


sdks/android/Mentat/app/src/main/java/com/mozilla/mentat/Mentat.java
```diff
@@ -0,0 +1,100 @@
...
+        } else {
+            return null;
+        }
+    }
+
+    public long entIdForAttribute(String attribute) {
+        return JNA.INSTANCE.store_entid_for_attribute(this.rawPointer, attribute);
+    }
+
+    public RustResult sync() {
```
rnewman: No syncing in this PR, please.

sdks/android/Mentat/app/src/main/java/com/mozilla/mentat/Mentat.java
```diff
@@ -0,0 +1,100 @@
...
+    public Mentat() {
+        this.rawPointer = JNA.INSTANCE.store_open("");
+    }
+
+    public Mentat(Pointer rawPointer) { this.rawPointer = rawPointer; }
+
+    public TxReport transact(String transaction) {
+        RustResult result = JNA.INSTANCE.store_transact(this.rawPointer, transaction);
+        if (result.isFailure()) {
+            Log.i("Mentat", result.err);
```
rnewman: Add TODOs when you need this, and remove them when you have exception support!

sdks/android/Mentat/app/src/main/java/com/mozilla/mentat/Mentat.java
```diff
@@ -0,0 +1,100 @@
...
+        return null;
+    }
+
+    public void registerObserver(String key, String[] attributes, TxObserverCallback callback) {
+        // turn string array into int array
+        long[] attrEntids = new long[attributes.length];
+        for(int i = 0; i < attributes.length; i++) {
+            attrEntids[i] = JNA.INSTANCE.store_entid_for_attribute(this.rawPointer, attributes[i]);
+        }
+        Log.i("Mentat", "Registering observer {" + key + "} for attributes:");
```
rnewman: …

sdks/android/Mentat/app/src/main/java/com/mozilla/mentat/AttributeList.java
```diff
@@ -0,0 +1,46 @@
...
+public class AttributeList extends Structure implements Closeable {
+    public static class ByReference extends AttributeList implements Structure.ByReference {
+    }
+
+    public static class ByValue extends AttributeList implements Structure.ByValue {
+    }
+
+    public IntByReference attributes;
+    public int numberOfItems;
+    // Used by the Swift counterpart, JNA does this for us automagically.
```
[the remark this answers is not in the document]
fluffyemily: Because otherwise the fields don't match. Have added comment to that effect

sdks/android/Mentat/app/src/main/java/com/mozilla/mentat/InProgressBuilder.java
```diff
@@ -0,0 +1,10 @@
+package com.mozilla.mentat;
```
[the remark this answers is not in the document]
fluffyemily: Because when you convert an app to a library, it doesn't rename the folder. Have renamed.

fluffyemily:
@rnewman  I have just documented the F*** out of this. Hope that helps!

ffi/src/lib.rs
```diff
@@ -8,6 +8,66 @@
...
+//! the format it is expecting.
+//! This interface is designed to ensure that nothing unsafe ever leaves this module
+//! and enters Mentat proper.
+//!
+//! Structs defined with `#[repr(C)]` are guaranteed to have a layout that is compatible
+//! with the platform's representation in C.
+//!
+//! This API passes pointers in two ways, depending on the lifetime of the object and
+//! what object owns it.
+//! Pointers to values that can guarenteed to live beyond the lifetime of the function,
```
ncalexan: nit: "are guaranteed"?

ffi/src/lib.rs
```diff
@@ -8,6 +8,66 @@
...
+//! Functions that are available to other languages in this module are defined as
+//! extern "C" functions which allow them to be layed out correctly for the
+//! platform's C ABI. They all have a `#[no_mangle]` decorator to ensure
+//! Rust's name mangling is turned off, so that it is easier to link to.
+//!
+//! Mentat's FFI contains unsafe code. As it is an interface between foreign code
+//! and native Rust code, Rust cannot guarentee that the types and data that it is
+//! handling that have been passed to it from another language are present and in
+//! the format it is expecting.
+//! This interface is designed to ensure that nothing unsafe ever leaves this module
```
ncalexan: nit: leaves?  This module is filtering unsafe inputs, correct?  Can we find a word that suggests that filtering -- perhaps, "nothing unsafe passes through this module and enters Mentat proper."?

ffi/src/lib.rs
```diff
@@ -8,6 +8,66 @@
...
+//! value as *const TypedValue
+//! ```
+//! Pointers to values that cannot be guarenteed to live beyond the lifetime of the function
+//! are first Boxed so that they live on the heap, and the raw pointer passed this way.
+//! ```
+//! Box::into_raw(Box::new(value))
+//! ```
+//!
+//! The memory for values that are moved onto the heap before being passed over the FFI
+//! are no longer managed by Rust, but Rust still owns the value. Therefore the pointer
```
ncalexan: nit: the plural shits here.  "The memory ... are".  Try "The memory ... is"?

ffi/src/lib.rs
```diff
@@ -8,6 +8,66 @@
...
+//! must be returned to Rust in order to be released. To this effect a number of `destructor`
+//! function are provided for each Rust value type that is passed, and a catch all destructor
```
ncalexan: nit: functions.

ncalexan: And maybe: "as is a catch all ...".

ffi/src/lib.rs
```diff
@@ -8,6 +8,66 @@
...
+//! define_destructor!(query_builder_destroy, QueryBuilder);
+//! ```
+//!
+//! Passing a pointer to released memory will cause Mentat to crash, so callers have to be
+//! careful to ensure they manage their pointers properly. Failure to call a destructor for a
+//! value on the heap will cause a memory leak.
+//!
+//! Most of the functions exposed in this FFI have a direct mapping to existing Mentat APIs.
+//! Application logic has been kept to a minumum in order to provide the greatest flexibility
+//! for callers using the interface, however there are a one exception where several steps
```
ncalexan: nit: "there are a one"...

Rather than fixing the grammar, I think we should say something more general: "This module is generally a direct mapping to existing Mentat APIs, because [rationale].  However, in some instances a single convenience ... because ... [rationale for the class of exceptional cases]."

Let's not have this be out of date quite so quickly!

ffi/src/lib.rs
```diff
@@ -8,6 +8,66 @@
...
+//! have been wrapped into a single call in order to make the interface easier to use.
+//! `store_register_observer` takes a single native callback function is wrapped inside a
+//! Rust closure and added to a `TxObserver` struct that this then used to register the
+//! observer with the store.
+//!
+//! Result and Option Rust types have `repr(C)` structs that mirror them. This is to provide a more
```
ncalexan: Nit: `Result`, `Option`.

ffi/src/lib.rs
```diff
@@ -66,9 +126,20 @@ pub use utils::strings::{
 
 pub use utils::log;
 
+// type aliases for iterator types.
 pub type TypedValueIterator = vec::IntoIter<TypedValue>;
 pub type TypedValueListIterator = vec::IntoIter<Vec<TypedValue>>;
 
+/// A C representation of the a change provided by the transaction observers
```
ncalexan: nit: "the a".

ffi/src/lib.rs
```diff
@@ -66,9 +126,20 @@ pub use utils::strings::{
...
+/// from a single transact.
+/// Holds a transaction identifier, the changes as a set of affected attributes
```
ncalexan: Huh.  I understand (I think!) why the history points to exposing affected attributes -- that's what was needed for caching.  But why is this what we're exposing to consumers?

Datomic provides all the datoms transacted.  That's expensive for us, but maybe we should talk about why we don't do that and how to achieve it?  Maybe we should make it an option for transaction observers?  Maybe it's a special type of transaction observer that filters the affected attribute set and then pulls the relevant datoms?

ffi/src/lib.rs
```diff
@@ -66,9 +126,20 @@ pub use utils::strings::{
...
+/// and the length of the list of changes.
+///
+/// #Safety
+///
+/// Callers are responsible for managing the memory for the return value.
```
ncalexan: Can we hyperlink in Rust docs, so that we don't have this verbosity?  Or better yet, can we make a (mostly dummy) trait that encodes this memory management style, so that it's obvious _from the code_ what the consumer is responsible for?  I.e., for types `A`, `B`, `C`, they all implement `ConsumerDestroys`, which is the `destroy` implementation...  (Maybe this is how your macro works; I will find out.)

ffi/src/lib.rs
```diff
@@ -77,13 +148,30 @@ pub struct TransactionChange {
...
+/// A destructor `destroy` is provided for releasing the memory for this
+/// pointer type.
 #[repr(C)]
 #[derive(Debug)]
 pub struct TxChangeList {
     pub reports: Box<[TransactionChange]>,
     pub len: usize,
 }
 
+/// A C representation Rust's Option.
```
ncalexan: Nit -- lots of `Option` `Some` `None` quotes.

Is this really the standard way to declare this stuff for FFIs?  That surprises me, because it means every crate will  have its own, subtly different, external `Option` representation.  I thought "the point" of Rust's `Option` representation was that it _transparently_ was null or the pointer as appropriate across the FFI boundary.  Or is it that you're allocating in a way that needs special handling in Rust?

ffi/src/lib.rs
```diff
@@ -124,11 +222,17 @@ impl<T, E> From<Result<T, E>> for ExternResult where E: std::error::Error {
...
-// A store cannot be opened twice to the same location.
-// Once created, the reference to the store is held by the caller and not Rust,
-// therefore the caller is responsible for calling `destroy` to release the memory
-// used by the Store in order to avoid a memory leak.
-// TODO: Start returning `ExternResult`s rather than crashing on error.
+/// A store cannot be opened twice to the same location.
+/// Once created, the reference to the store is held by the caller and not Rust,
+/// therefore the caller is responsible for calling `destroy` to release the memory
+/// used by the Store in order to avoid a memory leak.
+/// TODO: Start returning `ExternResult`s rather than crashing on error.
```
ncalexan: There's a pattern for this type of `Result` across the FFI boundary in mozilla-central, but the best reference I can find says that the situation is more complicated than I thought: https://groups.google.com/d/topic/mozilla.dev.platform/udd-DrIDYP4/discussion.

I think we should talk to a Rust/C expert to figure out a good pattern for this.  I doubt we want to proliferate `Extern*` types on the Rust FFI side of the house, but maybe I'm wrong.

ffi/src/lib.rs
```diff
@@ -146,6 +250,18 @@ pub extern "C" fn store_open(uri: *const c_char) -> *mut Store {
...
+/// # Safety
+///
+/// Callers must ensure that the pointer to the `Store` is not dangling and that
+/// the C string provided to `transaction` is valid.
+///
+/// Callers are responsible for managing the memory for the return value.
+/// A destructor `destroy` is provided for releasing the memory for this
+/// pointer type.
+///
+/// TODO: Document the errors that can result from transact
```
ncalexan: One reason this is hard is that the transactor uses `error-chain` -- badly.  Most of the errors are string values, which of course isn't helpful.  Making this better requires work, which might be part of #586, or that I might do independently.

ffi/src/lib.rs
```diff
@@ -159,18 +275,35 @@ pub unsafe extern "C" fn store_transact(store: *mut Store, transaction: *const c
     Box::into_raw(Box::new(result.into()))
 }
 
+/// Fetches the `tx_id` for the given `TxReport`.
+///
+/// # Safety
```
ncalexan: Assuming you mean "points to a Rust TxReport" by "not dangling", I don't think these safety blocks help in this case.  Garbage in, garbage out -- and there's nothing to be done our side if the consumer is just wrong.

ffi/src/lib.rs
```diff
@@ -159,18 +275,35 @@ pub unsafe extern "C" fn store_transact(store: *mut Store, transaction: *const c
...
     tx_report.tx_instant.timestamp() as c_longlong
 }
 
+/// Fetches the `Entid` assigned to the `tempid` during the transaction represented
+/// by the given `TxReport`.
+///
+/// # Safety
+///
+/// Callers must ensure that the pointer to the `TxReport` is not dangling and that
+/// the C string provided to `tempid` is valid.
```
ncalexan: What does "valid" mean here?  I see that we are manually handling `None => null` -- I rather hoped that `Option` would do that for us more directly.

And we're allocating memory to turn the `char *` into a Rust owned `String`, which we immediately borrow to look up a value in `tempids`.  What does it take to borrow the `char *` for the `tempids` lookup directly (crashing if we're given garbage)?  I'm pretty sure the `BTreeMap` under the hood here is designed for this type of flexibility when looking at the keys...

ffi/src/lib.rs
```diff
@@ -193,20 +339,43 @@ pub unsafe extern "C" fn store_query<'a>(store: *mut Store, query: *const c_char
...
+/// Binds a `TypedValue::Long` to a `Variable` with the given name.
+///
+/// # Safety
+///
+/// Callers must ensure that the pointer to the `QueryBuilder` is not dangling and that
+/// the C string provided to `var` is valid.
 #[no_mangle]
 pub unsafe extern "C" fn query_builder_bind_long(query_builder: *mut QueryBuilder, var: *const c_char, value: c_longlong) {
     let var = c_char_to_string(var);
     let query_builder = &mut*query_builder;
```
ncalexan: nit: the spacing looks a little off here.

ffi/src/lib.rs
```diff
@@ -193,20 +339,43 @@ pub unsafe extern "C" fn store_query<'a>(store: *mut Store, query: *const c_char
...
    query_builder.bind_long(&var, value);
```
ncalexan: Again, we allocate a `String` and then bind by reference.  I hope we can avoid a lot of this allocation with a little thought.

ffi/src/lib.rs
```diff
@@ -280,13 +496,39 @@ pub unsafe extern "C" fn query_builder_execute_scalar(query_builder: *mut QueryB
     Box::into_raw(Box::new(extern_result))
 }
 
+/// Executes a query and returns the results as a Coll.
```
ncalexan: Through-out: I think all of these will be richer if they refer to `QueryResults::Coll` and get hyperlinked.

ffi/src/lib.rs
```diff
@@ -313,162 +568,365 @@ fn unwrap_conversion<T>(value: Option<T>, expected_type: ValueType) -> T {
...
+///
+/// # Panics
+///
+/// If the `ValueType` of the `TypedValue` is not `ValueType::Boolean`.
+///
+/// # Safety
+///
+/// Callers must ensure that the pointer to the `TypedValue` is not dangling .
 #[no_mangle]
 pub unsafe extern "C" fn typed_value_as_boolean(typed_value: *mut TypedValue) -> i32 {
```
ncalexan: Generally "into" means "consume" and "as" means "take a reference".  This looks like it's taking ownership (consuming), since that's what `from_raw` does.  That is, it's not OK to do:
```
void* foo = some_function_that_gives_a_Rust_TypedValue();
if (typed_value_as_boolean(foo)) {
 // do something.
}
... later ...
if (!typed_value_as_boolean(foo)) {
 // do something else.
}
```
and that's not suggested by the name.  Can we either produce a reference to the internal value (I bet we can't, since some of them can't be exposed directly to the FFI as C-level references, for example, `DateTime<Utc>` probably can't) _or_ follow the Rust convention and use `into`?

ffi/src/lib.rs
```diff
@@ -313,162 +568,365 @@ fn unwrap_conversion<T>(value: Option<T>, expected_type: ValueType) -> T {
...
 
-//value_type
+/// Returns the `ValueType` of this `TypedValue`.
+///
+/// # Safety
+///
+/// Callers must ensure that the pointer to the `TypedValue` is not dangling .
 #[no_mangle]
 pub unsafe extern "C" fn typed_value_value_type(typed_value: *mut TypedValue) ->  ValueType {
     let typed_value = &*typed_value;
```
ncalexan: Is this mangling (`&*`) necessary?  That's syntactically a no-op, so I'm surprised the compiler needs it.

nit: there seem to be extra spaces before the return types `->  ValueType`, perhaps throughout.  Copy-paste error?

ffi/src/lib.rs
```diff
@@ -313,162 +568,365 @@ fn unwrap_conversion<T>(value: Option<T>, expected_type: ValueType) -> T {
...
+///
+/// # Safety
+///
+/// Callers must ensure that the pointer to the `Vec<Vec<TypedValue>>` is not dangling .
+///
+/// Callers are responsible for managing the memory for the return value.
+/// A destructor `typed_value_result_set_iter_destroy` is provided for releasing the memory for this
+/// pointer type.
 #[no_mangle]
 pub unsafe extern "C" fn rows_iter(rows: *mut Vec<Vec<TypedValue>>) ->  *mut TypedValueListIterator {
```
ncalexan: Again, `into_rows_iter`?

ffi/src/lib.rs
```diff
@@ -313,162 +568,365 @@ fn unwrap_conversion<T>(value: Option<T>, expected_type: ValueType) -> T {
...
+/// Callers must ensure that the pointer to the `Vec<TypedValue>` is not dangling .
 #[no_mangle]
 pub unsafe extern "C" fn value_at_index_as_entid(values: *mut Vec<TypedValue>, index: c_int) ->  Entid {
     let result = &*values;
     let value = result.get(index as usize).expect("No value at index");
     unwrap_conversion(value.clone().into_entid(), ValueType::Ref)
 }
 
-// as kw
+/// Returns the value of the `TypedValue` at `index` as a keyword `String`.
```
ncalexan: `String` seems wrong here, since it's _not_ a Rust `String`.

Throughout, is there a reason we're not using Rust's `CString` (for owned C-strings) and `CStr` (for references to C-strings)?  I feel like I'm seeing a bunch of deviations from what I expect in the API, but I just don't know enough to be evaluate if those deviations are justified.

Enlighten me!

ffi/src/lib.rs
```diff
@@ -545,29 +1061,31 @@ pub unsafe extern "C" fn tx_change_list_entry_at(tx_report_list: *mut TxChangeLi
...
-    let store = &mut*store;
-    let user_uuid = c_char_to_string(user_uuid);
-    let server_uri = c_char_to_string(server_uri);
-    let res = store.sync(&server_uri, &user_uuid);
-    Box::into_raw(Box::new(res.into()))
-}
-
+/// destroy function for releasing the memory for `repr(C)` structs.
 #[no_mangle]
 pub unsafe extern "C" fn destroy(obj: *mut c_void) {
```
ncalexan: Suppose a consumer calls `destroy(null)`.  We'll silently accept that, where-as `free null` crashes in the real world.  Are we doing consumers a favour?  Something is horribly wrong, and we're not telling them in the way that is most immediately actionable -- with a crash.  Explain why this is the right choice.  (Note that consumers can always wrap `destroy`, if they want to avoid nulls themselves.  And I suppose that consumers can also force a crash if they want it, so perhaps this is moot.  But still, I'd like to understand our rationale.  Is this a convention in Rust FFI implementations?)

ffi/src/lib.rs
```diff
@@ -545,29 +1061,31 @@ pub unsafe extern "C" fn tx_change_list_entry_at(tx_report_list: *mut TxChangeLi
...
     if !obj.is_null() {
         let _ = Box::from_raw(obj);
     }
 }
 
+/// Creates a function with a given `$name` that releases the memroy for a type `$t`.
```
ncalexan: A trait and a standard implementation would be nicer, I think.  Since we own the trait, we can define for all of these types coming from other crates; and we can handle `Vec<T>` generically, so we don't require specialization...

ffi/src/lib.rs
fluffyemily: The reason we provide a list of the affected attributes is to enable consumers to decide which actions it wants to perform based on which attributes are affected. If I have an observer that is registered to listen to changes to `:foo/name` and `:foo/due_date`, I might want to do something different when if an `:foo/name` changes (just update the UI to reflect the change) to `:foo/due_date` (fire a notification to a user to let them know that the due date on an item has changed`. As we expand observers to further filter by entity id and acceptance criteria, this might become more important (name changed on entity a and due date changed on entity b in the same transaction:).

ffi/src/lib.rs
fluffyemily: I was approaching this from the point of view of someone who was coming to the FFI and reading the docs wanting to implement their own native client using it. I felt that they would want to read about the safety without having to go to the code in order to figure it out and a trait implementation would obscure that. There might be a more centralized way of doing it though - maybe implementing the trait and referencing it from the crate docs. I'll noodle on it.

ffi/src/lib.rs
fluffyemily: Yup, you're right. We are doing unecessary work here. I'll see how we can improve this - string handling has been on my list to look at for a little while now.

ffi/src/lib.rs
fluffyemily: Yes, the `&*` is necessary. Dereference the raw pointer (`*mut TypedValue`) and then create a rust reference to it. This is the standard way of referencing an raw pointer directly without reclaiming ownership.

ffi/src/lib.rs
fluffyemily: Convention with FFI states that you start the function name with the type of the pointer that the function acts on. With that in mind, this should probably be `typed_value_result_into_iter` or something. Will rename.

ffi/src/lib.rs
fluffyemily: no convention. I'll revert to crashing on null - we leave it up to the consumer to manage their pointers in every other case so we should do that here too.

ffi/src/lib.rs
fluffyemily: I think this may be possible. Then I can define the `destroy` function only and call obj.destroy(). The only question I have is how possible it is to deal with generics over FFI, but I will experiment.

fluffyemily: Right, it turns out that you cannot `#[no_mangle]` a generic function. In order to use a generic in the function definition `#[no_mangle]` would have to be removed, which would make the function name very difficult to determine in order to call externally. 

Which means, in order to use the trait I would have to create a destroy function for each type anyway so I can call the trait function on that type.

All the ways I have tried to get around this result in a level of confusion that defeats the point of doing this.

ffi/src/lib.rs
ncalexan: Thanks!  TIL.

ffi/src/lib.rs
ncalexan: Your choice.

ffi/src/lib.rs
```diff
@@ -506,7 +506,7 @@ fn unwrap_conversion<T>(value: Option<T>, expected_type: ValueType) -> T {
 ///
 /// If the [ValueType](mentat::ValueType) of the [TypedValue](mentat::TypedValue) is not [ValueType::Long](mentat::ValueType::Long).
 #[no_mangle]
-pub unsafe extern "C" fn typed_value_as_long(typed_value: *mut TypedValue) ->  c_longlong {
+pub unsafe extern "C" fn typed_value_into_long(typed_value: *mut TypedValue) ->  c_longlong {
```
ncalexan: The commit message is off -- it should be "-> `into`.".

ffi/src/utils.rs
```diff
@@ -21,18 +21,17 @@ pub mod strings {
...
 
     pub fn string_to_c_char<T>(r_string: T) -> *mut c_char where T: Into<String> {
         CString::new(r_string.into()).unwrap().into_raw()
     }
 
-    pub fn kw_from_string(mut keyword_string: String) -> NamespacedKeyword {
-        let attr_name = keyword_string.split_off(1);
+    pub fn kw_from_string(keyword_string: &'static str) -> NamespacedKeyword {
+        let attr_name = keyword_string.trim_left_matches(":");
         let parts: Vec<&str> = attr_name.split("/").collect();
```
ncalexan: This will silently accept ":one/two/three" and produced ":one/two".  `panic!` instead?

ffi/src/lib.rs
ncalexan: > All the ways I have tried to get around this result in a level of confusion that defeats the point of doing this.

Thanks, that's super valuable!  I'm happy to take this bit of wisdom and roll with it.  Perhaps in the future we'll learn something better.  Maybe a marker trait (like `Sync` or `Sized`) might be possible at some point.

.gitignore
```diff
@@ -54,3 +54,28 @@ pom.xml.asc
 /fixtures/*.db-shm
 /fixtures/*.db-wal
 /query-parser/out/
+## Build generated
+build/
```
ncalexan: This isn't right -- `build/version.rs` exists in the tree.  Is this written by XCode?  Can we make it not be written by XCode?  Or should we move our build script?

fluffyemily:
@ncalexan it's not the Keyword changes that have bitrotted this, it's the move to `Binding` and `RelResult`. Having to rethink the way we iterate `RelResult` over FFI and doing a fair bit of work to update everything to work with `Binding` rather than `TypedValue`. 

+1 on trying to get the FFI tests running. I'm thinking of creating some Rust tests for this, just so we can catch when things that change break the FFI.

.gitignore
fluffyemily: No, this was my mistake. This should be specific for Xcode `build/` directory, not the root one.

PR #648, opened 2018-04-20
Optimize NamespacedKeyword (and NamespacedSymbol)

thomcc:
Currently NamespacedKeyword (and NamespacedSymbol) is effectively a tuple of two `String`s. This means that each instance of it will require two allocations, reading the data out of them will typically require two cache misses, etc. There are a number of ways this could be improved.

One way is interning, and the comments indicate in several places that we'd like *should* to doing this, however, the way Mentat is currently structured, I think this is not trivial<sup>[1](#f1)</sup>.

Another is just to store it in a single string, and remember the boundaries. That's what these patches do.

Aside from some additional code complexity (which is localized to the module implementing this optimization), there's no real downside here, and the upside is you'll have better cache locality for code that touches both the name and namespace, fewer allocations for most code, reduced memory usage for objects holding `NamespacedKeywords` (e.g. `std::mem::size_of::<NamespacedKeyword>()` is lower), cheaper creation of these keywords, and etc.

In theory at least, to be sure about that benchmarks are required, so the first patch fixes some issues in the tx-parser benchmark, and makes it parse the transactions with the seattle data from our fixtures. While using the tx-parser to benchmark this is not great representation of mentat as a whole<sup>[2](#f2)</sup>, it is probably reasonably accurate for a benchmark of code that uses keywords heavily, and (most importantly) it was already there.

Anyway, there's a little bit of `unsafe` inside `NamespacedName` now, but the code is still totally safe. There are two variants we need to maintain for safety, and both are maintained by `NamespacedName::new`, and reflected in comments for the `boundary` property.

One note: Do we intend to support the `serde_support` feature (in the `edn` crate)? It's not used on most types, and was certainly the hardest part of this patch to get working properly (note: using `derive(Deserialize)` on a type that requires that its fields maintain certain invariants for the sake of memory safety is a very bad idea). Regardless, this *does* work now, is safe, and has tests...

### Benchmarks

Anyway, the end result of this for me is quite a decent speedup on the tx-parser benchmark:

Before (no optimization -- it does have the first patch which fixes and extends the benchmarks though):
```
test bench_parse1 ... bench:       2,856 ns/iter (+/- 580)
test bench_parse2 ... bench:      51,548 ns/iter (+/- 8,617)
test bench_parse3 ... bench:     109,473 ns/iter (+/- 50,330)
test bench_parse4 ... bench:  11,773,658 ns/iter (+/- 1,575,618)
```

After (with optimization):
```
test bench_parse1 ... bench:       2,392 ns/iter (+/- 901)
test bench_parse2 ... bench:      44,803 ns/iter (+/- 16,559)
test bench_parse3 ... bench:      97,035 ns/iter (+/- 52,726)
test bench_parse4 ... bench:   9,439,930 ns/iter (+/- 1,370,286)
```

This varied across runs but these were relatively representative for me.

It would be nice to know how much that helps in more general cases, but it's actually better than I had expected. Of course, real workloads are going to be dominated by sqlite time, so this somewhat silly to work on (I actually waffled about submitting this PR for this reason, but it seems better to have than not), but it's bugged me for a while, and the kinds of things I work on during the day don't give me many opportunities to micro-optimize any more.

Anyway, you can try yourself with `cargo +nightly bench -p mentat_tx_parser` if you have a nightly build installed in rustup and feel like it. (Sadly, cargo bench is [going away](https://github.com/rust-lang/rfcs/pull/2287) soon, but supposedly something similar will replace it, and not require nightly).

---

<b id="f1">1</b> In particular, mentat treats keywords as value types and creates and extracts them all over the place. In the places we do intern things, we use reference counting, mostly to make the lifetimes bearable (I can think of ways to make this work with references, but none of them are as simple as Rc). Sadly, this is hardly ideal, since `Rc<String>` has a very real cost when compared with `&str`.

Rustc uses both [references](https://github.com/rust-lang/rust/blob/8830a0304327ba8c983555ac5d42cec0569c31bb/src/libsyntax_pos/symbol.rs#L364-L376) and [numeric handles to interned strings](https://github.com/rust-lang/rust/blob/8830a0304327ba8c983555ac5d42cec0569c31bb/src/libsyntax_pos/symbol.rs#L108) depending on the case, but it's worth noting that the lifetime there is a lie, and could result in UAF if used wrong (given that Mentat connections might be long lived, the "leak everything until task cleanup" strategy seems dodgy to me, unless we spawned a task for each query/transaction, which... dunno, might work?). Anyway, nothing prevents us from doing both, and the current representation seemed needlessly expensive to me.

<b id="f2">2</b> In general, it would be really good to get more benchmarks (both ones that use sqlite and ones that don't). There are certainly comments that say things about performance that are [dubious](https://github.com/mozilla/mentat/blob/0c31fc7875294fdf80f4caf253590999a9214e0d/edn/src/types.rs#L45-L47), and not having numbers to back this up doesn't help... Also, things like using `Rc` to avoid copy is not always a win, especially when the type it holds onto is small -- most keywords and symbols are quite short, and they are Rced in several places -- this might be the right choice, but it also could be bad for a lot of reasons. Measuring would remove doubt here.

### DIFF
--- a/edn/Cargo.toml
+++ b/edn/Cargo.toml
@@ -20,6 +20,10 @@ uuid = { version = "0.5", features = ["v4", "serde"] }
 serde = { version = "1.0", optional = true }
 serde_derive = { version = "1.0", optional = true }
 
+[dev-dependencies]
+serde_test = "1.0"
+serde_json = "1.0"
+
 [features]
 serde_support = ["serde", "serde_derive"]
 
--- a/edn/src/lib.rs
+++ b/edn/src/lib.rs
@@ -22,6 +22,8 @@ extern crate serde;
 #[macro_use]
 extern crate serde_derive;
 
+// Intentionally not pub
+mod namespaced_name;
 pub mod symbols;
 pub mod types;
 pub mod pretty_print;
--- a/edn/src/namespaced_name.rs
+++ b/edn/src/namespaced_name.rs
@@ -1,0 +1,220 @@
+// Copyright 2016 Mozilla
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software distributed
+// under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+use std::cmp::{Ord, PartialOrd, Ordering};
+use std::fmt;
+
+#[cfg(feature = "serde_support")]
+use serde::{
+    de::{self, Deserialize, Deserializer},
+    ser::{Serialize, Serializer}
+};
+
+// Data storage for both NamespacedKeyword and NamespacedSymbol. This gets it's own module because
+// modifying it's fields is `unsafe`, and so we'd like to minimize the amount of code that can
+// touch it.
+#[derive(Clone, Eq, Hash, PartialEq)]
+pub struct NamespacedName {
+    // The bytes that make up the namespace followed directly by those
+    // that make up the name.
+    ns_and_name: String,
+
+    // The index (in bytes) into `ns_and_name` where the namespace ends and
+    // name begins.
+    //
+    // Important: The following invariants around `boundary` must be maintained
+    // for memory safety.
+    //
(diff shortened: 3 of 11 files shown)


rnewman:
I’m wrapping this up and taking it further to support aliasing in pull.

ncalexan:
> I’m wrapping this up and taking it further to support aliasing in pull.

I'm glad to see this landing.  Could this help when consuming `StructuredMap`?  Right now the keys are `ValueRc<NamespacedKeyword>`, which is frustrating to create.

rnewman:
Obsoleted by #682.

Issue #378, opened 2017-03-20
[tx] Implement Datomic's `:db.fn/retractEntity` and DataScript's `:db.fn/retractAttribute`


ncalexan:
This is #46 but for the Rust implementation.

Issue #21, opened 2016-08-04
Implement excision


ncalexan:
See http://docs.datomic.com/excision.html.

rnewman:
Requirements/description:

Excision cannot remove vocabulary, the bootstrap schema, or any other built-ins. This includes the `:db/excise` attributes themselves.

Excision records datoms that define the excision itself, and additionally removes the affected datoms from history. Indeed, recording these datoms is how one triggers an excision, though I'm sure we'll wrap this in a nice API.

Excision targets _either_ an entity or an attribute. (If you want to treat an attribute as an entity, you need to specify attributes to excise.)

If an entity is targeted, all _inbound_ and _outbound_ datoms are removed, as are any component entities linked by an outbound datom.

An optional set of attributes can be specified to filter the set of affected datoms.

An optional timestamp or transaction ID filter can be specified, similarly.

If an attribute is targeted, it affect all entities, subject to filtering as above.

It's important that excision is deterministic: for a given database state, a given excision input should always remove the exact same set of datoms.

rnewman:
Implementation choices:

This is likely to hook into [the same spot in the transactor that picks up schema changes](https://github.com/mozilla/mentat/blob/master/db/src/tx.rs#L715):

```rust
        if tx_might_update_metadata {
            // Extract changes to metadata from the store.
            let metadata_assertions = self.store.committed_metadata_assertions(self.tx_id)?;
```

I suggest that at that point we synchronously compute an 'excision operation' that can be persisted in the database (ensuring that all excisions eventually and atomically take place), in a separate table, and then process that table after the transact commits. This is inline with Datomic's background behavior. We can expose a synchronous 'purge' call that blocks until all pending excisions are processed.

Excision will mutate the datoms table, the log, potentially the fulltext values table (Datomic doesn't support excision of fulltext, but I think we should), and update the cache. It'll have no interaction with syncing _except_ that certain sync operations — particularly merges — should only proceed against a store with no pending excisions, for obvious determinism reasons.

ncalexan:
> It'll have no interaction with syncing except that certain sync operations — particularly merges — should only proceed against a store with no pending excisions, for obvious determinism reasons.

[...]

PR #650, opened 2018-04-20
(tx) Implement :db/retractEntity. (#378)

ncalexan:
Datomic used to expose `:db.fn/retractEntity` and
`:db.fn/retractAttribute`, but there latest Cloud offering has only
:db/retractEntity.  Since that's what I am interested in using, that's
all I've implemented.

This transformation doesn't follow the existing pattern of inserting
into the temp.*_searches table.  It instead populates
temp.search_results directly from datoms, and in so doing it makes
some assumptions about how the searches tables will be accessed.

It might be even more efficient to have an entirely new temporary
table just for these retractions.  One advantage with the current
scheme is that indexing restrictions placed on the search results
table will apply to the datoms retracted by :db/retractEntity as well.

There are a few remaining items TODO.

TODO: ensure that we mark the schema as potentially modified when we
:db/retractEntity.  It's not clear to me how to do this efficiently.

TODO: ensure that transaction watchers get the correct transacted
datom stream.  I didn't try to address this yet because it appears to
me that the existing watcher implementation isn't quite correct: it
tells watchers about datoms that are potentially to be transacted but
not about datoms that actually are transacted.  This, of course,
matters when watching :db/retractEntity entities being transacted.

### DIFF
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -549,6 +549,9 @@ pub trait MentatStoring {
     fn insert_non_fts_searches<'a>(&self, entities: &'a [ReducedEntity], search_type: SearchType) -> Result<()>;
     fn insert_fts_searches<'a>(&self, entities: &'a [ReducedEntity], search_type: SearchType) -> Result<()>;
 
+    // :db/retractEntity the given entids.
+    fn insert_retract_entities<'a>(&self, entids: &'a [Entid]) -> Result<()>;
+
     /// Finalize the underlying storage layer after a Mentat transaction.
     ///
     /// Use this to finalize temporary tables, complete indices, revert pragmas, etc, after the
@@ -980,6 +983,48 @@ impl MentatStoring for rusqlite::Connection {
             .chain_err(|| "Could not drop FTS search ids!")?;
 
         results.map(|_| ())
+    }
+
+    /// Insert datoms corresponding to :db/retractEntity entities into the search results.
+    fn insert_retract_entities(&self, entids: &[Entid]) -> Result<()> {
+        let max_vars = self.limit(Limit::SQLITE_LIMIT_VARIABLE_NUMBER) as usize;
+        let bindings_per_statement = 2;
+
+        let chunks: itertools::IntoChunks<_> = entids.into_iter().chunks(max_vars / bindings_per_statement);
+
+        // We'd like to flat_map here, but it's not obvious how to flat_map across Result.
+        let results: Result<Vec<()>> = chunks.into_iter().map(|chunk| -> Result<()> {
+            let mut count = 0;
+            let mut params: Vec<&ToSql> = chunk.flat_map(|e| {
+                count += 1;
+                once(e as &ToSql)
+            }).collect();
+
+            // Two copies, first for testing e, then for testing v.
+            // TODO: perhaps we can reference the same named parameter multiple times, making this
+            // more efficient.
+            let mut params2 = params.clone();
+            params.append(&mut params2);
+
+            let values: String = repeat_values(count, 1);
+
+            // Note that the value for flags (-1) is nonsense.  Since these rows are being removed
+            // from datoms, flags/flags0 is not referenced and this is fine.  The value for the
+            // search type is also ignored.
+            let s: String = format!(r#"
+              INSERT INTO temp.search_results
+              SELECT d.e, d.a, d.v, d.value_type_tag, 0, -1, ':db.cardinality/many', d.rowid, d.v
+              FROM datoms AS d
+              WHERE d.e IN ({}) OR (d.value_type_tag = 0 AND d.v IN ({}))
+              "#, values, values);
+
+            let mut stmt = self.prepare_cached(s.as_str())?;
+            stmt.execute(&params)
+                .map(|_c| ())
+                .chain_err(|| "Could not retract entities!")
+        }).collect::<Result<Vec<()>>>();
+
+        results.and(Ok(()))
     }
 
     fn commit_transaction(&self, tx_id: Entid) -> Result<()> {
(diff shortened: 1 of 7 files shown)


ncalexan:
@rnewman @fluffyemily this is ready for initial review.

I have an open question about transaction watching, and don't see any documentation.  Right now, it appears that a transaction watcher will be told about any datom flowing through the system, whether it is or is not transacted.  That is, a duplicate datom that is dropped because it's already asserted will be provided to a transaction watcher.  That seems wrong to me, but I'm not sure what the intention actually is.

This matters for `:db/retractEntity` because it's non-trivial work to surface the set of datoms that are actually retracted.

rnewman:
Re tx watchers: yes, to avoid reading the transacted datoms back out of the store, we just use the input to the transactor. Naturally we might not get around reading the transacted datoms (or some aspects of them) for synced transactions, and for `retractEntity` we probably just want a way to tell the listener that an entity was retracted. The idea is that a watcher can always go to the store with the tx-id and interrogate it; the notification itself is as much detail as the transactor can cheaply provide.

db/src/db.rs
```diff
@@ -982,6 +985,48 @@ impl MentatStoring for rusqlite::Connection {
...
+            // from datoms, flags/flags0 is not referenced and this is fine.  The value for the
+            // search type is also ignored.
+            let s: String = format!(r#"
+              INSERT INTO temp.search_results
+              SELECT d.e, d.a, d.v, d.value_type_tag, 0, -1, ':db.cardinality/many', d.rowid, d.v
+              FROM datoms AS d
+              WHERE d.e IN ({}) OR (d.value_type_tag = 0 AND d.v IN ({}))
+              "#, values, values);
+
+            let mut stmt = self.prepare_cached(s.as_str())?;
```
rnewman: You're preparing the statement here but only using it once.

Your entity IDs are always safe to inline in the SQL, so I think it's better to do that (and thus avoid the `max_vars` check) unless you're reusing the prepared statement.

db/src/db.rs
```diff
@@ -982,6 +985,48 @@ impl MentatStoring for rusqlite::Connection {
...
+        // We'd like to flat_map here, but it's not obvious how to flat_map across Result.
+        let results: Result<Vec<()>> = chunks.into_iter().map(|chunk| -> Result<()> {
+            let mut count = 0;
+            let mut params: Vec<&ToSql> = chunk.flat_map(|e| {
+                count += 1;
+                once(e as &ToSql)
+            }).collect();
+
+            // Two copies, first for testing e, then for testing v.
+            // TODO: perhaps we can reference the same named parameter multiple times, making this
```
rnewman: You can: `?1` is the first supplied parameter, `?2` the second, and so on. (Yes, 1-indexed.)

db/src/db.rs
```diff
@@ -982,6 +985,48 @@ impl MentatStoring for rusqlite::Connection {
...
+            params.append(&mut params2);
+
+            let values: String = repeat_values(count, 1);
+
+            // Note that the value for flags (-1) is nonsense.  Since these rows are being removed
+            // from datoms, flags/flags0 is not referenced and this is fine.  The value for the
+            // search type is also ignored.
+            let s: String = format!(r#"
+              INSERT INTO temp.search_results
+              SELECT d.e, d.a, d.v, d.value_type_tag, 0, -1, ':db.cardinality/many', d.rowid, d.v
```
rnewman: Is `search_type` really an ident string?! Can we please have an issue to use a `TINYINT` instead? That string will be duplicated for every row…

db/src/db.rs
```diff
@@ -982,6 +985,48 @@ impl MentatStoring for rusqlite::Connection {
         results.map(|_| ())
     }
 
+    /// Insert datoms corresponding to :db/retractEntity entities into the search results.
+    fn insert_retract_entities(&self, entids: &[Entid]) -> Result<()> {
```
rnewman: I don't think we want to allow retracting:

- Core vocabulary.
- Attributes.
- Transactions.

Could you add validation here?

db/src/db.rs
```diff
@@ -2387,6 +2432,71 @@ mod tests {
...
+        assert_transact!(conn, "[[:db/retractEntity 999]]");
+        assert_matches!(conn.last_transaction(),
+                        "[[888 :test/dangling 999 ?tx false]
+                          [?tx :db/txInstant ?ms ?tx true]]");
+
+        // Let's make sure we actually retracted from the datoms table.
+        assert_matches!(conn.datoms(),
+                        "[[333 :db/ident :test/dangling]
+                          [333 :db/valueType :db.type/ref]]");
+    }
```
rnewman: Add a test to make sure I can't retract a tx-id or an attribute.

db/src/tx.rs
```diff
@@ -677,6 +689,12 @@ impl<'conn, 'a, W> Tx<'conn, 'a, W> where W: TransactWatcher {
                         (true, true) => fts_many.push(reduced),
                     }
                 },
+
+                Term::RetractEntity(KnownEntid(e)) => {
+                    // TODO: Might update metadata?
```
rnewman: We shouldn't allow retracting vocabulary, so this might update idents… but that's a can of worms.

db/src/tx.rs
```diff
@@ -677,6 +689,12 @@ impl<'conn, 'a, W> Tx<'conn, 'a, W> where W: TransactWatcher {
...
+                    // TODO: Watching/counting datoms?
```
rnewman: Yep, the observer service needs to be extended to be told about retracted entities.

db/src/tx.rs
```diff
@@ -705,6 +723,8 @@ impl<'conn, 'a, W> Tx<'conn, 'a, W> where W: TransactWatcher {
             self.store.insert_fts_searches(&fts_many[..], db::SearchType::Exact)?;
         }
 
+        self.store.insert_retract_entities(&retract_entities[..])?;
```
rnewman: Do this first, if only for conceptual clarity?

ncalexan:
I realize that this also is not recursing to retract `:db/isComponent` attributes.  In fact, we don't retract `:db/isComponent` attributes at all right now, and that's not clearly messaged.

I ended up building some substantial work on top of this (#673) and then, in the course of that work, started to doubt this implementation strategy.  Fundamentally, this represents `RetractEntity` as an `Entity` (high-level) _and_ as a `Term` (low-level).  That's awkward for the term building apparatus, and I think that awkwardness speaks to an interface mismatch: maybe we shouldn't be exposing high-level concepts to the low-level interface?  If it's better to keep `Term` low-level, then the implementation of `RetractEntity` might instead query the store to determine relevant terms as part of the transactor pipeline from `Entity` to `Term`, rather than pushing the work to a lower level.  That's probably slightly less efficient, but it might simplify the expression of the transactor.  I'm particularly interested in the same question for excision (#21).

In light of this uncertainty, I'm not going to try to push this across the line just now.

Issue #655, opened 2018-04-23
Add variable bindings to `transact`


ncalexan:
The more queries and transactions I write _as a library author_ building on top of Mentat, the more similar the query and transact syntax feels, and the more I want to bind variables in transact strings.  That is, I want to be able to transact
```edn
[:db/add [:credential/id "id"] :credential/title ?title]
```
and bind `?title` to `"A title coming from user-space."` independently.  This is just like the query syntax and like SQL, etc.  We already have some support this pattern in the term builder; this would just push it, in a more uniform manner, to the transactor.

This is similar to https://github.com/mozilla/mentat/issues/631, but simpler.

PR #397, opened 2017-03-31
Use Rc for TypedValue and Variable. (#395) r=nalexander

rnewman:
Part 1, core: use Rc for String and Keyword.
Part 2, query: use Rc for Variable.
Part 3, sql: use Rc for args in SQLiteQueryBuilder.
Part 4, query-algebrizer: use Rc.
Part 5, db: use Rc.
Part 6, query-parser: use Rc.
Part 7, query-projector: use Rc.
Part 8, query-translator: use Rc.
Part 9, top level: use Rc.

Issue #395, opened 2017-03-29
Make ConjoiningClauses feasible to fork


rnewman:
CC includes bindings provided by the user, extensive mappings from `Variable`s, and other large heap-allocated data. 

The processing of some patterns (e.g., `or`) is most conveniently expressed as a partial copy of the CC, conventional application, then a merge back. In order for this to be efficient in Rust we need to avoid copying all of these strings. 

We could do this by inserting `Rc` into some fundamental type definitions. We could also use `Cow` or similar to allow a derived CC to share state with its parent, at the cost of adding a lifetime to CC. There are other potential solutions. 

Choosing one and implementing it is a necessary precursor to implementing `or`.

rnewman:
Slipped in interning of `PatternNonValuePlace::Ident` and `PatternValuePlace::IdentOrKeyword`, too.

a5023c7

PR #656, opened 2018-04-24
refactoring Rc->Arc

mmacedoeu:
Refactoring to Arc will get Send and ability for use on multi-threaded environments like webapps

### DIFF
--- a/sql/src/lib.rs
+++ b/sql/src/lib.rs
@@ -16,6 +16,7 @@ extern crate rusqlite;
 extern crate mentat_core;
 
 use std::rc::Rc;
+use std::sync::Arc;
 
 use std::collections::HashMap;
 
@@ -103,7 +104,7 @@ pub struct SQLiteQueryBuilder {
     // Instead we track byte and String arguments separately, mapping them to their argument name,
     // in order to dedupe. We'll add these to the regular argument vector later.
     byte_args: HashMap<Vec<u8>, String>,             // From value to argument name.
-    string_args: HashMap<Rc<String>, String>,        // From value to argument name.
+    string_args: HashMap<Arc<String>, String>,        // From value to argument name.
     args: Vec<(String, Rc<rusqlite::types::Value>)>, // (arg, value).
 }
 
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -31,7 +31,7 @@ use std::collections::{
 };
 
 use std::fmt;
-use std::rc::Rc;
+use std::sync::Arc;
 
 use enum_set::EnumSet;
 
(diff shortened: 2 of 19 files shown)


rnewman:
I agree that this needs to be addressed. However:

We almost certainly don't want to do all of these — e.g., query parser internals shouldn't escape across thread boundaries, so they don't need to be `Arc`.

This is also an opportunity to do better encapsulation; we shouldn't be leaking that strings and keywords are `Rc` or `Arc`. Laziness on my part!

More broadly, I'm not convinced that directly switching `Rc` to `Arc` in a bunch of places is the right thing to do: it's the easiest way to add that marker, yes, but at significant performance cost.

We use `TypedValue`s in four main ways:

- As part of query and transact parsing. These don't cross thread boundaries, and so don't need to be `Send`.
- In the cache. This needs to be `Send` and/or `Sync` in order for stores to be shared across threads, but I think we really want stores to be `Sync`, because we realistically don't want to be passing stores by value.
- In query results, which need to be handed back to callers.
- In query inputs, which need to be passed from the caller into the query engine, where they will be cloned around and chewed on. (This line item generalizes to #655.)

The reason we introduced `Rc` in the first place was for the latter (see #395 / #397). We will also need cheap cloning/reuse of values in query results when I finish building pull (#638).

My suspicion is that we need to introduce some kind of _context_ against which `TypedValue`s are valid. We'd have one for `QueryInputs`, one for `QueryResult`, and one for `AttributeCaches`.

sql/src/lib.rs
```diff
@@ -103,7 +104,7 @@ pub struct SQLiteQueryBuilder {
     // Instead we track byte and String arguments separately, mapping them to their argument name,
     // in order to dedupe. We'll add these to the regular argument vector later.
     byte_args: HashMap<Vec<u8>, String>,             // From value to argument name.
-    string_args: HashMap<Rc<String>, String>,        // From value to argument name.
+    string_args: HashMap<Arc<String>, String>,        // From value to argument name.
```
rnewman: I see no reason why this would need to change. Did you just replace every `Rc` in the project, or did you make the changes necessary to pass some particular test?

mmacedoeu: Some changes to Arc simplified the refactoring as most code with clone or cloned could remain the same instead of implementing some convert from Arc->Rc which would require to clone the inner value, so just cloning Arc potentially is cheaper then cloning the inner value into a new Rc.

rnewman:
I think I managed to boil this PR down to a one-line change in `types.rs`.

I'd like to get a little better understanding of the implications before I make that one-line change, and some other work needs to happen before Mentat is truly usable across threads (#660, for example).

rnewman:
Fixed in #678.

PR #658, opened 2018-04-24
Pre: eliminate some occurrences of Rc, largely through the magic of Into.

rnewman:
I figured this little polish would make work like #656 easier to read and experiment with. (@mmacedoeu, you might want to take a look.)

PR #659, opened 2018-04-25
Arc, refactored

rnewman:
Informally I don't see significant performance differences with this. More timing to do; you can switch between `Rc` and `Arc` by switching the commenting for a few lines in `types.rs`. 

This PR is like #656 but reproduced independently on top of #658. There are a few changes that should go into #658, and not all the code builds yet, but `cargo test -p mentat --test query` will run.


core/src/types.rs
```diff
@@ -72,6 +72,11 @@ impl<T> FromRc<T> for Arc<T> where T: Sized + Clone {
     }
 }
 
+//
+// Use Rc for values.
```
ncalexan: This comment isn't great.  At first I thought it was just wrong, but now I see that this isn't an alias, it's picking one of `Rc<T>` and the implicitly not chosen `Arc<T>`.

rnewman:
Comments improved, FFI works, and we can now use `Box`, `Rc`, and `Arc`.

@fluffyemily, please post-review https://github.com/mozilla/mentat/pull/659/commits/a6ec3f227230dff557959437b78e5606bcfc421c if this lands during your nighttime.

Our test times are incredibly stable, with no variance visible between the types.

`Box`:
```
cargo test -p mentat  14.38s user 13.03s system 172% cpu 15.888 total
```

`Rc`:
```
cargo test -p mentat  14.43s user 13.03s system 173% cpu 15.858 total
```

`Arc`:
```
cargo test -p mentat  14.43s user 12.97s system 172% cpu 15.854 total
```

PR #662, opened 2018-04-26
FFI Entity Building and Transactions

fluffyemily:
There are two ways to create each builder, directly from a `Store` or from an `InProgress`. Creating from `Store` will perform two actions, creating a new `InProgress` and then returning a builder from that `InProgress`. In the case of `store_entity_builder_with_entid` and `store_entity_builder_from_tempid`, the function goes a step further and calls `describe` or `describe_tempid` from the created `InProgressBuilder` and returning the `EntityBuilder` that results. These two functions are replicated on `InProgress`. This has been done to reduce the overhead of objects being passed over the FFI boundary.

The decision to do this enables us to go from something like

```
in_progress  = store_begin_transaction(store);
builder = in_progress_builder(in_progress);
entity_builder = in_progress_builder_describe(builder, entid);
```
to
```
entity_builder = store_entity_builder_from_entid(store);
```

There is an `add_*` and `retract_*` function specified for each `TypedValue` type for both `InProgressBuilder` and `EntityBuilder`.

To enable `transact` on `EntityBuilder` and `InProgressBuilder`, a new `repr(C)` struct has been created that contains a pointer to an `InProgress` and a pointer to a `Result<TxReport>` to allow passing the tuple result returned from `transact` on those types over the FFI.

Commit is possible from both builders and `InProgress`.


ffi/src/lib.rs
```diff
@@ -247,17 +255,737 @@ pub extern "C" fn store_open(uri: *const c_char) -> *mut Store {
 
 // TODO: begin_read
 
-// TODO: begin_transaction
+/// Starts a new transaction to allow multiple transacts to be
+/// performed together. This is more efficient than performing
+/// a large set of individual commits.
+///
+/// Returns as [Result<TxReport>](mentat::TxReport) as an [ExternResult](ExternResult).
```
ncalexan: nit: first "as" should be "a".

ffi/src/lib.rs
```diff
@@ -247,17 +255,737 @@ pub extern "C" fn store_open(uri: *const c_char) -> *mut Store {
...
+/// Callers must ensure that the pointer to the [Store](mentat::Store) is not dangling.
+///
+/// Callers are responsible for managing the memory for the return value.
+/// A destructor `tx_report_destroy` is provided for releasing the memory for this
+/// pointer type.
+///
+/// TODO: Document the errors that can result from begin_transaction
+#[no_mangle]
+pub unsafe extern "C" fn store_begin_transaction(store: *mut Store) -> *mut ExternResult {
+    let store = &mut*store;
```
ncalexan: Is this spacing our standard?  I would not write `&mutfoo` (it doesn't parse!) so I would expect `&mut *store`.

ffi/src/lib.rs
```diff
@@ -247,17 +255,737 @@ pub extern "C" fn store_open(uri: *const c_char) -> *mut Store {
...
+    Box::into_raw(Box::new(store.begin_transaction().into()))
+}
 
-/// Performs a single transaction against the store.
+/// Perform a single transact operation using the current in progress
+/// transaction. Takes edn as a string to transact.
+///
+/// Returns as [Result<TxReport>](mentat::TxReport) as an [ExternResult](ExternResult).
```
ncalexan: Ditto -- "returns as" should be "returns a".

ffi/src/lib.rs
```diff
@@ -247,17 +255,737 @@ pub extern "C" fn store_open(uri: *const c_char) -> *mut Store {
...
+    let store = &mut*store;
+    let temp_id = c_char_to_string(temp_id);
+    let result = store.begin_transaction().and_then(|in_progress| {
+        Ok(in_progress.builder().describe_tempid(&temp_id))
+    });
+    Box::into_raw(Box::new(result.into()))
+}
+
+/// Starts a new transaction and creates a builder for an entity with `entid`
+/// using the that transaction to allow for programmatic assertion of values for that entity.
```
ncalexan: "the that" (many places).

ffi/src/lib.rs
```diff
@@ -247,17 +255,737 @@ pub extern "C" fn store_open(uri: *const c_char) -> *mut Store {
...
+    Box::into_raw(Box::new(result.into()))
+}
+
+/// Uses `builder` to assert `value` for `kw` on entity `entid`.
+///
+/// # Errors
+///
+/// If `entid` is not present in the store.
+/// If `kw` is not a valid attribute in the store.
+/// If the type of `kw` is not `:db.type/string`.
```
ncalexan: This type is a schema-level consideration; it doesn't map directly to a concrete type in Rust, and it definitely doesn't in the FFI.  Do you mean it should be `TypedValue::String`?  I don't think so since you're interpreting it as a `char*`.  And even then you're turning it into a `TypedValue::Keyword`... did you mean `value` should be a string?

ncalexan: Yes, I think you did mean `value` should be of a certain type, here and many other places.  But the notes on the type still apply -- you either want a real, Rust, concrete type; or you want to talk about coercion into the value space in whatever way makes sense for FFI consumers.

ffi/src/lib.rs
```diff
@@ -657,7 +1386,7 @@ pub unsafe extern "C" fn typed_value_list_into_iter(values: *mut Vec<Binding>) -
 /// A destructor `typed_value_destroy` is provided for releasing the memory for this
 /// pointer type.
 #[no_mangle]
-pub unsafe extern "C" fn typed_value_list_iter_next(iter: *mut BindingIterator) -> *mut Binding {
+pub unsafe extern "C" fn typed_value_list_iter_next(iter: *mut BindingIterator) ->  *mut Binding {
```
ncalexan: I personally found this style (two spaces after `->`) odd, but now I see you're doing it as a change.  Why do you want the two spaces?

ffi/src/lib.rs
```diff
@@ -905,6 +1634,18 @@ pub unsafe extern "C" fn changelist_entry_at(tx_report: *mut TransactionChange,
     tx_report.changes[index].clone()
 }
 
+/// Destructor for releasing the memory of [InProgressBuilder](mentat::InProgressBuilder).
```
ncalexan: Why aren't we using `define_destructor!`?  Is it the life times?  You can define multiple rules for a macro to handle additional cases...

sdks/android/Mentat/library/src/main/java/com/mozilla/mentat/InProgressTransactionResult.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Override
+    protected List<String> getFieldOrder() {
+        return Arrays.asList("inProgress", "result");
+    }
+
+    public InProgress getInProgress() {
+        return new InProgress(this.inProgress);
+    }
+    public TxReport getReport() {
```
ncalexan: nit: newline.

rnewman:
> Apps can individually decide which threads should be used to execute which actions, rather than us making that decision for them. The FFI is then as thread safe as Mentat itself and the app that is accessing it.

I do think we should allow callers to decide where callbacks happen by dispatching to a thread of their choice, and that we should run queries and writes on the calling thread to allow the caller to choose where those happen, too.

But that doesn't mean we can ignore thread safety. If the FFI allows pointers to escape, then the calling code has to play nice.

E.g., if two threads both borrow a `Conn`, we must ensure that they respect the mutex for access to the schema and cache. Only one thread can use a SQLite connection handle at a time, and so only a single thread is allowed to use an `InProgress`. Presumably there are some values that are only valid within the scope of a callback, and shouldn't be retained beyond it. And so on. 

Those things need to be documented, and potentially annotated where Java allows us to.

ffi/src/lib.rs
fluffyemily: Changed to ```If the `:db/type` of the attribute described by `kw` is not``` instead

ffi/src/lib.rs
fluffyemily: No, this was a rebase issue. I fixed it in the parent branch and I think the rebase of this branch onto this caused it and I missed it when I reviewed.

ffi/src/lib.rs
fluffyemily: It was the lifetimes. My macro foo is not very strong but I shall try it out.

Issue #664, opened 2018-04-26
[tx] :db/tx is only partially supported in the transactor


ncalexan:
#225 added partial support for `:db/tx` in the transactor, but it's not quite complete; this ticket tracks completing it by allowing `:db/tx` in the entity column (so that additional metadata about the transaction can be recorded) and in the value column (so that data can refer to the current transaction). 

`:db/tx` is rather special, since it's a reference whose resolved entid varies with the transaction.  That's similar to, but in the implementation very different to, `:db/txInstant`, which is an attribute.  In particular, since `:db/txInstant` is in the core schema, we never need to worry about resolving it specially.

`:db/tx`, however, is a special value that doesn't exist in the (core) schema.  The initial support in #225 added it as a special tempid, and added special parsing support to produce it from EDN input.  It's not a bad approach, but it doesn't accommodate the entity builder, which bypasses EDN input entirely.  And I don't think that the tests are comprehensive, so I'm not convinced it is correctly resolved in both the `a` and `v` columns even when coming from EDN.

ncalexan:
Closed by #665.

PR #665, opened 2018-04-26
(tx) Replace :db/tx with (current-tx) transaction function and broade…

ncalexan:
…n support. (#664)

:db/tx (and Datomic's version, :datomic/tx) suffer from the same
ambiguities that [a v] lookup references do -- determining the type of
the result is context sensitive.  (In this case, is :db/tx a reference
to the current transaction ID, or is it a valid keyword?)  This commit
addresses the ambiguity by introducing a notion of a transaction
functions, and provides a little scaffolding for adding more (should
the need arise).  I left the scaffolding in place rather than handling
just (current-tx) because I started trying to
implement (current-txInstant) as well, which is more difficult -- see
the comments.

It's worth noting that this approach generalizes more or less directly
to ?input variables, since those can be eagerly bound like the
implemented transaction function (current-tx).

### DIFF
--- a/db/src/tx.rs
+++ b/db/src/tx.rs
@@ -282,17 +282,19 @@ impl<'conn, 'a, W> Tx<'conn, 'a, W> where W: TransactWatcher {
                         Ok(Either::Left(e))
                     },
 
-                    // Special case: current tx ID.
-                    entmod::EntidOrLookupRefOrTempId::TempId(TempId::Tx) => {
-                        Ok(Either::Left(self.tx_id))
-                    },
-
                     entmod::EntidOrLookupRefOrTempId::TempId(e) => {
                         Ok(Either::Right(LookupRefOrTempId::TempId(self.intern_temp_id(e))))
                     },
 
                     entmod::EntidOrLookupRefOrTempId::LookupRef(ref lookup_ref) => {
                         Ok(Either::Right(LookupRefOrTempId::LookupRef(self.intern_lookup_ref(lookup_ref)?)))
+                    },
+
+                    entmod::EntidOrLookupRefOrTempId::TxFunction(ref tx_function) => {
+                        match tx_function.op.0.as_str() {
+                            "transaction-tx" => Ok(Either::Left(self.tx_id)),
+                            unknown @ _ => bail!(ErrorKind::NotYetImplemented(format!("Unknown transaction function {}", unknown))),
+                        }
                     },
                 }
             }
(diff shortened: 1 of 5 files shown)


db/src/tx.rs
```diff
@@ -412,6 +421,26 @@ impl<'conn, 'a, W> Tx<'conn, 'a, W> where W: TransactWatcher {
...
+                                    "current-tx" => TypedValue::Ref(self.tx_id),
+                                    unknown @ _ => bail!(ErrorKind::NotYetImplemented(format!("Unknown transaction function {}", unknown))),
+                                };
+
+                                // Here we do schema-aware typechecking: we assert that the computed
+                                // value is in the attribute's value set.  If and when we have
+                                // transaction functions that produce numeric values, we'll have to
+                                // be more careful here, because a function that produces an integer
+                                // value can be used where a double is expected.  See also
+                                // `SchemaTypeChecking.to_typed_value(...)`.
```
rnewman: The algebrizer function stuff has some coercions that might be useful to extract.

tx/src/entities.rs
```diff
@@ -69,12 +66,30 @@ pub struct LookupRef {
     pub v: edn::Value, // An atom.
 }
 
+/// A "transaction function" that exposes some value determined by the current transaction.  The
+/// prototypical example is the current transaction ID, `(current-tx)`.
+///
+/// A natural next step might be to expose the current transaction instant `(current-txInstant)`,
+/// but that's more difficult: the transaction itself can set the transaction instant (with some
+/// restrictions), so the transaction function must be late-binding.  Right now, that's difficult to
```
rnewman: Perhaps you can work around that by simply defining `(current-txInstant)` as the timestamp that the transactor would use _by default_.

That would allow callers to do fun things like impose a timestamp but keep the transactor's timestamp around in a separate property, or correct clock drift, or whatever.

You could define the transactor's behavior as having a default

```edn
[:db/add (current-tx) :db/txInstant (current-txInstant)]
```

… which 'feels' better if you call it `(current-instant)` or `(transact-instant)`.

db/src/tx.rs
ncalexan: Thanks for the quick response.  I do expect there's some shared typechecking stuff between query and parse; they're getting closer, not further apart :)

tx/src/entities.rs
ncalexan: I think, in fact, that `(transaction-tx)` and `(transaction-instant)` might be better than any reference to "current", since current might be reasonably interpreted to mean "the latest transaction committed" or "the timestamp when I built this data structure".  I find that `(current-txInstant)` blurs that considerably.

As for allowing an eager `(current-txInstant)` while allowing to set `:db/txInstant`... it just seems so prone to error.  Follow-up, for sure.

Issue #616, opened 2018-04-03
Create FFI for caching for iOS and Android


fluffyemily:
Caching needs to be usable from native code, not just Rust code. An FFI interface needs to be created.

PR #667, opened 2018-04-27
Expose Caching over FFI

fluffyemily:
This exposes an FFI function for each direction of caching, `Forward`, `Reverse` and `Both`. This is to make is as clear as possible to consumers which direction they are caching their attributes in. 

The `CacheDirection` enum is replicated on each of the native sides to provide a usable interface, but the FFI layer is more explicit to avoid easy to make errors.

https://github.com/mozilla/mentat/issues/616


ffi/src/lib.rs
```diff
@@ -1245,9 +1246,55 @@ pub unsafe extern "C" fn tx_report_entity_for_temp_id(tx_report: *mut TxReport,
     }
 }
 
-// TODO: cache
+/// Adds an attribute to the cache.
+/// `store_cache_attribute_forward` caches values for an attribute keyed by entity
+/// (i.e. find values and entities that have this attribute, or find values of attribute for an entity)
+///
+/// # Safety
```
ncalexan: My comments on the earlier PR, about the relatively low value of these particular safety blocks, are relevant here.

ffi/src/lib.rs
```diff
@@ -1245,9 +1246,55 @@ pub unsafe extern "C" fn tx_report_entity_for_temp_id(tx_report: *mut TxReport,
     }
 }
 
-// TODO: cache
+/// Adds an attribute to the cache.
+/// `store_cache_attribute_forward` caches values for an attribute keyed by entity
```
ncalexan: nit: full sentences.  I think it would be more useful to lead with the `i.e.`, preferably in the form "for fixed `a`, cache the map from `e` to `v` such that `[e a v]` is a datom in the store".  Below, you'd have "cache the map `v` to `e`."

ffi/src/lib.rs
```diff
@@ -1245,9 +1246,55 @@ pub unsafe extern "C" fn tx_report_entity_for_temp_id(tx_report: *mut TxReport,
...
+/// (i.e. find entities that have a particular value for an attribute).
+///
+/// # Safety
+///
+/// Callers must ensure that the pointer to the `Store` is not dangling and that
+/// the C string provided to `attribute` is valid.
+#[no_mangle]
+pub extern "C" fn store_cache_attribute_reverse(store: *mut Store, attribute: *const c_char) -> *mut ExternResult {
+    let store = unsafe { &mut *store };
+    let kw = kw_from_string(c_char_to_string(attribute));
```
ncalexan: Again, similar comments to the earlier (documentation) PR -- can we avoid this allocation?  Maybe we need to figure out how to find internal string references so that we can look up `NamespacedKeyword` by `(&str, &str)` (and therefore, hopefully, by `(CStr, CStr)`).  That is, can we make `(&str, &str) impl Borrow<&NamespacedKeyword>`?  I don't know if that's possible without changing the representation of `NamespacedKeyword`.  But it's going to be very common, so we might want to do this.

sdks/swift/Mentat/Mentat/Mentat.swift
```diff
@@ -56,11 +62,36 @@ class Mentat: RustObject {
         self.init(raw: store_open(storeURI))
     }
 
+    /**
+     Add an attribute to the cache. The {@link CacheDirection} determines how that attribute can be
+     looked up.
+
+     - Parameter attribute: The attribute to cache
+     - Parameter direction: The direction the attribute should be keyed.
+        `FORWARD` caches values for an attribute keyed by entity
```
ncalexan: Is the allcaps style a Swiftdoc thing?

sdks/swift/Mentat/MentatTests/MentatTests.swift
```diff
@@ -1175,5 +1175,70 @@ class MentatTests: XCTestCase {
...
+
+        XCTAssertNoThrow(try mentat.query(query: query).run { (result) in
+            cachedTimer.end()
+            XCTAssertNotNil(result)
+        })
+
+        let timingDifference = uncachedTimer.duration() - cachedTimer.duration()
+        print("Cached query is \(timingDifference) nanoseconds faster than the uncached query")
+
+        XCTAssertLessThan(cachedTimer.duration(), uncachedTimer.duration())
```
ncalexan: This seems optimistic, but lets see how it does in the wild.

sdks/android/Mentat/library/src/androidTest/java/com/mozilla/mentat/FFIIntegrationTest.java
```diff
@@ -1241,5 +1245,77 @@ public void handleRow(TupleResult row) {
...
+        long timingDifference = uncachedTimer.duration() - cachedTimer.duration();
+        Log.d("testCaching", "Cached query is "+ timingDifference +" nanoseconds faster than the uncached query");
+
+        assert cachedTimer.duration() < uncachedTimer.duration();
+
+    }
+
+}
+
+class QueryTimer {
```
ncalexan: Is this even legal Java?  Two classes at top-level?  Make this an inner class, or (if you think it'll be used) make it an exposed class.  Perhaps you want to make it easier to run `.query()` with a given timer?

sdks/android/Mentat/library/src/main/java/com/mozilla/mentat/Mentat.java
```diff
@@ -178,6 +231,15 @@ public EntityBuilder entityBuilder(long entid) {
         return null;
     }
 
+    /**
+     * Creates a new transaction ({@link InProgress}) and returns an {@link EntityBuilder} for a new
+     * entity with `tempId` for that transaction.
+     *
+     * TODO: Throw an exception if the result contains an error.
```
ncalexan: Do we have a ticket on file for this error handling discipline?  Let's link to it, please.

sdks/android/Mentat/library/src/androidTest/java/com/mozilla/mentat/FFIIntegrationTest.java
fluffyemily: Two classes are allowed at the top level if only one of them is public. It's just for tests so I didn't think it would matter too much, but I'll make it an inner class to be sure.

sdks/android/Mentat/library/src/main/java/com/mozilla/mentat/Mentat.java
fluffyemily: https://github.com/mozilla/mentat/issues/700

Issue #663, opened 2018-04-26
[tx] `transact_builder` is not type-checking input values


ncalexan:
It appears that `transact_builder` is not type-checking the provided values, so that, say:
```
builder.add(sl.clone(), in_progress.get_entid(&ATTR).expect("attr"), TypedValue::typed_string("test")).expect("to add");
```
and
```
builder.add(sl.clone(), in_progress.get_entid(&ATTR).expect("attr"), TypedValue::Keyword(kw!(:some/value).into())).expect("to add");
```
successfully transact.  It's not clear to me where in the stack the error is:
- is it that `transact` in general is not type checking correctly?
- is it that `transact_builder`, which was added long after the initial `transact` implementation, is inserting itself into the stack after the type checking is expected to have been done?

In either case this is critical!

ncalexan:
Oh dear.  `transact_builder` invokes `transact_terms`

https://github.com/mozilla/mentat/blob/ab957948b4f642e9d4e5e6ac8169c8b382295573/db/src/tx.rs#L785

which invokes `transact_simple_terms`

https://github.com/mozilla/mentat/blob/ab957948b4f642e9d4e5e6ac8169c8b382295573/db/src/tx.rs#L544

which skips the first stage of the transact pipeline that turns `entities` into `terms`... and along the way does all of the schema-aware typechecking

https://github.com/mozilla/mentat/blob/ab957948b4f642e9d4e5e6ac8169c8b382295573/db/src/tx.rs#L532

It's probably not hard to address this, with some slight duplication of work, by growing typechecking assertions in the `SchemaTypeChecking` trait and invoking them as part of the term processing loop.  It will be a little awkward that passing an EDN `:keyword/value` ident as the value to an attribute of type `:db.type/ref` will look up the entid for you, but (as written) passing a term `TypedValue::Keyword(...)` as the value to an attribute of type `:db.type/ref` won't look up the entid for you... but that's the advantage and disadvantage of using a lower level interface.

ncalexan:
Addressed by https://github.com/mozilla/mentat/commit/46c2a0801f450f4f62482ee067ed272e1057a747.

Issue #535, opened 2018-01-23
Use a more sophisticated representation of datoms to insert in order to detect errors


rnewman:
#531 is the original issue. #533 made it so that an error is thrown, but it's not a useful error: it's very internal. It would be better to detect the two datoms that failed.

One way to do this is to build a trie of `e -> a -> v` instead of a vector of `ReducedEntity`, which not only eliminates duplication and perhaps gives us a more efficient time with insertion, but also lets us exactly detect duplicates. If we can build this trie as part of the last transact generation pass, then this might be cheap or even a saving.

PR #673, opened 2018-05-01
Improve type checking and constraint checking; treat input as set.

ncalexan:
This addresses #532 and should address #663.  It's ready for initial review but it's not quite there yet -- part 2 needs additional tests to verify that it really does address #663, and part 3 needs a list of routine tests so that issues are found in a dedicated portion of the test suite, and so that we verify that the error output is as desired.

### DIFF
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -146,16 +146,7 @@ lazy_static! {
         r#"CREATE UNIQUE INDEX idx_datoms_unique_value ON datoms (a, value_type_tag, v) WHERE unique_value IS NOT 0"#,
 
         r#"CREATE TABLE transactions (e INTEGER NOT NULL, a SMALLINT NOT NULL, v BLOB NOT NULL, tx INTEGER NOT NULL, added TINYINT NOT NULL DEFAULT 1, value_type_tag SMALLINT NOT NULL)"#,
-
-        // This index serves two purposes:
-        //
-        // - First, it allows to query specific transactions efficiently, and then to filter by
-        //   attribute in a transaction efficiently.
-        //
-        // - Second, it provides a layer of assurance that the transactor is correct by asserting
-        //   that a single datom `[e a v]` is added or retracted at most once in a single
-        //   transaction.
-        r#"CREATE UNIQUE INDEX idx_transactions_unique_tx_a_e_v ON transactions (tx, a, e, v)"#,
+        r#"CREATE INDEX idx_transactions_tx ON transactions (tx, added)"#,
 
         // Fulltext indexing.
         // A fulltext indexed value v is an integer rowid referencing fulltext_values.
@@ -524,14 +515,12 @@ fn search(conn: &rusqlite::Connection) -> Result<()> {
 fn insert_transaction(conn: &rusqlite::Connection, tx: Entid) -> Result<()> {
     // Mentat follows Datomic and treats its input as a set.  That means it is okay to transact the
     // same [e a v] twice in one transaction.  However, we don't want to represent the transacted
-    // datom twice, so we exploit the idx_transactions_unique_tx_a_e_v on transactions to IGNORE
-    // duplicated datoms coming from our search results.  This is a choice: we could also ensure
-    // that the search results satisfied the uniqueness constraints, or we could SELECT DISTINCT
-    // from the search results.  We're doing this since we need an index no transactions anyway, so
-    // we might as well make it a UNIQUE index and exploit it for this purpose too.
+    // datom twice.  Therefore, the transactor unifies repeated datoms, and in addition we add
+    // indices to the search inputs and search results to ensure that we don't see repeated datoms
+    // at this point.
 
     let s = r#"
-      INSERT OR IGNORE INTO transactions (e, a, v, tx, added, value_type_tag)
+      INSERT INTO transactions (e, a, v, tx, added, value_type_tag)
       SELECT e0, a0, v0, ?, 1, value_type_tag0
       FROM temp.search_results
       WHERE added0 IS 1 AND ((rid IS NULL) OR ((rid IS NOT NULL) AND (v0 IS NOT v)))"#;
@@ -542,7 +531,7 @@ fn insert_transaction(conn: &rusqlite::Connection, tx: Entid) -> Result<()> {
         .chain_err(|| "Could not insert transaction: failed to add datoms not already present")?;
 
     let s = r#"
-      INSERT OR IGNORE INTO transactions (e, a, v, tx, added, value_type_tag)
+      INSERT INTO transactions (e, a, v, tx, added, value_type_tag)
       SELECT e0, a0, v, ?, 0, value_type_tag0
       FROM temp.search_results
       WHERE rid IS NOT NULL AND
@@ -581,11 +570,11 @@ fn update_datoms(conn: &rusqlite::Connection, tx: Entid) -> Result<()> {
     // Insert datoms that were added and not already present. We also must expand our bitfield into
     // flags.  Since Mentat follows Datomic and treats its input as a set, it is okay to transact
     // the same [e a v] twice in one transaction, but we don't want to represent the transacted
-    // datom twice in datoms.  Hence we IGNORE repeated datoms here.  This is a choice: we could
-    // also push the uniqueness constraints into the search results themselves, or SELECT DISTINCT
-    // from the search results.  This exploits the existing UNIQUE indices on the `datoms` table.
+    // datom twice in datoms.  The transactor unifies repeated datoms, and in addition we add
+    // indices to the search inputs and search results to ensure that we don't see repeated datoms
+    // at this point.
     let s = format!(r#"
-      INSERT OR IGNORE INTO datoms (e, a, v, tx, value_type_tag, index_avet, index_vaet, index_fulltext, unique_value)
+      INSERT INTO datoms (e, a, v, tx, value_type_tag, index_avet, index_vaet, index_fulltext, unique_value)
       SELECT e0, a0, v0, ?, value_type_tag0,
              flags0 & {} IS NOT 0,
              flags0 & {} IS NOT 0,
@@ -700,6 +689,11 @@ impl MentatStoring for rusqlite::Connection {
                added0 TINYINT NOT NULL,
                flags0 TINYINT NOT NULL)"#,
 
+            // It is fine to transact the same [e a v] twice in one transaction, but the transaction
+            // processor should unify such repeated datoms.  This index will cause insertion to fail
+            // if the transaction processor incorrectly tries to assert the same (cardinality one)
+            // datom twice.  (Sadly, the failure is opaque.)
+            r#"CREATE UNIQUE INDEX IF NOT EXISTS temp.inexact_searches_unique ON inexact_searches (e0, a0) WHERE added0 = 1"#,
             r#"DROP TABLE IF EXISTS temp.search_results"#,
             // TODO: don't encode search_type as a STRING.  This is explicit and much easier to read
             // than another flag, so we'll do it to start, and optimize later.
@@ -713,6 +707,13 @@ impl MentatStoring for rusqlite::Connection {
                search_type STRING NOT NULL,
                rid INTEGER,
                v BLOB)"#,
+            // It is fine to transact the same [e a v] twice in one transaction, but the transaction
+            // processor should identify those datoms.  This index will cause insertion to fail if
+            // the internals of the database searching code incorrectly find the same datom twice.
+            // (Sadly, the failure is opaque.)
+            //
+            // N.b.: temp goes on index name, not table name.  See http://stackoverflow.com/a/22308016.
+            r#"CREATE UNIQUE INDEX IF NOT EXISTS temp.search_results_unique ON search_results (e0, a0, v0, value_type_tag0)"#,
         ];
 
         for statement in &statements {
@@ -2541,7 +2542,7 @@ mod tests {
         ]"#);
 
         // Observe that "foo" and "zot" upsert to the same entid, and that doesn't cause a
-        // cardinality constraint, because we treat the input with set semantics and accept
+        // cardinality conflict, because we treat the input with set semantics and accept
         // duplicate datoms.
         let report = assert_transact!(conn, r#"[
             [:db/add "bar" :page/id "z"]
--- a/db/src/upsert_resolution.rs
+++ b/db/src/upsert_resolution.rs
@@ -71,8 +71,8 @@ pub(crate) struct Generation {
     upserts_ev: Vec<UpsertEV>,
 
     /// Entities that look like:
-    /// - [:db/add TEMPID b OTHERID]. where b is not :db.unique/identity;
-    /// - [:db/add TEMPID b v].  b may be :db.unique/identity.
+    /// - [:db/add TEMPID b OTHERID].  b may be :db.unique/identity if it has failed to upsert.
+    /// - [:db/add TEMPID b v].  b may be :db.unique/identity if it has failed to upsert.
     /// - [:db/add e b OTHERID].
     allocations: Vec<TermWithTempIds>,
 


core/src/types.rs
```diff
@@ -121,7 +121,7 @@ impl<T: Clone> Cloned<T> for Box<T> where T: Sized + Clone {
 /// This type must implement `FromRc` and `Cloned`, and a `From` implementation must exist for
 /// `TypedValue`.
 ///
-pub type ValueRc<T> = Rc<T>;
+pub type ValueRc<T> = Arc<T>;
```
rnewman: Can we make this its own tiny PR?

rnewman: (And comment about it in #656)

db/src/db.rs
```diff
@@ -146,7 +146,16 @@ lazy_static! {
...
         r#"CREATE TABLE transactions (e INTEGER NOT NULL, a SMALLINT NOT NULL, v BLOB NOT NULL, tx INTEGER NOT NULL, added TINYINT NOT NULL DEFAULT 1, value_type_tag SMALLINT NOT NULL)"#,
-        r#"CREATE INDEX idx_transactions_tx ON transactions (tx, added)"#,
+
+        // This index serves two purposes:
+        //
+        // - First, it allows to query specific transactions efficiently, and then to filter by
+        //   attribute in a transaction efficiently.
+        //
+        // - Second, it provides a layer of assurance that the transactor is correct by asserting
+        //   that a single dato `[e a v]` is added or retracted at most once in a single
```
rnewman: datom

db/src/db.rs
```diff
@@ -513,8 +522,16 @@ fn search(conn: &rusqlite::Connection) -> Result<()> {
 ///
 /// See https://github.com/mozilla/mentat/wiki/Transacting:-entity-to-SQL-translation.
 fn insert_transaction(conn: &rusqlite::Connection, tx: Entid) -> Result<()> {
+    // Mentat follows Datomic and treats its input as a set.  That means it is okay to transact the
+    // same [e a v] twice in one transaction.  However, we don't want to represent the transacted
+    // datom twice, so we exploit the idx_transactions_unique_tx_a_e_v on transactions to IGNORE
+    // duplicated datoms coming from our search results.  This is a choice: we could also ensure
+    // that the search results satisfied the uniqueness constraints, or we could SELECT DISTINCT
+    // from the search results.  We're doing this since we need an index no transactions anyway, so
```
rnewman: index on

db/src/db.rs
```diff
@@ -146,7 +146,16 @@ lazy_static! {
...
+
+        // This index serves two purposes:
+        //
+        // - First, it allows to query specific transactions efficiently, and then to filter by
+        //   attribute in a transaction efficiently.
+        //
+        // - Second, it provides a layer of assurance that the transactor is correct by asserting
+        //   that a single dato `[e a v]` is added or retracted at most once in a single
+        //   transaction.
+        r#"CREATE UNIQUE INDEX idx_transactions_unique_tx_a_e_v ON transactions (tx, a, e, v)"#,
```
rnewman: Good opportunity to write a disk-level migration as an example, and avoid breaking any test DBs…

db/src/db.rs
```diff
@@ -561,10 +578,14 @@ fn update_datoms(conn: &rusqlite::Connection, tx: Entid) -> Result<()> {
...
         .chain_err(|| "Could not update datoms: failed to retract datoms already present")?;
 
-    // Insert datoms that were added and not already present. We also must
-    // expand our bitfield into flags.
+    // Insert datoms that were added and not already present. We also must expand our bitfield into
+    // flags.  Since Mentat follows Datomic and treats its input as a set, it is okay to transact
+    // the same [e a v] twice in one transaction, but we don't want to represent the transacted
+    // datom twice in datoms.  Hence we IGNORE repeated datoms here.  This is a choice: we could
+    // also push the uniqueness constraints into the search results themselves, or SELECT DISTINCT
+    // from the search results.  This exploits the existing UNIQUE indices on the `datoms` table.
```
rnewman: Another approach to doing this is one I remember outlining in #535 — if you build a trie instead of a collection of terms, you have a perfect opportunity to dedupe, and you don't need to do it here. Thoughts?

db/src/db.rs
```diff
@@ -2422,6 +2436,55 @@ mod tests {
...
+        let mut conn = TestConn::default();
+
+        assert_transact!(conn, r#"[
+            {:db/ident :page/id :db/valueType :db.type/string :db/index true :db/unique :db.unique/identity}
+            {:db/ident :page/ref :db/valueType :db.type/ref :db/index true :db/unique :db.unique/identity}
+            {:db/ident :page/title :db/valueType :db.type/string :db/cardinality :db.cardinality/many}
+        ]"#);
+
+        // Observe that "foo" and "zot" upsert to the same entid, and that doesn't cause a
+        // cardinality constraint, because we treat the input with set semantics and accept
```
rnewman: cardinality conflict

db/src/tx_checking.rs
```diff
@@ -44,3 +52,62 @@ where I: IntoIterator<Item=&'a AttributedTerm<'a>> {
...
+
+    for term in terms.into_iter() {
+        match term {
+            &Term::AddOrRetract(op, e, (a, attribute), ref v) => {
+                let map = match op {
+                    OpType::Add => &mut added,
+                    OpType::Retract => &mut retracted,
+                };
+                map
+                    .entry(e).or_insert(BTreeMap::default())
```
rnewman: Join to previous line

db/src/tx_checking.rs
```diff
@@ -44,3 +52,62 @@ where I: IntoIterator<Item=&'a AttributedTerm<'a>> {
...
+/// - add any attribute to an entity that is also the subject of a `:db/retractEntity` term
+///
+/// We try to be maximally helpful by yielding every malformed set of datoms, rather than just the
+/// first set, or even the first conflict.  In the future, we might change this choice, or allow the
+/// consumer to specify the robustness of the cardinality checking desired.
+pub(crate) fn cardinality_conflicts<'a, I>(terms: I) -> Vec<CardinalityConflict>
+where I: IntoIterator<Item=&'a AttributedTerm<'a>> {
+    let mut errors = vec![];
+
+    let mut added: BTreeMap<Entid, BTreeMap<(Entid, bool), BTreeSet<&TypedValue>>> = BTreeMap::default();
```
rnewman: You're basically building a trie here. If you slightly tweak your approach, at this point you can:

1. Look up attributes, find the value type once, and make sure each incoming value is of the right type.
2. Implement the one/many distinction in two tries, using the presence of an `Entry` as a cardinality failure.
3. Drop `Term`s at this point and use the tries instead.

It seems wasteful to do all of this work and then not use the output for other things!

Also: it would be worth considering whether a `BTreeMap` and `BTreeSet` are the right data structures to use. Perhaps switch between `Hash*` and `BTree*` and use `cargo cli` to `.i fixtures/cities.schema`, `.i fixtures/all_seattle.edn` and see which is faster? I suspect the hash versions will be better.

db/src/upsert_resolution.rs
```diff
@@ -64,8 +71,8 @@ pub(crate) struct Generation {
     upserts_ev: Vec<UpsertEV>,
 
     /// Entities that look like:
-    /// - [:db/add TEMPID b OTHERID], where b is not :db.unique/identity;
-    /// - [:db/add TEMPID b v], where b is not :db.unique/identity.
+    /// - [:db/add TEMPID b OTHERID]. where b is not :db.unique/identity;
```
rnewman: s/where//

db/src/upsert_resolution.rs
```diff
@@ -212,36 +219,83 @@ impl Generation {
         temp_id_avs
     }
 
-    /// After evolution is complete, yield the set of tempids that require entid allocation.  These
-    /// are the tempids that appeared in [:db/add ...] entities, but that didn't upsert to existing
-    /// entids.
-    pub(crate) fn temp_ids_in_allocations(&self) -> BTreeSet<TempIdHandle> {
+    /// After evolution is complete, yield the set of tempids that require entid allocation are the
```
rnewman: Finish rewriting this sentence :)

db/src/upsert_resolution.rs
```diff
@@ -212,36 +219,83 @@ impl Generation {
...
-                    // other upserts (or they fail the transaction).
-                },
+                // [:db/retract ...] entities never allocate entids; they have to resolve due to
+                // other upserts (or they fail the transaction).
+                &Term::AddOrRetract(OpType::Retract, _, _, _) => unreachable!(),
             }
         }
 
-        temp_ids
+        // Now we union-find all the known tempids.  Two tempids are unioned if they both appears as
```
rnewman: both appear

db/src/db.rs
ncalexan: Yeah, I'm happy to do this.  There are a few transformations that would improve the efficiency of this code:

- not collecting `Vec<>` as frequently -- much easier now that `impl Trait` is stable
- turning the main (hard) loop into an iterator, again avoiding a `collect`
- turning `final_terms` and the type/cardinality checking into a single pass that collects a better data structure.

It's not clear that a trie really wins here -- would you be happy with nested maps, like I'm using in the cardinality checking phase?

rnewman: I mostly mean a trie in the sense of shared prefixing, so yes.

Consider that the transactor always requires a fixed attribute. That becomes our root (just as it is for the cache!).

The entities and values can be lookup refs, functions, tempids, or known values.

We can build, for each attribute, entity to single value mappings for cardinality-one attributes, entity to multiple values mappings for cardinality-many. We can track value to entity reverse mappings for unique attributes. Intra-transaction cardinality violations and duplicates become obvious — they're hash collisions. Uniqueness violations/collisions/smushing, similarly. Perhaps two-tempids-are-the-same is as simple as "oh, the entity for this unique value is another tempid, they must be the same".

I think it should be possible to rephrase the various resolution transitions in the transactor phases as movements between these maps, or replacements of entries with one kind of `Either` for another.

We lose some amount of clarity, and might get a little bit less type system support, but in exchange we would get different kinds of collisions, and of course prefix compression, for free.

core/src/types.rs
ncalexan: https://github.com/mozilla/mentat/pull/678

db/src/tx_checking.rs
```diff
@@ -0,0 +1,46 @@
...
+///
+/// We try to be maximally helpful by yielding every malformed datom, rather than only the first.
+/// In the future, we might change this choice, or allow the consumer to specify the robustness of
+/// the type checking desired, since there is a cost to providing helpful diagnostics.
+pub(crate) fn type_disagreements<'a, I>(terms: I) -> BTreeMap<(Entid, Entid, TypedValue), ValueType>
+where I: IntoIterator<Item=&'a AttributedTerm<'a>> {
+    let errors: BTreeMap<_, _> = terms.into_iter().filter_map(|term| {
+        match term {
+            &Term::AddOrRetract(_, e, (a, attribute), ref v) => {
+                if attribute.value_type != v.value_type() {
```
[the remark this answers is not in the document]
ncalexan: I've reworked this to use a trie, and to get rid of `AttributedTerm` altogether.   (I was trying to mix `filter_map` with `Result`, which is hard -- hence `AttributedTerm`.  Without `filter_map`, no need for the transform in the middle.)  I've kept the informative errors; it's easy to add a reporting mode that bails more eagerly later.

db/src/db.rs
ncalexan: Meh, I don't really care to do this.  I'm going to leave as is and think more about what indexing we really want post-trie.

ncalexan:
@rnewman ok, back to you.  This is a little wonky, 'cuz I dropped the part about `AttributedTerm` instances, and then I built the trie structure on top of what was already there.  I think the end of all that is to fold everything down and accept that it's a pretty big blob :/

core/src/types.rs
rnewman: Needs a rebase on `master`…

db/src/db.rs
```diff
@@ -1242,6 +1191,29 @@ mod tests {
             Ok(report)
         }
 
+        fn transact_simple_terms<I>(&mut self, terms: I, tempid_set: InternSet<TempId>) -> Result<TxReport> where I: IntoIterator<Item=TermWithTempIds> {
+            let details = {
+                // The block scopes the borrow of self.sqlite.
+                // We're about to write, so go straight ahead and get an IMMEDIATE transaction.
+                let tx = self.sqlite.transaction_with_behavior(TransactionBehavior::Immediate)?;
+                // Applying the transaction can fail, so we don't unwrap.
+                let details = transact_terms(&tx, self.partition_map.clone(), &self.schema, &self.schema, NullWatcher(), terms, tempid_set)?;
```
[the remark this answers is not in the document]
ncalexan: Nah, this is test only.

ncalexan:
> I'm going to need to read this change closely, but figured I'd hit "Request changes" first 'cos you've missed a bunch of the nits I left in the last review — just read through the diff and take a look!

I looked through the nits pretty closely, and have a final commit addressing the ones I thought needed to be addressed.  Can you reflag at least a couple so I can see what I missed?

db/src/db.rs
```diff
@@ -146,7 +146,16 @@ lazy_static! {
...
+
+        // This index serves two purposes:
+        //
+        // - First, it allows to query specific transactions efficiently, and then to filter by
+        //   attribute in a transaction efficiently.
+        //
+        // - Second, it provides a layer of assurance that the transactor is correct by asserting
+        //   that a single datom `[e a v]` is added or retracted at most once in a single
+        //   transaction.
+        r#"CREATE UNIQUE INDEX idx_transactions_unique_tx_a_e_v ON transactions (tx, a, e, v)"#,
```
rnewman: Nit: I think you noted that you wanted to get rid of this.

db/src/db.rs
rnewman: Nit: index no -> index on

db/src/db.rs
rnewman: Nit: constraint -> conflict

db/src/upsert_resolution.rs
rnewman: Nit: s/where//

db/src/tx.rs
```diff
@@ -695,63 +698,45 @@ impl<'conn, 'a, W> Tx<'conn, 'a, W> where W: TransactWatcher {
         // We need to ensure that callers can't blindly transact entities that haven't been
         // allocated by this store.
 
+        let errors = tx_checking::type_disagreements(&aev_trie);
```
rnewman: I'm a bit surprised why this isn't a side-effect of insertion. After all, you know the attribute at the moment you're finalizing that datom, and so you also know which type the value must be…

db/src/tx.rs
```diff
@@ -695,63 +698,45 @@ impl<'conn, 'a, W> Tx<'conn, 'a, W> where W: TransactWatcher {
...
+        if !errors.is_empty() {
+            bail!(ErrorKind::SchemaConstraintViolation(errors::SchemaConstraintViolation::TypeDisagreements { conflicting_datoms: errors }));
+        }
+
+        let errors = tx_checking::cardinality_conflicts(&aev_trie);
```
rnewman: Same here. You know there are two assertions because you encounter the other one while you're inserting…

db/src/tx.rs
```diff
@@ -859,3 +844,59 @@ pub fn transact_terms<'conn, 'a, I, W>(conn: &'conn rusqlite::Connection,
     let report = tx.transact_simple_terms(terms, tempid_set)?;
     conclude_tx(tx, report)
 }
+
+fn extend_aev_trie<'schema, I>(schema: &'schema Schema, terms: I, trie: &mut AEVTrie<'schema>) -> Result<()>
+where I: IntoIterator<Item=TermWithoutTempIds>
+{
+    for Term::AddOrRetract(op, KnownEntid(e), a, v) in terms.into_iter() {
```
rnewman: Consider making this a peekable iterator. If the next datom uses the same attribute, save some work. Even better if the next datom also uses the same entity…

db/src/tx.rs
```diff
@@ -859,3 +844,59 @@ pub fn transact_terms<'conn, 'a, I, W>(conn: &'conn rusqlite::Connection,
...
+        let attribute: &Attribute = schema.require_attribute_for_entid(a)?;
```
rnewman: At this moment you have the `Attribute`, so you can type-check `v`. (And if you do so at this point, you can probably not bother keeping the `Attribute` ref around — you only need it to decide multi/single and FTS/non.)

db/src/tx_checking.rs
```diff
@@ -0,0 +1,84 @@
...
+use internal_types::{
+    AEVTrie,
+};
+
+/// Map from found [e a v] to expected type.
+pub(crate) type TypeDisagreements = BTreeMap<(Entid, Entid, TypedValue), ValueType>;
+
+/// Ensure that the given terms type check.
+///
+/// We try to be maximally helpful by yielding every malformed datom, rather than only the first.
```
rnewman: I would rather see this done by receiving an error on insert of each datom and collecting the errors, continuing anyway. It won't catch _every_ error, but it'll catch most, and it's a streaming approach that fails early in the common case.

db/src/tx_checking.rs
```diff
@@ -0,0 +1,84 @@
...
+///
+/// We try to be maximally helpful by yielding every malformed set of datoms, rather than just the
+/// first set, or even the first conflict.  In the future, we might change this choice, or allow the
+/// consumer to specify the robustness of the cardinality checking desired.
+pub(crate) fn cardinality_conflicts<'schema>(aev_trie: &AEVTrie<'schema>) -> Vec<CardinalityConflict> {
+    let mut errors = vec![];
+
+    for (&(a, attribute), evs) in aev_trie {
+        for (&e, ref ars) in evs {
+            if !attribute.multival && ars.add.len() > 1 {
```
rnewman: I opted to instead have separate single/multi-val data structures when building the cache. Naturally that only works if you don't need to keep multiple values for cardinality-one with late error reporting!

db/src/upsert_resolution.rs
```diff
@@ -241,7 +271,44 @@ impl Generation {
...
+        // contiguous labels possible.
+        let mut tempid_map: BTreeMap<TempIdHandle, usize> = BTreeMap::default();
+
+        let mut dense_labels: indexmap::IndexSet<usize> = indexmap::IndexSet::default();
+
+        // We want to produce results that are as deterministic as possible, so we allocate labels
+        // for tempids in sorted order.  This has the effect of making "a" allocate before "b",
+        // which is pleasant for testing.
+        for (tempid, tempid_index) in temp_ids {
+            let rep = uf.find_mut(tempid_index);
```
rnewman: Does this handle transitive equivalence? That is, if `"foo"` is equivalent to `"bar"` because of `:foo/x 123`, and `"bar"` is equivalent to `"baz"` because of `:foo/y 456`, do `foo`, `bar, and `baz` all end up in the same bucket here?

db/src/errors.rs
```diff
@@ -56,6 +85,20 @@ impl ::std::fmt::Display for SchemaConstraintViolation {
                 }
                 Ok(())
             },
+            &TypeDisagreements { ref conflicting_datoms } => {
+                write!(f, "type disagreements:\n")?;
+                for (ref datom, expected_type) in conflicting_datoms {
+                    write!(f, "  expected value of type {} but got datom [{} {} {:?}]\n", expected_type, datom.0, datom.1, datom.2)?;
```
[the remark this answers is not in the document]
ncalexan: Fixed throughout.

db/src/errors.rs
```diff
@@ -43,6 +61,17 @@ pub enum SchemaConstraintViolation {
...
         conflicting_upserts: BTreeMap<TempId, BTreeSet<KnownEntid>>,
     },
+
+    /// A transaction tried to assert a datom or datoms with the wrong value `v` type(s).
+    TypeDisagreements {
+        /// The key (`[e a v]`) has an invalid value `v`: it is not of the expected value type.
+        conflicting_datoms: BTreeMap<(Entid, Entid, TypedValue), ValueType>
+    },
+
+    /// A transaction tried to assert a datoms that don't observe the schema's cardinality constraints.
```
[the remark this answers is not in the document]
ncalexan: Fixed.

db/src/tx.rs
ncalexan: This is a choice.  As I said (on Slack?) we really want to compute in two different monads: one that exits early and one that collects all errors, for the best consumer experience.  I think it's harder to go from one-error to collecting-errors than the reverse, so I implemented the global one first.

db/src/tx.rs
ncalexan: I think there will come a time when such micro-optimizations are worth the effort, but I don't think it is now.

db/src/upsert_resolution.rs
ncalexan: Yes, but let me make sure there's a test of exactly this.  You want:
```
[[:db/add "foo" :foo/x 123]
 [:db/add "bar" :foo/y 456]
 {:db/id "baz" :foo/x 123 :foo/y 456}]
```
to unify all of "foo", "bar", "baz"?  I guess we should test with those datoms present and not present.  I've added a test that covers this.

rnewman: In this PR you're making it so that

```
[:db/add "foo" :foo/x 123]
[:db/add "bar" :foo/x 123]
```

correctly unify; the next step is to make sure that


```
[:db/add "foo" :foo/x 123]
[:db/add "bar" :foo/x 123]
[:db/add "bar" :foo/y 456]
[:db/add "baz" :foo/y 456]
```

works. The subsequent insert itself will fail if it's not possible for all three entities to be the same, I think.

PR #682, opened 2018-05-08
Pull improvements

rnewman:
This PR allows aliasing of pull expressions and the use of `:db/id` for the entity ID itself:

```edn
(pull ?person [
    [:db/id :as :person/id]
    [:person/name :as :person/fullName]])
```

I wanted to flatten `NamespacedKeyword`, but soon realized that #648 still hasn't been wrapped up and landed, so I'll do that tomorrow. I plan to merge `NamespacedKeyword` and `Keyword`, with the only difference being a non-zero division offset and a different result from a predicate; we can preserve the restriction that all attribute names are namespaced while normalizing the API. That will make aliasing more useful.

### DIFF
--- a/query-parser/src/parse.rs
+++ b/query-parser/src/parse.rs
@@ -67,6 +67,7 @@ use self::mentat_query::{
     Order,
     OrJoin,
     OrWhereClause,
+    NamedPullAttribute,
     NotJoin,
     Pattern,
     PatternNonValuePlace,
@@ -191,6 +192,7 @@ def_parser!(Query, order, Order, {
 def_matches_plain_symbol!(Query, the, "the");
 def_matches_plain_symbol!(Query, pull, "pull");
 def_matches_plain_symbol!(Query, wildcard, "*");
+def_matches_keyword!(Query, alias_as, "as");
 
 pub struct Where<'a>(std::marker::PhantomData<&'a ()>);
 
@@ -303,11 +305,28 @@ def_parser!(Query, aggregate, Aggregate, {
          })
 });
 
+def_parser!(Query, pull_concrete_attribute_ident, PullConcreteAttribute, {
+    forward_keyword().map(|k| PullConcreteAttribute::Ident(::std::rc::Rc::new(k.clone())))
+});
+
+def_parser!(Query, pull_aliased_attribute, PullAttributeSpec, {
+    vector().of_exactly(
+        (Query::pull_concrete_attribute_ident()
+            .skip(Query::alias_as()),
+        forward_keyword().map(|alias| Some(::std::rc::Rc::new(alias.clone()))))
+            .map(|(attribute, alias)|
+                PullAttributeSpec::Attribute(
+                    NamedPullAttribute { attribute, alias })))
+});
+
 def_parser!(Query, pull_concrete_attribute, PullAttributeSpec, {
-    forward_keyword().map(|k|
-        PullAttributeSpec::Attribute(
-            PullConcreteAttribute::Ident(
-                ::std::rc::Rc::new(k.clone()))))
+    Query::pull_concrete_attribute_ident()
+        .map(|attribute|
+            PullAttributeSpec::Attribute(
+                NamedPullAttribute {
+                    attribute,
+                    alias: None,
+                }))
 });
 
 def_parser!(Query, pull_wildcard_attribute, PullAttributeSpec, {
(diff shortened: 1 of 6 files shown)


query-parser/src/parse.rs
```diff
@@ -303,11 +305,28 @@ def_parser!(Query, aggregate, Aggregate, {
          })
 });
 
+def_parser!(Query, pull_concrete_attribute_ident, PullConcreteAttribute, {
```
ncalexan: I would have expected to be expressed as an optional `:as foo`, but I guess this works as well.  The multiple try blocks should be a little less efficient than the optional block, I think.

query-pull/src/lib.rs
```diff
@@ -175,16 +176,27 @@ impl Puller {
...
-                        attrs.insert(entid.into());
+                &PullAttributeSpec::Attribute(NamedPullAttribute {
+                    ref attribute,
+                    ref alias,
+                }) => {
+                    let alias = alias.as_ref()
+                                     .map(|ref r| r.to_value_rc());
+                    match attribute {
+                        &PullConcreteAttribute::Ident(ref i) => {
+                            if let Some(entid) = schema.get_entid(i) {
```
ncalexan: Again, I think this should fail if we can't look up the `entid` for an ident, since it's probably an error to pull a non-attribute.

query/src/lib.rs
```diff
@@ -499,12 +499,26 @@ pub enum PullConcreteAttribute {
...
+}
+
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub enum PullAttributeSpec {
     Wildcard,
-    Attribute(PullConcreteAttribute),
+    Attribute(NamedPullAttribute),
     // PullMapSpec(Vec<…>),
-    // AttributeWithOpts(PullConcreteAttribute, …),
     // LimitedAttribute(PullConcreteAttribute, u64),  // Limit nil => Attribute instead.
```
ncalexan: `PullConcreteAttribute` has been superceded, so update these comments to `NamedPullAttribute`.

query-parser/src/parse.rs
```diff
@@ -67,6 +67,7 @@ use self::mentat_query::{
     Order,
     OrJoin,
     OrWhereClause,
+    NamedPullAttribute,
```
ncalexan: I'd vaguely prefer `Aliased` to `Named`, since that's what we're using throughout, but I'm not hung up on it.

query-pull/src/lib.rs
```diff
@@ -143,6 +145,7 @@ pub struct Puller {
     // The range is the set of aliases to use in the output.
     attributes: BTreeMap<Entid, ValueRc<NamespacedKeyword>>,
     attribute_spec: cache::AttributeSpec,
+    db_id_alias: Option<ValueRc<NamespacedKeyword>>,
```
ncalexan: Definitely wants a comment here, since this is a) not Datomic-alike and b) subtle.

query-pull/src/errors.rs
```diff
@@ -22,6 +22,11 @@ error_chain! {
             description("unnamed attribute")
             display("attribute {:?} has no name", id)
         }
+
```
ncalexan: The commit message is missing a "be": "to BE mentioned as a pull attribute."

query-pull/src/lib.rs
```diff
@@ -232,9 +232,7 @@ impl Puller {
         // - Recursing. (TODO: we'll need AttributeCaches to not overwrite in case of recursion! And
         //   ideally not do excess work when some entity/attribute pairs are known.)
         // - Building a structure by walking the pull expression with the caches.
-        //   TODO: aliases.
         // TODO: limits.
-        // TODO: fts.
```
ncalexan: I didn't see any obvious FTS extractions in the tests.  Are there some?

query-parser/src/parse.rs
rnewman: Yep — Datomic's grammar is wrong. I will fix.

query-parser/src/parse.rs
rnewman: I agreed with you, but it turns out that these are not always aliases; this is simply a pull attribute and its name, which sometimes differs from the attribute keyword itself.

query-pull/src/lib.rs
rnewman: There are now!

rnewman:
Rebased on top of #689: `rnewman/pull-keyword`.

I'm about to add non-namespaced aliases, but it'll have to wait for Monday!

Issue #684, opened 2018-05-09
[query] Query projector is not handling NULL for aggregates over 0 rows


ncalexan:
I have a query that looks like:
```edn
[:find
 (max ?at) .
 :in
 ?id
 :where
 [?c :credential/id ?id]
 [?l :login/credential ?c]
 [?l :login/at ?at]
]
```
That gets turned into the following SQL:
```
SELECT max(`?at`) AS `(max ?at)` FROM (SELECT DISTINCT `datoms02`.v AS `?at` FROM `datoms` AS `datoms00`, `datoms` AS `datoms01`, `datoms` AS `datoms02` WHERE `datoms00`.a = 65537 AND `datoms00`.v = $v0 AND `datoms01`.a = 65543 AND `datoms02`.a = 65542 AND `datoms02`.tx > 268435461 AND `datoms00`.e = `datoms01`.v AND `datoms01`.e = `datoms02`.e) LIMIT 1
```
The `ScalarProjector` has the following spec:
```
FindScalar(Aggregate(Aggregate { func: QueryFunction(PlainSymbol("max")), args: [Variable(var(?at))] }))
```
I get
```
Error(ProjectorError(DbError(BadSQLValuePair(Null, 4)))
```
and I surmise that the scalar projector either isn't handling zero rows correctly, or isn't handling a `NULL` result for aggregates correctly.  Presumably `NULL` means no result for scalars, but it's less clear to me what should happen for a `NULL` aggregate for the other query types.

ncalexan:
This one is a lot easier to actually run:
```
.q [:find [(the ?a) (max ?b)] :where [_ ?a _ ?c] [?c :db/txInstant ?b] [(< ?b #inst "2016-01-01T11:00:00.000Z")]]
```

ncalexan:
> Presumably NULL means no result for scalars, but it's less clear to me what should happen for a NULL aggregate for the other query types.

The issue is that SQL(ite) yields `NULL` for an aggregate over 0 rows.  Datomic either does or would yield `nil` in this situation, but Mentat doesn't expose the equivalent in `TypedValue`.  It's not clear to me what the set of `ValueType`s are that Mentat will aggregate over, but it might be that there is a sensible unit (i.e. 0 for longs, perhaps the epoch for instants, perhaps `TX0` for transaction refs) in each case that we can coalesce `NULL` to.

rnewman:
`the` is a special case, in that it itself is a kind of aggregate.

[...]

PR #688, opened 2018-05-10
[query] Handle SQL NULL for aggregates over 0 rows. (#684)

ncalexan:


rnewman:
This seems like a perfectly fine 'code' solution to the problem.

I believe another solution is available in SQL.

```edn
[:find (max ?x) (count ?x) :where [_ :db/txInstant ?x]]
```

turns to 

```sql
SELECT max(`?x`) AS `(max ?x)`, count(`?x`) AS `(count ?x)` 
FROM (
  SELECT DISTINCT `datoms00`.v AS `?x`
  FROM `datoms` AS `datoms00`
  WHERE `datoms00`.a = 3
)
```

and SQLite says:

> Aggregate `max()` returns `NULL` if and only if there are no non-`NULL` values in the group. 

so we should handle that case of no `?x`s being bound.

If we just tag on the end there…

```sql
WHERE `(max ?x)` IS NOT NULL
```

then we don't need to make any changes to the results handling code.

We'd do this in `re_project`, collecting the relevant (might-be-null) columns just as we collect `group_by_cols` in `CombinedProjection`.

What do you think, @ncalexan?

ncalexan:
> What do you think, @ncalexan?

Yeah, this is much better.  Wish I'd waited before doing the most obvious thing :/

ncalexan:
> WHERE `(max ?x)` IS NOT NULL

Sadly, this doesn't seem valid: both
```
WHERE `(max ?x)` IS NOT NULL
```
and
```
WHERE `max(?x)` IS NOT NULL
```
yield "misuse of aggregate: sum()" from SQLite.

rnewman:
In both of those the backticks enclose the identifier. Try

```sql
WHERE max(`?x`) IS NOT NULL
```

rnewman:
Oh, never mind; that's not valid. The original formulation I suggested should work:

```sql
sqlite> create table foo ( x integer );
sqlite> select max(x) from foo;
max(x)
----------

sqlite> select max(x) from foo where max(x) is not null;
Error: misuse of aggregate: max()
sqlite> select `max(?x)` from (select max(x) as `max(?x)` from foo) where `max(?x)` is not null;
sqlite>
```

We already use that inner query technique for `GROUP BY`.

rnewman:
We always group for aggregates, so use that sub select?

ncalexan:
> We always group for aggregates, so use that sub select?

This is not true, for reasons I'm still trying to understand.  This isn't quite as straight-forward as it looks, because we need to use `HAVING` instead of `WHERE`; in the course of experimenting, I discovered that we don't always `GROUP BY`, and that `HAVING` requires `GROUP BY`.  I'm trying to figure out what no `GROUP BY` means when we have two aggregates right now.

rnewman:
We group when there are non-aggregate bindings returned. The tests for aggregates in the top-level query tests should explain the differences between aggregate-only, aggregate with var, aggregate with `:with`, and aggregate with `the`.

My point was that we have the mechanism in place to generate that subquery in order to handle the need for grouping; we can extend that to grouping and/or null-dropping.

ncalexan:
> My point was that we have the mechanism in place to generate that subquery in order to handle the need for grouping; we can extend that to grouping and/or null-dropping.

In some way, yes -- but as I'm discovering, this is not nearly as simple as we initially thought.  Newest challenge is to figure out how to use `HAVING` where there's no `GROUP BY` and the aggregates are independent.  I think it's easier to just use _another_ `SELECT` in this case, like
```
SELECT * FROM (...) WHERE `(max ?a)` IS NOT NULL
```

ncalexan:
@rnewman this isn't elegant but it's simple.  What do you think?

query-translator/src/translate.rs
```diff
@@ -427,14 +427,47 @@ fn re_project(mut inner: SelectQuery, projection: Projection) -> SelectQuery {
...
-        distinct: outer_distinct,
-        projection: projection,
-        from: FromClause::TableList(TableList(vec![TableOrSubquery::Subquery(Box::new(inner))])),
-        constraints: vec![],
-        group_by: group_by,
-        order: order_by,
-        limit: limit,
+    use self::Projection::*;
+
+    let having = match &projection {
```
rnewman: Change the name away from `having`.

query-translator/tests/translate.rs
```diff
@@ -1091,28 +1091,35 @@ fn test_project_aggregates() {
...
-                      WHERE `datoms00`.a = 99)");
+                     (SELECT max(`?t`) AS `(max ?t)` \
+                      FROM \
+                      (SELECT DISTINCT \
+                       `datoms00`.v AS `?t`, \
+                       `datoms00`.e AS `?e` \
+                       FROM `datoms` AS `datoms00` \
+                       WHERE `datoms00`.a = 99)\
+                      ) \
+                     WHERE `(max ?t)` IS NOT NULL");
```
rnewman: I’d like two more tests:

- that we don’t null check `count`. That test will fail, then you must decide what to change!
- that we don’t null-check `the` or grouped vars. Same.

ncalexan: > that we don’t null check count. That test will fail, then you must decide what to change!

I thought about this, and the null check is harmless in general.  A tiny bit of SQLite overhead.  

Doing better is irritating: we're fixing this up _way_ after the point we should be, due to us handling aggregates very early and then dropping all the information.  I thought of renaming `::Expression` to be more particular to aggregates, and then splitting that case based on the type of aggregate... but I really don't think this is worth it.

`the` and grouped vars would change other tests, and they don't produce `::Expression` enums, so this won't fail.  I'm happy to add a test specifically here to show that it doesn't happen.

I'll give the split a stab and if I like it I'll land it; if not, this'll get unblock me.

query-translator/src/translate.rs
```diff
@@ -427,14 +427,47 @@ fn re_project(mut inner: SelectQuery, projection: Projection) -> SelectQuery {
...
+    if nullable.is_empty() {
+        outer
+    } else {
+        SelectQuery {
+            distinct: false,
+            projection: Projection::Star,
+            from: FromClause::TableList(TableList(vec![TableOrSubquery::Subquery(Box::new(outer))])),
+            constraints: nullable,
+            group_by: vec![],
+            order: vec![],
```
rnewman: Gosh. I don’t actually know if the inner ordering is respected when returned in this way; there are databases in which it is not. Worth a little research and a test…

query-translator/src/translate.rs
```diff
@@ -455,19 +456,43 @@ fn re_project(mut inner: SelectQuery, projection: Projection) -> SelectQuery {
...
-        SelectQuery {
-            distinct: false,
-            projection: Projection::Star,
-            from: FromClause::TableList(TableList(vec![TableOrSubquery::Subquery(Box::new(outer))])),
-            constraints: nullable,
-            group_by: vec![],
-            order: vec![],
-            limit: Limit::None,
-        }
+    // `ORDER BY` is relevant when we have a `LIMIT`.  We therefore lift the `ORDER BY` if there’s
```
rnewman: Worth mentioning that we lift it because SQL doesn't guarantee ordering unless it's at the outermost level.

query-translator/src/translate.rs
```diff
@@ -455,19 +456,43 @@ fn re_project(mut inner: SelectQuery, projection: Projection) -> SelectQuery {
...
+            }
+        },
+        _ => {
+            SelectQuery {
+                distinct: outer_distinct,
+                projection: projection,
+                from: FromClause::TableList(TableList(vec![TableOrSubquery::Subquery(Box::new(inner))])),
+                constraints: vec![],
+                group_by: group_by,
+                order: order_by.clone(),
```
rnewman: Suggest phrasing without the repetition, as

```rust
SelectQuery {
    distinct: outer_distinct,
    projection,
    from: FromClause::…,
    constraints: Vec::new(),
    group_by,
    order: match &limit {
        &Limit::None => Vec::new(),
        &Limit::Fixed(_) | &Limit::Variable(_) => order_by.clone(),
    },
    limit,
}
```

query-translator/src/translate.rs
```diff
@@ -455,19 +456,43 @@ fn re_project(mut inner: SelectQuery, projection: Projection) -> SelectQuery {
...
+                group_by: group_by,
+                order: vec![],
+                limit: Limit::None,
+            }
+        },
+        _ => {
+            SelectQuery {
+                distinct: outer_distinct,
+                projection: projection,
+                from: FromClause::TableList(TableList(vec![TableOrSubquery::Subquery(Box::new(inner))])),
```
rnewman: Can we avoid the reuse of the name `inner`?

query-translator/src/translate.rs
```diff
@@ -455,19 +456,43 @@ fn re_project(mut inner: SelectQuery, projection: Projection) -> SelectQuery {
...
+    };
+
+    SelectQuery {
+        distinct: false,
+        projection: Projection::Star,
+        from: FromClause::TableList(TableList(vec![TableOrSubquery::Subquery(Box::new(inner))])),
+        constraints: nullable,
+        group_by: vec![],
+        order: order_by,
+        limit: Limit::None, // Any limit comes from the internal query.
```
rnewman: "Any limiting…" would be less confusing, I think.

PR #689, opened 2018-05-11
Combine NamespacedKeyword and Keyword

rnewman:
This incorporates @thomcc's changes, my review comments, some other improvements from me, and then:

- Mechanically rename constructors for `PlainSymbol`, `Keyword`, and `NamespacedKeyword` to not collide.
- Merge the latter two, which is mostly mechanical with some manual adding of `is_namespaced()` checks.

Tests all pass.

### DIFF
--- a/edn/src/namespaceable_name.rs
+++ b/edn/src/namespaceable_name.rs
@@ -83,8 +83,36 @@ impl NamespaceableName {
         }
     }
 
+    fn dwim<N, T>(namespace: Option<N>, name: T) -> Self where N: AsRef<str>, T: AsRef<str> {
+        if let Some(ns) = namespace {
+            Self::namespaced(ns, name)
+        } else {
+            Self::plain(name.as_ref())
+        }
+    }
+
     pub fn is_namespaced(&self) -> bool {
         self.boundary > 0
+    }
+
+    #[inline]
+    pub fn is_backward(&self) -> bool {
+        self.name().starts_with('_')
+    }
+
+    #[inline]
+    pub fn is_forward(&self) -> bool {
+        !self.is_backward()
+    }
+
+    pub fn to_reversed(&self) -> NamespaceableName {
+        let name = self.name();
+
+        if name.starts_with('_') {
+            Self::dwim(self.namespace(), &name[1..])
+        } else {
+            Self::dwim(self.namespace(), &format!("_{}", name))
+        }
     }
 
     #[inline]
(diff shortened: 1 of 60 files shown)


edn/src/symbols.rs
```diff
@@ -25,8 +25,8 @@ pub struct PlainSymbol(pub String);
 pub struct NamespacedSymbol {
     // We derive PartialOrd, which implements a lexicographic based
     // on the order of members, so put namespace first.
-    pub namespace: String,
-    pub name: String,
+    _namespace: String,
```
ncalexan: I think this is going to be address in later commits, but it's not common in Rust code to prepend `_` for privacy, since Rust has a compiler with rich visibility support.

edn/src/namespaced_name.rs
```diff
@@ -0,0 +1,220 @@
...
+use std::fmt;
+
+#[cfg(feature = "serde_support")]
+use serde::{
+    de::{self, Deserialize, Deserializer},
+    ser::{Serialize, Serializer}
+};
+
+// Data storage for both NamespacedKeyword and NamespacedSymbol. This gets it's own module because
+// modifying it's fields is `unsafe`, and so we'd like to minimize the amount of code that can
```
ncalexan: s/it's/its/

edn/src/namespaced_name.rs
```diff
@@ -0,0 +1,220 @@
...
+// touch it.
+#[derive(Clone, Eq, Hash, PartialEq)]
+pub struct NamespacedName {
+    // The bytes that make up the namespace followed directly by those
```
ncalexan: I would really like this to include any slash, so that we internally store `:foo/bar` or `foo/bar` or whatever.  That gives us a hope of implementing `AsRef<str>`.  If we don't store the leading colon and/or internal slash, that's out.

edn/src/namespaceable_name.rs
```diff
@@ -83,10 +83,38 @@ impl NamespaceableName {
         }
     }
 
+    fn dwim<N, T>(namespace: Option<N>, name: T) -> Self where N: AsRef<str>, T: AsRef<str> {
```
ncalexan: This is a pretty poor name; can we do better?  Maybe... `new`?  Also, should this be `Into<Option<AsRef<str>>>`?

edn/src/namespaced_name.rs
rnewman: Filed #692.

PR #709, opened 2018-05-23
Update README.md

cdbfoster:
Grammar -- the subjunctive is appropriate here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ We've observed that data storage is a particular area of difficulty for software
 
 ## Comparison to DataScript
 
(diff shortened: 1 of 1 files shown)


cdbfoster:
Oh dear, I just realized that this is a direct quote from the DataScript readme.  So perhaps you don't want this change.  The sentence is still wrong, though.

rnewman:
Perhaps open a PR on the DataScript repo?

ncalexan:
> Oh dear, I just realized that this is a direct quote from the DataScript readme. So perhaps you don't want this change. The sentence is still wrong, though.

Yeah, we'll take this.  Thanks!

PR #717, opened 2018-05-29
Min SDK 16, bump dependency versions, update gradle & wrapper, fix linter error

### LINKED ISSUE #716: Android SDK should be minSdkVersion:21 (or earlier)
Currently it's at 26. API 21 is what our [android-components](https://github.com/mozilla-mobile/android-components) project is targeting, so mentat should be at the very least compatible with that.

grigoryk:
Fixes #716 and updates/cleans up things a little.

API 16 is chosen somewhat arbitrarily (this is what fennec currently uses).

### DIFF
--- a/sdks/android/Mentat/library/src/main/java/com/mozilla/mentat/InProgressTransactionResult.java
+++ b/sdks/android/Mentat/library/src/main/java/com/mozilla/mentat/InProgressTransactionResult.java
@@ -41,7 +41,7 @@ public class InProgressTransactionResult extends Structure implements Closeable 
     
     public TxReport getReport() {
         if (this.result.isFailure()) {
-            Log.e("InProgressTransactionResult", this.result.err);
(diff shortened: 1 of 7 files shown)


sdks/android/Mentat/library/src/main/java/com/mozilla/mentat/InProgressTransactionResult.java
```diff
@@ -41,7 +41,7 @@ public InProgress getInProgress() {
     
     public TxReport getReport() {
         if (this.result.isFailure()) {
-            Log.e("InProgressTransactionResult", this.result.err);
+            Log.e("InProgressTransactRes", this.result.err);
```
fluffyemily: What was the reason for this log change?

rnewman: Android log tags are limited to 23 characters in length. The original tag was 28.

Issue #708, opened 2018-05-16
[edn] Support serde serialization and deserialization for `edn::Value`


ncalexan:
It's quite convenient to be able to turn arbitrary things into EDN for debugging, and we have many places in the tree where we do just that.  For example, this is used extensively in the `mentat_db` crate to `assert_matches!` EDN representations of datoms and transactions.

We could do this a little more generally if we made it easy to serialize into `edn::Value`.

victorporof:
Do we intend to do support anything other than debugging? Adding serde support is extremely simple for the default case, but as soon as we want to do more interesting things such as customising the actual syntax for textual formats, custom serialisation logic needs to be written.

An example where this matters is supporting json output: my experience is that serde's defaults are (although still human readable) extremely verbose and not something one would naturally write.

ncalexan:
> Do we intend to do support anything other than debugging? Adding serde support is extremely simple for the default case, but as soon as we want to do more interesting things such as customising the actual syntax for textual formats, custom serialisation logic needs to be written.

I don't have strong opinions.  I do know that I like dumping structures (datoms, transactions, query results) as EDN and pretty-printing them (using the pretty-printer you built!).  I don't really care about serializing arbitrary Rust structures to and from EDN, since I think JSON is much more likely to be used in the wild.

> An example where this matters is supporting json output: my experience is that serde's defaults are (although still human readable) extremely verbose and not something one would naturally write.

Is this in the encoding?  EDN is richer than JSON (e.g., lists, vectors, and sets are all syntactically differentiated) and keywords are more natural than some of the "tagging" encodings I saw the `serde-json` writing out.

I'd be pretty interested in seeing if we could make simple structures that `#[derive(Serialize)]` serialize to reasonable EDN cheaply.

PR #726, opened 2018-05-31
Edn serde support

victorporof:
Fixes https://github.com/mozilla/mentat/issues/708

Simple naive implementation of serde serialization/deserialization for edn values. This is a naive approach, relying on serde's defaults for the internal shape.

We can take this much further, but I'm bit wary of allocating more time on this without an active consumer. Points of improvement are around making the json representation of edn more ergonomic, and adding some helpers to stringify/parse edn into/out of json without having to go through serde_json.

I had to update the num crate to use its serde support. Splitting it into num-traits and num-bigint ended up the way I did it, but there's probably a way to rely on the num mothership crate if necessary.

I also enabled serde-supported tests on travis, they weren't before.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
 language: rust
 script:
   - cargo test --verbose --all
+  - cargo test --features edn/serde_support --verbose --all
--- a/edn/Cargo.toml
+++ b/edn/Cargo.toml
@@ -13,7 +13,8 @@ readme = "./README.md"
 [dependencies]
 chrono = "0.4"
 itertools = "0.7"
-num = "0.1"
+num-bigint = "0.2"
+num-traits = "0.2"
 ordered-float = "0.5"
 pretty = "0.2"
 uuid = { version = "0.5", features = ["v4", "serde"] }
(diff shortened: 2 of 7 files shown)


victorporof:
@ncalexan IIRC I had to update num to get access to its serde support for bigint. I think I can update num everywhere to get this if desired. Splitting the num dependency into the 2 smaller crates isn't required, just how I ended up doing it.

If updating num everywhere is acceptable, I believe it's trivial to revive this PR. Otherwise we can shelve it.

ncalexan:
> @ncalexan IIRC I had to update num to get access to its serde support for bigint. I think I can update num everywhere to get this if desired. Splitting the num dependency into the 2 smaller crates isn't required, just how I ended up doing it.

OK -- thanks, that's helpful.

> If updating num everywhere is acceptable, I believe it's trivial to revive this PR. Otherwise we can shelve it.

Let's shelve this for now.  As you say, there's no pressing consumer; and it's a little awkward until `chrono` rolls forward.  Thanks for teaching me some things, though!

PR #728, opened 2018-05-31
Parse queries using rust-peg.

ncalexan:
This is the equivalent of #681 but for the query parser.  Faster and simpler, although a lot of things needed to get moved around.

I'll follow-up to remove the `tx` and `query-parser` crates, since they're just pass-throughs at this point.


core/src/lib.rs
```diff
@@ -38,10 +38,17 @@ pub use chrono::{
...
 pub use edn::{
+    Cloned,
     FromMicros,
+    FromRc,
     Keyword,
     ToMicros,
     Utc,
+    ValueRc,
+};
+pub use edn::parse::{
```
rnewman: Nit: newline before.

core/src/lib.rs
```diff
@@ -38,10 +38,17 @@ pub use chrono::{
...
+    query as parse_query,
```
rnewman: If we're eliminating the query parser, I think I'd rather see the consumers of query parsing import `edn::parse::query` directly. It's the most clear expression of what's happening. Otherwise `core` is just going to be a bottleneck crate.

edn/src/edn.rustpeg
```diff
@@ -252,6 +282,210 @@ pub entity -> Entity<ValueAndSpan>
...
+pattern_non_value_place -> query::PatternNonValuePlace
+    = v:value {? query::PatternNonValuePlace::from_value(&v).ok_or("expected pattern_non_value_place") }
+
+pattern -> query::WhereClause
+    = __ "["
+      src:src_var?
+      e:pattern_non_value_place
+      a:pattern_non_value_place
+      v:pattern_value_place?
+      tx:pattern_non_value_place?
```
rnewman: I haven't looked closely, but I wonder if this sequence of optionals will always work correctly, and never leave `v` unbound while binding `tx`…

edn/src/edn.rustpeg
```diff
@@ -252,6 +282,210 @@ pub entity -> Entity<ValueAndSpan>
...
 
 pub entities -> Vec<Entity<ValueAndSpan>>
     = __ "[" __ es:(entity*) __ "]" __ { es }
+
+// Query parser starts here.
+//
+// We expect every rule except the `raw_*` rules to eat whitespace
+// (with `__`) at its start and finish.  That means that every string
+// pattern (say "[") should be bracketed on either side with either a
+// whitespace-eating rule or an explicit whitespace eating `__`.
```
rnewman: Do we also correctly 'eat' commas and comments?

edn/src/query.rs
```diff
@@ -0,0 +1,1253 @@
...
+}
+
+impl std::fmt::Display for QueryFunction {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum Direction {
```
rnewman: We are definitely reaching the point where the `edn` crate should be renamed `parsers` or something.

edn/src/query.rs
```diff
@@ -0,0 +1,1253 @@
+// Copyright 2016 Mozilla
```
rnewman: I would have expected this file to track via `git mv`… maybe replay that work to make sure we don't lose blame?

edn/src/query.rs
rnewman: And doing so might allow us to have domain types like these as a dependency of `parsers`, rather than being _part of_ it.

edn/src/edn.rustpeg
```diff
@@ -252,6 +282,210 @@ pub entity -> Entity<ValueAndSpan>
...
+
+query_function -> query::QueryFunction
```
grigoryk: Do we have a documented grammar definition somewhere for reference? It would be much easier to skim through for reference than this syntax (which I realize isn't intended for direct human consumption).

edn/src/query.rs
```diff
@@ -0,0 +1,1253 @@
...
+pub enum NonIntegerConstant {
+    Boolean(bool),
+    BigInteger(BigInt),
+    Float(OrderedFloat<f64>),
+    Text(ValueRc<String>),
+    Instant(DateTime<Utc>),
+    Uuid(Uuid),
+}
+
+// impl NonIntegerConstant {
```
grigoryk: Clean this up, maybe?

edn/src/query.rs
```diff
@@ -0,0 +1,1253 @@
...
+            alias: None,
+        }
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum PullAttributeSpec {
+    Wildcard,
+    Attribute(NamedPullAttribute),
+    // PullMapSpec(Vec<…>),
```
grigoryk: Cleanup

edn/src/edn.rustpeg
```diff
@@ -252,6 +282,210 @@ pub entity -> Entity<ValueAndSpan>
...
+    = pattern
+    / or_join_clause
+    / or_clause
+    / not_join_clause
+    / not_clause
+    / type_annotation
+    / pred
+    / where_fn
+
+query_part -> query::QueryPart
```
grigoryk: Although, this is pretty straightforward.

edn/src/edn.rustpeg
```diff
@@ -252,6 +282,210 @@ pub entity -> Entity<ValueAndSpan>
...
+// whitespace-eating rule or an explicit whitespace eating `__`.
+
+query_function -> query::QueryFunction
+    = __ n:$(symbol_name) __ {? query::QueryFunction::from_symbol(&PlainSymbol::plain(n)).ok_or("expected query function") }
+
+fn_arg -> query::FnArg
+    = v:value {? query::FnArg::from_value(&v).ok_or("expected query function argument") }
+    / __ "[" args:fn_arg+ "]" __ { query::FnArg::Vector(args) }
+
+find_elem -> query::Element
```
grigoryk: This grammar definition does get a little cryptic though, for a casual reader.

edn/src/query.rs
```diff
@@ -0,0 +1,1253 @@
...
+pub struct ParsedFindQuery {
+    pub find_spec: FindSpec,
+    pub default_source: SrcVar,
+    pub with: Vec<Variable>,
+    pub in_vars: Vec<Variable>,
+    pub in_sources: BTreeSet<SrcVar>,
+    pub limit: Limit,
+    pub where_clauses: Vec<WhereClause>,
+    pub order: Option<Vec<Order>>,
+    // TODO: in_rules;
```
grigoryk: Issue number?

query-algebrizer/src/types.rs
```diff
@@ -691,6 +693,22 @@ impl Debug for EmptyBecause {
     }
 }
 
+
+/// A `FindQuery` represents a valid query to the query algebrizer.
+#[allow(dead_code)]
+#[derive(Debug, Eq, PartialEq)]
+pub struct FindQuery {
```
grigoryk: It'd be nice to have a duplicate comment here, similar to one above ParsedFindQuery

query-algebrizer/src/types.rs
```diff
@@ -691,6 +693,22 @@ impl Debug for EmptyBecause {
     }
 }
 
+
+/// A `FindQuery` represents a valid query to the query algebrizer.
```
grigoryk: I personally would like "ValidFindQuery" a bit more - or perhaps the more specific "AlgebrizableFindQuery". To me the current naming is very vague in context of the split from ParsedFindQuery.

grigoryk: Especially given that this is conceptually a "sub type" of the more generic ParsedFindQuery, right? In a sense that it encompasses a smaller number of possibilities.

core/src/lib.rs
ncalexan: Fixed.

core/src/lib.rs
ncalexan: This was done specifically for `query-algebrizer`, which doesn't import `edn` directly.  I'd prefer to keep it.

edn/src/edn.rustpeg
ncalexan: I do not think it is possible, because everything that can come in a non-value place can also come in a value-place.  I'm not sure it's worth trying to be more careful here.

edn/src/edn.rustpeg
ncalexan: We do, but I'll add a few tests such that it's clear.

edn/src/query.rs
ncalexan: Yes, I concur.  Follow-up, either as I land (now) or as real follow-up.

edn/src/query.rs
ncalexan: I did this with a `Part 3b` that is just the file move.  I thought `git` would determine this automatically, but apparently it didn't.

edn/src/edn.rustpeg
ncalexan: We run off of https://docs.datomic.com/on-prem/query.html#grammar -- except where it's subtly wrong.  And we do have some extensions.

edn/src/query.rs
ncalexan: These are specifically placeholders for the future, so I'm going to leave it.

edn/src/query.rs
ncalexan: Historical, pretty sure it's untracked.  I think I'll remove it, since I very much doubt we'll ever implement Datalog rules.

query-algebrizer/src/types.rs
ncalexan: Sure.

query-algebrizer/src/types.rs
ncalexan: I don't want to change the name `FindQuery`, so I'll change `ParsedFindQuery` to `ParsedQuery`.

edn/src/query.rs
ncalexan: This is just a little more effort than I want to roll into this right now -- #734.

PR #730, opened 2018-06-01
Use failure instead of error_chain

### LINKED ISSUE #586: Use failure instead of error_chain

grigoryk:
This is an early WIP to get feedback.

So far the only non-mechanical bit was conversion of `InvalidBinding`, since it wraps a `BindingError`.

Eventually fixes #586.


query-algebrizer/src/clauses/convert.rs
```diff
@@ -27,10 +29,7 @@ use clauses::{
     ConjoiningClauses,
 };
 
-use errors::{
-    ErrorKind,
-    Result,
-};
+use errors::AlgebrizerError;
```
ncalexan: I'm not against declaring `Result<T> = std::result::Result<T, errors::AlgebrizerError>` to reduce the changes.  Or even implementing `bail!` in each crate -- it's easy.

The `Err(...)?` pattern is odd to me -- I prefer the `return Err(...)` expression.

rnewman: Using `?` adds an implicit `into` to do conversion of errors, which is a little more compact than `return Err(some_err.into())` or `return some_result.map_err(|e| e.into())`.

grigoryk:
This is bulk of the work. Needs another cleanup pass, and rebasing on top of `master` - query-parser is gone, algebrizer changed a bit, and #736 is about to land.

grigoryk:
Tests pass! I'm not super happy with how parts of this turned out, but it's not a good idea to let this fall behind the other work. We can iterate on the particulars once/if we hit problems along the line.

grigoryk:
@ncalexan do you mind giving this a lookover before I have to rebase it again? :sob:

PR #731, opened 2018-06-01
Closes #660 Store management

### LINKED ISSUE #660: Management of open stores
It's important that a particular store is only opened once at a time (see also #547).

Further, it's convenient for FFI to have a static place to root open stores.

The obvious solution is a manager, [just like rkv's](https://github.com/mozilla-prototypes/rkv/blob/master/src/manager.rs#L36).

But what do we manage?

I propose:

- Make `Store`'s `conn` field `Arc`. This gives us the ability to share `Conn`s and thus use `Store` as a pair of a connection and a `Conn` when there are multiple open SQLite connections.
- Add a method to `Store`, perhaps named `fork`, that yields a new store with cloned `conn` using a provided `rusqlite::Connection`. This allows us to get a new copy of a store that's already open.
- Add a static method to `Store` to create an in-memory store. The empty path will now be confusing. (I believe the FFI already makes a distinction…)
- Implement a manager of `Store`s, just as in rkv. It's important to canonicalize file paths. "Stores" might be a good name.
- Expose these methods:
  - Is a store open? `fn is_open(path: PathBuf) -> bool`
  - Given a path, give me a mutable reference to a suitable `Store`, opening it if necessary. `fn open(path: PathBuf) -> Result<&mut Store>`
  - Given a path, give me a (mutable) reference if the store is open. `fn {get,get_mut}(path: PathBuf) -> Option<{&mut,&} Store>`
  - Given a path, give me a `Store` instance with its own SQLite connection. I understand that I now have to work with SQLite's concurrency model (_e.g._, I can encounter `SQLITE_BUSY`), but Mentat will serialize writes to `Metadata`. `fn connect(path: PathBuf) -> Result<Store>`
  - Close a store by path. This needs to check the reference count on `Store.conn` and fail if it's more than 1. It can only be more than 1 if a caller has cloned the store, either explicitly or via `connect`.

The usual pattern for working with a store then changes from:

```rust
        let mut store: Store = Store::open("/some/path").expect("opened");
        let mut in_progress = store.begin_transaction().expect("began");
        in_progress.import(fixture_path("cities.schema")).expect("transacted schema");
```

to

```rust
        let store: &mut Store = Stores::open("/some/path").expect("opened");
        let mut in_progress = store.begin_transaction().expect("began");
        in_progress.import(fixture_path("cities.schema")).expect("transacted schema");
```

— a one-character change and a change in type.

fluffyemily:
Add Store manager to ensure that we only have one open handle to each database.

This implementation has gone through several iterations. The considerations that formed the current design are as follows:

1. `rusqlite::Connection` is `Send` but not `Sync` and therefore cannot be stored in a static var. This is why we only store the `Conn` inside `Stores`.
2. `Conn` is stored as a `Weak` reference because otherwise every time we return a `Store` we increase the reference count from one to two, forcing every mutable operation performed on the `Conn` to clone rather than return a mutable reference, dramatically increasing the number of Conn references proliferating the system. By storing `Conn` as a `Weak`, for as long as there is only one Store we can perform mutable operations without increasing the reference count on `Conn`.
3. By storing `Conn` as a `Weak` reference, if we cease to hold any strong references to the `Conn`, by not having any active `Store`'s, the `Weak` will no longer be upgradable, but the entry will remain in the map of open `Conns`. We therefore must recreate the `Conn` if a request for a previously opened store is received. We do this in the background without consumers being aware.
4. The original implementation had a `thread_local` store of open `Rc<rusqlite::Connection>`'s mapped by the same key as `Conn`'s. However, due to `rusqlite::Connection` not being `Clone`, this meant that `Rc::make_mut` could not be used, which resulted in the code being unable to get a mutable reference to the `rusqlite::Connection` if there was more than one reference. As the map retained a strong reference, that ensured that every returned Store had a reference count of at least 2 on it's `rusqlite::Connection` which resulted in all mutable operations on the `Store` failing. I therefore made the decision to create a new `rusqlite::Connection` for every Store returned. This contravenes the [original spec](#660), but given the constraints something had to give and I decided on this one.
5. Writing tests with in memory stores mapped by path ("") with a static `Store` lead to a problem where the same `Conn` was shared between tests and the assertions around reference counts on stored `Conn`'s were unreliable. In order to get around this I introduced the concept of the `named_in_memory_store` which assigns a name to an in memory store ensuring that the same `Conn` was not shared between tests. This is the requirement that lead to the decision to store keys as `String`'s rather than `PathBuf`.

### DIFF
--- a/src/conn.rs
+++ b/src/conn.rs
@@ -88,10 +88,6 @@ use edn::entities::{
     OpType,
 };
 
-use mentat_tolstoy::Syncer;
-
-use uuid::Uuid;
-
 use entity_builder::{
     InProgressBuilder,
     TermBuilder,
@@ -162,48 +158,7 @@ pub struct Conn {
     // TODO: maintain cache of query plans that could be shared across threads and invalidated when
     // the schema changes. #315.
     tx_observer_service: Mutex<TxObservationService>,
-}
-
-/// A convenience wrapper around a single SQLite connection and a Conn. This is suitable
-/// for applications that don't require complex connection management.
-pub struct Store {
-    conn: Conn,
-    sqlite: rusqlite::Connection,
-}
-
-impl Store {
-    /// Open a store at the supplied path, ensuring that it includes the bootstrap schema.
-    pub fn open(path: &str) -> Result<Store> {
-        let mut connection = ::new_connection(path)?;
-        let conn = Conn::connect(&mut connection)?;
-        Ok(Store {
-            conn: conn,
-            sqlite: connection,
-        })
-    }
-
-    /// Returns a totally blank store with no bootstrap schema. Use `open` instead.
-    pub fn open_empty(path: &str) -> Result<Store> {
-        if !path.is_empty() {
-            if Path::new(path).exists() {
-                bail!(ErrorKind::PathAlreadyExists(path.to_string()));
-            }
-        }
-
-        let mut connection = ::new_connection(path)?;
-        let conn = Conn::empty(&mut connection)?;
-        Ok(Store {
-            conn: conn,
-            sqlite: connection,
-        })
-    }
-
-    pub fn transact(&mut self, transaction: &str) -> Result<TxReport> {
-        let mut ip = self.begin_transaction()?;
-        let report = ip.transact(transaction)?;
-        ip.commit()?;
-        Ok(report)
-    }
+    pub(crate) tx_observer_service: Arc<Mutex<TxObservationService>>,
 }
 
 pub trait Queryable {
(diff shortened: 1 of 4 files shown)


fluffyemily:
@ncalexan This is the version without connection pooling.

sdks/android/Mentat/gradle.properties
```diff
@@ -10,6 +10,7 @@
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
 org.gradle.jvmargs=-Xmx1536m
+org.gradle.configureondemand=false
```
ncalexan: Is this really necessary?  This sucks, because always configuring all projects is very expensive as sub-projects multiply.

sdks/android/Mentat/library/src/main/java/org/mozilla/mentat/Mentat.java
```diff
@@ -36,7 +36,7 @@ public Mentat(String dbPath) {
     }
 
     /**
-     * Open a connection to an in-memory Store.
+     * Open a connection to an anonymous in-memory Store.
```
ncalexan: This is very misleading, and is really the _problem_ here -- your store isn't anonymous, it's the "empty string", hence it always open a _new_ store.  Why not make this function actually anonymous by randomizing the string and drop all the hand-coded names?

PR #736, opened 2018-06-06
Fast-forward syncing, rebased.

ncalexan:
Here's a sliced-and-diced version of #563, with the simple stuff pulled forward.

I'll dig into the larger restructure for atomic metadata updates now, but perhaps you could stamp the easy bits @grigoryk?


grigoryk:
Work continues in #563 .

Issue #738, opened 2018-06-13
Run SQLCipher tests in Travis


thomcc:
This is a follow up to https://github.com/mozilla/mentat/pull/737, but it ended up being a bit too furry of a yak for me to shave right now, and probably shouldn't block that code landing.

I do think it's worth doing, to avoid a case where nobody has tested support for SQLCipher a while. This is especially because of `cargo test --all` not working with it (specifying individual features to use for sub-crates doesn't work the way we would need – it only applies the feature to the root crate – and will be completely disallowed soon because of this).

This is the .travis.yml I had before deciding it's not worth figuring this out fully before moving on: https://gist.github.com/thomcc/b48c38560d950a253b86ed45fa31a3c9.

This builds, but fails when run due to failing to find the unicode61 tokenizer we request here: https://github.com/mozilla/mentat/blob/3d5ae797b2f79b27c9d6afce77160ede5b9cdff7/db/src/db.rs#L160. AFAICT travis uses Ubuntu 14.04, which uses the following version of https://packages.ubuntu.com/trusty/libsqlcipher-dev... which is from 2013. And so it corresponds to a Sqlite version that's too old for the features we need.

I guess the recommended way to use a newer Ubuntu on Travis is to use Docker, which... Sounds like a huge pain. Maybe we could just point apt at a newer package repository instead?

It's also possible we could use `os: osx`, and then install sqlcipher somehow (`brew install sqlcipher --with-fts` works great for me, but I don't know if you can easily use Homebrew in travis on osx). I've heard a lot of people complaining about Travis's osx support being spotty though, so this seems less ideal.

PR #741, opened 2018-06-13
Separate `Store` from `Conn`.

ncalexan:
This is a Pre: part extracted from #660.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,13 @@
 language: rust
+rust:
+  - 1.25.0  # Must align with `build/version.rs`.
+  - stable
+  - beta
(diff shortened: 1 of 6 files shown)


thomcc:
This accidentally broke sqlcipher support, since it didn't move the impl Store that was behind `#[cfg(feature = "sqlcipher")]` (will fix)

ncalexan:
> This accidentally broke sqlcipher support, since it didn't move the impl Store that was behind #[cfg(feature = "sqlcipher")] (will fix)

Sorry -- I should have realized that #738 meant my testing locally and relying on Travis wasn't sufficient.  Please correct me.

PR #742, opened 2018-06-14
Avoid using 1.26.0-only features when using sqlcipher, and move the sqlcipher Store support to the correct file

thomcc:
This fixes the issue with the sqlcipher build that was introduced by #741, and makes sure that the work done in f041dfe applies to the sqlcipher feature as well.


thomcc:
I have a fix for #738 that is running on travis (it passed before I just updated). If it continues to pass, I'll PR it. That should avoid issues like this in the future.

thomcc:
(Sorry about the commit name here, that's what I get for using a mobile app to merge the PR)

PR #745, opened 2018-06-15
Fix known leaks in both swift and android SDKs

thomcc:
This was easier for the Swift one than the Android one, although I'm more confident that the android one is completely good now -- In the swift, I'm not 100% certain that TypedValues are being cleaned up, since we should have seen some of the same crashes that I did in android when things were freeing values returned by value_at_index.  

Anyway, the Android one also has several changes that were needed in order to make it less likely to cause issues in the future. I'd like to go further here and make the rawPointer on RustObject private, but I stopped before doing so.


thomcc:
In general, I have a couple fears about the FFI code that I'm going to take this comment to express:

Manual memory management in languages that don't help you out is fairly error prone, and getting it wrong can lead to double frees or use-after-frees, both of which lead to security vulnerabilities (and are the types of things that Rust was designed to avoid).

The changes I've made don't help here<sup id="a0">[0](#f0)</sup> -- leaking memory isn't unsafe (look at `std::mem::forget`) but once you stop leaking the memory you open yourself up to all sorts of nastiness. Unfortunately, I don't think it's realistic to think we could ship while leaking memory very frequently.

My biggest fear is that one of our consumers (Fenix? Lockbox?) ends up with a security vulnerability due to some memory safety error in the FFI boundary. I think this would be *extremely* embarrassing for all involved, and is completely possible.

The only solution I can think of is for us to run with either something like msan/asan/valgrind (which IDK if you can use for e.g. Java on Android), or possibly a debug heap that tracks that you're calling all the correct `destroy_*` functions exactly once, ensures nothing that we didn't allocate ever is freed with them, and that we never access memory that has been freed (e.g. it unmaps the pages where items were located after they're freed).

Even this would only be a half solution though, since it wouldn't catch calling a function other than a destructor with the wrong type. This would still be an exploitable memory safety error. ExternResult<sup id="a1">[1](#f1)</sup> makes this sort of thing sound very plausible to me, and the Android code's heavy use of Pointer even more so<sup id="a2">[2](#f2)</sup>.

<span id="f0">[0](#a0)</span>: Well, some of them do. Removing heap allocated ExternResults, should make things a bit better. Same goes for some of the changes made to the android FFI, which make an effort to prevent double frees or use after frees by making sure you discard RustObject's rawPointer after it's freed. 

<span id="f1">[1](#a1)</span>: Presumably, doing something like `#[repr(C)] struct ExternResult<T> { ok: *mut T, err: *mut c_char }` is our best option, especially since we hand write our .h/JNA.java files. We'd then write `FooResult`, and `BarResult` for each `ExternResult<Foo>`/`ExternResult<Bar>` in the `store.h`.

<span id="f2">[2](#a2)</span>: I gather the way to get more type safety for some of this is to define PointerTypes? JNI also is a bit better here since we'd be implementing a lot more of the bindings in Rust, where we could be confident about not making mistakes.

scripts/ios_build.sh
```diff
@@ -1,4 +1,5 @@
 cd ffi
 cargo lipo --release
 cd ..
+mkdir -p sdks/swift/Mentat/External-Dependencies/
```
ncalexan: Thank you, I had done this locally but apparently I didn't get it upstream.

sdks/swift/Mentat/Mentat/Extensions/String+Destroying.swift
```diff
@@ -0,0 +1,21 @@
...
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License. */
+
+
+import Foundation
+import MentatStore
+
+public extension String {
+    /** Helper to construct a String from a mentat string without leaking it */
```
ncalexan: nit: capital "Mentat", trailing period.

sdks/swift/Mentat/Mentat/Extensions/UnsafeMutablePointerResult+Unwrap.swift
```diff
@@ -11,37 +11,43 @@
 import Foundation
 import MentatStore
 
-public extension Result {
+extension UnsafeMutablePointer where Pointee == Result {
     /**
-     Force unwraps a result.
-     Expects there to be a value attached and throws an error is there is not.
-
+     Unwraps an optional result, yielding either a successful value or a nil.
```
ncalexan: nit: Just "or nil".  I don't think there are multiple `nil` instances, even in Swift.

sdks/swift/Mentat/Mentat/Extensions/UnsafeMutablePointerResult+Unwrap.swift
```diff
@@ -11,37 +11,43 @@
...
+     
+     Frees the Result afterwards, but not the object held in the `ok` property (which is what you want)
```
ncalexan: How about: "... in the `ok` property.  Usually, this is what you want: in the success case, you extract the `ok` property and don't want it to vanish beneath you, but you do want the container to be released."

sdks/swift/Mentat/Mentat/Query/Query.swift
```diff
@@ -246,10 +246,12 @@ open class Query: OptionalRustObject {
      */
     open func run(callback: @escaping (RelResult?) -> Void) throws {
         let result = query_builder_execute(try! self.validPointer())
+        defer { destroy(result); }
```
ncalexan: Oh my, you are really right -- this is just fragile.

sdks/swift/Mentat/Mentat/Extensions/UnsafeMutablePointerResult+Unwrap.swift
```diff
@@ -11,37 +11,43 @@
...
         }
         return success
     }
-
+    
     /**
      Unwraps an optional result, yielding either a successful value or a nil.
-
+     
+     Frees the Result afterwards, but not the object held in the `ok` property (which is what you want)
```
ncalexan: This surprised me at first, but I see that things that are called `try_unwrap` do consume `self` in the Rust standard library.  +1 for documenting this fact.

sdks/android/Mentat/build.gradle
```diff
@@ -10,7 +10,7 @@ buildscript {
     ext.build = [
             compileSdkVersion: 27,
             targetSdkVersion: 27,
-            minSdkVersion: 16
+            minSdkVersion: 19
```
ncalexan: Can I get an explanation for why this is necessary?  We're not fixed _yet_, but I'd like to understand what's driving our choices.

sdks/android/Mentat/library/src/main/java/org/mozilla/mentat/AttributeList.java
```diff
@@ -39,10 +38,5 @@
         return Arrays.asList("attributes", "numberOfItems", "len");
     }
 
-    @Override
-    public void close() {
-        if (this.getPointer() != null) {
-            JNA.INSTANCE.destroy(this.getPointer());
-        }
-    }
+    // TODO: where is this object returned, who owns it's memory, who should be freeing it?
```
ncalexan: nit: its.  This is terrifying me.

sdks/android/Mentat/library/src/main/java/org/mozilla/mentat/Mentat.java
```diff
@@ -67,14 +67,15 @@ public void cache(String attribute, CacheDirection direction) {
         switch (direction) {
             case FORWARD:
                 result = JNA.INSTANCE.store_cache_attribute_forward(this.rawPointer, attribute);
+                break;
```
ncalexan: I'm going to assume we weren't (aren't?) testing any of this, 'cuz this should fail pretty hard.

sdks/android/Mentat/library/src/main/java/org/mozilla/mentat/RustIterator.java
```diff
@@ -0,0 +1,69 @@
...
+// others in the future). This code is a bit error-prone so it's worth avoiding the duplication.
+abstract class RustIterator<E extends RustObject> extends RustObject implements Iterator<E> {
+    // We own this if it is not null!
+    private Pointer nextPointer;
+
+    RustIterator(Pointer pointer) {
+        super(pointer);
+    }
+
+    /** Implement by calling `JNI.INSTANCE.whatever_iter_next(this.validPointer());` */
```
ncalexan: Maybe "Implementations generally should call `JNI....`"

sdks/android/Mentat/library/src/main/java/org/mozilla/mentat/RustIterator.java
```diff
@@ -0,0 +1,69 @@
...
+        if (next == null) {
+            throw new NoSuchElementException();
+        }
+        return this.constructItem(next);
+    }
+
+    @Override
+    public void close() {
+        if (this.nextPointer != null) {
+            // clean up the next pointer by delegating to the iterated item -- this simplifies
```
ncalexan: Nit: capital "// Clean up ...".

sdks/android/Mentat/library/src/main/java/org/mozilla/mentat/RustIterator.java
```diff
@@ -0,0 +1,69 @@
...
+        if (this.nextPointer == null) {
+            throw new NullPointerException(this.getClass() + " nextPointer already consumed");
+        }
+        Pointer p = this.nextPointer;
+        this.nextPointer = null;
+        return p;
+    }
+
+    @Override
+    public boolean hasNext() {
```
ncalexan: Is this semantically allowed by Java iterators?  (I.e., that calling `hasNext()` might invoke `next()`, potentially consuming a resource or whatever?

sdks/android/Mentat/library/src/main/java/org/mozilla/mentat/RustObject.java
```diff
@@ -10,21 +10,32 @@
...
  * subclasses to implement it. This ensures that all classes that inherit from RustObject
  * will have their {@link Pointer} destroyed when the Java wrapper is destroyed.
  */
-abstract class RustObject implements Closeable {
-    Pointer rawPointer;
+abstract class RustObject implements AutoCloseable {
+    // This should probably be private to let us better prevent usage mistakes (which lead to
+    // memory-unsafety).
+    protected Pointer rawPointer;
+    // If this is set to true, we don't destroy this object. Right now this is used for TypedValueAtIndex
```
ncalexan: I don't see another usage of `notOwned` in _this_ patch, but perhaps it is used in others.  Can we improve this, in any case, perhaps by changing `TypedValueAtIndex`?

sdks/android/Mentat/library/src/main/java/org/mozilla/mentat/TupleResult.java
```diff
@@ -136,8 +129,8 @@ public Date asDate(Integer index) {
      * @return  The {@link String} at that index.
      */
     public String asString(Integer index) {
-        this.validate();
-        return JNA.INSTANCE.value_at_index_into_string(this.rawPointer, index);
+            return getAndConsumeMentatString(
```
ncalexan: Indentation looks funky here.

ffi/src/lib.rs
```diff
@@ -132,6 +132,14 @@ pub use utils::log;
 pub type BindingIterator = vec::IntoIter<Binding>;
 pub type BindingListIterator = std::slice::Chunks<'static, mentat::Binding>;
 
+/// Helper macro for asserting one or more pointers are not null at the same time.
+#[macro_export]
+macro_rules! assert_not_null {
+    ($($e:expr),+ $(,)*) => ($(
+        assert!(!$e.is_null(), concat!("Unexpected null pointer: ", stringify!($e)));
```
ncalexan: `stringify!` arbitrary expression might lead to some odd reports, but I assume that `assert!` captures enough information (file/line) to be useful.

ffi/src/lib.rs
```diff
@@ -1730,14 +1836,17 @@ pub unsafe extern "C" fn store_register_observer(store: *mut Store,
             reports: extern_reports.into_boxed_slice(),
             len: len,
         };
-        callback(string_to_c_char(obs_key), &reports);
+        let s = string_to_c_char(obs_key);
+        callback(s, &reports);
```
ncalexan: One wonders what happens if `callback` is badly behaved.

ffi/src/lib.rs
```diff
@@ -1801,7 +1913,9 @@ macro_rules! define_destructor (
     ($name:ident, $t:ty) => (
         #[no_mangle]
         pub unsafe extern "C" fn $name(obj: *mut $t) {
-            let _ = Box::from_raw(obj);
+            if !obj.is_null() {
```
ncalexan: I argued against null-checking destructors once before, so I'll do it again.  I contend that if we're double freeing (or freeing null instances!) that something is horribly wrong and a quick crash report is the best possible outcome.  (All the asserts this commit adds should generate exactly those crash reports, hopefully with actionable stacks.)
 
However, both you (@thomcc) and I think @fluffyemily wanted to do this null-checking, and this is definitely not my area of expertise, so can you come to a decision and inform me of what our pattern is?  Thanks!

sdks/android/Mentat/library/src/main/java/org/mozilla/mentat/RustObject.java
```diff
@@ -33,8 +33,6 @@
 
     RustObject(Pointer p) {
         rawPointer = p;
-        Log.d("RustObject", "Created " + this.getClass() + " " + p.toString());
```
ncalexan: Please fold this into your stack before landing.

sdks/swift/Mentat/Mentat/Query/Query.swift
```diff
@@ -250,8 +250,7 @@ open class Query: OptionalRustObject {
         self.raw = nil
 
         if let err = result.pointee.err {
-            let message = String(cString: err)
-            destroy_mentat_string(err);
+            let message = String(destroyingMentatString: err)
```
ncalexan: And fold this back into the relevant patch as well.

sdks/android/Mentat/library/src/main/java/org/mozilla/mentat/RustObject.java
```diff
@@ -27,9 +27,7 @@
 abstract class RustObject implements AutoCloseable {
     // This should probably be private to let us better prevent usage mistakes (which lead to
     // memory-unsafety).
-    protected Pointer rawPointer;
-    // If this is set to true, we don't destroy this object. Right now this is used for TypedValueAtIndex
-    boolean notOwned = false;
+    private Pointer rawPointer;
```
ncalexan: Ah, you addressed my concern.  Please fold this follow-up back into your stack before landing as well.

sdks/android/Mentat/build.gradle
grigoryk: At the very least, this series makes `RustObject` use `AutoCloseable` which is API19+. This allows for a much prettier pattern around try-with-resources - `close` is automatically called when exiting the `try` block.

sdks/android/Mentat/library/src/main/java/org/mozilla/mentat/RustObject.java
```diff
@@ -35,7 +42,25 @@ void validate() {
         }
     }
 
-    public Pointer getPointerForUUID(UUID uuid) {
+    Pointer validPointer() {
+        this.validate();
+        return this.rawPointer;
+    }
+
+    boolean isConsumed() {
```
grigoryk: nit: Perhaps use `isConsumed` in `validate`, since concept of "valid" is defined in terms of "consumed"?

grigoryk: And perhaps rename `validate` into something like `assertValidPointer`. Would also be good to establish a pattern (in a docstring) as to when we should throw on null pointer, and when we simply check.

sdks/android/Mentat/library/src/main/java/org/mozilla/mentat/TypedValue.java
```diff
@@ -44,8 +40,7 @@ public TypedValue(Pointer pointer) {
      */
     public Long asLong() {
         if (!this.isConsumed()) {
-            this.value = JNA.INSTANCE.typed_value_into_long(this.rawPointer);
-            this.rawPointer = null;
+            this.value = JNA.INSTANCE.typed_value_into_long(this.consumePointer());
```
grigoryk: (off-topic) I realize this isn't part of your changes, but the casting pattern we establish in this class is pretty error-prone. The API lets you do things like `typedValue.asString(); typedValue.asDouble()` and expect something sensible to happen - but I think we'll just crash instead!

sdks/android/Mentat/library/src/main/java/org/mozilla/mentat/TupleResult.java
```diff
@@ -88,8 +85,7 @@ public Long asEntid(Integer index) {
      * @return  The keyword at that index.
      */
     public String asKeyword(Integer index) {
-        this.validate();
-        return JNA.INSTANCE.value_at_index_into_kw(this.rawPointer, index);
+        return getAndConsumeMentatString(JNA.INSTANCE.value_at_index_into_kw(this.validPointer(), index));
```
grigoryk: (off-topic) this seems pretty brittle to me - we're implicitly relying on presence of a `clone()` in Rust to not crash when these kinds of methods are called more than once for a given index.

ffi/src/lib.rs
thomcc: As discussed in IRC, because we can't panic across FFI boundaries doing null checks is probably the best option.

sdks/android/Mentat/library/src/main/java/org/mozilla/mentat/RustIterator.java
thomcc: I think so.

1. This object should be the only thing that has access to the iterator pointer, so nobody should be able to tell if we advance it, so long as we appear to not have advanced it
2. Calling hasNext multiple times will not advance it multiple times. We advance it once just to see if we're at the end or not.
3. Calling next() after hasNext has the same behavior as calling it without.

That is to say, in the API this class exposes, hasNext is a pure function, and next is the only impure one.

Under the hood, hasNext may fetch the item that will be returned by next() the next time you call it, but if you didn't call hasNext, next will just do that.

thomcc:
The latest two commits are a very different approach with the following benefits:

1. The type returned by FFI functions are now explicit. This mitigates issues where the documentation claims something returns an ExternResult that holds one type, but it actually holds another type. This was especially error prone given that we would convert from Result<T, E> to ExternResult with into(), which means we did so without ever typing out the actual return type!
2. Java is now much more type safe, while still using JNA. This was actually not as hard as I had feared. If we use this technique it should go a long way towards preventing passing pointers to functions that take different types.
3. The Swift SDK is still not type safe, but it's no worse than before. I will admit the ergonomics did suffer here, but I think benefits 1 and 2 make up for it. I had hoped that I'd be able to have an opaque T in a `Unsafe{,Mutable}Pointer<T>`, but no such luck.

I still haven't responded to all review comments

ffi/src/lib.rs
```diff
@@ -1783,17 +1999,27 @@ pub unsafe extern "C" fn tx_change_list_entry_at(tx_report_list: *mut TxChangeLi
...
+    assert_not_null!(tx_report);
     let tx_report = &*tx_report;
-    let index = index as usize;
-    tx_report.changes[index].clone()
+    assert!(0 <= index && (index as usize) < (tx_report.changes_len as usize));
+    std::ptr::read(tx_report.changes.offset(index as isize))
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn destroy_mentat_string(s: *mut c_char) {
```
fluffyemily: Can this be `c_string_destroy` instead? That will keep the pattern of `<type>_destroy` we have for all of the other destructors and `<type>_action` that we have for all the FFI functions and de-mentatifies a function that applies to all C Strings and not just Mentat generated ones.

sdks/swift/Mentat/Mentat/Extensions/RustError+Unwrap.swift
```diff
@@ -0,0 +1,48 @@
...
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License. */
+
+import Foundation
+import MentatStore
+
+extension RustError {
+
+    @discardableResult
+    public static func unwrap(_ callback: (UnsafeMutablePointer<RustError>) throws -> OpaquePointer?) throws -> OpaquePointer {
```
fluffyemily: I agree, this pattern is unpleasant. I will have a think about how to do it better.

sdks/android/Mentat/library/src/main/java/org/mozilla/mentat/Query.java
```diff
@@ -281,22 +284,23 @@ public void run(final CollResultHandler handler) {
      * @param handler   the handler to call with the results of this query
      */
     public void run(final TupleResultHandler handler) {
-        RustResult result = JNA.INSTANCE.query_builder_execute_tuple(this.consumePointer());
+        RustError.ByReference error = new RustError.ByReference();
```
fluffyemily: When we finally get around to throwing exceptions on error in Java rather than just logging and moving on, is this going to affect this pattern so we have to have something similar to Swift?

sdks/android/Mentat/library/src/main/java/org/mozilla/mentat/RustIterator.java
fluffyemily: `JNA` not `JNI`

ffi/src/lib.rs
thomcc: > de-mentatifies a function that applies to all C Strings and not just Mentat generated ones.

I did this mostly because I wanted to avoid name collisions, but we can probably address them if/when they become an issue.

thomcc:
> I think we can do something with typing in Swift. Possibly involving better annotations in the C header to give the opaque pointers a named type value and then casting to an UnsafePointer of that type. I can take a look at that if you want to move on though. As you say, it's no worse than it was before.

...

> Now, you can cast an OpaquePointer to an typed UnsafePointer in Swift. I don't know what the result of this would be here, but I can experiment...

The issue is there's no `T` here, since it was an incomplete type in C. I thought about doing `struct Foo {};` in the header, which would let us use `UnsafePointer<T>`/`UnsafeMutablePointer<T>`, and we could have all the types return the real pointer types and get type safety. Then swift would be back to having more type safety than the java code, since it understands const/non-const.

The reason I didn't do this is because it's very much not the layout of these structs, and it would allow swift code to create values of these types without getting them through from rust. Doing so would always be a bug.

There might be a way to signal this appropriately like `struct Foo { void *__internal_fake_data; }` or something though, where if you do `Foo(__internal_fake_data: UnsafeRawMutablePointer(bitPattern: 0))` (or something similar) in Swift you know you're doing something very dodgy. I'd be fine with this, but it would be a strange pattern.

sdks/swift/Mentat/Mentat/Extensions/String+Destroying.swift
```diff
@@ -13,7 +13,7 @@ import Foundation
 import MentatStore
 
 public extension String {
-    /** Helper to construct a String from a mentat string without leaking it */
+    /** Helper to construct a String from a Mentat string without leaking it. */
     public init(destroyingMentatString mentatCString: UnsafeMutablePointer<CChar>) {
```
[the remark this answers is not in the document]
thomcc: I've made it `destroyingRustString` to emphasize that you need to only pass strings allocated on the rust heap to this, and not on the C heap.

sdks/android/Mentat/library/src/main/java/org/mozilla/mentat/Query.java
thomcc: I think the error.logOrWhatever() will just be `error.throwIfError()`. The lack of lightweight first class functions in java means that IDK if it's worth trying to abstract it away, especially given that the abstraction would increase the amount of code callers have to type compared to doing it directly.

thomcc:
So, some further thinking is that while the "type safe swift" approach I linked in the previous comment (e.g. https://github.com/thomcc/mentat/tree/type-safe-swift, but the only difference is the single patch https://github.com/thomcc/mentat/commit/85322bebd2256c905e795c075cca167140f8f5a0) has some drawbacks (If you create a `RawBlah(_internal_fake_data: ...)` from swift it will be a memory safety issue basically 100% of the time), this actually seems much easier to audit and prevent than ensuring everybody passes their OpaquePointers to the correct functions.

If we do this we'll have type safety from Swift *and* Java, with the caveat that it depends on getting the declarations in the `store.h`/`JNA.java` equivalents correct. This is a big caveat, since we had mistakes in ours, but it also seems *a lot* easier to address with tooling.

Issue #697, opened 2018-05-14
Generate Swiftdoc, JavaDoc and FFI Rustdoc in Integration


fluffyemily:
Add a step to Travis build to generate documentation for iOS, Android and Rust FFI.

fluffyemily:
Alternatively, investigate whether https://readthedocs.org/ can give us what we need.

PR #754, opened 2018-06-21
Mentat documentation website using Jekyll

fluffyemily:
Basic website using Jekyll to create a site for accessing documentation, blog posts etc. 

It doesn't look very pretty right now, but it is there. This will tie into https://github.com/mozilla/mentat/issues/697 to ensure that we keep the documentation up to date.


grigoryk:
- We should not include documentation for our Rust dependencies
- It works for now, but "Posts" are probably not a good way to represent content such as "worked examples" and "tutorials" and "guidance on usage" etc. I think pages and some basic structure around that are a better fit
- `about` page right now is a Jekyll placeholder; maybe change it to something simple that's mentat-specific?

A more general comment is on how keeping this stuff up-to-date. We have two tiers of content: hand-written docs, tutorials, guidance, etc (which are a well suited to live in the tree as long as we can remember to keep them up-to-date) and generated content (apis, docstrings, etc).

Generated docs are _already_ out of date (I see 'error_chain' in the Rust docs, which isn't relevant anymore). Dumping bunch of generated files without any automation, or git hooks, or ... into the tree seems like the worst way to ensure they're up do date? Any change that involves any of the APIs, or docstrings, will need to included re-generated static files.

Could we host the generated files someplace where they're hosted and automatically updated from our release tags? Could we use something like readthedocs for the generated stuff, and link to relevant pages from our "overview" docs?

We don't have to figure that out in this PR, I suppose - as long as we periodically bump our documentation forward until we do... But if that's the path, let's at least file a follow-up.

grigoryk:
Ah! I didn't read the linked issue #697! Ok, so the plan is to have travis auto-generate docs and commit them to the tree for us whenever commits happen? I didn't realize travis build bot has commit rights to the repository. Or were you thinking about having it upload generated files someplace?

I really think hosting this stuff externally is a better fit if we can help it, if we can find something that works for rust, java and swift.

fluffyemily:
I'm investigating writethedocs for external hosting and will add it to #697 as a possibility

fluffyemily:
Landing this so we can iterate.

Steps to building docs locally:

1. [Install Jekyll](https://jekyllrb.com/docs/installation/)
2. `cd docs`
3. `bundle exec jekyll serve --incremental`
4. open local docs site at [http://127.0.0.1:4000/](http://127.0.0.1:4000/)

PR #766, opened 2018-06-25
Implement excision. (#21)

ncalexan:
My WIP towards excision.


db/src/excision.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    for (&e, avs) in eav_trie.iter() {
+        for (&(_a, _attribute), ars) in avs {
+            if !ars.retract.is_empty() {
+                bail!(DbError::BadExcision("retraction".into())); // TODO: more details.
+            }
+        }
+
+        let target = avs.get(&pair(entids::DB_EXCISE)?)
+            .and_then(|ars| ars.add.iter().next().cloned())
```
grigoryk: Excising multiple targets would require multiple excisions, right? E.g. there's no way to do something like `[{:db/excise 42 43 44}]`?

db/src/excision.rs
```diff
@@ -0,0 +1,129 @@
...
+            .ok_or_else(|| DbError::BadExcision("no :db/excise".into()))?; // TODO: more details.
+
+        if schema.get_ident(target).is_some() {
+            bail!(DbError::BadExcision("cannot mutate schema".into())); // TODO: more details.
+        }
+
+        let partition = partition_map.partition_for_entid(target)
+            .ok_or_else(|| DbError::BadExcision("target has no partition".into()))?; // TODO: more details.
+        // Right now, Mentat only supports `:db.part/{db,user,tx}`, and tests hack in `:db.part/fake`.
+        if partition == ":db.part/db" || partition == ":db.part/tx" {
```
grigoryk: Explicitly testing for partitions that are allowed in excisions seems more future-proof.

db/src/excision.rs
```diff
@@ -127,3 +133,108 @@ pub(crate) fn excisions<'schema>(partition_map: &'schema PartitionMap, schema: &
...
+        if let Some(attrs) = excision.attrs {
+            // println!("attrs {:?}", attrs);
+            for attr in attrs {
+                stmt2.execute(&[&entid, &attr])?;
+            }
+        }
+    }
+
+    // TODO: filter by attrs.
+    let mut stmt: rusqlite::Statement = conn.prepare(format!("WITH ids AS (SELECT d.rowid FROM datoms AS d, excisions AS e WHERE e.status > 0 AND (e.target IS d.e OR (e.target IS d.v AND d.a IS NOT {}))) DELETE FROM datoms WHERE rowid IN ids", entids::DB_EXCISE).as_ref())?;
```
grigoryk: What's the reasoning behind splitting excision into two steps? e.g. first delete stuff from datoms, then on a separate call from transactions. It seems backwards to me (but perhaps understanding the reason for the split will help here) - e.g. we first clean out a materialization of our transactions (datoms), and only then the "source of truth". I'd expect a two-step operation to start with the source, and finish at the edges?

db/src/excision.rs
```diff
@@ -127,3 +133,108 @@ pub(crate) fn excisions<'schema>(partition_map: &'schema PartitionMap, schema: &
         Ok(Some(excisions))
     }
 }
+
+pub(crate) fn enqueue_pending_excisions(conn: &rusqlite::Connection, schema: &Schema, tx_id: Entid, excisions: ExcisionMap) -> Result<()> {
```
grigoryk: While this is called "enqueue", the function actually mutates state beyond just making a note of an excision-to-be-processed. Not in a permanent sense (we can always derive erased datoms if the `ensure_no_pending_excisions` step doesn't run, but it still does more than it lets on, I think.

db/src/excision.rs
```diff
@@ -134,42 +136,71 @@ pub(crate) fn excisions<'schema>(partition_map: &'schema PartitionMap, schema: &
...
 
+pub(crate) fn enqueue_pending_excisions(conn: &rusqlite::Connection, schema: &Schema, tx_id: Entid, excisions: &ExcisionMap) -> Result<()> {
     let mut stmt1: rusqlite::Statement = conn.prepare("INSERT INTO excisions VALUES (?, ?, ?, ?)")?;
     let mut stmt2: rusqlite::Statement = conn.prepare("INSERT INTO excision_attrs VALUES (?, ?)")?;
 
     for (entid, excision) in excisions {
-        stmt1.execute(&[&entid, &excision.target, &excision.before_tx, &excision.before_tx.unwrap_or(tx_id)])?; // XXX
-        if let Some(attrs) = excision.attrs {
-            // println!("attrs {:?}", attrs);
+        let status = excision.before_tx.unwrap_or(tx_id);
```
grigoryk: The column name `status` seems a bit misleading and its use lossy. Its use as a "do i need to process anything" means that we don't record the "beforeT" attribute (like we do for other excision parameters like attributes). It _is_ in the transaction table, yes, but the difference in handling seems inconsistent to me.

The "misleading" bit - `status` implies "do i have something to do?", but here its also used to mean "what do i need to do?" at the same time, and the meaning of the column changes as we step through the steps.

I'm guessing you're optimizing away an extra column, essentially?

db/src/excision.rs
ncalexan: > Excising multiple targets would require multiple excisions, right? E.g. there's no way to do something like [{:db/excise [42 43 44]}]?

No, although I see now that I may have implemented this incorrectly.  `:db/excise` is `:db/cardinality :db.cardinality/many` (which I assume we culted from Datomic), so the expression you give is accepted by the transactors.  I will go add tests to ensure I'm doing reasonable things -- I feel like I might have assumed that the entity corresponds to only one target, which isn't true if this is cardinality many.

db/src/excision.rs
ncalexan: You're right.  I'll grow an `excision_allowed()` flag.

db/src/excision.rs
ncalexan: I did implement this incorrectly.  To capture things correctly I had to make `target` be `targets` and change the underlying materialized view.

db/src/excision.rs
ncalexan: I've updated to use this.

db/src/db.rs
```diff
@@ -247,6 +259,18 @@ lazy_static! {
         r#"CREATE INDEX idx_schema_unique ON schema (e, a, v, value_type_tag)"#,
         // TODO: store entid instead of ident for partition name.
         r#"CREATE TABLE parts (part TEXT NOT NULL PRIMARY KEY, start INTEGER NOT NULL, idx INTEGER NOT NULL)"#,
+
+        // Excisions are transacted as data, so they end up in the transactions table; this
+        // materializes that view.  Excisions are never removed from the transactions (or datoms)
+        // table so it's not required to include the `added` column, but let's do it so that if we
```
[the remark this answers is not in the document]
ncalexan: You're right.

db/src/db.rs
```diff
@@ -2839,4 +2901,118 @@ mod tests {
...
+             [300 :test/many 2001]
+             [300 :test/many 2002]
+             [301 :test/one 1001]
+             [301 :test/ref 300]]"#);
+
+        let report = assert_transact!(conn, r#"[
+            {:db/id "e" :db/excise 300}
+        ]"#);
+        // This is implementation specific, but it should be deterministic.
+        assert_matches!(tempids(&report),
```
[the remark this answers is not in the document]
ncalexan: I copied the marker comment; I think it's clear enough, should we ever wish to adjust things.

db/src/excision.rs
ncalexan: This is a choice, suggested by Richard in https://github.com/mozilla/mentat/issues/21#issuecomment-372730559.

Excision is very expensive, so we want to be able to amortize the cost over time.  The initial implementation won't do much toward amortizing, but I want to put the framework in place for it.  The reason to start at the materialization (`datoms`, the edge) is that immediately after an excision the world presented to queries should change.  You could imagine an alternate implementation where the expensive writes to the `datoms` table are themselves queued, and instead the query engine grows knowledge of "excised entities" and filters them from query results as appropriate.  You could even do this with a SQLite view: if you had an index that made it fast to look up whether a particular entity was pending excision, like
```sql
CREATE INDEX idx_excisions_pending ON excisions (target) WHERE status IS NOT 0
```
then you could
```sql
CREATE VIEW datoms AS
SELECT * FROM datoms_before_excisions
WHERE NOT EXISTS
 (SELECT 1 FROM excisions
  WHERE
  excisions.status IS NOT 0 AND
  ((datoms.e IS excisions.target) OR (datoms.v IS excisions.target AND datoms.value_type_tag = 13))
```

That might be a good idea, since most of the time the `idx_excisions_pending` partial index will be empty and the view will be as performant (I think) as querying the underlying datoms directly.

However, the _really_ expensive part is rewriting all of the transaction log (and, later, rewriting the shared transaction log on a remote service).  We want to be able to do that a little at a time; hence the machinery.  So, really, we're splitting excision into `datoms` and `transactions`, and then splitting `transactions` into lots of little re-writing steps that can proceed incrementally.

Only time will tell which of these approaches is best, and whether the machinery is worth the effort.

db/src/excision.rs
ncalexan: This is a fair comment.  Maybe `begin_excisions`?

ncalexan:
> You're doing a bunch of stuff with SQL as separate statements, all wrapped in individual implicit transactions.
> 
> I expected to see as many of these individual operations wrapped together in a single transaction so that the whole thing is atomic. Is there a reason that's not feasible?

The layer that makes the whole thing atomic is higher up the chain -- it's the `Conn` that manages all of this.  The `Conn` also needs to be involved to handle transaction watchers and cache invalidation.  That's coming shortly.

db/src/excision.rs
ncalexan: > I'm guessing you're optimizing away an extra column, essentially?

Yes.

I pushed this up before revisiting this with a helpful comment.  You deduced most of the assumptions and asked questions about the rest (great!); but I'll update this comment and add a class comment as well.

Issue #711, opened 2018-05-24
Display entity names during AttributeValidation errors, not just `entid`


grigoryk:
During attribute validation, it's helpful to see actual entity names in the error message, and not just their `entid`.

So, instead of ...
`BadSchemaAssertion(":db/fulltext true without :db/index true for entid: 65537")`

... it'd be nice to bail out with ...
`BadSchemaAssertion(":db/fulltext true without :db/index true for :page/title (entid: 65537)")`

... or just drop `entid` entirely, as it's not providing useful information when debugging a vocabulary:
`BadSchemaAssertion(":db/fulltext true without :db/index true for :page/title")`

PR #773, opened 2018-06-28
Low hanging dependency fruit

thomcc:
This does #711, which is the most meaningful. Removing unused `num` dependency from core is also worth doing, even if just for cleanliness.

I also noticed yesterday that we were using `num` to iterate backwards over a range in `remove_every`, which probably should have been `(0..self.len()).rev()` if it were going to stay.

But this was still `O(n^2)`, since remove is `O(n)`. Using `retain` removes all of them in one pass, and IMO is clearer (or at least less code). I also fixed the test for this function to actually check that return the count we had before, which we weren't doing. (I also killed the comment that said to replace it with a function that AFAICT isn't going to do the same thing).

FWIW removing the num deps probably doesn't effect the lib size, since we still depend on it in `edn`.


query-sql/src/lib.rs
```diff
@@ -559,15 +558,20 @@ impl QueryFragment for FromClause {
...
 
+/// `var` is something like `?foo99-people`.
+/// Trim the `?` and escape the rest. Prepend `i` to distinguish from
+/// the inline value space `v`.
+fn format_select_var(var: &str) -> String {
+    let without_question = var.split_at(1).1;
+    let replaced_iter = without_question.chars().map(|c|
+        if c.is_ascii_alphanumeric() { c } else { '_' });
+    // Prefix with `i` (Avoiding this copy is probably not worth the trouble but whatever).
+    ['i'].into_iter().cloned().chain(replaced_iter).collect()
```
ncalexan: nit: use `std::iter::once`?

ncalexan:
Landed in 7d2fe8c62528e96d04c51513e94e167612a328bb.

PR #775, opened 2018-06-28
Closes #767 - Fix Broken documentation links.

### LINKED ISSUE #767: [docs] API links are broken
I think the initial landing of the `docs/` directory and Jekyl integration landed with version numbers, but the landing page at https://mozilla.github.io/mentat/ is full of broken links because the version numbers are missing.

I'd like to see:
- the links corrected
- some kind of "latest" and "v0.x" incorporated
- version numbers of the APIs to be at the top-level, like `mentat/0.7/{swift,rust,...}` rather than `mentat/{swift,rust}/0.7`
- version numbers to agree with the main Mentat version number (so 0.7 throughout, or whatever is appropriate)
- the stanza `bundle exec jekyll serve --incremental` somewhere in the documentation itself

@fluffyemily can you take some or all of those items?  Thanks!  (P.S.  Your instructions worked well for me locally, and the site looks good!)

fluffyemily:
Addressing comments and issues raised in #767 

- Fix broken API documentation links.
- Unify SDKs and Mentat under same version number.
- Reorganize API documentation under version number.
- Add `latest` symlink to point to the latest API documentation version.
- Update `.gitignore` to ignore documentation related directories and files.
- Add README to `/docs` to include local build instructions.
- Update API documentation.


fluffyemily:
So it turns out that you have to create the symlink inside the directory where the link it so be created in order for git to understand the link, so `ln -s docs/0.7 docs/latest` won't work but `cd docs && ln -s 0.7 latest` will.

Also, I stupidly didn't check in the correct editing of the link in `index.md`. Sorry.

Issue #674, opened 2018-05-02
[tx] Consider replacing `Term` building with `Entity` building


ncalexan:
This is a ticket for discussion more than implementation.  Right now we have three entry points to the transactor:
- input as a string of EDN
- input as `Vec<Entity>`, structured but with `edn::ValueAndSpan` values
- input as `Vec<Term>`, greatly restricted (and built of `TypedValue` values)

All of these are a little awkward to use.  We have tickets for:
- accepting placeholders in EDN strings (#655)
- not skipping critical validation staps in the term builder interface (#663)
- producing richer inputs with the term builder interface (#666)

I think we should discuss removing the third entry point -- the lowest level entry point as `Vec<Term>` -- entirely, and making it possible for programmatic consumers to build the richer `Entity` types directly.  As a consumer, there are lots of things that the `Term` interface doesn't support (for instance, for Sync 1.5 logins I want `(current-tx)`, and that's too high an abstraction for `Term`).  And as an implementer, I want to be confident that we're not introducing bugs by entering the pipeline partway through, violating invariants enforced by earlier stages.

@rnewman, @fluffyemily -- what do you think?  Richard, you implemented the initial term builder interface.  It seems that you were motivated by what was most accessible at that time:

Way back when the entity builder was introduced, @rnewman observed (see https://github.com/mozilla/mentat/commit/812f10b3e42bd1face73eecd9d17c57bf860848d#diff-8cfa3ec371250108d34a8687bd4ab9bdR51) that

> Entity -- the input to the transactor -- is intimately tied to EDN and to spanned values.

That's not really true.  I worked up a little patch to make `Entity` vary over it's value type `V`, although I stopped short of producing the trait that will be needed to abstract the schema-aware type coercion and value manipulation that will be necessary for the transactor to work with some abstraction of an EDN (spanned) value.

@fluffyemily, you are most responsible for the FFI interface.  Is there a difference to FFI consumers if the various builder interfaces produce `Entity<TypedValue>` rather than `Term`?

PR #778, opened 2018-07-05
Build `Entity` instances, not `Term*` instances.  (#674)

ncalexan:
I haven't tried to rename `EntityBuilder` (or remove it), and I haven't tried to rename `TermBuilder` to `EntityBuilder`.  Those can come later.

This also doesn't intern `TempId` handles at parse time; we can get to that.

This addresses #674.

### DIFF
--- a/edn/src/entities.rs
+++ b/edn/src/entities.rs
@@ -13,10 +13,28 @@
 use std::collections::BTreeMap;
 use std::fmt;
 
+use value_rc::{
+    ValueRc,
+};
+
 use symbols::{
     Keyword,
     PlainSymbol,
 };
+
+use types::{
+    ValueAndSpan,
+};
+
+/// `EntityPlace` and `ValuePlace` embed values, either directly (i.e., `ValuePlace::Atom`) or
+/// indirectly (i.e., `EntityPlace::LookupRef`).  In order to maintain the graph of `Into` and
+/// `From` relations, we need to ensure that `{Value,Entity}Place` can't match as a potential value.
+/// (If it does, the `impl Into<T> for T` default conflicts.) This marker trait allows to mark
+/// acceptable values, thereby removing `{Entity,Value}Place` from consideration.
+pub trait TransactableValueMarker {}
+
+/// `ValueAndSpan` is the value type coming out of the entity parser.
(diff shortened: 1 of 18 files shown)


ncalexan:
> You say "we haven't observed" - which I assume to mean "at some point we've profiled, and didn't find the added cost of Arc to be an issue given our usage patterns"?

Correct.  That happened in #659 and friends; see https://github.com/mozilla/mentat/pull/659#issuecomment-384437713.  This is more analogy than science in this instance.
 
> There is a cost, but I don't have a good enough sense of the flow of things to guess if/when it'll be a problem.

Exactly.

> What about somehow shifting the burden to the somewhat rare (once things application code stabilizes in its ways) error cases, maybe? e.g. clone whatever's is in Rc's if things go wrong, maybe?

I'm happy to do that at some point, and considered it for this -- cloning names out of `TempId` instances while building errors wouldn't be onerous, for example.  Right now, in a number of places, we expose our guts (`ValueRc<String>`, say, rather than `String` or `&str`) directly to consumers.  It's not clear if that's foolish quite yet.

edn/src/entities.rs
```diff
@@ -22,6 +22,19 @@ use symbols::{
     PlainSymbol,
 };
 
+use types::{
+    ValueAndSpan,
+};
+
+/// Entity and value places embed values, either directly (i.e., `ValuePlace::Atom`) or indirectly
+/// (i.e., `EntityPlace::LookupRef`).  In order to maintain the graph of `Into` and `From`
+/// relations, we need to ensure that `{Value,Entity}Place` can't match as a potential value.  This
```
grigoryk: On first read, the dual meaning of "value" in this comment makes the web a little confusing at first read - but perhaps that's just my unfamiliarity with the subject. Unsure what you'd call "embedded values" other than values though.

Perhaps this can be talked about in terms of terminating value resolution? E.g. "{Entity,Value}Place need to resolve to a concrete value, which could happen either directly (i.e. ValuePlace::Atom, value is embedded in the enum) or indirectly (i.e. EntityPlace::LookupRef, value is a hop away). We need to ensure that the resolution is guaranteed to terminate, etc etc".

PR #782, opened 2018-07-11
Add a top-level "syncable" feature.

grigoryk:
Extracted out of #563.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,9 +16,10 @@ version = "0.8.1"
 build = "build/version.rs"
 
 [features]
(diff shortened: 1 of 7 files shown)


db/Cargo.toml
```diff
@@ -17,6 +18,9 @@ log = "0.4"
 ordered-float = "0.5"
 time = "0.1"
 petgraph = "0.4.12"
+serde = { version = "1.0", optional = true }
+serde_json = { version = "1.0", optional = true }
```
ncalexan: Do you actually need `serde_json`?

PR #783, opened 2018-07-12
Basic timelines support

grigoryk:
This PR adds support for basic timelines to the transactor along with ability to move ranges of transactions off-of the `main timeline`.

- Default timeline for all local transactions is the `main timeline`, `0`.
- A timeline of a `transaction` is indicated by an additional single field (same type as `e`).
- Moving of transactions off of the main timeline is implemented as a "rewind": assertions are reversed and transacted in a descending order in order to achieve correct state of materialized views

### DIFF
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -465,7 +465,7 @@ pub(crate) fn read_ident_map(conn: &rusqlite::Connection) -> Result<IdentMap> {
 pub(crate) fn read_attribute_map(conn: &rusqlite::Connection) -> Result<AttributeMap> {
     let entid_triples = read_materialized_view(conn, "schema")?;
     let mut attribute_map = AttributeMap::default();
-    metadata::update_attribute_map_from_entid_triples(&mut attribute_map, entid_triples, ::std::iter::empty())?;
+    metadata::update_attribute_map_from_entid_triples(&mut attribute_map, entid_triples, vec![])?;
     Ok(attribute_map)
 }
 
@@ -1015,14 +1015,30 @@ pub fn update_metadata(conn: &rusqlite::Connection, _old_schema: &Schema, new_sc
                      &[])?;
     }
 
+    // Populate the materialized view directly from datoms.
+    // It's possible that an "ident" was removed, along with its attributes.
+    // That's not counted as an "alteration" of attributes, so we explicitly check
+    // for non-emptiness of 'idents_altered'.
 
-    let mut stmt = conn.prepare(format!("INSERT INTO schema SELECT e, a, v, value_type_tag FROM datoms WHERE e = ? AND a IN {}", entids::SCHEMA_SQL_LIST.as_str()).as_str())?;
-    for &entid in &metadata_report.attributes_installed {
-        stmt.execute(&[&entid as &ToSql])?;
+    // TODO expand metadata report to allow for better signaling for the above.
+
+    if !metadata_report.attributes_installed.is_empty()
+        || !metadata_report.attributes_altered.is_empty()
+        || !metadata_report.idents_altered.is_empty() {
+
+        conn.execute(format!("DELETE FROM schema").as_str(),
+                     &[])?;
+        // NB: we're using :db/valueType as a placeholder for the entire schema-defining set.
+        let s = format!(r#"
+            WITH s(e) AS (SELECT e FROM datoms WHERE a = {})
+            INSERT INTO schema
+            SELECT s.e, a, v, value_type_tag
+            FROM datoms, s
+            WHERE s.e = datoms.e AND a IN {}
+        "#, entids::DB_VALUE_TYPE, entids::SCHEMA_SQL_LIST.as_str());
+        conn.execute(&s, &[])?;
     }
 
-    let mut delete_stmt = conn.prepare(format!("DELETE FROM schema WHERE e = ? AND a IN {}", entids::SCHEMA_SQL_LIST.as_str()).as_str())?;
-    let mut insert_stmt = conn.prepare(format!("INSERT INTO schema SELECT e, a, v, value_type_tag FROM datoms WHERE e = ? AND a IN {}", entids::SCHEMA_SQL_LIST.as_str()).as_str())?;
     let mut index_stmt = conn.prepare("UPDATE datoms SET index_avet = ? WHERE a = ?")?;
     let mut unique_value_stmt = conn.prepare("UPDATE datoms SET unique_value = ? WHERE a = ?")?;
     let mut cardinality_stmt = conn.prepare(r#"
@@ -1035,9 +1051,6 @@ SELECT EXISTS
         left.v <> right.v)"#)?;
 
     for (&entid, alterations) in &metadata_report.attributes_altered {
-        delete_stmt.execute(&[&entid as &ToSql])?;
-        insert_stmt.execute(&[&entid as &ToSql])?;
-
         let attribute = new_schema.require_attribute_for_entid(entid)?;
 
         for alteration in alterations {
@@ -1523,7 +1536,10 @@ mod tests {
     fn test_db_install() {
         let mut conn = TestConn::default();
 
-        // We can assert a new attribute.
+        // We're missing some tests here, since our implementation is incomplete.
+        // See https://github.com/mozilla/mentat/issues/797
+
+        // We can assert a new schema attribute.
         assert_transact!(conn, "[[:db/add 100 :db/ident :test/ident]
                                  [:db/add 100 :db/valueType :db.type/long]
                                  [:db/add 100 :db/cardinality :db.cardinality/many]]");
@@ -1560,10 +1576,32 @@ mod tests {
                           [101 :test/ident -9 ?tx true]
                           [?tx :db/txInstant ?ms ?tx true]]");
 
-        // Cannot retract a characteristic of an installed attribute.
+        // Cannot retract a single characteristic of an installed attribute.
         assert_transact!(conn,
                          "[[:db/retract 100 :db/cardinality :db.cardinality/many]]",
                          Err("bad schema assertion: Retracting attribute 8 for entity 100 not permitted."));
+
+        // Cannot retract a single characteristic of an installed attribute.
+        assert_transact!(conn,
+                         "[[:db/retract 100 :db/valueType :db.type/long]]",
+                         Err("bad schema assertion: Retracting attribute 7 for entity 100 not permitted."));
+
+        // Cannot retract a non-defining set of characteristics of an installed attribute.
+        assert_transact!(conn,
+                         "[[:db/retract 100 :db/valueType :db.type/long]
+                         [:db/retract 100 :db/cardinality :db.cardinality/many]]",
+                         Err("bad schema assertion: Retracting defining attributes of a schema without retracting its :db/ident is not permitted."));
+
+        // See https://github.com/mozilla/mentat/issues/796.
+        // assert_transact!(conn,
+        //                 "[[:db/retract 100 :db/ident :test/ident]]",
+        //                 Err("bad schema assertion: Retracting :db/ident of a schema without retracting its defining attributes is not permitted."));
+
+        // Can retract all of characterists of an installed attribute in one go.
+        assert_transact!(conn,
+                         "[[:db/retract 100 :db/cardinality :db.cardinality/many]
+                         [:db/retract 100 :db/valueType :db.type/long]
+                         [:db/retract 100 :db/ident :test/ident]]");
 
         // Trying to install an attribute without a :db/ident is allowed.
         assert_transact!(conn, "[[:db/add 101 :db/valueType :db.type/long]
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -426,14 +426,10 @@ impl TypedSQLValue for TypedValue {
 /// store.
 pub(crate) fn read_materialized_view(conn: &rusqlite::Connection, table: &str) -> Result<Vec<(Entid, Entid, TypedValue)>> {
     let mut stmt: rusqlite::Statement = conn.prepare(format!("SELECT e, a, v, value_type_tag FROM {}", table).as_str())?;
-    let m: Result<Vec<(Entid, Entid, TypedValue)>> = stmt.query_and_then(&[], |row| {
-        let e: Entid = row.get_checked(0)?;
-        let a: Entid = row.get_checked(1)?;
-        let v: rusqlite::types::Value = row.get_checked(2)?;
-        let value_type_tag: i32 = row.get_checked(3)?;
-        let typed_value = TypedValue::from_sql_value_pair(v, value_type_tag)?;
-        Ok((e, a, typed_value))
-    })?.collect();
+    let m: Result<Vec<_>> = stmt.query_and_then(
+        &[],
+        row_to_datom_assertion
+    )?.collect();
     m
 }
 
@@ -515,14 +511,20 @@ pub trait MentatStoring {
     fn insert_non_fts_searches<'a>(&self, entities: &'a [ReducedEntity], search_type: SearchType) -> Result<()>;
     fn insert_fts_searches<'a>(&self, entities: &'a [ReducedEntity], search_type: SearchType) -> Result<()>;
 
-    /// Finalize the underlying storage layer after a Mentat transaction.
+    /// Prepare the underlying storage layer for finalization after a Mentat transaction.
     ///
     /// Use this to finalize temporary tables, complete indices, revert pragmas, etc, after the
     /// final `insert_non_fts_searches` invocation.
-    fn commit_transaction(&self, tx_id: Entid) -> Result<()>;
+    fn materialize_mentat_transaction(&self, tx_id: Entid) -> Result<()>;
 
-    /// Extract metadata-related [e a typed_value added] datoms committed in the given transaction.
-    fn committed_metadata_assertions(&self, tx_id: Entid) -> Result<Vec<(Entid, Entid, TypedValue, bool)>>;
+    /// Finalize the underlying storage layer after a Mentat transaction.
+    ///
+    /// This is a final step in performing a transaction.
+    fn commit_mentat_transaction(&self, tx_id: Entid) -> Result<()>;
+
+    /// Extract metadata-related [e a typed_value added] datoms resolved in the last
+    /// materialized transaction.
+    fn resolved_metadata_assertions(&self) -> Result<Vec<(Entid, Entid, TypedValue, bool)>>;
 }
 
 /// Take search rows and complete `temp.search_results`.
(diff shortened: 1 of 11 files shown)


db/src/db.rs
```diff
@@ -2778,4 +2779,97 @@ mod tests {
         // Run a basic test as a sanity check.
         run_test_add(TestConn::with_sqlite(sqlite));
     }
+
+    #[test]
+    fn test_timeline_move() {
+        let mut conn = TestConn::default();
+
+        assert_transact!(conn, r#"[
```
ncalexan: You can make this pass by manually allocating in the `:db.part/user` partition, like:
```rust
        assert_eq!((65536..65539),
                   conn.partition_map.allocate_entids(":db.part/user", 3));
```
(Well, pass up until the actual broken part.  See below.)

db/src/timelines.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    let timeline = current_timeline(conn, smallest_tx, tx_ids)?;
+
+    // TODO stil need to return schema and partitionmap...
+    if timeline != MAIN_TIMELINE {
+        bail!(DbErrorKind::TimelinesNotOnMain);
+    }
+
+    let reversed_terms = reversed_terms_after(conn, smallest_tx - 1)?;
+    let lowest_e = match reversed_terms.last() {
```
ncalexan: Fundamentally, it's not possible to determine the set of known partitions, or the current "next index" of a partition, by examining `datoms` or any fixed length set of `transactions`.  You could have:
- tx 1 could allocate 1000 entids
- tx 2 ... N only reference the first, say, 10 entids
- tx N + 1 retracts entid 1000
You'll never "discover" the real range of valid entids.

Thus, the transactor needs to be able to tell you exactly the state of the partition map at each transaction.  I've started working on this and will push something up when I can, but you can also start thinking about this, if you'd like.

grigoryk:
Tests are failing only on 1.25.0 since I'm pattern matching a reference w/ a non-reference pattern.

db/src/db.rs
```diff
@@ -1565,6 +1577,17 @@ mod tests {
...
 
+        // Cannot retract a characteristic of an installed attribute.
+        assert_transact!(conn,
+                         "[[:db/retract 100 :db/valueType :db.type/long]]",
+                         Err("bad schema assertion: Retracting attribute 7 for entity 100 not permitted."));
+
+        // Can retract all of characterists of an installed attribute in one go.
+        assert_transact!(conn,
+                         "[[:db/retract 100 :db/cardinality :db.cardinality/many]
+                         [:db/retract 100 :db/valueType :db.type/long]
```
ncalexan: nit: indentation.

db/src/metadata.rs
```diff
@@ -105,7 +105,38 @@ impl MetadataReport {
 /// Returns a report summarizing the mutations that were applied.
 pub fn update_attribute_map_from_entid_triples<A, R>(attribute_map: &mut AttributeMap, assertions: A, retractions: R) -> Result<MetadataReport>
```
ncalexan: Are these always `Vec`, or could they always be `Vec`?

db/src/metadata.rs
```diff
@@ -105,7 +105,38 @@ impl MetadataReport {
...
     where A: IntoIterator<Item=(Entid, Entid, TypedValue)>,
-          R: IntoIterator<Item=(Entid, Entid, TypedValue)> {
+          R: Iterator<Item=(Entid, Entid, TypedValue)> {
+
+    // Process retractions of schema attributes first. It's allowed to retract a schema attribute
+    // if all of the schema-defining schema attributes are being retracted.
+    // A defining set of attributes is :db/ident, :db/valueType, :db/cardinality.
+    let retractions: Vec<(Entid, Entid, TypedValue)> = retractions.collect();
```
ncalexan: Let's put this into the code, next to `SCHEMA_SQL_LIST` -- say, `SCHEMA_RETRACTION_SQL_LIST`.

ncalexan: Here you're saying `:db/ident`, but you're not consider `:db/ident` below.  Why not?

db/src/metadata.rs
```diff
@@ -105,7 +105,38 @@ impl MetadataReport {
 /// Returns a report summarizing the mutations that were applied.
 pub fn update_attribute_map_from_entid_triples<A, R>(attribute_map: &mut AttributeMap, assertions: A, retractions: R) -> Result<MetadataReport>
     where A: IntoIterator<Item=(Entid, Entid, TypedValue)>,
-          R: IntoIterator<Item=(Entid, Entid, TypedValue)> {
+          R: Iterator<Item=(Entid, Entid, TypedValue)> {
+
+    // Process retractions of schema attributes first. It's allowed to retract a schema attribute
```
ncalexan: For excision, I built a little EAV trie -- see https://github.com/mozilla/mentat/commit/03dd4a37a35602c5ee0256be8504f9672393d744#diff-929f073706d06052e936ffa3c71e523aR217.  I'd like to standardize these types of filtering and searching operations around `AEVTrie` and `EAVTrie`.  Take a look.  I think it would make this a little more pleasant.

db/src/db.rs
```diff
@@ -1565,6 +1577,17 @@ mod tests {
                          "[[:db/retract 100 :db/cardinality :db.cardinality/many]]",
                          Err("bad schema assertion: Retracting attribute 8 for entity 100 not permitted."));
 
+        // Cannot retract a characteristic of an installed attribute.
```
ncalexan: nit: "a _single_ characteristic", or "a _lone_ characteristic".

db/src/db.rs
```diff
@@ -1015,14 +1015,29 @@ pub fn update_metadata(conn: &rusqlite::Connection, _old_schema: &Schema, new_sc
...
-    let mut stmt = conn.prepare(format!("INSERT INTO schema SELECT e, a, v, value_type_tag FROM datoms WHERE e = ? AND a IN {}", entids::SCHEMA_SQL_LIST.as_str()).as_str())?;
-    for &entid in &metadata_report.attributes_installed {
-        stmt.execute(&[&entid as &ToSql])?;
+    // TODO expand metadata report to allow for better signaling for the above.
+
+    if !metadata_report.attributes_installed.is_empty()
+        || !metadata_report.attributes_altered.is_empty()
+        || !metadata_report.idents_altered.is_empty() {
+
+        conn.execute(format!("DELETE FROM schema").as_str(),
```
ncalexan: Maybe a little comment that we're using `:db/valueType` as a placeholder for the entire schema-defining set.

db/src/db.rs
```diff
@@ -515,12 +515,19 @@ pub trait MentatStoring {
     fn insert_non_fts_searches<'a>(&self, entities: &'a [ReducedEntity], search_type: SearchType) -> Result<()>;
     fn insert_fts_searches<'a>(&self, entities: &'a [ReducedEntity], search_type: SearchType) -> Result<()>;
 
-    /// Finalize the underlying storage layer after a Mentat transaction.
-    ///
+    /// Prepare the underlyin storage layer for finalization after a Mentat transaction.
```
ncalexan: nit: underlyin_g_.

db/src/db.rs
```diff
@@ -515,12 +515,19 @@ pub trait MentatStoring {
...
+    /// 
     /// Use this to finalize temporary tables, complete indices, revert pragmas, etc, after the
     /// final `insert_non_fts_searches` invocation.
+    fn materialize_transaction(&self, tx_id: Entid) -> Result<()>;
```
ncalexan: I got tripped up earlier this week, so let's call this `materialize_mentat_transaction`.

db/src/db.rs
```diff
@@ -515,12 +515,19 @@ pub trait MentatStoring {
...
+
+    /// Finalize the underlying storage layer after a Mentat transaction.
+    ///
+    /// This is a final step in performing a transaction.
     fn commit_transaction(&self, tx_id: Entid) -> Result<()>;
```
ncalexan: `commit_mentat_transaction`.

db/src/db.rs
```diff
@@ -515,12 +515,19 @@ pub trait MentatStoring {
...
 
+    fn resolved_metadata_assertions(&self, tx_id: Entid) -> Result<Vec<(Entid, Entid, TypedValue, bool)>>;
```
ncalexan: docstring?

db/src/db.rs
```diff
@@ -946,23 +953,30 @@ impl MentatStoring for rusqlite::Connection {
...
+        Ok(())
+    }
+
+    fn materialize_transaction(&self, tx_id: Entid) -> Result<()> {
+        search(&self)?;
         update_datoms(&self, tx_id)?;
         Ok(())
     }
 
+    // TODO it's not clear that we need this distinction at all (other than 'comitted' is faster).
```
ncalexan: nit: co_m_mitted.

db/src/db.rs
```diff
@@ -995,6 +1009,57 @@ pub fn update_partition_map(conn: &rusqlite::Connection, partition_map: &Partiti
...
+                    ) ORDER BY e, a, v, value_type_tag, added"#,
+                    entids::METADATA_SQL_LIST.as_str(), entids::METADATA_SQL_LIST.as_str()
+                )
+            )
+        }
+    };
+    
+    // TODO: use concat! to avoid creating String instances.
+    let mut stmt = conn.prepare_cached(&sql_stmt)?;
+    let m: Result<Vec<_>> = stmt.query_and_then(&params[..], |row| -> Result<(Entid, Entid, TypedValue, bool)> {
```
ncalexan: Let's extract this to a helper function -- `db::row_to_transaction_assertion`.  Maybe add `db::row_to_datom` as well (without `added`).

db/src/tx.rs
```diff
@@ -130,6 +130,11 @@ use watcher::{
     TransactWatcher,
 };
 
+pub(crate) enum TransactorAction {
```
ncalexan: This is where the nice comment explaining what is happening in the two cases goes :)

db/src/db.rs
```diff
@@ -995,6 +1009,57 @@ pub fn update_partition_map(conn: &rusqlite::Connection, partition_map: &Partiti
     Ok(())
 }
 
+fn metadata_assertions(conn: &rusqlite::Connection, tx_id: Entid, comitted: bool) -> Result<Vec<(Entid, Entid, TypedValue, bool)>> {
```
ncalexan: This isn't a function with a `bool` parameter, it's two functions.

db/src/db.rs
```diff
@@ -995,6 +1009,57 @@ pub fn update_partition_map(conn: &rusqlite::Connection, partition_map: &Partiti
...
+    let (params, sql_stmt) = match comitted {
+        true => {
+            (
+                vec![&tx_id as &ToSql],
+                format!(r#"
+                    SELECT e, a, v, value_type_tag, added
+                    FROM transactions
+                    WHERE tx = ? AND a IN {}
```
ncalexan: This is a table walk, since we don't have an index on `transactions.tx`.  (Or maybe we do?)  I have been thinking of using a different table format, but need to benchmark some things first, so I'm okay with this for now.  (It'll get better in the newer format.)

db/src/timelines.rs
```diff
@@ -0,0 +1,442 @@
...
+/// Get terms for tx_id, reversing them in meaning (swap add & retract).
+fn reversed_terms_for(conn: &rusqlite::Connection, tx_id: Entid) -> Result<Vec<TermWithoutTempIds>> {
+    let mut stmt = conn.prepare("SELECT e, a, v, value_type_tag, tx, added FROM timelined_transactions WHERE tx = ? AND timeline = ? ORDER BY tx DESC")?;
+    let mut rows = stmt.query_and_then(&[&tx_id, &MAIN_TIMELINE], |row| -> Result<TermWithoutTempIds> {
+        let op = match row.get_checked(5)? {
+            true => OpType::Retract,
+            false => OpType::Add
+        };
+        Ok(Term::AddOrRetract(
+            op,
```
ncalexan: This is where we could use that row helper, although I see you're mapping as well.

db/src/timelines.rs
```diff
@@ -0,0 +1,442 @@
...
+
+    ensure_on_tail_of(conn, &sorted_tx_ids, MAIN_TIMELINE)?;
+
+    let mut last_schema = None;
+    for tx_id in sorted_tx_ids {
+        let reversed_terms = reversed_terms_for(conn, tx_id)?;
+
+        // Rewind schema and datoms.
+        let (_, _, new_schema, _) = transact_terms_with_action(
+            conn, partition_map.clone(), schema, schema, NullWatcher(),
```
ncalexan: Interesting question: what does a `Conn` listener see as this is happening?  Surely it needs to know to update caches, etc...

db/src/timelines.rs
```diff
@@ -0,0 +1,442 @@
...
+
+        assert_matches!(conn.datoms(), r#"
+            [[?e1 :db/ident :test/one]
+             [?e1 :db/valueType :db.type/long]
+             [?e1 :db/cardinality :db.cardinality/one]
+             [?e2 :db/ident :test/many]
+             [?e2 :db/valueType :db.type/long]
+             [?e2 :db/cardinality :db.cardinality/many]]
+        "#);
+        assert_matches!(conn.transactions(), r#"
```
ncalexan: Does this do the right thing for the index flags?  Probably need to do a custom query against `datoms` to check this.

db/src/timelines.rs
```diff
@@ -0,0 +1,442 @@
...
+        Borrow,
+    };
+
+    use debug::{
+        TestConn,
+    };
+
+    use bootstrap;
+
+    // TODO I don't think "moving timelines" belongs in a conn (which would handle this),
```
ncalexan: This is exactly the kind of thing that would live in `TestConn`.  You want rewinding to do the right thing, schema and partition map and all.  (I'm not sure it would live in top-level `Conn`, since it's Tolstoy specific.  Maybe.)

db/src/timelines.rs
```diff
@@ -0,0 +1,442 @@
+// Copyright 2016 Mozilla
```
ncalexan: nit: 2018.

db/src/timelines.rs
```diff
@@ -0,0 +1,442 @@
...
+    conn.execute(&format!(
+        "UPDATE timelined_transactions SET timeline = {} WHERE tx IN ({})",
+            new_timeline,
+            ::repeat_values(tx_ids.len(), 1)
+        ), &(tx_ids.iter().map(|x| x as &rusqlite::types::ToSql).collect::<Vec<_>>())
+    )?;
+    Ok(())
+}
+
+/// Get terms for tx_id, reversing them in meaning (swap add & retract).
```
ncalexan: nit: `tx_id`.

db/src/timelines.rs
```diff
@@ -0,0 +1,442 @@
...
+    Ok(())
+}
+
+fn move_transactions_to(conn: &rusqlite::Connection, tx_ids: &[Entid], new_timeline: Entid) -> Result<()> {
+    // Move specified transactions over to a specified timeline.
+    conn.execute(&format!(
+        "UPDATE timelined_transactions SET timeline = {} WHERE tx IN ({})",
+            new_timeline,
+            ::repeat_values(tx_ids.len(), 1)
+        ), &(tx_ids.iter().map(|x| x as &rusqlite::types::ToSql).collect::<Vec<_>>())
```
ncalexan: Ranges would make this nicer too.

db/src/internal_types.rs
```diff
@@ -181,6 +181,15 @@ impl TermWithTempIds {
     }
 }
 
+// TODO this seems like a sign that I'm doing something wrong...
```
ncalexan: Not really.  I think this is just the `Into` implementation chaining `TermWithoutTempIds -> TermWithTempIds -> TermWithTempIdsAndLookupRefs`.  Break that out as a little Pre:, please.

db/src/timelines.rs
```diff
@@ -0,0 +1,442 @@
...
+            op,
+            KnownEntid(row.get_checked(0)?),
+            row.get_checked(1)?,
+            TypedValue::from_sql_value_pair(row.get_checked(2)?, row.get_checked(3)?)?,
+        ))
+    })?;
+
+    let mut terms = vec![];
+
+    while let Some(row) = rows.next() {
```
ncalexan: `rows.collect()` will let you turn `Vec<Result>` into `Result<Vec>`.

db/src/timelines.rs
```diff
@@ -0,0 +1,442 @@
...
+        assert_matches!(conn.datoms(), r#"
+            [[?e1 :db/ident :test/one]
+             [?e1 :db/valueType :db.type/long]
+             [?e1 :db/cardinality :db.cardinality/one]
+             [?e2 :db/ident :test/many]
+             [?e2 :db/valueType :db.type/long]
+             [?e2 :db/cardinality :db.cardinality/many]]
+        "#);
+        assert_matches!(conn.transactions(), r#"
+            [[[?e1 :db/ident :test/one ?tx1 true]
```
ncalexan: nit: indentation looks off here as well.

db/src/timelines.rs
```diff
@@ -0,0 +1,442 @@
...
+    }
+
+    #[test]
+    fn test_pop_in_sequence() {
+        let mut conn = TestConn::default();
+        conn.sanitized_partition_map();
+
+        let partition_map_after_bootstrap = conn.partition_map.clone();
+
+        assert_eq!((65536..65538),
```
ncalexan: It's a shame to have to do this allocation in these tests.  I have a plan to improve this, however: our matcher is rich, and knows what it matched.  We just need to grow `assert_matches!` to _also_ match the matcher, so that we can feed the tempids into the `assert_matches!` and verify that we really got what we expected.  Then it's all symbolic :)

db/src/lib.rs
```diff
@@ -63,6 +63,8 @@ pub use bootstrap::{
     USER0,
 };
 
+pub static TIMELINE_MAIN: i64 = 0;
```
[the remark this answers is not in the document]
grigoryk: Yeah, this stuff needs some cleanup. I was moving code across commits and things got lost in the noise.

db/src/db.rs
grigoryk: Splitting functions around queries (where db operation is a common bit) makes more sense than the current approach, I think.

db/src/db.rs
```diff
@@ -288,6 +289,35 @@ pub fn create_empty_current_version(conn: &mut rusqlite::Connection) -> Result<(
     Ok((tx, DB::new(bootstrap_partition_map, bootstrap_schema)))
 }
 
+/// Creates a partition map view for the main timeline based on partitions
+/// defined in 'known_parts'.
+fn create_current_partition_view(conn: &rusqlite::Connection) -> Result<()> {
```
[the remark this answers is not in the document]
grigoryk: I'm using it simple as an alias for that query. The nice thing about having an actual view is that we won't need to do the work necessary to create the query every time. The downside is that whenever partitions will change (if they will, in the future) we'd need to change the view as well.

db/src/timelines.rs
```diff
@@ -0,0 +1,442 @@
...
+    TermWithoutTempIds,
+};
+
+use watcher::{
+    NullWatcher,
+};
+
+pub static MAIN_TIMELINE: Entid = 0;
+
+fn ensure_on_tail_of(conn: &rusqlite::Connection, sorted_tx_ids: &[Entid], timeline: Entid) -> Result<()> {
```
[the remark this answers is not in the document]
grigoryk: Great suggestion! Enforcing that at an interface layer is good.

db/src/timelines.rs
grigoryk: I didn't want to hide the nature of this function behind a test interface. E.g. it explicitly _doesn't_ update a conn (which would live layers above, anyway...), so being explicit about it in tests seemed useful. But, having this stuff in `TestConn` is handy.

This is an interesting question, it touches on how we split stuff across crates (e.g. `TestConn` present in `db`...)

db/src/db.rs
```diff
@@ -245,8 +245,9 @@ lazy_static! {
         r#"CREATE INDEX idx_idents_unique ON idents (e, a, v, value_type_tag)"#,
         r#"CREATE TABLE schema (e INTEGER NOT NULL, a SMALLINT NOT NULL, v BLOB NOT NULL, value_type_tag SMALLINT NOT NULL)"#,
         r#"CREATE INDEX idx_schema_unique ON schema (e, a, v, value_type_tag)"#,
+
         // TODO: store entid instead of ident for partition name.
-        r#"CREATE TABLE parts (part TEXT NOT NULL PRIMARY KEY, start INTEGER NOT NULL, end INTEGER NOT NULL, idx INTEGER NOT NULL, allow_excision SMALLINT NOT NULL)"#,
+        r#"CREATE TABLE known_parts (part TEXT NOT NULL PRIMARY KEY, start INTEGER NOT NULL, end INTEGER NOT NULL, allow_excision SMALLINT NOT NULL)"#,
```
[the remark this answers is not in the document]
grigoryk: Yeah, I thought of doing that and agree that this is what we'd like ideally - but it seemed just complicated enough (given my understanding of the internals) to punt on that work right now.

This would open up a rabbit hole down the "user-defined partitions" path which I'm not ready to follow yet.

db/src/db.rs
grigoryk: Everything looks aligned on my end;

db/src/metadata.rs
grigoryk: I tied going down that path initially, but couldn't make the ownership and lifetimes work.

grigoryk:
One thing that might be a bit dicey is making sure transactions stack up well on the target timeline.

The way we're currently thinking of this - as, essentially, a one-way scratch pad for purposes of sync - this isn't necessary. In fact, it's probably not possible to "make sure transactions stack up well", since we're moving just partial slices of the transaction log. What does it mean for one slice to be compatible (i.e. non-conflicting) with another?

Perhaps a good follow-up would be to ensure we can't move transactions onto a non-empty timeline, since there's a good chance _that_ would be quite meaningless. That change would firm up this a little.

I'll address this in the work that will start using timelines.

grigoryk:
@ncalexan I've addressed your feedback in a few separate "review" commits; take a look when you get a chance.

grigoryk:
Err, tests are _passing_, failure on `rust beta` seems to be a fluke of sorts.

db/src/types.rs
```diff
@@ -144,6 +144,9 @@ impl DB {
 /// Used to represent lookup-refs and [TEMPID a v] upserts as they are resolved.
 pub type AVPair = (Entid, TypedValue);
 
+/// Used to represent assertions and retractions.
+pub type EAV = (Entid, Entid, TypedValue);
```
ncalexan: This shouldn't be `pub`, right?

Issue #784, opened 2018-07-13
[tx] Transactor shouldn't treat `:db/doc` as defining schema attributes


ncalexan:
It's valid to use `:db/doc` to describe an entity, not just schema attributes.

ncalexan:
Landed in https://github.com/mozilla/mentat/commit/e9cddd63e446461884641db20e01f02628871bb8.

PR #785, opened 2018-07-13
[tx] Don't treat :db/doc as defining a schema attribute. (#784)

ncalexan:


db/src/db.rs
```diff
@@ -1498,6 +1498,22 @@ mod tests {
...
 
+    #[test]
+    fn test_db_doc_is_not_schema() {
+        let mut conn = TestConn::default();
+
+        // Neither transaction below is defining a new attribute.  That is, it's fine to use :db/doc
+        // to describe any entity in the system, not just attributes.  And in particular, including
+        // :db/doc shouldn't make the transactor consider the entity a schema attribute.
+        assert_transact!(conn, r#"
+            [{:db/doc "test"}]
```
grigoryk: What meaning does this actually have?

ncalexan: It's really from the Lisp (and later, Python, and other languages) tradition of having "docstrings" that self-describe code.  So it makes sense to hang a docstring off just about anything; in this case, this allocates a new entity with only one assertion about it: `[e :db/doc "test"]`.

PR #788, opened 2018-07-17
Enforce partition integrity & PartitionMap newtype

grigoryk:
Peeling off bits and pieces off of timelines work.

Partition integrity stuff helps us ensure we're doing sane things when moving an index in irregular ways. PartitionMap newtype conversion is non-controversial and helps clean up things a little bit.


db/src/db.rs
```diff
@@ -1108,9 +1103,7 @@ impl PartitionMap {
 mod tests {
     extern crate env_logger;
 
-    use std::borrow::{
-        Borrow,
-    };
+    use std::borrow::Borrow;
```
ncalexan: Eh?  Generally I've been trying to migrate to the `{\n}` style, since it's consistent.

grigoryk: Sure!

PR #790, opened 2018-07-18
WIP Run Android and iOS tests in CI

fluffyemily:


fluffyemily:
@ncalexan I was going to ask you about Android in travis and whether it was needed any longer. I'm glad it's not because travis doesn't have enough disk space for a full build and test of all targets!

iOS is nearly done in travis - just a couple of niggles to get through, plus failing iOS build & tests which I have addressed here: https://github.com/mozilla/mentat/pull/803

PR #798, opened 2018-07-23
Make sure mentat error types preserve failure backtraces (wip)

thomcc:
This adds a lot of boilerplate which could be simplified by macros.

I was planning on cleaning this up before pushing it but here it is!

### DIFF
--- a/query-algebrizer/src/errors.rs
+++ b/query-algebrizer/src/errors.rs
@@ -21,6 +21,12 @@ use mentat_core::{
 use self::mentat_query::{
     PlainSymbol,
 };
+use std::fmt;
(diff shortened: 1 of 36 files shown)


query-algebrizer/src/errors.rs
```diff
@@ -31,6 +37,49 @@ macro_rules! bail {
     )
 }
 
+#[derive(Debug)]
+pub struct AlgebrizerError(Box<Context<AlgebrizerErrorKind>>);
```
grigoryk: It's not clear to me why you're boxing the `Context`. Can you explain why?

PR #799, opened 2018-07-26
Android SDK: Remove unused "app_name" string.

pocmo:
It seems like this string is unused (although I'm unable to build the library locally) and a library overriding this default string can cause funny problems. :)


ncalexan:
Thanks -- this is done in https://github.com/mozilla/mentat/commits/build-android-sdk, which I'm about to land.

PR #804, opened 2018-08-08
Adding iOS Build and Test to CI

fluffyemily:


ncalexan:
It looks like the build directories with all of the Rust targets are too big for Travis CI's build caching.  Should we `cargo clean`, so that we keep VCS checkouts but drop object files?  Can we split targets across jobs?

.travis.yml
```diff
@@ -24,6 +34,14 @@ script:
     for manifest in $(find . -type f -name Cargo.toml); do
       cargo test --manifest-path $manifest --verbose --no-default-features --features sqlcipher
     done
+  - |
+    if [[ "$TRAVIS_RUST_VERSION" = "1.25.0" ]]; then
+      cd ffi &&
```
ncalexan: Consider `pushd` and `popd`.

.travis.yml
```diff
@@ -13,6 +13,16 @@ matrix:
   allow_failures:
     - rust: nightly
   fast_finish: true
+install:
+  - |
+    if [[ "$TRAVIS_RUST_VERSION" = "1.25.0" ]]; then
+      rustup target add aarch64-apple-ios armv7-apple-ios armv7s-apple-ios x86_64-apple-ios i386-apple-ios &&
```
ncalexan: That's a lot of targets.  Do we actually run tests on all of these targets?

.travis.yml
```diff
@@ -13,6 +13,16 @@ matrix:
...
+      # Cargo-lipo needs to be Beta version available on master and not crates version to handle building in subdirectories
+      cargo install --git https://github.com/TimNN/cargo-lipo &&
```
ncalexan: Is it possible to pin this version to a SHA?  We shouldn't be fetching something random off the Web, basically ever.

.travis.yml
```diff
@@ -13,6 +13,16 @@ matrix:
...
+      # If we don't list the devices available then when we come to pick on during the test run travis can't find our device.
```
ncalexan: nit: s/pick on/pick one/.
nit: full sentence with punctuation, here and above.

.travis.yml
```diff
@@ -24,6 +34,14 @@ script:
     for manifest in $(find . -type f -name Cargo.toml); do
       cargo test --manifest-path $manifest --verbose --no-default-features --features sqlcipher
     done
+  - |
+    if [[ "$TRAVIS_RUST_VERSION" = "1.25.0" ]]; then
```
ncalexan: Why are we only doing this for one Rust version?

.travis.yml
fluffyemily: We build a universal library that contains all of the targets so we do need to add all of them.

PR #805, opened 2018-08-08
Move MentatError away from the top-level crate

grigoryk:
This PR is in preparation for landing some of the work in #563 - specifically, the Pre work which splits `transaction`-like functionality away from the top-level into its own inner crate.

`Pre` commits move any trait or type that is necessary to compose an error out of a given crate, and into a "leaf" crate, which use the `*-traits` naming convention (e.g. `db` and `db-traits`). This pattern is similar to how Servo structures their internal inter-dependencies, although at a weaker level (since I'm retrofitting this pattern onto the current world, and going all-in on it would be quite a bit of work!).

This move is needed so that we can use a single Error type in our public API.

Without these changes, the Pre patch above would have to leak a `TransactionError` into mentat's public API space, which is a subset of the `MentatError`. If the `MentatError` is defined in an inner crate, it can now be referenced in other inner crates, in absence of circular dependencies.

This pattern is useful for crates which:
- tie a lot of internal functionality together
- are exposed as part of the public API
- themselves don't introduce new error types

A prime example is the "mentat transaction state manager" (the `transaction` crate in #563), and the upcoming `tolstoy` which makes heavy use of the `transaction` crate.


grigoryk:
This introduces a temporary inconsistency in how crates are named - the new crates I've added don't use the "mentat_" prefix. The intention is to remove that prefix entirely, but that will be done once work that depends on this - in #563 - lands, to avoid an even more annoying rebase.

src/errors.rs
```diff
@@ -72,13 +72,13 @@ pub enum MentatError {
     InvalidVocabularyVersion,
 
     #[fail(display = "vocabulary {}/{} already has attribute {}, and the requested definition differs", _0, _1, _2)]
-    ConflictingAttributeDefinitions(String, ::vocabulary::Version, String, Attribute, Attribute),
```
ncalexan: In all of these messages, ensure that all of these `u32` values are printed with "version {}" so that it's clear what the integer means.  So
```     #[fail(display = "vocabulary {}/version {} already has attribute {}, and the requested definition differs", _0, _1, _2)] ```

grigoryk:
I don't _expect_ the public API to have changed - but it's probably a fair statement that it did in some subtle way, and we currently don't have much in place to ensure that it didn't change. Regarding the lack of `src/` in the subcrates, it just seemed a bit neater to me.

PR #807, opened 2018-08-20
Split "mentat transaction" logic into its own crate

grigoryk:
Splitting some more "pre" work from #563.

Paraphrased from the discussion in that PR:

Splitting high level metadata logic away from the top level is needed in some form to enable nicer patterns for the internal consumers (well, just `tolstoy` for now). Further refactoring of this logic should be easier from this point on, and the consumers building on top of the `transaction` crate won't be affected significantly - likely just a few import changes.

As it stands, `transaction` right now is a thin layer on top of `db`, doing some bookkeeping one layer above it. I'm not a fan of having Store and Conn at the top level, either - it makes writing tests that want to use a real Conn awkward, forcing them to be at the top level, away from the functionality that's being tested. The TestConn surrogate is a smell of that. So, there's like more re-shuffling to be done! My hope is that little steps like this will help inform us what the "ideal" world might look like.


grigoryk:
Carrying over r+ from @ncalexan in #563.

PR #808, opened 2018-08-20
Allow 'sqlcipher' feature for all uses of rusqlite

grigoryk:
This also patches our CI test script to only run "--feature sqlcipher"
tests on sub-crates which expose this feature (i.e. themselves rely on rusqlite).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,10 @@ script:
   # We can't pick individual features out with `cargo test --all` (At the time of this writing, this
   # works but does the wrong thing because of a bug in cargo, but its fix will be to disallow doing
   # this all-together, see https://github.com/rust-lang/cargo/issues/5364 for more information). To
-  # work around this, we run individual tests for each subcrate individually.
+  # work around this, we run tests individually for subcrates that rely on `rusqlite`.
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -19,10 +19,10 @@ script:
   # We can't pick individual features out with `cargo test --all` (At the time of this writing, this
   # works but does the wrong thing because of a bug in cargo, but its fix will be to disallow doing
   # this all-together, see https://github.com/rust-lang/cargo/issues/5364 for more information). To
-  # work around this, we run individual tests for each subcrate individually.
+  # work around this, we run tests individually for subcrates that rely on `rusqlite`.
   - |
-    for manifest in $(find . -type f -name Cargo.toml); do
-      cargo test --manifest-path $manifest --verbose --no-default-features --features sqlcipher
+    for crate in "" "db" "db-traits" "ffi" "public-traits" "query-projector" "query-projector-traits" "query-pull" "sql" "tolstoy" "tolstoy-traits" "tools/cli"; do
```
thomcc: This is somewhat fragile but I don't see an alternative :(

PR #817, opened 2018-08-30
add method to return current schema to android bindings

sc13-bioinf:
Needs a review regarding safety/correctness. I am pretty new to rust / JNA.


grigoryk:
Hi, and thanks for the PR!

Could you explain a bit what your use case is for this?

sc13-bioinf:
It is useful for repl development. Also it is useful if you are trying to copy data to datascript.

PR #819, opened 2018-08-31
Make sure double retractions are not inserted. Fixes #818

### LINKED ISSUE #818: Double retractions in the transaction log
Some fun transactor behaviour:
- (given cardinality one, unique/identity :test/ident)
- first transact: `{:test/ident "One"}` - say, it gets assigned e=1
- then transact:
```
[[:db/add 1 :test/ident "Two"][:db/retract 1 :test/ident "One"]]
```
- then observe transaction log:
```
[[1 :test/ident "Two" 1]
[1 :test/ident "One" 0]
[1 :test/ident "One" 0]]
```

This is likely happening due to the way retractions are processed out of temp_search_results and into the log. We check for change in values, and synthesize a retraction datom, but we also actually have a retraction datom as well, so we end up with a duplicate retraction when there should have been just one.

grigoryk:


grigoryk:
@ncalexan this is failing tests on 1.25.0 for reasons seemingly unrelated to the patch itself.

ncalexan (commented):
lgtm.  I would have preferred to ensure that we only insert into the search table once, but c'est la vie.

PR #820, opened 2018-09-01
Add method to get the ValueType of a TypedValue as a keyword to Android API

sc13-bioinf:
Also remove a println (leftover from debug?)

This makes life easier when dealing with query results, as it allows them to be extracted without knowing the types in advance. This makes use from clojure easier.

Returning standard data structures and types instead (commented on elsewhere) is something that I think is also worth a look.

I would be interested in helping out with this project so I would be happy to chat sometime.

### DIFF
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -1556,7 +1556,6 @@ pub unsafe extern "C" fn typed_value_into_long(typed_value: *mut Binding) -> c_l
 pub unsafe extern "C" fn typed_value_into_entid(typed_value: *mut Binding) -> Entid {
     assert_not_null!(typed_value);
     let typed_value = Box::from_raw(typed_value);
-    println!("typed value as entid {:?}", typed_value);
     unwrap_conversion(typed_value.into_entid(), ValueType::Ref)
 }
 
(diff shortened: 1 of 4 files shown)


ffi/src/lib.rs
```diff
@@ -1666,6 +1666,14 @@ pub unsafe extern "C" fn typed_value_value_type_kw(typed_value: *mut Binding) ->
     string_to_c_char(value.clone())
 }
 
+/// Returns the size of a `row` as usize
```
ncalexan: nit: full sentences please, so: trailing period.

ffi/src/lib.rs
```diff
@@ -1666,6 +1666,14 @@ pub unsafe extern "C" fn typed_value_value_type_kw(typed_value: *mut Binding) ->
...
+#[no_mangle]
+pub unsafe extern "C" fn typed_value_list_size(values: *mut Vec<Binding>) -> c_int {
+    assert_not_null!(values);
+    let values = &*values;
+    values.len () as c_int
+}
+
 /// Returns the value at the provided `index` as a `Vec<ValueType>`.
```
ncalexan: nit: no space between function and arguments, so: `.len()`.

ncalexan (commented):
I think this works, but we're allocating memory on both sides of the FFI boundary.  If we turned the underlying `int` into a static Java `String` on the Java side of the FFI boundary, we'd both have a simpler memory-safety story, and we'd allocate less.  How do you feel about doing that?

ffi/src/lib.rs
sc13-bioinf: The formatting is done. I am happy to try and implement your suggestion but I don't understand. Please could you walk me through it? What should all the return types be? Should the java class size method return a String?

sc13-bioinf: Ok, my current idea would be to do this by passing a direct allocated static final byteBuffer from java. To do this it seems easiest if I have access to JNIEnv, is there a reason why you don't seem to use this style?
https://mozilla.github.io/firefox-browser-architecture/experiments/2017-09-21-rust-on-android.html

Would need to use synchronise(this) around the call to rust within the constructor.

